PR #10, opened 2015-01-27
Add plot with black background in examples

lesteve:
in order to show the impact of commit d5a5507cd9d4b2ba4af0bf4cf42bbda043853cb5.

Maybe the matplotlib example can be simplified but it does the job...

For completeness, here how the the saved image looks like:

### Before the fix
sphinx-gallery-black-background-before-fix

### After the fix
sphinx-gallery-black-background-after-fix


agramfort:
thanks @lesteve !

PR #16, opened 2015-02-05
add support for mayavi

agramfort:


agramfort:
I fear a rebase mess here. No time to do it.

please take over.

sphinxgallery/gen_rst.py
```diff
@@ -20,6 +20,7 @@
 import posixpath
 import subprocess
 import warnings
+from copy import deepcopy
```
Titan-C: this deepcopy is not used?

sphinxgallery/gen_rst.py
```diff
@@ -987,6 +1040,9 @@ def embed_code_links(app, exception):
         'numpy': 'http://docs.scipy.org/doc/numpy-1.6.0',
         'scipy': 'http://docs.scipy.org/doc/scipy-0.11.0/reference',
     }
+    if has_mayavi:
```
Titan-C: The new version of sphinx-gallery handles this urls as user input. Shall this considered as default documented package?

sphinxgallery/gen_rst.py
agramfort: yes please remove it.

sphinxgallery/gen_rst.py
agramfort: not a default, just optional. I'll need it for mne-python.

sphinxgallery/gen_rst.py
```diff
@@ -118,6 +129,31 @@ def _get_data(url):
...
+    """Scale image such that width <= max_width
+    """
+    try:
+        from PIL import Image
+    except ImportError:
+        import Image
+    img = Image.open(in_fname)
+    width_in, height_in = img.size
+
+    if width_in <= max_width:
```
Titan-C: this is possibly the only extra clause that distinguishes this function, from the function that does the thumbnails.
Is there a reason to use this instead than the `make_thumbnail` with different input parameters and add there a clause like 
```
if scale >1.0:
    return
```

agramfort:
closing in favor or #17

PR #17, opened 2015-02-05
add mayavi support

Titan-C:
Based on the PR of @agramfort to incorporate mayavi support as default.

to use the documentation references in the example scripts
the sphinx `conf.py` needs to include
```python
    sphinxgallery_conf = {
        'doc_module'        : 'yourmodule',      # Your module
        'resolver_urls'     : {                     # External python modules documentation websites
            'matplotlib': 'http://matplotlib.org',
            'numpy': 'http://docs.scipy.org/doc/numpy-1.6.0',
            'scipy': 'http://docs.scipy.org/doc/scipy-0.11.0/reference',
            'mayavi': 'http://docs.enthought.com/mayavi/mayavi'}
```
more info in: http://sphinx-gallery.readthedocs.org/en/latest/advanced_configuration.html#linking-to-external-documentations

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -18,15 +18,6 @@ Contents:
    getting_started
    auto_examples/index
 
-.. include:: modules/generated/numpy.linspace.examples
-.. raw:: html
-
-    <div style='clear:both'></div>
-
-.. include:: modules/generated/numpy.exp.examples
-.. raw:: html
-
-    <div style='clear:both'></div>
 
 Indices and tables
 ==================
--- a/examples/plot_strings.py
+++ b/examples/plot_strings.py
@@ -1,0 +1,16 @@
+# -*- coding: utf-8 -*-
+"""
+==============================
+Constrainded Text output frame
+==============================
+
+This example captures the stdout and includes it in the
+example. If output is too long it becomes automatically
+framed into a text area.
+
+"""
+
+# Code source: Óscar Nájera
+# License: BSD 3 clause
+
+print('This is a long test Output\n'*800)
--- a/sphinxgallery/_static/gallery.css
+++ b/sphinxgallery/_static/gallery.css
@@ -71,3 +71,9 @@ div.thumbnailContainer div p {
 div.thumbnailContainer div a {
     display: block; /* To have a better hover behavior */
 }
+
+.max-height pre {
+    /* For very long outpt: restrict the max height */
+    overflow-y: scroll;
+    max-height: 30em;
+}
--- a/sphinxgallery/gen_gallery.py
+++ b/sphinxgallery/gen_gallery.py
@@ -57,6 +57,7 @@ gallery_conf = {
     'examples_gallery'  : 'auto_examples',
     'mod_generated'     : 'modules/generated',
     'doc_module'        : 'sphinxgallery',
+    'use_mayavi'        : False,
     'resolver_urls'     : {
         'matplotlib': 'http://matplotlib.org',
         'numpy': 'http://docs.scipy.org/doc/numpy-1.9.1',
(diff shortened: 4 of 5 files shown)


agramfort:
travis complains

LGTM but @GaelVaroquaux was suggesting to allow not using mayavi even if it is available. e.g. using a conf.py flag

Titan-C:
Indeed travis complains in the anaconda python3 environment. I notice it also complains in your pull request, but I'm unable to understand why it fails.

I'm also for having it as something you can configure, or an extension. But I'll need help on that, I can't think of such a solution now. I wanted to complete the work today as I have the code fresh in my mind. And give you time to test it for a while.

agramfort:
>
> Indeed travis complains in the anaconda python3 environment. I notice it
> also complains in your pull request, but I'm unable to understand why it
> fails.
>
does it fail for you with a py3k setup?


> I'm also for having it as something you can configure, or an extension.
> But I'll need help on that, I can't think of such a solution now. I wanted
> to complete the work today as I have the code fresh in my mind. And give
> you time to test it for a while.
>

an enable_mayavi param in conf.py ?

GaelVaroquaux:
> LGTM but @GaelVaroquaux was suggesting to allow not using mayavi even if it is
> available. e.g. using a conf.py flag

Yes. The import of mayavi is slow, and might trigger all kind of problems
(for instance it needs an X server), so it should not be on by default.

agramfort:
you need to rebase.

any plan to merge this soon?

I'd like to switch mne-python soon.

GaelVaroquaux:
Once again: we should not try/except the import of mayavi if not requested in the config file. Indeed, first it is going to be very slow. Second, this import might might segfault.

So, we need to add a switch in the config file before this gets merged.

Titan-C:
Is it a good thing to globally import a module from a function? I would propose to make it like this

```python
def config_mayavi(gallery_config):
  if gallery_config['use_mayavi']:
      global mlab
      try:
          from mayavi import mlab
          return True
      except Exception as e:
          try:
              from enthought.mayavi import mlab
              return True
          except Exception as e:
              raise
    else:
        return False
```

agramfort:
sounds good to me

GaelVaroquaux:
> sounds good to me

+1

Titan-C:
This is done, not sure if it works as I don't have mayavi. It does not break the old build.
It would be good if you give a check on the pull request of the config dictionary, since I have changed some naming.

agramfort:
> This is done, not sure if it works as I don't have mayavi. It does not break the old build.
> It would be good if you give a check on the pull request of the config dictionary, since I have changed some naming.

conda install mayavi?
sudo apt-get install mayavi?

please test and let us know if it works.

Titan-C:
No it doesn't work. It manages to go through the gallery and execute all examples but in step when sphinx writes the output there is an error. The function compliaining is the get_tokens_unprocessed in pygments_highlighter.py from sphinx.
The longer sphinx log, traces a long chain of called functions between sphinx pygments and docutils. I don't understand the log to follow the bug.

agramfort:
sorry. I cannot look into it. No time :(

agramfort:
this becomes a pressing need for mne folks during the sprint.

shall I take over?

Titan-C:
The code has changes a lot since this last pull request. Re branching from master might be better than rebasing.
PR #36 is also waiting to get review and pushed in. But it is so nicely modular and includes notebook like rendering of the examples(Execution by blocks). I would branch from there. In that case the code for rendering mayavi would need to be attached into the function `save_figures` `in gen_rst.py`

Titan-C:
Yes please take over on this feature. Also we shall plan me joining the sprint, is it the whole week right?

agramfort:
cool thanks.

Yes please join even today. We're at La Paillasse 226 rue de St Denis
(métro strasbourg st denis)

I'd love to team up with you to integrate sphinx-gallery to MNE

lesteve:
closed in favor of #42.

Issue #18, opened 2015-02-16
Unicode support


Titan-C:
I came to notice that if the examples docstring have an unicode character, sphinx-gallery fails in python2. It is not the case if the unicode characters are outside the main docstring, because the encoding header of the file makes the scripts work. The reason the files having my foreign accented name still work.

this PR https://github.com/scikit-learn/scikit-learn/pull/3777 didn't help with my issue.

PR #19, opened 2015-02-16
Including a minimal unicode support for examples docstrings

Titan-C:
This is proposed as a short solution to address issue #18. Problem never arises in python 3 as unicode is default string. For dealing with python2 the use of the codecs module is included.

### DIFF
--- a/examples/plot_strings.py
+++ b/examples/plot_strings.py
@@ -13,4 +13,4 @@ framed into a text area.
 # Code source: Óscar Nájera
 # License: BSD 3 clause
 
-print('This is a long test Output\n'*50)
+print('This is a long test Output\n' * 50)
--- a/examples/plot_strings.py
+++ b/examples/plot_strings.py
@@ -13,4 +13,4 @@ framed into a text area.
 # Code source: Óscar Nájera
 # License: BSD 3 clause
 
-print('This is a long test Output\n'*800)
+print('This is a long test Output\n'*100)
--- a/examples/plot_strings.py
+++ b/examples/plot_strings.py
@@ -13,4 +13,4 @@ framed into a text area.
 # Code source: Óscar Nájera
 # License: BSD 3 clause
 
-print('This is a long test Output\n'*100)
+print('This is a long test Output\n'*50)
--- a/examples/plot_unicode.py
+++ b/examples/plot_unicode.py
@@ -15,7 +15,7 @@ import matplotlib.pyplot as plt
 
 
 N = 50
-x = np.linspace(-1,2,50)
+x = np.linspace(-1, 2, 50)
 colors = np.random.rand(N)
 area = np.pi * (15 * np.random.rand(N))**2
 for i in range(5):
--- a/examples/plot_unicode.py
+++ b/examples/plot_unicode.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+"""
+=======================================
+Únîcödè support in documentation string
+=======================================
+
+In this example we include Unicode characters that are so required this days.✓
+"""
+
+# Code source: Óscar Nájera
+# License: BSD 3 clause
+
+import numpy as np
+import matplotlib.pyplot as plt
+
+
+N = 50
+x = np.linspace(-1,2,50)
+colors = np.random.rand(N)
+area = np.pi * (15 * np.random.rand(N))**2
+for i in range(5):
+    plt.scatter(x, -np.cos(i*x), s=area, c=colors, alpha=0.5)
+plt.show()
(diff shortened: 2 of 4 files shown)


examples/plot_strings.py
```diff
@@ -13,4 +13,4 @@
 # Code source: Óscar Nájera
 # License: BSD 3 clause
 
-print('This is a long test Output\n'*800)
\ No newline at end of file
+print('This is a long test Output\n'*50)
```
agramfort: pep8 spaces around *

agramfort:
besides LGTM

I assume you tested :)

examples/plot_unicode.py
```diff
@@ -0,0 +1,23 @@
...
+
+# Code source: Óscar Nájera
+# License: BSD 3 clause
+
+import numpy as np
+import matplotlib.pyplot as plt
+
+
+N = 50
+x = np.linspace(-1,2,50)
```
lesteve: Another PEP8 here

GaelVaroquaux: I am one who thinks that Unicode in source code is a bad thing. The reason is that Unicode characters are characters that not everybody can read or type. In a sens, not supporting it forces people not to use it, as they are never something necessary for an example.

Titan-C: > I am one who thinks that Unicode in source code is a bad thing.

I think this more a matter of timing. As I said with python 3 having unicode as default implies the need for foreign languages support and that it is quite an issue. And this bug fix is not even necessary in python 3.
The reason I'm proposing this is not because of having unicode in the code. That already worked as you can find by checking that my examples included my Spanish accented name. And I agree that the source code should be the most plain text possible. I propose this because it is the main docstring of the example that could not include foreign characters. Docstrings are aimed for a broader audience than the programmers, so that they are most certainly are in the target audience most conforming setup, there every body would be able to read.

GaelVaroquaux: I still don't see the value of non ASCII characters in the docstring. You
and I have a non ASCII character in our name. But we accept perfectly the
fact that we might get email without it.

Inversely, if you get an email, that says 'здравствуйте', it makes it
really hard for you to search your mailbox with this email. It makes it
hard to fix a typo in this mail... I am against the use of non-ASCII
character in a whole software project that is not specificaly about
localization.

GaelVaroquaux: TODO items before merge:

* joblib must be replaced by shelve

Remarks for later (not preventing merge in nilearn)

* _path_static should be renamed path_static, as it is used in the conf.py

* the css names should be harmonized. In particular,  they should be all given w prefix, such as sphx-glr-...

* functions should not be defined in closures: _select_block, _parse_dict_structure must be moved out of the function where they are defined

* no @author in docstrings, but author as a comment in the beginning of a file,  with also the full name of the author, the original authors if any,  and "license: BSD"

* the code isn't completely pep8: there should be two empty lines between first level functions

* later: this code needs tests

GaelVaroquaux: Sorry. I sent this to the wrong issue. That's the problem with reviewing
PRs on the phone. I'll put it elsewhere.

G

On Wed, Feb 25, 2015 at 09:26:54AM +0100, Gael Varoquaux wrote:
> TODO items before merge:

> * joblib must be replaced by shelve

> Remarks for later (not preventing merge in nilearn)

> * _path_static should be renamed path_static, as it is used in the conf.py

> * the css names should be harmonized. In particular,  they should be all given
> w prefix, such as sphx-glr-...

> * functions should not be defined in closures: _select_block,
> _parse_dict_structure must be moved out of the function where they are defined

> * no @author in docstrings, but author as a comment in the beginning of a
> file,  with also the full name of the author, the original authors if any,  and
> "license: BSD"

> * the code isn't completely pep8: there should be two empty lines between first
> level functions

> * later: this code needs tests

> Sent from my phone. Please forgive brevity and mis spelling

> On Feb 17, 2015, at 15:39, Gael Varoquaux <gael.varoquaux@normalesup.org>
> wrote:

>     I still don't see the value of non ASCII characters in the docstring. You
>     and I have a non ASCII character in our name. But we accept perfectly the
>     fact that we might get email without it.

>     Inversely, if you get an email, that says 'здравствуйте', it makes it
>     really hard for you to search your mailbox with this email. It makes it
>     hard to fix a typo in this mail... I am against the use of non-ASCII
>     character in a whole software project that is not specificaly about
>     localization.


-- 
    Gael Varoquaux
    Researcher, INRIA Parietal
    Laboratoire de Neuro-Imagerie Assistee par Ordinateur
    NeuroSpin/CEA Saclay , Bat 145, 91191 Gif-sur-Yvette France
    Phone:  ++ 33-1-69-08-79-68
    http://gael-varoquaux.info            http://twitter.com/GaelVaroquaux

PR #21, opened 2015-03-07
Use of shelve

Titan-C:
The main update is the removal of the joblib dependency in order to use the python shelve module instead

To this update I added:
- the license files and comments
- use of dictionary update for the gallery configuration dictionary
- update path_static function name


sphinxgallery/tests/test_gen_gallery.py
```diff
@@ -0,0 +1,33 @@
...
+Testing the rst files generator
+"""
+from __future__ import division, absolute_import, print_function
+import sphinxgallery.docs_resolv as sg
+import tempfile
+
+
+def _get_data(fid, shelve_file, test_string):
+    data = sg.get_data(fid, shelve_file)
+    assert data == test_string
```
agramfort: use assert_equal from nose

agramfort:
are you done here?

Titan-C:
>are you done here?
What do you mean? I'm not sure to really understand you.

agramfort:
:)

I mean do you consider this branch ready to merge or work in progress?

Titan-C:
this is already working. I thought I wrote this, but I wanted feedback on the way I'm dealing with the string types and unicode. I was quite a pain, because the shelve in py2 can only have strings as keys, but sphinx gives unicode. Whereas in py3 unicode is the default string, so unicode function did not exist anymore, but you had to transform unicode to bytes to use the write to file.

In the end I decided to try: py2 except: py3

agramfort:
PR looks clean.

I guess to know if one should merge is does not work on realistic doc such as nilearn?

we should then talk about using sphinx-gallery for mne

sphinxgallery/docs_resolv.py
```diff
@@ -42,12 +46,28 @@ def _get_data(url):
...
-get_data = mem.cache(_get_data)
+
+def get_data(url, cached_file='_build/searchindex'):
+    """Persistent dictionary usage to retrieve the search indexes"""
+
+    try:
+        if isinstance(url, unicode):
+            url = url.encode('utf-8')
+    except NameError:
+        pass
```
lesteve: This seems a bit convoluted. I have to admit I didn't know off the top of my head that unicode wasn't a builtin in python 3.

I would do something like that instead:

```python
# shelve keys need to be str in python 2
if six.PY2 and isinstance(url, unicode):
    url = url.encode('utf-8')
```

Also this should be done outside of this PR and is not really high-priority but it'd be nice to have a cache function in sphinx-gallery (similar to jolib.Memory.cache) because you may want to cache other stuff than url content at some point in the future.

sphinxgallery/tests/test_gen_gallery.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def test_shelve():
+    test_string = 'test information'
+    shelve_file = 'shelve_file'
+    with tempfile.NamedTemporaryFile() as fid:
+        try:
+            fid.write(test_string)
+        except TypeError:
+            fid.write(test_string.encode())
```
lesteve: This seems to work for me:

```python

with tempfile.NamedTemporaryFile('w') as fid:
    fid.write(test_string)
```

lesteve: > This seems to work for me:

i.e. for both python 2 and 3

sphinxgallery/tests/test_gen_gallery.py
```diff
@@ -0,0 +1,34 @@
...
+"""
+Testing the rst files generator
+"""
+from __future__ import division, absolute_import, print_function
+import sphinxgallery.docs_resolv as sg
+import tempfile
+from nose.tools import assert_equals
+
+
+def _get_data(fid, shelve_file, test_string):
```
lesteve: I am not sure this helper function helps that much:
- naming isn't great, same function exists in docs_resolv and it doesn't give you a hint that this actually tests something
- when you are glancing at test_shelve it is not immediately clear where the assert actually happens

Basically I would just replace the function call by its code in test_shelve.

sphinxgallery/tests/test_gen_gallery.py
```diff
@@ -0,0 +1,34 @@
...
+    with tempfile.NamedTemporaryFile() as fid:
+        try:
+            fid.write(test_string)
+        except TypeError:
+            fid.write(test_string.encode())
+
+        fid.seek(0)
+        _get_data(fid.name, shelve_file, test_string)
+
+    _get_data(fid.name, shelve_file, test_string)
```
lesteve: Just curious, why is this line the exact same one as two lines above?

sphinxgallery/tests/test_gen_gallery.py
```diff
@@ -0,0 +1,34 @@
...
+    try:
+        _get_data(unicode(fid.name), shelve_file, test_string)
+    except NameError:
+        pass
```
lesteve: Maybe something like this is simpler to grok:

```python
if six.PY2:
    _get_data(unicode(fid.name), shelve_file, test_string)
```

Titan-C: Is six default or an extra dependency?

sphinxgallery/tests/test_gen_gallery.py
Titan-C: because it is outside the with, the temporary file does not exist anymore. This was to test I'm actually reading from the shelve and not from file anymore.

sphinxgallery/tests/test_gen_gallery.py
lesteve: I assumed you already had six in your dependencies to do the python2 python3 compatibility apparently I was wrong. For six.PY2 it's very easy to have something equivalent with sys.version_info.

sphinxgallery/tests/test_gen_gallery.py
lesteve: Got you, you may want to add comments to explain shortly what each assert is testing.

lesteve:
LGTM merging.

PR #23, opened 2015-03-28
Naming convention for configuration

Titan-C:
This pull request includes my proposal for the names in the configuration/setup of the sphinx-gallery.
* I decided to keep the configuration in a dictionary, and a nested dictionary in it for the reference urls of other documentations. Keywords refering to directories have the _dir suffix.
* Keyword `doc_module` is a string or list of stings containing the local modules for which examples are provided, and one later wants to link when documenting modules.
example: http://sphinx-gallery-local.readthedocs.org/en/config/advanced_configuration.html#establishing-local-references-to-examples

* Keyword `reference_url` is a new dictionary containing module name and url to online documentation of that same module. This is the one that allows to put the links in each command of the python scripts presented in the gallery to the online reference 

The default configuration dictionary is stored in `gen_gallery.py` and the extension of it, used for the sphinx-gallery documentation is in the `conf.py`. The documentation has been extended to match the new setup.

This pull request is build at http://sphinx-gallery-local.readthedocs.org/en/config/

### DIFF
--- a/doc/advanced_configuration.rst
+++ b/doc/advanced_configuration.rst
@@ -3,7 +3,7 @@ Advanced Configuration
 ======================
 
 Here are the personal configurations that you can modify within Sphinx-Gallery.
-What is writen from here on are the default values.
+
 
 Changing default directories
 ============================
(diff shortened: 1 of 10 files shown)


doc/advanced_configuration.rst
```diff
@@ -13,51 +13,70 @@ Within your Sphinx ``conf.py`` file you need to add a configuration dictionary:
...
 
     sphinxgallery_conf = {
-        'doc_module'        : 'sphinxgallery',      # Your module
-        'resolver_urls'     : {                     # External python modules documentation websites
-            'matplotlib': 'http://matplotlib.org',
-            'numpy': 'http://docs.scipy.org/doc/numpy-1.9.1',
-            'scipy': 'http://docs.scipy.org/doc/scipy-0.15.1/reference'}
+        'reference_url':  {
+                 # Modules you locally document use an empty string
+                'sphinxgallery': '',
```
lesteve: Maybe a None rather than an empty string is more appropriate?

sphinxgallery/docs_resolv.py
```diff
@@ -324,12 +324,16 @@ def embed_code_links(app, exception):
...
-                                                     relative=True)
 
-    for this_module, url in gallery_conf['resolver_urls'].items():
+    for this_module, url in gallery_conf['reference_url'].items():
         try:
-            doc_resolvers[this_module] = SphinxDocLinkResolver(url)
+            if url == '':
+                doc_resolvers[this_module] = SphinxDocLinkResolver(
+                                                            app.builder.outdir,
+                                                            relative=True)
```
lesteve: Doesn't look very PEP8 to me

doc/advanced_configuration.rst
```diff
@@ -13,51 +13,70 @@ Within your Sphinx ``conf.py`` file you need to add a configuration dictionary:
...
+        # path where to store your example linker templates
+        'mod_example_dir'     : 'modules/generated',
 
-Then within your sphinx documentation files you include this lines::
+        # Your documented modules. You can use a string or a list of strings
+        'doc_module'          : ('sphinxgallery', 'numpy')}
+
+The path you specified will get populated with the links to examples using your
+module and their methods. Then within your sphinx documentation files you
+include this lines to include this links::
```
lesteve: this -> these (2 times)

sphinxgallery/gen_rst.py
```diff
@@ -439,7 +439,7 @@ def identify_names(code):
     return example_code_obj
 
 
-def generate_file_rst(fname, target_dir, src_dir, root_dir, gallery_conf, plot_gallery):
+def generate_file_rst(fname, target_dir, src_dir, gallery_conf, plot_gallery):
     """ Generate the rst file for a given example.
 
     Returns the set of sklearn functions/classes imported in the example.
```
lesteve: I guess the sklearn mention should be removed.

doc/advanced_configuration.rst
```diff
@@ -13,51 +13,70 @@ Within your Sphinx ``conf.py`` file you need to add a configuration dictionary:
...
 
+
 Linking to external documentations
 ==================================
 
-If you want to hyperlink commands in your example scripts you can. Again within
-your Sphinx ``conf.py`` file you need to add a configuration dictionary:
+Sphinx-Gallery enables you to hyperlink commands in your example scripts to the
+matching location in their online documentation . Have a look at this in action
+in our example :ref:`example_plot_gallery_version.py`.
```
lesteve: This sounds quite vague to me and it actually took me a while to figure out what that actually meant, i.e. that some functions in the example html were clickable to access the online documentation of packages used in the example, e.g. numpy, matplotlib, etc .... Maybe it would help to add a few words why this is useful or mention an explicit line to click on in the example.

Loosely related it doesn't help that clickable functions are not visible in the example. I agree that for readability underline is not necessarily the best option. Maybe some CSS tricks on hover would be good enough.

Titan-C: Indeed I have a hard time explaining this. I'll try to put a clickable code snippet bellow to exemplify this.

The CSS for such links it is a matter of choice how to put them. The read the docs css removes any style to hyperlinks, so it shall be enforced on top of it. My local sphinx default theme, colours blue all links and makes the underline upon hover. I see nilearn is always underlined plus drastic colour change upon hover, scikit-learn is always underline and slight colour change upon hover. Would there be a preferred one? I also guess it needs to be address in the webpage CSS rather than in the gallery CSS, right?

Titan-C:
What do you think about this new changes? This time CSS changes are enforced in the code blocks within the gallery and clickable functions have a full grey background. Have a look at
http://sphinx-gallery-local.readthedocs.org/en/config/auto_examples/plot_gallery_version.html#example-plot-gallery-version-py
The way this was exemplified in the documentation is putting directly the resulting html code. It might be a nice thing for future work to add this feature to extend the sphinx directives codeblock and literalinclude, instead as it is done now. Scanning the output html gallery directory finding the patterns of functions and just then including the links.

lesteve:
> What do you think about this new changes?

For some reason in Firefox links are still not visible in the gallery although the one with np.sin [here](http://sphinx-gallery-local.readthedocs.org/en/config/advanced_configuration.html#linking-to-external-documentations) works fine :confused:

Titan-C:
> For some reason in Firefox links are still not visible in the gallery although the one with np.sin here works fine :confused:

that is strange as I use firefox and it works on both. Specially as the short np.sin example is a direct copy of the generated html. Maybe do a refresh? I'm thinking now that this background I put is specific to modify sphinxgallery readthedocs theme. But would not be enforced anywhere else. Although by default sphinx does hightlight link in codeblocks.

lesteve:
> Maybe do a refresh?

Yep that was just me being thick, sorry for the noise.

lesteve:
@Titan-C feel free to merge this one when you feel like it.

PR #24, opened 2015-03-30
Update into a unified sphinx-gallery namespace

Titan-C:
Including a cleaner css file with the sphx-glr prefix for all item used in sphinx-gallery.


agramfort:
LGTM

PR #26, opened 2015-04-11
scale image in stead of make thumbnail and independent backrefereces function

Titan-C:
Here I refactor in 3 things.
- First I include the change from the mayavi branch that turns make thumbnail into scale image
- Second I split the backreferece file generation into a separate function
- Last the fhindex is no more moved around the gen_dir_rst functions it is now a return value


sphinxgallery/gen_rst.py
```diff
@@ -244,7 +241,25 @@ def _thumbnail_div(subdir, full_dir, fname, snippet):
     return out
 
 
-def generate_dir_rst(directory, fhindex, examples_dir, gallery_dir, gallery_conf, plot_gallery, seen_backrefs):
+def write_backreferces(backrefs, seen_backrefs, gallery_conf,
```
agramfort: write_backreferces ? I don't understand.

sphinxgallery/backreferences.py
```diff
@@ -0,0 +1,167 @@
...
+    if ref_name.startswith('._'):
+        ref_name = ref_name[2:]
+
+    out = """
+.. raw:: html
+
+    <div class="sphx-glr-thumbContainer" tooltip="{}">
+
+.. figure:: /{}
+    :target: /{}.html
```
Titan-C: This is the target directive to put links to the figure. It would be great if it could take references and not links urls

Titan-C:
@lesteve this is the broken or almost working branch. If you build the documentation you will notice that in the gallery all the links on the images are broken, but not their caption. Nevertheless if you run a webserver on _build/html these links work. This is the problem of the target directive. See comment in code.

What does this have to do with backreferences? Backrefences are the exact same thumbnails of the gallery but put at the end of the modules documentation API(Feature only used by scikit-learn), so one has to link to the same example website but from different locations(the Gallery and the documented API) The scikit-learn solution was to make _thumbnail_div take 2 input dirs. One having the directory of the gallery and the other containing the relative path from the API documentation to the gallery. But that requires to hardcode those relative locations in the script. And that shall not be done for a generic module, as every user would locate their files to his convenience.

The problem of the back references is how to keep track of the location of the gallery example files and images and how to link to them from an arbitrary location. My solution was to use sphinx *absolute paths* to the sphinx source directory. So I would only need to know the example files location, sphinx does the rest. This can be used for the figure directive which searches within sphinx, but the target directive takes urls. So if a webserver is used on the _build/html or any output directory, then all absolute paths work.

sphinxgallery/backreferences.py
lesteve: Probably very naive: but it seems that removing the leading `/` in the target fixes the link without running a webserver. Is there a reason why you need it?

Same thing goes for the `/` in figure, it doesn't look like it is needed.

Titan-C: > Probably very naive: but it seems that removing the leading / in the target fixes the link without running a webserver. Is there a reason why you need it?

It doesn't solve it for me. Which sphinx version are you using? I have 1.2.3
I use the leading `/` because it allows me to link to the image from any location in the documentation. And if used for the target when running the web server it also links properly from anywhere.

lesteve: I guess I was right, it was a very naive attempt! I think it may have fixed the link in the backreferences but not in the gallery or the other way around. I'll try to take a closer look.

FWIW the [documentation](http://docutils.sourceforge.net/docs/ref/rst/directives.html#image) says the `target` can take reference but I was unable to get it to work, maybe because it is not supposed to work across documents.

In general, it would be a lot better if a webserver wasn't needed to look at the documentation.

GaelVaroquaux: > In general, it would be a lot better if a webserver wasn't needed to look at
> the documentation.

+1000

Titan-C: Yes, I also agree on that. Although once published online I doesn't do much of a difference. The thing I fear is that it will require to extended sphinx to accept the :doc: or :ref: directive inside the figure environment. at target can only take urls or same page references. And I'm still trying to figure out how does sphinx work at this level.

lesteve:
> The scikit-learn solution was to make _thumbnail_div take 2 input dirs. One having the directory of the gallery and the other containing the relative path from the API documentation to the gallery. But that requires to hardcode those relative locations in the script.

Why can't we keep doing the same thing with two directories arguments? In particular it would be great if you could point me to the hardcoding needed in the scikit-learn code?

GaelVaroquaux:
> Why can't we keep doing the same thing with two directories arguments?

+1

Titan-C:
> Why can't we keep doing the same thing with two directories arguments? In particular it would be great if you could point me to the hardcoding needed in the scikit-learn code?

So the _thumbnail_div processes the directories separately here: https://github.com/scikit-learn/scikit-learn/blob/master/doc/sphinxext/gen_rst.py#L584-590

but when one calls if for a gallery thumbnail one uses: https://github.com/scikit-learn/scikit-learn/blob/master/doc/sphinxext/gen_rst.py#L633

whereas for the backreferences. One codes where: 
* the examples are stored: https://github.com/scikit-learn/scikit-learn/blob/master/doc/sphinxext/gen_rst.py#L643
* then you give the relative location of one to the other: https://github.com/scikit-learn/scikit-learn/blob/master/doc/sphinxext/gen_rst.py#L653-654
* finally you only call the example file from a rst file in the same folder

GaelVaroquaux:
Aren't these architectural problems that should be fixed, rather than
introducing fragile workarounds?

On Mon, May 04, 2015 at 12:16:34PM -0700, Óscar Nájera wrote:
>     Why can't we keep doing the same thing with two directories arguments? In
>     particular it would be great if you could point me to the hardcoding needed
>     in the scikit-learn code?

> So the _thumbnail_div processes the directories separately here: https://
> github.com/scikit-learn/scikit-learn/blob/master/doc/sphinxext/gen_rst.py#
> L584-590

> but when one calls if for a gallery thumbnail one uses: https://github.com/
> scikit-learn/scikit-learn/blob/master/doc/sphinxext/gen_rst.py#L633

> whereas for the backreferences. One codes where:

>   • the examples are stored: https://github.com/scikit-learn/scikit-learn/blob/
>     master/doc/sphinxext/gen_rst.py#L643
>   • then you give the relative location of one to the other: https://github.com
>     /scikit-learn/scikit-learn/blob/master/doc/sphinxext/gen_rst.py#L653-654
>   • finally you only call the example file from a rst file in the same folder

> —
> Reply to this email directly or view it on GitHub.*


-- 
    Gael Varoquaux
    Researcher, INRIA Parietal
    NeuroSpin/CEA Saclay , Bat 145, 91191 Gif-sur-Yvette France
    Phone:  ++ 33-1-69-08-79-68
    http://gael-varoquaux.info            http://twitter.com/GaelVaroquaux

Titan-C:
> Aren't these architectural problems that should be fixed, rather than introducing fragile workarounds?

yes. That's why I wanted to use absolute paths. But I run short of the capabilities of sphinx when the figure instruction can only take links as urls or local references, and not cross references across documents like ref or absolute paths to rst files like doc.

Titan-C:
The backreferences are solved. This was a quite painful task, to get to work along with sphinx. In the end the solution looks is very simple and elegant.  I first extend the figure directive with a custom one to hold the gallery thumbnails. Put a placeholder in the link acting for the image and let sphinx resolve the links. Then when writing the html I copy the resolved link an put it onto the image. This way all paths in the rendered html are relative to each other. No need of the webserver to see all working.
This works well on this branch of sphinxgallery. On scikit-learn in PR https://github.com/scikit-learn/scikit-learn/issues/4658 that uses call the local installed version(git version of this branch) also works.

lesteve:
Looks like it is working indeed, good work ! My gut feeling is that this solution is slightly overkill but at the same time you know sphinx way better than I do by now so you probably tried all the simpler solutions first.

Something I realised recently is that the naming of the thumbnails is not great, maybe their name should contain 'thumb' because once they are copied in _build/html/_images they look almost the same as the non thumbnails png. For example:

```
$ ls -ltrh _build/html/_images/sphx_glr_plot_grow*
-rw-rw-r-- 1 le243287 le243287 20K May  6 08:46 _build/html/_images/sphx_glr_plot_grow_001.png
-rw-rw-r-- 1 le243287 le243287 21K May  6 08:46 _build/html/_images/sphx_glr_plot_grow_002.png
-rw-rw-r-- 1 le243287 le243287 12K May  6 08:46 _build/html/_images/sphx_glr_plot_grow.png
```

sphx_glr_plot_grow.png is the thumbnail and the other two are the standard png generated by running the example.

doc/reference.rst
```diff
@@ -8,3 +8,20 @@ Sphinx-Gallery Reference
 
 .. autodata:: sphinxgallery.__version__
     :annotation: Sphinx Gallery current version
+
+.. include:: modules/generated/sphinxgallery.__version__.examples
+.. raw:: html
+
+        <div style='clear:both'></div>
```
lesteve: Is this clearing needed and if yes should it be added to the .examples file directly?

Also maybe not in this PR but it'd be great to use sphinx templates as scikit-learn is doing in order to introduce the .examples automatically and have an example within sphinx-gallery showing this feature.

sphinxgallery/__init__.py
```diff
@@ -5,4 +5,4 @@
 
 def path_static():
     """Returns path to packaged static files"""
-    return os.path.abspath(os.path.dirname(__file__))+'/_static'
+    return os.path.join(os.path.abspath(os.path.dirname(__file__)), '_static')
```
lesteve: Super minor niggle, it feels more natural to have the abspath as the outer call (i.e. `abspath(join(...))` rather than `join(abspath(...))`) even if the result is the same ...

sphinxgallery/backreferences.py
```diff
@@ -0,0 +1,162 @@
...
+    :tooltip: {}
+
+    :ref:`example_{}`
+
+""".format(thumb, snippet, ref_name)
+
+    return out
+
+
+def write_backreferces(seen_backrefs, gallery_conf,
```
lesteve: typo: write_backreferces -> write_backreferences

sphinxgallery/gen_gallery.py
```diff
@@ -46,12 +48,14 @@ def generate_gallery_rst(app):
     if not plot_gallery:
         return
 
-    build_image_dir = os.path.join(app.builder.outdir, '_images')
-    clean_gallery_out(build_image_dir)
+    #clean_gallery_out(app.builder.outdir)
```
lesteve: was that commented out on purpose, i.e. don't we want to remove the 'sphx-glr*.png' from build/_images?

doc/Makefile
```diff
@@ -61,6 +61,9 @@ html:
 	@echo
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
 
+livehtml:
+	sphinx-autobuild -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
+
```
lesteve: we probably don't want this change in this PR

doc/reference.rst
Titan-C: It is required maybe 90% because one is floating the thumbnails and one has to signalize when that ends. 10% when it is not needed is because the theme of the website is simple enough it has no conflict with the floats. The basic/default sphinx theme, which you render locally is immune. The read the docs theme, and scikit-learn where many other divs are floated need this. But it is all your choice on the theme you use.
I'm not sure if it should be added to the .example file directly. If one uses the sphinx templates for documenting modules and classes(as in scikit-learn) I guess it is better to add it manually there and once. All because it is more a theme and style instruction. If you change the CSS of sphinx-gallery and its floats, then this becomes irrelevant or more the user choice.

sphinxgallery/backreferences.py
```diff
@@ -0,0 +1,162 @@
...
+    thumb = os.path.join(full_dir, 'images', 'thumb',
+                         'sphx_glr_%s.png' % fname[:-3])
+    ref_name = os.path.join(full_dir, fname).replace(os.path.sep, '_')
+    if ref_name.startswith('._'):
+        ref_name = ref_name[2:]
+
+    out = """
+.. glr_thumb:: /{}
+    :tooltip: {}
+
```
lesteve: Is this blank line necessary? Quickly glancing at it it looks almost like the old way of doing it where the ref wasn't part of the figure directive.

sphinxgallery/gen_gallery.py
Titan-C: This is a new thing. It was commented on purpose but not related to this PR. I was testing If this instruction is at all needed. I was noticing, at least for sphinx gallery documentation, that if you don't clear any image in sphinx 1.2.3. On a new build sphinx wont copy the images again. So maybe this has might not be needed anymore. But It shall be addressed in another PR. For the moment I'll keep deleting images.

sphinxgallery/backreferences.py
Titan-C: Yes the blank line is necessary. The ref is part of the figure caption and not a part of the figure attributes, because is a link with the example file title. I want to keep this behaviour. glr_thumb is an extended version of figure directive, it has the enclosing div with the css attribute selector tooltip, and the :target: option is not used anymore because it is automatically attached from the resolved ref.

doc/reference.rst
lesteve: Out of interest, what is the worse that can happen if you always use the clearing trick, even when not needed ?

.example files are automatically generated so I don't really see a problem to add the clearing there. At least it ensures that just doing an include will just work and you don't have to remember to add the raw html additional content.

Titan-C: > Out of interest, what is the worse that can happen if you always use the clearing trick, even when not needed ?

I have no idea. It is CSS.

Titan-C: > .example files are automatically generated so I don't really see a problem to add the clearing there

I'll put the clearing

sphinxgallery/backreferences.py
lesteve: I see thanks for the details.

GaelVaroquaux:
> In the end the solution looks is very simple and elegant. I first
> extend the figure directive with a custom one to hold the gallery
> thumbnails.

I really don't like the extension of a directive. The whole design of the
original gen_rst.py is to avoid this. Sphinx directives have prooved to
be fragile and to break.

Please absolutely avoid creating sphinx directives and absolute imports.

Titan-C:
@GaelVaroquaux would you take over this PR. I really don't see a more simpler solution that gets the job done. Besides relying on absolute paths and running always on a webserver. Indeed having an extra directive, although simple, does feel like an overshoot, but I could not get a better solution.

GaelVaroquaux:
> @GaelVaroquaux would you take over this PR. I really don't see a more simpler
> solution that gets the job done. Besides relying on absolute paths and running
> always on a webserver. Indeed having an extra directive, although simple, does
> feel like an overshoot, but I could not get a better solution.

You have no idea how overbooked I am. There is no chance that I can take
over this PR.

I don't know how we got to this situation, but the first versions of
gen_rst.py did not need such hacks. I suspect that it's a new feature
that is calling for those hacks. Well, maybe it's time to cut our loses
and decide that we don't want this feature. All I am saying is that those
hacks should not get in the project because they will make the codebase
too fragile.

Titan-C:
This can not get any simpler, I should have gotten this idea in the first place. Instead of putting a link to the image in the thumbnail. The anchor in the caption is extended in CSS to cover the complete thumbnail area. So the references get properly resolved by sphinx, and the link is not only over the text but on the whole thumbnail. No sphinx extensions, only CSS tricks.

lesteve:
This seems to be working fine indeed, good stuff!

Out of interest, can you explain how it works? My question is mostly how come you can now click anywhere on the thumbnail to go to the example page? In other words why the target of the <a ..> element is available on one of its parent element?

GaelVaroquaux:
could you try to use only relative imports when imprint things from sphinxgallery to make it easier to embed the code

GaelVaroquaux:
I had a quick look at this PR. Good job fixing the back reference difficulty! I like your solution very much!

Titan-C:
> could you try to use only relative imports when imprint things from sphinxgallery to make it easier to embed the code

are not relative imports being dropped in python3?

Titan-C:
> Out of interest, can you explain how it works? My question is mostly how come you can now click anywhere on the thumbnail to go to the example page? In other words why the target of the element is available on one of its parent element?

Well, that it why it did not occur to me in the beginning. In CSS you have to think of element arranged in layers, So it is not that the target of the element if available for its parent. But that the element is on a higher layer than the parent and it extends itself over the whole parent covering its entire surface.

GaelVaroquaux:
>     could you try to use only relative imports when imprint things from
>     sphinxgallery to make it easier to embed the code

> are not relative imports being dropped in python3?

No, it's rather the oppositive: imports are no longer relative by
default. In other words "import os" will never be equivalent to "from .
import os".

lesteve:
I think this PR can go in. The CSS thing seems a little bit fragile but way more transparent than the sphinx directive and the backreference functionality is a great thing to have.

Titan-C:
Sorry for the timeout, I wasn't around. I include some minor fixes based on the comments.

PR #27, opened 2015-04-18
Pep8

Titan-C:
This includes the remarks for pep8 in https://github.com/nilearn/nilearn/pull/399
also the rename of make_thumbnail to scale image

### DIFF
--- a/sphinxgallery/__init__.py
+++ b/sphinxgallery/__init__.py
@@ -1,7 +1,7 @@
 """Sphinx Gallery
 """
 import os
-__version__ = '0.0.7'
+__version__ = '0.0.9'
 
 def path_static():
     """Returns path to packaged static files"""
--- a/sphinxgallery/__init__.py
+++ b/sphinxgallery/__init__.py
@@ -5,4 +5,4 @@ __version__ = '0.0.9'
 
 def path_static():
     """Returns path to packaged static files"""
-    return os.path.abspath(os.path.dirname(__file__))+'/_static'
+    return os.path.join(os.path.abspath(os.path.dirname(__file__)), '_static')
--- a/sphinxgallery/gen_rst.py
+++ b/sphinxgallery/gen_rst.py
@@ -308,9 +308,10 @@ def generate_dir_rst(directory, fhindex, examples_dir, gallery_dir, gallery_conf
     """)  # clear at the end of the section
 
 
-def make_thumbnail(in_fname, out_fname, width, height):
-    """Make a thumbnail with the same aspect ratio centered in an
-       image with a given width and height
+def scale_image(in_fname, out_fname, max_width, max_height):
+    """Scales an image with the same aspect ratio centered in an
+       image with a given max_width and max_height
+       if in_fname == out_fname the image can only be scaled down
     """
     # local import to avoid testing dependency on PIL:
     try:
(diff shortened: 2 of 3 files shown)


sphinxgallery/__init__.py
```diff
@@ -1,8 +1,8 @@
 """Sphinx Gallery
 """
 import os
-__version__ = '0.0.7'
+__version__ = '0.0.9'
```
lesteve: I am not sure about the origin of this diff but I can see that sphinx-gallery 0.0.8 has already been released on pypi. 

I am going to hazard a wild guess and assume that you released from a local modification which should be avoided. Tagging the release point with git is also a good idea for the future releases. I can see that you did that for a few releases in the past.

sphinxgallery/gen_rst.py
```diff
@@ -308,9 +310,10 @@ def generate_dir_rst(directory, fhindex, examples_dir, gallery_dir, gallery_conf
     """)  # clear at the end of the section
 
 
-def make_thumbnail(in_fname, out_fname, width, height):
-    """Make a thumbnail with the same aspect ratio centered in an
-       image with a given width and height
+def scale_image(in_fname, out_fname, max_width, max_height):
+    """Scales an image with the same aspect ratio centered in an
+       image with a given max_width and max_height
+       if in_fname == out_fname the image can only be scaled down
```
lesteve: Minor style comment: the short description of a function is supposed to fit on one line, see PEP 257 and/or [numpydoc](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) convention.

sphinxgallery/gen_rst.py
```diff
@@ -319,14 +322,17 @@ def make_thumbnail(in_fname, out_fname, width, height):
...
+    scale_h = max_height / float(height_in)
 
-    if height_in * scale_w <= height:
+    if height_in * scale_w <= max_height:
         scale = scale_w
     else:
         scale = scale_h
 
+    if scale >= 1.0 and in_fname == out_fname:
+        return
```
lesteve: Just curious, what's the purpose of this change ?

Titan-C: The change of scale_image is more an import of the mayavi branch pull request. I do find it more convenient to have it like this than the old make_thumbnail function and it is something I started working early on for this new release start.
Why this last line? It is actually the work of @agramfort

sphinxgallery/__init__.py
Titan-C: Yes version 0.0.8 was an emergency release. Because there was a conflict in the CSS hierarchy when merging release 0.0.7 to nilearn(There is a big difference in css if you have div.class or .class I only matters if earlier in you CSS you have defined some directives like div.class, which is the case for some sphinx themes), but I could not make a release with the same version number 0.0.7 in pypi. Thus the release to pypi was a mere local change. Then I missed to update sphinx-gallery with the 0.0.8 version number and make the release.

sphinxgallery/gen_rst.py
lesteve: Just a remark for the future: it's actually a good idea to keep unrelated work in separate branches. I can see that you have the same commit in different PR branches. A good rule of thumb when you start working on a new feature is to do it in a new branch you created off master.

Titan-C:
This PR became old and irrelevant

PR #30, opened 2015-04-28
Shelve now stored in gallery_dir

### LINKED ISSUE #29: Fails on non-standard doc build location
The get_data() function in docs_resolv.py tries to open a cache file at ``_build/searchindex``. If a project uses a non-standard output location the ``_build`` directory may not present. As a result the build fails with:

````
DBNoSuchFileError: (2, 'No such file or directory')
````

caused by ``shelve.open(...)``

At present, there is no way to tweak this default location from the outside. As a workaround it is sufficient to create a dummy ``_build`` directory.

Titan-C:
This is established to fix #29 saving the shelve file of documentation api indexes is the gallery_dir path set by the user in his configuration file. The shelve does not get copied by sphinx into the final output html directory.

### DIFF
--- a/sphinxgallery/docs_resolv.py
+++ b/sphinxgallery/docs_resolv.py
@@ -51,13 +51,14 @@ def _get_data(url):
     return data
 
 
-def get_data(url, cached_file='_build/searchindex'):
+def get_data(url, gallery_dir):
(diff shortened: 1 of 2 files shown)


GaelVaroquaux:
Looks good to me. Thanks!

agramfort:
LGTM

Issue #8, opened 2015-01-04
competition plot directive matplotlib/sphinx and the sphinx gallery


Titan-C:
Because sphinx never deletes empties images in this directory. Sphinx in order to avoid overwriting images in this directory saves new images with a different name. As such html/_images grows larger every time, gen_rst has a hack to empty images in html/_images on every execution.
The problem arises if within the documentation one uses the plot directive from matplotlib `.. plot:: file.py`
in this case the code is executed once again and a new image is generated with a new filename. But once gen_rst comes to execution it deletes the gallery images(which are regenerated) and the images from the documentation directive. As such it happens that the end documentation does not display images from the plot directive.
Of course a simple `make clean` walks around the problem.

agramfort:
this issue has bugged me many times especially when make doc takes a
while... if you have a fix you will deserve :beers:

GaelVaroquaux:
How about we store a list of images generated by sphinx-gallery and delete only these. I suspect it's slightly harder than what I just said, but that's the spirit :)

lesteve:
This should now be fixed since #31 has been merged.

Issue #5, opened 2014-12-11
How to add custom image to example


hugadams:
Hi,

I love that examples built with "plot_" prefix will autogenerate a plot.  I'm wondering what's the recommended approach to add a custom image to an example that doesn't render a plot.  For example, we have a code that returns HTML tables, or a code that launches an application, and I want to put a screenshot of this at the top of the example in the same place where the plot appears in the examples named "plot_".  Do you have any suggestions?

lesteve:
One work-around with the caveat that the gallery won't show any thumbnail for your example: you could just use a image directive in the docstring of your module.

I modified just_code.py like this:

```diff
diff --git a/examples/just_code.py b/examples/just_code.py
index e76c92f..fe5944b 100644
--- a/examples/just_code.py
+++ b/examples/just_code.py
@@ -7,6 +7,9 @@ A short python script
 A script that is not executed when gallery is generated but nevertheless
 gets included as an example.
 Doing a list
+
+.. image:: ../../plot_classifier_comparison_1.png
+
 """
```

and I have plot_classifier_comparison_1.png in `../..` compared to `doc/auto_examples`. It seems to do the trick:

![just_code_with_plot](https://cloud.githubusercontent.com/assets/1680079/5946799/c15463ca-a738-11e4-929b-48a519797878.png)

trbedwards:
Is there anyway to get the image directive hack to replace the thumbnail?

lesteve:
Yep: http://sphinx-gallery.readthedocs.io/en/latest/advanced_configuration.html#using-a-custom-default-thumbnail-image

lesteve:
This is not a hack but a feature by the way.

trbedwards:
Yes, but this means setting a single default thumbnail for all examples. What I'd like is a different thumbnail for each example, but without having to execute the example (since some examples rely on user input)

lesteve:
Not at the moment then. Just curious, what would the purpose of the image and how would it be different from example to example?

trbedwards:
There are some example scripts that are either expensive to run or require user input, so it's difficult to automatically produce the plots for those. What I'd like to do is manually run the example myself, get a screenshot or something, then use that for the thumbnail.

lesteve:
How long does a single example take ?

We haven't put too much thought about examples where you need user input (I am guessing you mean GUI-like interaction).

[...]

PR #31, opened 2015-04-29
Image clean

Titan-C:
This change was first motivated by #29 to remove the explicit use of the _build directory path in any instruction in sphinx-gallery(PR #30 also related).

The hack in sphinx-gallery that cleans the images is now more specific as it now scans for png images with the sphx_glr prefix that should be only created by us. Is solves the competition with the maplotlib plot directive #8 at least in my known case.

### DIFF
--- a/sphinxgallery/gen_rst.py
+++ b/sphinxgallery/gen_rst.py
@@ -214,7 +214,8 @@ def line_count_sort(file_list, target_dir):
 
 def _thumbnail_div(subdir, full_dir, fname, snippet):
     """Generates RST to place a thumbnail in a gallery"""
-    thumb = os.path.join(full_dir, 'images', 'thumb', fname[:-3] + '.png')
+    thumb = os.path.join(full_dir, 'images', 'thumb',
+                         'sphx_glr_%s.png' % fname[:-3])
     link_name = os.path.join(full_dir, fname).replace(os.path.sep, '_')
     ref_name = os.path.join(subdir, fname).replace(os.path.sep, '_')
     if ref_name.startswith('._'):
--- a/sphinxgallery/gen_rst.py
+++ b/sphinxgallery/gen_rst.py
@@ -445,7 +445,7 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf, plot_gallery):
     Returns the set of functions/classes imported in the example.
     """
     base_image_name = os.path.splitext(fname)[0]
-    image_fname = '%s_%%03d.png' % base_image_name
+    image_fname = 'sphx_glr_%s_%%03d.png' % base_image_name
 
     this_template = rst_template
     last_dir = os.path.split(src_dir)[-1]
@@ -473,7 +473,7 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf, plot_gallery):
                                'stdout_%s.txt' % base_image_name)
     time_path = os.path.join(image_dir,
                              'time_%s.txt' % base_image_name)
-    thumb_file = os.path.join(thumb_dir, base_image_name + '.png')
+    thumb_file = os.path.join(thumb_dir, 'sphx_glr_%s.png' % base_image_name)
     time_elapsed = 0
     if plot_gallery and fname.startswith('plot'):
         # generate the plot as png image if file name
--- a/sphinxgallery/gen_gallery.py
+++ b/sphinxgallery/gen_gallery.py
@@ -46,7 +46,6 @@ def generate_gallery_rst(app):
     if not plot_gallery:
         return
 
-    dir(app.builder)
     build_image_dir = os.path.join(app.builder.outdir, '_images')
     clean_gallery_out(build_image_dir)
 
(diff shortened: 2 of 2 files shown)


GaelVaroquaux:
Looks good to me!

sphinxgallery/gen_rst.py
```diff
@@ -445,7 +445,7 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf, plot_gallery):
     Returns the set of functions/classes imported in the example.
     """
     base_image_name = os.path.splitext(fname)[0]
-    image_fname = '%s_%%03d.png' % base_image_name
+    image_fname = 'sphx_glr_%s_%%03d.png' % base_image_name
```
lesteve: Why the sphx_glr_ prefix ? I don't think it adds anything since the image name already contains the name of the example script.

Also this is going to break the nilearn doc because we manually includes generated images (do `git grep -P 'plot.*.png' doc` to get all of them). I am guessing this is the same for scikit-learn.

Titan-C: The sphx-glr prefix is because scripts that don't plot don't start with plot and it is also given to the thumbnails.

Indeed the extra prefix would break the manual calls one does on the documentation of nilearn.

Another thought is if this whole hack of cleaning the images has to be done in the gallery scripts. I see nilearn and scikit-learn(https://github.com/scikit-learn/scikit-learn/blob/master/doc/Makefile#L40) clean the _build/html/_images with an instruction from the Makefile every time one uses `make html`, which seems more robust as it is done before sphinx starts doing anything. I wonder if it then would be best to just remove this hack from the gallery script and write in the Sphinx-gallery documentation to users to insert the extra instruction in their makefile.

lesteve: The right thing to do would be to implement what Gaël was saying in #5.

> How about we store a list of images generated by sphinx-gallery and delete only these. I suspect it's slightly harder than what I just said, but that's the spirit :)

If that's too hard, I guess we can live with this PR as a stop-gap solution/hack.

Stupid questions: 
* where is the code that copies the images from doc/auto_examples/images to _build/html/_images ? Is that inside sphinx or inside sphinx-gallery?
* where does the matplotlib plot directive stores its output images, is that _build/html/_images ?
* what are the proper variable names in sphinx-gallery for both of these folders?

Titan-C: > If that's too hard, I guess we can live with this PR as a stop-gap solution/hack.

what is stop-gap solution?

> Stupid questions:
> 
>     where is the code that copies the images from doc/auto_examples/images to _build/html/_images ? Is that inside sphinx or inside sphinx-gallery?

This is a sphinx feature. All images are copied to `_build/html/_images` no folder structure is kept. This is the reason why sphinx renames images with repeated names preventing overwrites and the reason new images keep increasing in the build folder as it never gets cleared. Until the next `make clean`

>     where does the matplotlib plot directive stores its output images, is that _build/html/_images ?

I don't know if it has a temporary cached dir, but in the end the matplotlib plot directive gets overtaken by sphinx and it sends the final images to `_build/html/_images` with the same script name. There is when the conflict starts if one uses scripts from the gallery, so one image gets renamed and sphinx keeps track of the naming. But on subsequent `make html` images are cleared at some point by the hack in sphinx-gallery the images of the gallery are copied again and linked properly by sphinx, but ones of the plot directive don't get copied again or get deleted after being set, so the output html has lost images but sphinx does not complain because it is sure to have copied the image to `_build/html/_images`. It is a synchronization and name conflict problem.

>     what are the proper variable names in sphinx-gallery for both of these folders?

Sphinx-gallery creates recursively two folders inside the `gallery_dir` one for images `image_dir` and one inside it for the thumbnails `thumb_dir`, and this structure is keep for every sub-folder of the `examples_dir`. Sphinx-gallery links to this images locally from the rst files, and when generating the html it copies individually the image to `_build/html/_images`(renaming it if necessary) and establishes the correct html link. The directory `_build/html` is a sphinx variable found in `app.builder.outdir` and the `_images` sub-folder is pure managed by sphinx

lesteve: > what is stop-gap solution?

temporary/short-term solution.

Titan-C: > Another thought is if this whole hack of cleaning the images has to be done in the gallery scripts. I see nilearn and scikit-learn(https://github.com/scikit-learn/scikit-learn/blob/master/doc/Makefile#L40) clean the _build/html/_images with an instruction from the Makefile every time one uses make html, which seems more robust as it is done before sphinx starts doing anything. I wonder if it then would be best to just remove this hack from the gallery script and write in the Sphinx-gallery documentation to users to insert the extra instruction in their makefile.

It is no good option as it over deletes images that sphinx wont rebuild. If you are using pngmath instead of mathjax or if the matplotlib plot directive notices no change in the scripts it will assume the images still resides in `_build/html/_images` and it wont be copied again.

One has to keep track the images generated by sphinx-gallery but not where we save them but in the target directory of sphinx after it is renamed. A longer term fix would be to use the same procedure sphinx has to track images and only copy the ones that changed and not the whole gallery on every build.

lesteve: Let's go for the sphx_glr prefix for now then and try to come up with a better long-term solution. 

It's easy enough to do a search and replace in the nilearn rst docs.

PR #35, opened 2015-06-07
Path class

Titan-C:
This is more of a nice feature to have(mostly inspired by the work done in plot2rst). But that I could test and allowed me to discover the bug fixed in #34 . It extends the string class for paths, given some ease of use when working with paths.

### DIFF
--- a/sphinxgallery/__init__.py
+++ b/sphinxgallery/__init__.py
@@ -12,9 +12,6 @@ def glr_path_static():
 class Path(str):
     """Path object for manipulating directory and file paths."""
 
-    def __init__(self, path):
-        super(Path, self).__init__(path)
-
     @property
     def isdir(self):
         return os.path.isdir(self)
--- a/sphinxgallery/__init__.py
+++ b/sphinxgallery/__init__.py
@@ -32,6 +32,7 @@ class Path(str):
     def makedirs(self):
         if not self.exists:
             os.makedirs(self)
+        return self
 
     def listdir(self):
         return os.listdir(self)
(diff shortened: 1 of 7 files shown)


lesteve:
@Titan-C, is that still something we want?

GaelVaroquaux:
> @Titan-C, is that still something we want?

What is it actually? I didn't get it from the description of the PR.

lesteve:
Not entirely sure, but it looks like adding a Path class for more convenient path manipulations ...

Titan-C:
> What is it actually? I didn't get it from the description of the PR.

Can't really explain it better.
- It's an internal feature, not for the user. It was inspired by the fact that the gallery has to deal with multiple paths(examples_source_dir, Gallely_dir, modules_dir, file_path, image_paths, ...)
- The idea is to store this paths not only as strings, but in an object that is aware they are paths and provide inherent features of this. Like, joining paths, creating dir, listing dirs, checking it the path exists or is a directory.

It remains a nice feature to have which I was very eager to use, back when #36 was only aimed at refactoring the code. Now it needs an update from master.

GaelVaroquaux:
I always frown at adding abstractions, because abstractions make life
easier for their creators, and harder for people who have to discover
them.

If you really want to go down that path (sic), you can use the
semi-standard 'path.py' existing Python package:
https://pypi.python.org/pypi/path.py

For this, you'll need a to embed a version, and thus add a note to the
license.

Titan-C:
No more need for this anymore.

PR #36, opened 2015-06-07
[Review] Code refactoring, block execution of scripts

Titan-C:
I'm still unsure which the final goal of this clean up is. By now it breaks up some statements into simpler functions.
One big advantage is that now the re-execution of a example script is evaluated a lot earlier and one saves a lot of if..else, so less branching and less deeper indentation.
- [x]  docstrings for this functions
- [x] try to test on them


sphinxgallery/gen_rst.py
```diff
@@ -52,6 +51,7 @@ def execfile(filename, global_vars=None, local_vars=None):
     # matplotlib
     import matplotlib
     matplotlib.use('Agg')
+    import matplotlib.pyplot as plt
```
lesteve: Just curious, what's the goal of this added line?

sphinxgallery/gen_rst.py
```diff
@@ -310,81 +339,57 @@ def execute_script(image_dir, thumb_file, image_fname, base_image_name,
     if os.path.exists(time_path):
         time_elapsed = float(open(time_path).read())
 
-    if not os.path.exists(first_image_file) or \
-       os.stat(first_image_file).st_mtime <= os.stat(src_file).st_mtime:
-        # We need to execute the code
-        print('plotting %s' % fname)
-        t0 = time()
-        import matplotlib.pyplot as plt
```
Titan-C: @lesteve 
I'm actually moving it out of here. I asked in scikit-learn if there was a good reason to keep this import so nested in the function and not moving it to the top of the file, no answer yet.

lesteve: Makes sense, I misread the diff.

jnothman: Its purpose would be to not require matplotlib to compile documentation
that used the gallery without plotting...

Titan-C:
This one is a big change. A lot of reordering of the code went into this, now I find it more readable and modular. There are some notable changes.

* Example script is executed by code blocks. It is now possible to give the example file a notebook structure with rst text followed by example code.
    * Example script of this included
    * Rendered layout changes organization from
1)  Module docstring 2) Output figures 3) execution stdout 4) download link 5) script code 
to
a) Docstring b) code block c) code output figures then stdout d) repeat from (a) until end of file e) download link

* There is no gallery header imposed by sphinx-gallery. This mean the user has to define the header of his gallery is his own `README.txt` in his examples directory.
* Sorting of the examples is by name and not by line count.
* A full build of the docs is done in travis for testing(until a better solution is found)

Titan-C:
Maybe not part of this PR. But have we agreed on the naming convention for the references?
* keep them as  `_example_full_path_to_file.py`?
* more similar to image naming `sphx_glr_full_path_to_file.py`?
* just `full_path_to_file.py`?
* rename the gallery_dir to something different than auto_examples?

sphinxgallery/gen_rst.py
```diff
@@ -292,157 +271,169 @@ def scale_image(in_fname, out_fname, max_width, max_height):
...
+    return fhindex
+
+
+def execute_script(code_block, example_globals, image_path, fig_count, src_file):
+    """Executes the code block of the example file"""
+    time_elapsed = 0
+    stdout = ''
+
+    # We need to execute the code
+    print('plotting %s' % src_file)
```
jnothman: This is now imprecise

sphinxgallery/gen_rst.py
```diff
@@ -292,157 +271,169 @@ def scale_image(in_fname, out_fname, max_width, max_height):
...
+        # horizontal list or a single rst call to 'image'.
+        if len(figure_list) == 1:
+            figure_name = figure_list[0]
+            image_list = SINGLE_IMAGE % figure_name.lstrip('/')
+        else:
+            image_list = HLIST_HEADER
+            for figure_name in figure_list:
+                image_list += HLIST_IMAGE_TEMPLATE % figure_name.lstrip('/')
+
+    except:
```
jnothman: Always use `except Exception`, or you capture `KeyboardInterrupt` making it hard for the user to break out of the whole process.

lesteve:
This sounds like a great improvement but I won't have time to look at it more closely before a week or so.

It'd be great to agree on the reference naming indeed. From what I remember I think this is the only thing that is blocking the https://github.com/nilearn/nilearn/pull/592.

Titan-C:
- [ ] Agree on the reference and image naming

sphinxgallery/gen_rst.py
GaelVaroquaux: > Its purpose would be to not require matplotlib to compile documentation that
> used the gallery without plotting...

I think that that's a good goal.

Titan-C: But I think it misses the point of not requiring matplotlib if we call it anyway to set the backend Agg before everything.

Titan-C:
@lesteve @GaelVaroquaux @agramfort 
This is ready, please have a look, it has become a huge PR and many things started going in at the same time.
The change log would be:
* Block processing and rendering of the example files.
* Notebook style capabilities respecting python code style. New text blocks have to be comment blocks. corresponding example is given)
* Update travis syntax for their new block container architecture. Build the doc on travis
* Multifolder Sphinx-Gallery. To have one for examples and one for tutorial(those benefit the most of notebook style)

Titan-C:
there is the readthedocs live version of this branch at

http://sphinx-gallery-local.readthedocs.org/en/refactor_split/

examples/plot_strings.py
```diff
@@ -14,3 +14,9 @@
 # License: BSD 3 clause
 
 print('This is a long test Output\n' * 50)
+
+####################################
+#One line out
```
agramfort: space missing after #

sphinxgallery/gen_rst.py
```diff
@@ -118,133 +78,155 @@ def flush(self):
...
+
+def save_figures(image_path, fig_count):
+    """Save all open matlplotlib figures of the example code-block
+
+    Parameters
+    ----------
+    image_path : str
+        Path where plots are saved (format string which accepts figure number)
+    fig_count : int
+        Previous figure number count. Figure number add from this number
```
agramfort: Returns is not documented

agramfort:
one last thing I would make the sphx_glr_ prefix configurable but that can be done later.

FYI it works like a charm with mne !

Looking forward to get this merged. Tomorrow morning? :)

Titan-C:
I just implemented my proposal for the reference naming.
Now that we support multiple galleries I believe it is better to rename the references to example scripts as:

`_sphx_glr_full_underscored_path.py`
then when referencing to rst examples would use ``:ref:`sphx_glx_auto_examples_plot_code.py` `` and its corresponding image would be `sphx_glx_auto_examples_plot_code_001.png` and its thumbnail  `sphx_glx_auto_examples_plot_code_thumb.png`. Then all this strings are similar and consistent to each other.

agramfort:
Fine with me

README.rst
```diff
@@ -41,8 +41,8 @@ To install::
     $ python setup.py develop
 
 
-Setting up your project
-=======================
+Setting up your project by the defaults
```
lesteve: Not sure what you mean by "by the defaults"

doc/advanced_configuration.rst
```diff
@@ -19,6 +19,24 @@ Within your Sphinx ``conf.py`` file you need to add a configuration dictionary:
...
+the sphinx configuration dictionary:
+
+.. code-block:: python
+
+    sphinxgallery_conf = {
+        'examples_dir'   : ['../examples', '../tutorials'],
+        'gallery_dir'    : ['auto_examples', 'tutorials'],
+    }
+
+Keep in mind that both list have to be of the same length.
```
lesteve: list -> lists

doc/syntax.rst
```diff
@@ -0,0 +1,38 @@
...
+Sphinx-Gallery syntax
+=====================
+
+For your examples to work you have to make them valid python code.
+
+
+Simple examples
+===============
+
+This are the base of a python script of module. The structure follows
```
lesteve: This are -> This is

doc/syntax.rst
```diff
@@ -0,0 +1,38 @@
...
+a leading docstring that must reStructuredText syntax and a header.
```
lesteve: missing verb between must and reStructuredText. Maybe "must follow reStructuredText".

doc/syntax.rst
```diff
@@ -0,0 +1,38 @@
...
+Then the executable code of your script.
+
+For a quick referece have a look at the example
```
lesteve: typo: reference

README.rst
Titan-C: Meaning you name your folder examples and the gallery auto examples. And that you don't have to set the sphinx-gallery configuration dictionary. So the defaults

doc/syntax.rst
```diff
@@ -0,0 +1,38 @@
...
+want to embed more documenting strings within your code and separating
+it by blocks. This is also possible, you follow standard python coding
+syntax but new documenting strings that will be rendered as reStructuredText
+have to be leaded by a line of hashes.
+
+For a clear example refer to the rendered example
+:ref:`sphx_glr_tutorials_plot_notebook.py` and compare it to the generating
+:download:`original python script <tutorials/plot_notebook.py>`
+
+In this multiblock structure. The website is rendered ordered by the
```
lesteve: "structure. The" -> "structure, the"

examples/plot_gallery_version.py
```diff
@@ -33,17 +35,26 @@ def bump(a):
...
-    fig = plt.figure()
+fig = plt.figure()
+d = layers(3, 100)
+try:
+    with plt.xkcd():
+        plt.stackplot(range(100), d.T, baseline='wiggle')
+        plt.annotate(
+            'Introducing Sphinx-Gallery ' + sphinxgallery.__version__,
+            xy=(12, 4), arrowprops=dict(arrowstyle='->'), xytext=(15, -4))
+except Exception:
```
lesteve: Just curious, it's always a good idea to keep examples as simple as possible. What's the reason of this "try ... catch" ? Supporting different matplotlib versions?

Titan-C: Yes this uses the matplotlib cartoon output which is no available in matplotlib 1.1 which is ubuntu 12.04 default. So travis was failing this build

README.rst
lesteve: Still not 100 % clear to me, sorry. Maybe just stick to "Setting up your project" ...

examples/plot_gallery_version.py
lesteve: I would be in favor of not using the "cartoon" output (by that I am guessing you mean xkcd) then, just to keep the example simple.

sphinxgallery/gen_rst.py
```diff
@@ -118,133 +78,161 @@ def flush(self):
...
+    f = open(source_file)
+
+    blocks = []
+    block = ''
+
+    has_header_string = 0
+    continue_text = False
+    for line in f:
+        # python docstring
+        if line.startswith('"""') and not has_header_string:
```
lesteve: Does that mean that we are only supporting `"""` for the example header ? This seems slightly fragile, there is no reason why people could not use `'''` or even `"` or `'`' for one-line headers. 

It seems to me that the old AST based solution was more robust even if more complicated.

sphinxgallery/gen_rst.py
```diff
@@ -118,133 +78,161 @@ def flush(self):
...
-                                     src_dir)
-    for fname in sorted_listdir:
-        if fname.endswith('py'):
-            generate_file_rst(fname, target_dir, src_dir, plot_gallery)
-            new_fname = os.path.join(src_dir, fname)
-            _, snippet, _ = extract_docstring(new_fname, True)
-            write_backreferences(seen_backrefs, gallery_conf,
-                               target_dir, fname, snippet)
+        # comment blocks
+        if line.startswith('#') and not continue_text:
```
lesteve: Just curious, what is the reason we switch to comments to separate blocks ? Can we still use strings or not ?

My hope was that using strings with the AST-based code would be more robust that this manual parsing. Also it seems that some projects use strings as block separator.

sphinxgallery/gen_rst.py
```diff
@@ -118,133 +78,161 @@ def flush(self):
...
+    """Test existence of image file and later touch time to source script"""
+
+    first_image_file = image_file.format(1)
+    needs_replot = (not os.path.exists(first_image_file) or
+              os.stat(first_image_file).st_mtime <= os.stat(src_file).st_mtime)
+    return not needs_replot
+
+
+def save_figures(image_path, fig_count):
+    """Save all open matlplotlib figures of the example code-block
```
lesteve: Typo in matplotlib

sphinxgallery/gen_rst.py
Titan-C: There are a few things to this. Indeed only a single docstring header will be supported and that uses the triple quotes.
Now this function is indeed quite fragile as is was designed yesterday and for our specific syntax case. Although the notebook example does provide some tricky syntax from time to time.
There was in between a new block splitter based on the tokenize, but that splitted based on strings

sphinxgallery/gen_rst.py
```diff
@@ -118,133 +78,161 @@ def flush(self):
...
+    fig_count : int
+        Previous figure number count. Figure number add from this number
+
+    Returns
+    -------
+    list of strings containing the full path to each figure
+    """
+    figure_list = []
+    # In order to save every figure we have two solutions :
+    # * iterate from 1 to infinity and call plt.fignum_exists(n)
```
lesteve: Is this part of the comment still useful? It doesn't seem like this plt.fignum_exists is used below so I would remove it.

sphinxgallery/gen_rst.py
```diff
@@ -292,157 +280,175 @@ def scale_image(in_fname, out_fname, max_width, max_height):
...
+    # We need to execute the code
+    print('plotting code blocks in %s' % src_file)
+
+    plt.close('all')
+    cwd = os.getcwd()
+    # Redirect output to stdout and
+    orig_stdout = sys.stdout
+
+    try:
+        # First CD in the original example dir, so that any file
```
lesteve: CD -> cd

sphinxgallery/gen_rst.py
Titan-C: This was more a @GaelVaroquaux statement that we don't want people writting unclean code, so free strings should not be lying around the code for the sake of being splitted into codeblocks for the gallery. So we would enforce the comment blocks and won't support free strings.

lesteve: OK fair enough.

FWIW unclean is a bit harsh, strings for multi-line comments was mentioned by Guido Van Rossum [here](https://twitter.com/gvanrossum/status/112670605505077248).

sphinxgallery/tests/test_get_rst.py
```diff
@@ -0,0 +1,28 @@
...
+Testing the rst files generator
+"""
+from __future__ import division, absolute_import, print_function
+import sphinxgallery.gen_rst as sg
+from nose.tools import assert_equals
+
+
+def test_split_code_and_text_blocks():
+    """Test if a known example file gets properly split"""
+
```
lesteve: It would be quite nice to include a more complicated example. You can create the file yourself with `with tempfile.NamedTemporaryFile():`

What I have in mind is more than two blocks testing a few edge cases like an inline string

```
"""This should not be part of a 'text' block
```
a function docstring
```python
def dummy():
    """This should not be part of a 'text' block'
```

a comment that is separated from a splitter by a new line something like:

```
######################################
# asdf

# this should not be part of a 'text' block
```

and possibly other edge cases that you think about.

Testing the content as well as the type of the block would be great too.

lesteve: Something I forgot: testing that 19 hashes in a row does not produce a split would be good too.

.travis.yml
```diff
@@ -12,12 +19,6 @@ env:
         - DISTRIB="conda" PYTHON_VERSION="3.4"
 
 before_install:
-    - if [ "$DISTRIB" == "ubuntu" ]; then
-         sudo apt-get update -qq;
-         sudo apt-get install -qq python-numpy python-matplotlib
-         python-setuptools;
-      fi
-
```
lesteve: Just curious, did you have to use sudo=false ? Does that mean that python-numpy and al. are installed for all other distribs including the conda ones ?

Titan-C: sudo:false is to use the new travis Container-Based Infrastructure which allows for faster builds. And all this apt-get packages are cached, so very fast install. Speed gain is really noticeable.

Yes these get installed in the conda builds as well but don't mix at all since conda has a virtual environment.

GaelVaroquaux: > Just curious, did you have to use sudo=false ? Does that mean that
> python-numpy and al. are installed for all other distribs including the
> conda ones ?

Yes, this is really the right way of doing it. Thanks a lot.

We can't do this in nilearn so far, because neurodebian has not been
added to the whitelist of authorized source.list

lesteve:
Something else I was wondering about. It'd be good if there was a better way to separate the output (as in stdout) from the input. At the moment, Just adding "Script output: " in front of the cell like it is done [here](http://sphinx-gallery-local.readthedocs.org/en/refactor_split/auto_examples/plot_strings.html#sphx-glr-auto-examples-plot-strings-py) is not optimal ...

lesteve:
Having said that, it could be done in a separate PR.

sphinxgallery/gen_rst.py
```diff
@@ -118,133 +78,155 @@ def flush(self):
...
-        start_row = 1
-    line_iterator = iter(lines)
-    tokens = tokenize.generate_tokens(lambda: next(line_iterator))
-    check_docstring = True
-    erow_docstring = 0
-    for tok_type, _, _, (erow, _), _ in tokens:
-        tok_type = token.tok_name[tok_type]
-        if tok_type in ('NEWLINE', 'COMMENT', 'NL', 'INDENT', 'DEDENT'):
+        # because we only allow for one single docstring
+        elif line.startswith('"""') and has_header_string == 1:
```
lesteve: Am I getting this right, that a docstring formatted like this:

```
"""First line
and there is another one"""
```
won't work because the second line does  not start with `""""` ?

Again, why not keep the previous code that does the AST magic for processing the docstring ?

Titan-C: You are not getting it right.
this only allows for a single docstring, the header docstring that means

```
"""First line of docstring
and the continuation is captured as docstring"""

"""This free string is a code comment and
will not be captured as comment and block breaker"""
```

The previous code with the tokenizer splits between free strings and not free strings(the code). I really have no idea on how to treat the new syntax using it.

lesteve: It looks like `ast.get_docstring(ast.parse(open('yourfile').read()))` might do what we want.

GaelVaroquaux:
> Having said that, it could be done in a separate PR.

+1

agramfort:
A script like this breaks things

```
# -*- coding: utf-8 -*-
"""
========================
Notebook styled examples
========================

Bla bla !
"""

##############################################################################
# Title

import numpy as np

##############################################################################
# End comments
#
# That's all folks !
```

for 2 reasons. You cannot start by a comment block after the header and you cannot end with a comment

sphinxgallery/gen_rst.py
lesteve: > You are not getting it right.

OK let's see what reality has to say about that. I tried using plot_strings.py and putting the second `"""` of the docstring in the same line as the last line of text i.e.:

```python
"""
=============================
Constrained Text output frame
=============================

This example captures the stdout and includes it in the
example. If output is too long it becomes automatically
framed into a text area."""
```

This is the output I get:

plot_strings

as you can see, the docstring ends up in a code cell rather than a text cell.

Titan-C:
Now it doesn't break anymore. Although it breaks not because of the parser, but because there is a literal_eval user in the rst rendering, which is used because comment strings used to be encapsulated in triple quotes. The last quick fix attaches those.

GaelVaroquaux:
Ideally, when you fix a problem like this, you should add a test, to avoid a regression.

agramfort:
Also something like this does not render fine

```
##############################################################################
# By default logging messages print to the console, but look at
# mne.set_log_file() to save output to a file.
#
# Access raw data
# ^^^^^^^^^^^^^^^
```

agramfort:
the reason is that the empty line before section title is removed when convert to rst

Titan-C:
My take on the Code splitter

Free verbatim strings are legal python  comments. They are recognized by the tokenizer in all python versions as complete blocks. We should support them(putting them back). Also is a good way to catch the module docstring

Hash leaded comments are recognized line by line and not taken as contiguous blocks. So the tokenizer is not usable, catching them by regular expressions is doable and I include the test set of the incurred fails. As an example in notebook style and a test. Test don't pass in python 2.6 as not all the regular expressions options are supported.

GaelVaroquaux:
> Free verbatim strings are legal python comments.

But they are horrible style, discouraged in every Python style guide. You
never see them in real code. We want to be encouraging people to write
beautiful code. People will copy-paste our examples. If our examples are
ugly, people will write ugly code. We shouldn't support them as a block
marker.

sphinxgallery/gen_rst.py
lesteve: That's pretty cool, the comment springs back to life if the diff becomes relevant again, I didn't know that.

lesteve:
> Hash leaded comments are recognized line by line and not taken as contiguous blocks. So the tokenizer is not usable

I am not sure I follow this completely. It feels like you can very well do it on a line by line basis. Basically the `20 *  '#'` starts the block, and on each subsequent line if there is a leading # you stay in the same 'text' block. As soon as you don't have a leading # you end the 'text' block.

Obviously you have worked on this a lot more closely than I have so there may be edge cases that I am just missing. Basically I'd be interested to have a more detailed understanding of them.

For extracting the docstring I put somewhere above an ast-based construct that was relatively simple (combination of ast.parse and ast.get_doctring IIRC)

Side note:

```python
###################
```

is not PEP8 (You need a space after the first #), so do we want to encourage a better block separator?

Titan-C:
I think this is ready

sphinxgallery/gen_rst.py
```diff
@@ -118,133 +79,134 @@ def flush(self):
...
-        if tok_type in ('NEWLINE', 'COMMENT', 'NL', 'INDENT', 'DEDENT'):
+
+    with open(source_file) as f:
+        source_lines = f.read()
+
+    doc_string = ast.get_docstring(ast.parse(source_lines))
+
+    blocks = [('text', doc_string)]
+    # Remove from source_lines the docstring
+    end_doc = [m.end() for m in re.finditer('"""', source_lines)][1]
```
lesteve: This is a bit fragile.

Looking at the ast.get_docstring code, I came up with a function that takes a filename and returns the docstring + the rest of the content of the file (feel free to tweak it any way you see fit):

```python
def get_module_docstring_and_rest(filename):
    with open(filename) as f:
        content = f.read()

    node = ast.parse(content)
    if not isinstance(node, ast.Module):
        raise TypeError("This function only supports modules. You provided {0}".format(node.__class__.__name__))
    if node.body and isinstance(node.body[0], ast.Expr) and \
       isinstance(node.body[0].value, ast.Str):
        docstring_node = node.body[0]
        docstring = docstring_node.value.s
        rest = content.split('\n', maxsplit=docstring_node.lineno)[-1]
        return docstring, rest
    else:
        raise ValueError(('Could not find docstring in file "{0}". '
                          'A docstring is required by sphinx-gallery').format(filename))
```

sphinxgallery/gen_rst.py
```diff
@@ -118,133 +79,134 @@ def flush(self):
...
+    source_lines = source_lines[end_doc:]
+
+    pattern = re.compile('^#{20,}.*\s((?:^#.*\s)*)', flags=re.M)
+    split_block = re.split(pattern, source_lines)
+    for split_code_block in split_block:
+        if split_code_block is None or split_code_block.strip() == '':
```
lesteve: I tried a few things but I couldn't figure out a way to get a None for split_code_block. Do you have one in mind ?

I would put a comment on top of this condition to make it more clear when it happens (e.g. newlines between two comment blocks).

Additional bonus points if you improve the naming for `split_block` and `split_code_blocks`. Maybe `for content in block_contents`? More than open for better suggestions.

sphinxgallery/gen_rst.py
```diff
@@ -118,133 +79,134 @@ def flush(self):
...
-    if not os.path.exists(os.path.join(src_dir, 'README.txt')):
-        print(80 * '_')
-        print('Example directory %s does not have a README.txt file' %
-              src_dir)
-        print('Skipping this directory')
-        print(80 * '_')
-        return ""  # because string is an expected return type
+        elif split_code_block.startswith('#'):
+            sub_pat = re.compile('^#', flags=re.M)
+            coment_block = dedent(re.sub(sub_pat, '', split_code_block))
```
lesteve: Missing 'm' in coment

sphinxgallery/gen_rst.py
```diff
@@ -118,133 +79,134 @@ def flush(self):
...
 
-            fhindex += _thumbnail_div(target_dir, fname, snippet)
-            fhindex += """
 
-.. toctree::
-   :hidden:
+def codestr2rst(codestr):
+    """Return reStructuredText code block from code string"""
+    code_directive = "\n.. code-block:: python\n\n"
+    indented_block = '    ' + codestr.replace('\n', '\n    ')
```
lesteve: You can use textwrap.indent here. I think this could be used in one or two other places in this file.

sphinxgallery/gen_rst.py
```diff
@@ -118,133 +79,134 @@ def flush(self):
...
 
-# clear at the end of the section
-    fhindex += """.. raw:: html\n
-    <div style='clear:both'></div>\n\n"""
 
-    return fhindex
+def extract_intro(filename):
+    """ Extract the first paragraph of module-level docstring. max:95 char"""
+
+    first_text = split_code_and_text_blocks(filename)[0][-1]
```
lesteve: It is a bit of a shame, that split_code_and_text_blocks is called again here to get the docstring given that you have already done the necessary work in generate_file_rst.

One way to fix this would be to return the docstring from generate_rst and pass it to make extract_info which would take a docstring instead of a filename. Better suggestions welcome!

sphinxgallery/gen_rst.py
```diff
@@ -118,133 +79,134 @@ def flush(self):
...
+
+    return first_par
+
+
+def _plots_are_current(src_file, image_file):
+    """Test existence of image file and later touch time to source script"""
+
+    first_image_file = image_file.format(1)
+    needs_replot = (not os.path.exists(first_image_file) or
+              os.stat(first_image_file).st_mtime <= os.stat(src_file).st_mtime)
```
lesteve: PEP8 under-indentation

sphinxgallery/gen_rst.py
```diff
@@ -292,157 +254,175 @@ def scale_image(in_fname, out_fname, max_width, max_height):
...
+              src_dir)
+        print('Skipping this directory')
+        print(80 * '_')
+        return ""  # because string is an expected return type
+
+    fhindex = open(os.path.join(src_dir, 'README.txt')).read()
+    if not os.path.exists(target_dir):
+        os.makedirs(target_dir)
+    sorted_listdir = [fname for fname in sorted(os.listdir(src_dir))
+                                if fname.endswith('py')]
```
lesteve: PEP8

sphinxgallery/gen_rst.py
```diff
@@ -292,157 +254,175 @@ def scale_image(in_fname, out_fname, max_width, max_height):
...
+
+        fhindex += _thumbnail_div(target_dir, fname, intro)
+        fhindex += """
+
+.. toctree::
+   :hidden:
+
+   /%s/%s\n""" % (target_dir, fname[:-3])
+
+
```
lesteve: PEP8 too many blank lines

sphinxgallery/gen_rst.py
```diff
@@ -292,157 +254,175 @@ def scale_image(in_fname, out_fname, max_width, max_height):
...
+    # clear at the end of the section
+    fhindex += """.. raw:: html\n
+    <div style='clear:both'></div>\n\n"""
+
+    return fhindex
+
+
+def execute_script(code_block, example_globals, image_path, fig_count, src_file):
```
lesteve: PEP8 line too long.

Pro tip: take some time setting up PEP8 checks in your editor of choice, it will be worth the investment in the long run.

sphinxgallery/gen_rst.py
```diff
@@ -118,133 +79,134 @@ def flush(self):
...
+
+    first_text = split_code_and_text_blocks(filename)[0][-1]
+
+    paragraphs = first_text.split('\n\n')
+    if len(first_text) > 1:
+        first_par = re.sub('\n', ' ', paragraphs[1])
+        first_par = ((first_par[:95] + '...')
+                     if len(first_par) > 95 else first_par)
+    else:
+        raise ValueError("Missing first paragraph."
```
lesteve: You might want a slightly better error message here. For example mentioning that you need a title (as a rst header) and a short explanation paragraph or something better.

sphinxgallery/gen_rst.py
```diff
@@ -292,157 +254,175 @@ def scale_image(in_fname, out_fname, max_width, max_height):
...
+            if blabel == 'code':
+                code_output, rtime, fig_count = execute_script(bcontent,
+                                                               example_globals,
+                                                               image_path,
+                                                               fig_count,
+                                                               src_file)
+
+                time_elapsed += rtime
+                # Single example style
+                if len(script_blocks) == 2:
```
lesteve: script_blocks length is not going to change in the `for blabel, bcontent in script_blocks` loop. You should put that in a nicely named boolean variable outside of the loop, for example:

```python

is_example_notebook_like = len(script_blocks) > 2

for blabel, bcontent in script_blocks:
    ...
    if has_example_notebook_style:
        ...
    else:
        ...
```

This way you can even get rid of the comments

sphinxgallery/gen_rst.py
```diff
@@ -292,157 +254,175 @@ def scale_image(in_fname, out_fname, max_width, max_height):
...
+
+    if not fname.startswith('plot'):
+        convert_func = dict(code=codestr2rst, text=text2string)
+        for blabel, bcontent in script_blocks:
+            example_rst += convert_func[blabel](bcontent)+'\n'
+    else:
+        example_globals = {}
+        fig_count = 0
+        for blabel, bcontent in script_blocks:
+            if blabel == 'code':
```
lesteve: slight niggle: you can take that out of the loop:

```python
code_blocks = [label, content for label, content in script_blocks if label == 'code']
for blabel, bcontent in code_blocks:
    ...
```

Now you can even use len(code_blocks) > 1 as the is_example_notebook_like condition which seems a tad less magical.

lesteve: Actually I was wrong, I didn't see the else clause below ...

lesteve:
OK I pushed a few commits to address my remaining comments so I am going to merge this one as is.

lesteve:
@Titan-C, great stuff btw, thanks!

agramfort:
> @Titan-C, great stuff btw, thanks!
> 
+1

GaelVaroquaux:
Nice!!!

Good job. Really good job.

lesteve:
@Titan-C BTW if you have five minutes, just double-check that nothing subtle has broken because of my recent commits: http://sphinx-gallery.readthedocs.org/en/latest/

Titan-C:
I think it looks fine.

lesteve:
> I think it looks fine.

Good to know, thanks for double-checking.

PR #49, opened 2015-08-04
demo bug

agramfort:
by just removing this empty line I break the rendering of the example.

@lesteve @Titan-C thanks for taking a look.


lesteve:
Bug confirmed:

sphinx_gallery_demo_bug

I'll take a closer look.

PR #51, opened 2015-08-05
Fix gen_rst.split_code_and_text_blocks with comment right after the docstring

lesteve:
plus test. Also not adding text block if it contains only whitespace hence the modification of reference_parse.txt.

Fix #49.


agramfort:
LGTM

thanks for the quick fix.

lesteve:
Merging then.

PR #62, opened 2015-10-04
Shell script for the local embeding

Titan-C:
This script is designed for updating sphinx-gallery in third party
packages. Performs a download from PyPI.

### DIFF
--- a/bin/copy_sphinxgallery.sh
+++ b/bin/copy_sphinxgallery.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+# Script to do a local install of sphinx-gallery
+rm -rf tmp sphinxgallery
+mkdir -p tmp/lib/python2.7/site-packages
+
+easy_install -Zeab tmp sphinx-gallery
+
+cp -vru tmp/sphinx-gallery/sphinxgallery/ .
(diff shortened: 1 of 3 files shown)


bin/copy_sphinxgallery.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+# Script to do a local install of sphinx-gallery
+rm -rf tmp sphinxgallery
+
+easy_install -Zeab tmp sphinx-gallery
+
+cp -vru tmp/sphinx-gallery/sphinxgallery/ .
+
+git add sphinxgallery
```
lesteve: Should the git part really be part of this script ?

Titan-C: Give that the last time I updated sphinx-gallery from PyPI I forgot to git add the new files, I do think is worth having this line. Even if it fails because the user does not use git, the local embedding is done and is one less thing to worry about.

https://github.com/scikit-learn/scikit-learn/pull/4658#discussion_r31616558

lesteve: I remembered that too. It's just that I would expect this kind of script to just get the newest sphinx-gallery without messing with my git status. Maybe you just want to test the latest sphinx-gallery version without any intent to commit the changes. 

Also it's not like missing a file is a something that we don't realise right away. IIRC it was caught during the review and it would have been caught next time someone tried to generate the doc anyway.

lesteve:
You need to rebase on master after the sphinx-gallery to sphinx_gallery renaming.

PR #66, opened 2015-10-27
[MRG] Rename sphinxgallery package to sphinx_gallery

### LINKED ISSUE #47: Rename package to sphinx_gallery rather than sphinxgallery?
A very minor annoyance, but I thought I would vent it out and ask for opinions, it seems to me that sphinx_gallery for the package name is more in line with sphinx-gallery as the project name.

Basically if people agree it would be wise to fix that sooner rather than later, to reduce the impact on potential users.

lesteve:
for consistency with the sphinx-gallery project name.

Fix #47.


Titan-C:
Now I'm wondering if we should have a change log file. I'm specially concerned as this next release will break some backwards compatibility. We change the configuration variables from `examples_dir` to `examples_dirs`, now is plural and we are also renaming the module.

GaelVaroquaux:
> Now I'm wondering if we should have a change log file.

That's a good idea. Do a CHANGES.txt file and 'include' it somewhere in
the docs. You can see how joblib does it for instance.

lesteve:
> Do a CHANGES.txt file and 'include' it somewhere in the docs

Created #67 about adding a whats_new file.

lesteve:
If nobody can see any issues with the sphinxgallery -> sphinx_gallery renaming, I'll merge this in a couple of hours.

agramfort:
besides the fact that MNE is released with the old import I am fine :)

lesteve:
> besides the fact that MNE is released with the old import I am fine :)

Don't you ship a copy of sphinx-gallery within MNE-python ?

agramfort:
nope due to avoid debian packaging issues. But don't worry we'll backport

lesteve:
> nope due to avoid debian packaging issues. But don't worry we'll backport

OK thanks.

Nobody complained loudly so I am merging this one.

GaelVaroquaux:
> Nobody complained loudly so I am merging this one.

There's a logical fallacy here: aside from @agramfort, your end-users are
not aware of this choice, so they cannot complain :)

lesteve:
> There's a logical fallacy here: aside from @agramfort, your end-users are not aware of this choice, so they cannot complain :)

Let's keep the philosopy/formal logic discussions for lunch, shall we ?

More seriously we can probably put together a backward-compatible mechanism if it turns out to be a real problem.

PR #72, opened 2015-11-26
STYLE: change the color of the output of a script

### LINKED ISSUE #44: Better way of separating the stdout in the notebook style examples
When #36 is merged, it would be good to find a better way to separate stdout from the input.

sphinx_gallery_output

Not sure about the best way how to do that at the moment, but just putting "Script output:" doesn't seem to cut it for me ...

GaelVaroquaux:
Fixes #44

### DIFF
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -55,7 +55,9 @@ div.sphx-glr-thumbContainer {
 
 .sphx-glr-script-out pre {
   overflow-y: auto;
-  max-height: 30em; }
+  max-height: 30em;
+  background-color: #FFFFFE;
+}
 
 ul.sphx-glr-horizontal {
   padding: 0px;
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -60,7 +60,8 @@ div.sphx-glr-script-out div.highlight {
 .sphx-glr-script-out pre {
   overflow: auto;
   max-height: 30em;
-  background-color: #FFFFFE;
+  background-color: #FAFAE2;
+  border: none;
   margin-left: 1ex;
   margin-top: 0px;
   padding-left: 1ex;
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -70,6 +70,19 @@ p.sphx-glr-horizontal {
     margin-bottom: 0px;
 }
 
+div.sphx-glr-download {
+    padding: 1ex;
+    margin: 1em auto 1ex auto;
+    border-radius: 4px;
+    max-width: 45ex;
+    background-color: #ffc;
+    border: 1px solid #C2C22D;
+}
+
+div.sphx-glr-download a {
+    color: #4B4600;
+}
+
 ul.sphx-glr-horizontal {
   padding: 0px;
   list-style: none; }
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -54,9 +54,16 @@ div.sphx-glr-thumbContainer {
     bottom: 58%; }
 
 .sphx-glr-script-out pre {
-  overflow-y: auto;
+  overflow: auto;
   max-height: 30em;
   background-color: #FFFFFE;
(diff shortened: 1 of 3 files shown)


sphinx_gallery/_static/gallery.css
```diff
@@ -55,7 +55,9 @@ div.sphx-glr-thumbContainer {
 
 .sphx-glr-script-out pre {
   overflow-y: auto;
-  max-height: 30em; }
+  max-height: 30em;
+  background-color: #FFFFFE;
```
lesteve: This is almost white and I thought you wanted it light yellow ... did you mean #FFFFEF instead?

GaelVaroquaux: It looks nice on my screen with my eyes. I don't mind it to be very light.
screenshot

Titan-C:
Back when I was thinking on issue #44 it came to my mind to have the script output with some indentation.

GaelVaroquaux:
OK, new iteration taking in account this comment. I think that this iteration is indeed better. I reduced also the vertical spacing.
screenshot

GaelVaroquaux:
The previous version disappearred, as it was in a diff that got overridden. I am putting it here for comparison.

screenshot_old

lesteve:
Not a CSS expert but this looks fine to me.

GaelVaroquaux:
One more iteration, to look better with the default sphinx CSS:
screenshot_default_css

GaelVaroquaux:
One last thing: I have made the download more visible (a bit of CSS, and adding a div to the HTML):
screenshot_download

GaelVaroquaux:
@Titan-C : how do you feel about this? Do you think that we can merge this, or do you have any further comments?

Titan-C:
I'll try to nitpick. It is probably fine in the old default of sphinx. But the new default where everything is in white: code output and the code are not so different. The same for the read the docs theme. The slight indent helps a bit.

I like making the download, something closer to a button. I think it would be very useful in that case to actually copy the CSS trick I implemented for the gallery thumbnails to make the link overtake the full surface of its container making everything clickable, and friendlier for screens in mobile devices.

Finally as I confessed in #73 I actually code in SCSS, but I can port back your CSS style changes at a later time.

GaelVaroquaux:
> I'll try to nitpick. It is probably fine in the old default of sphinx.
> But the new default where everything is in white: code output and the
> code are not so different. The same for the read the docs theme. The
> slight indent helps a bit.

OK, I'll have a lot. It's a pity that compiling the docs locally does not
show this.

> I like making the download, something closer to a button. I think it would be
> very useful in that case to actually copy the CSS trick I implemented for the
> gallery thumbnails to make the link overtake the full surface of its container
> making everything clickable, and friendlier for screens in mobile devices.

I'll look.

> Finally as I confessed in #73 I actually code in SCSS, but I can port
> back your CSS style changes at a later time.

I think that this is a problem. The files under version control should
always be editable. The way I work to do CSS is that I play with an
online editor (the chrome one, but it doesn't matter). However, I port
back manually the changes to the CSS. It helps having a clear and
readable CSS, which is important for long term maintenance.

GaelVaroquaux:
New version of the color, in a brighter yellow, to make a difference with the read the docs theme. The screenshot attached is with a choice of colors that mimics this theme.

new_screen_yellow

GaelVaroquaux:
> > I like making the download, something closer to a button. I think it
> > would be very useful in that case to actually copy the CSS trick I
> > implemented for the gallery thumbnails to make the link overtake the
> > full surface of its container making everything clickable, and
> > friendlier for screens in mobile devices.

> I'll look.

To do think, I would need to have only the download link, and not the
'Download Python source code' label. We could choose to keep only this
label, and loose the name of the file. The technical constrain is that I
cannot have 2 different styles in the button. 

I think that I prefer the current layout, that makes it clear what the
file name is. Indeed, on some system, the file just gets downloaded in
the background, and people might be looking for this file in the list of
files that they have downloaded.

GaelVaroquaux:
To summarize, I think that I have addressed your comments as well as I could for now, and I'd like a new review for merge.

agramfort:
LGTM

+1 for merge.

thanks @GaelVaroquaux

PR #74, opened 2015-11-28
ENH: sort the examples from shortest to longest

GaelVaroquaux:
In a gallery, we want the simplest examples in the beginning. A good
proxy for simplicity is the amount of code.

This commit adds sorting by the amount of code.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -482,10 +482,11 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
 
     script_blocks = split_code_and_text_blocks(example_file)
 
+    amount_of_code = sum([len(bcontent)
+                         for blabel, bcontent in script_blocks
+                         if blabel == 'code'])
+
     if _plots_are_current(src_file, image_path):
-        amount_of_code = sum([len(bcontent)
-                            for blabel, bcontent in script_blocks
-                            if blabel == 'code'])
         return amount_of_code
 
     time_elapsed = 0
@@ -529,10 +530,6 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
 
             else:
                 example_rst += text2string(bcontent) + '\n'
-
-    amount_of_code = sum([len(bcontent)
-                          for blabel, bcontent in script_blocks
-                          if blabel == 'code'])
 
     save_thumbnail(image_path, base_image_name, gallery_conf)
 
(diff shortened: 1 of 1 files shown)


agramfort:
nice !

GaelVaroquaux:
> nice !

I am restoring what we used to have in nilearn and scikit-learn.

sphinx_gallery/gen_rst.py
```diff
@@ -515,9 +530,14 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
             else:
                 example_rst += text2string(bcontent) + '\n'
 
+    amount_of_code = sum([len(bcontent)
+                          for blabel, bcontent in script_blocks
+                          if blabel == 'code'])
+
```
Titan-C: This block is repeated. I would put it only once before the `if _plots_are_current` clause

Titan-C:
I changed the default sorting to alphabetic as it is simpler (depending on your goals) to sort files at will depending on file name.
There is the issue #37 to provide a sorting mechanism.

sphinx_gallery/gen_rst.py
GaelVaroquaux: > This block is repeated. I would put it only once before the if _plots_are_current clause

Thanks. Addressed.

GaelVaroquaux:
> I changed the default sorting to alphabetic as it is simpler (depending
> on your goals) to sort files at will depending on file name.

It's a pretty bad default. It will not do anything close to "the right
thing" by default.

> There is the issue #37 to provide a sorting mechanism.

Sure it is possible to do something better, but in the mean time we need
something better than an alphabetic sort. For any major project with a
long list of examples it's important. This PR solves the problem for the
90% use case with a few lines of code and no configuration necessary from
the user.

sfias:
Sorting by default on the amount of code is probably the stupidest thing I've ever seen! It's a nice feature, but using it by default?! If I'm looking for a certain example in the gallery I know how to look for it alphabetically, I don't know how much code it's going to have... Moreover, shorter code isn't necessarily easier to understand. The easiest way would be to number your examples from easier to harder. And if I may say so, GaelVaroquaux, your comment on not doing "anything close to "the right thing" by default" is you just being a d*ck about it.

GaelVaroquaux:
> Sorting by default on the amount of code is probably the stupidest thing I've ever seen!

> And if I may say so, GaelVaroquaux, your comment on not doing "anything close to "the right thing" by default" is you just being a d*ck about it.

Thank you! I am flattered.

I guess that based on this comment, the sphinx-gallery development team
can expect a pull request from you that implements optional variants on
how examples are sorted, as discussed in:
https://github.com/sphinx-gallery/sphinx-gallery/issues/37

lesteve:
I would just add: @sfias maybe have a quick look at your github profile picture again and double-check that the `beAwesome()` line has not been optimized away by your internal compiler.

lesteve:
@sfias you may want to have a look at https://opensource.guide/how-to-contribute/#communicating-effectively and especially the "Above all, keep it classy" bit.

Issue #39, opened 2015-06-29
Wish: IPython Notebook Download


florian-wagner:
Since documenting strings are possible since PR #36. A possible feature would be an optional "Download as IPython Notebook" button for example scripts, when an IPython installation is found.

jnothman:
Unfortunately jupyter supports markdown natively, but not restructured text

florian-wagner:
I see, haven't thought about that. But it should be possible the other way round, i.e. allowing and integrating plot_example.ipynb files in examples folder.

lesteve:
Obviously this, or a variant of this, is something we'd like to do at some point in the future. 

There are multiple approaches but I think we want to keep examples as plain python files that are convertible into IPython notebooks (for example through simple markup), rather than allowing .ipynb examples.

GaelVaroquaux:
We now have notebook-style examples (which are awesome!). It's the first step to have IPython notebooks. @lesteve, @Titan-C any idea of the timeline for that?

http://sphinx-gallery.readthedocs.org/en/latest/tutorials/plot_notebook.html#sphx-glr-tutorials-plot-notebook-py

Titan-C:
closed with #75

PR #75, opened 2015-12-07
[MRG+1] IPython notebook parser

Titan-C:
This PR addresses #39 
Using the block splitting of the scripts introduced for the notebook styled example. It is now simple to construct an IPython notebook cell by cell.
I now include as a dependency the nbformat which defines the cell nodes of the Ipython notebook.
Code is tidy in the notebooks. The RST text is understood by the Notebook quite fine.

Now that I see on travis python2.6 fails

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,3 @@ matplotlib
 pillow
 sphinx
 nose
-nbformat
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ matplotlib
 pillow
 sphinx
 nose
+nbformat
--- a/sphinx_gallery/notebook.py
+++ b/sphinx_gallery/notebook.py
@@ -6,6 +6,7 @@ Parser for Jupyter notebooks
 
 from __future__ import division, absolute_import, print_function
 import nbformat
+import os
 import warnings
 
 notebook_skeleton = {
@@ -35,8 +36,9 @@ notebook_skeleton = {
 
 
 class notebook_file(object):
-    def __init__(self, nb_file_name):
-        self.file_name = nb_file_name
+    def __init__(self, file_name, target_dir):
+        self.file_name = file_name.replace('.py', '.ipynb')
+        self.write_file = os.path.join(target_dir, self.file_name)
         self.work_notebook = nbformat.from_dict(notebook_skeleton)
         self.add_code("%matplotlib inline")
 
@@ -58,5 +60,5 @@ class notebook_file(object):
         })
         self.work_notebook["cells"].append(markdown_cell)
 
-    def write_file(self):
-        nbformat.write(self.work_notebook, self.file_name)
+    def save_file(self):
+        nbformat.write(self.work_notebook, self.write_file)
--- a/sphinx_gallery/notebook.py
+++ b/sphinx_gallery/notebook.py
@@ -14,6 +14,7 @@ from __future__ import division, absolute_import, print_function
 from copy import deepcopy
 import json
 import os
+import re
 
 NOTEBOOK_SKELETON = {
     "cells": [],
@@ -88,6 +89,8 @@ class Notebook(object):
         code : str
             Cell content
         """
+        top_heading = re.compile(r'^=+$\s^([\w\s]+)^=+$', flags=re.M)
+        text = re.sub(top_heading, r'# \1', text)
         markdown_cell = {
             "cell_type": "markdown",
             "metadata": {},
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -3,3 +3,4 @@ matplotlib
 pillow
 sphinx
 nose
+nbformat
```
GaelVaroquaux: nbformat pulls in traitlets, ipython_genutils, jsonchema, and jupyter_core.

I worry particularly about jupyter_core: we could easily run into conflicts or break an IPython install by dragging in an incompatible version.

It seems like a big cost, given that writing a notebook is not much more than writing a json file. How difficult would it be to have our own code for writing such a file?

GaelVaroquaux: The failure on 2.6 is that traitlets is not 2.6 compatible.

Titan-C: > nbformat pulls in traitlets, ipython_genutils, jsonchema, and jupyter_core.

They are indeed a lot of dependencies, it's too much

> It seems like a big cost, given that writing a notebook is not much more than writing a json file. How difficult would it be to have our own code for writing such a file?

Not difficult at all. I started with nbformat, because that is the library for the notebook file and it provides their utilities. Now, I just used python standard json library. The file is not too ugly after enforcing indentation on save(Now that anyone might be willing to read it). It does not come out nicely ordered as IPython seems to prefer. But It the file is a valid JSON file and IPython seems to open it without problems. Upon saving it from the notebook it does recover the nice ordering.

GaelVaroquaux: > I started with nbformat, because that is the library for the notebook
> file and it provides their utilities. Now, I just used python standard
> json library.

Good thinking! That was definitely the best way to get quick turn
around times and short cycles.

Titan-C:
I have added an extra converter from rst to IPython markdown. Well it only parses the top heading and math. Other headings as they are not overlined work well.

The boundary case syntax used in the plot_parse example is not parsed. There are exotic undelines characters for headings, and heavy use of rst syntax. Is just rendered as text. For the rest of the examples in here they look nice

Titan-C:
Rebased to master, conflicts solved. Reviews needed

sphinx_gallery/notebook.py
```diff
@@ -0,0 +1,121 @@
...
+
+    math_eq = re.compile(r'^\.\. math::((?:.+)?(?:\n+^  .+)*)', flags=re.M)
+    text = re.sub(math_eq,
+                  lambda match: r'$${0}$$'.format(match.group(1).strip()),
+                  text)
+    inline_math = re.compile(r':math:`(.+)`')
+    text = re.sub(inline_math, r'$\1$', text)
+
+    return text
+
```
GaelVaroquaux: For PEP8 compatibility, you should have an extra empty line here (two empty lines between top-level blocks).

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -8,8 +8,10 @@
 import tempfile
 
 import sphinx_gallery.gen_rst as sg
+import sphinx_gallery.notebook as gnb
```
GaelVaroquaux: In general, those shorthands make the code harder to read. In my world, GNB means "Gaussian Naive Bayes" :).

Maybe you could simply use "from sphinx_gallery import notebook".

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -88,3 +90,21 @@ def test_extract_intro():
...
+        for blabel, bcontent in blocks:
+            if blabel == 'code':
+                example_nb.add_code_cell(bcontent)
+            else:
+                example_nb.add_markdown_cell(sg.text2string(bcontent))
+
+        example_nb.save_file()
+
+        f.flush()
+        assert_equal(gnb.json.load(f), example_nb.work_notebook)
```
GaelVaroquaux: It would be good to directly import json, and not use the one that is inherited from gnb. The reason is that, to the reader that doesn't know the entire codebase, it is not clear at all that "gnb.json" is the json module from the stdlib.

GaelVaroquaux:
Overall the code is nice, and it is really cool that you got math working! Scikit-image needed that for integration of the gallery.

I made a few cosmetic comments, but aside these, :+1: for merge.

lesteve:
This needs a rebase actually.

Titan-C:
Rebased to master and comments addressed.

GaelVaroquaux:
LGTM. +1 for merge.

agramfort:
works like a charm !

+1 for merge.

really nice @Titan-C !

GaelVaroquaux:
:beers: Yey! yey! yey!

agramfort:
:beers: !!!

Issue #64, opened 2015-10-23
check example scripts updates by md5 hash and not timestamp


Titan-C:

Titan-C:
fixed in #76

PR #76, opened 2015-12-10
[MRG+1] Check script update by md5 updates

Titan-C:
In response to #64 this change follows the code md5

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -240,7 +240,7 @@ def _plots_are_current(src_file, image_file):
     has_image = os.path.exists(first_image_file)
 
     with open(src_file, 'r') as src_data:
-        src_content = src_data.read()
+        src_content = src_data.read().encode('utf-8')
         src_md5 = hashlib.md5(src_content).hexdigest()
 
     src_md5_file = src_file + '.md5'
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -233,11 +233,8 @@ def extract_intro(filename):
     return first_paragraph
 
 
-def _plots_are_current(src_file, image_file):
-    """Test existence of image file and later touch time to source script"""
-
-    first_image_file = image_file.format(1)
-    has_image = os.path.exists(first_image_file)
+def get_md5sum(src_file):
+    """Returns md5sum of file"""
 
     with open(src_file, 'r') as src_data:
         src_content = src_data.read()
@@ -247,6 +244,13 @@ def _plots_are_current(src_file, image_file):
             src_content = src_content.encode('utf-8')
 
         src_md5 = hashlib.md5(src_content).hexdigest()
+    return src_md5
+
+
+def check_md5sum_change(src_file):
+    """Returns True if src_file has a different md5sum"""
+
+    src_md5 = get_md5sum(src_file)
 
     src_md5_file = src_file + '.md5'
     src_file_changed = True
@@ -259,6 +263,17 @@ def _plots_are_current(src_file, image_file):
     if src_file_changed:
         with open(src_md5_file, 'w') as file_checksum:
             file_checksum.write(src_md5)
+
+    return src_file_changed
+
+
+def _plots_are_current(src_file, image_file):
+    """Test existence of image file and no change in md5sum of
+    example"""
+
+    first_image_file = image_file.format(1)
+    has_image = os.path.exists(first_image_file)
+    src_file_changed = check_md5sum_change(src_file)
 
     return has_image and not src_file_changed
 
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -13,9 +13,6 @@ Files that generate images should start with 'plot'
 
 """
 from __future__ import division, print_function, absolute_import
-from . import glr_path_static
-from .backreferences import write_backreferences, _thumbnail_div
-from textwrap import dedent
 from time import time
 import ast
 import hashlib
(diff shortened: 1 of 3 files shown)


sphinx_gallery/gen_rst.py
```diff
@@ -236,10 +237,30 @@ def _plots_are_current(src_file, image_file):
...
+        with open(src_md5_file, 'r') as file_checksum:
+            ref_md5 = file_checksum.read()
+        if src_md5 == ref_md5:
+            src_file_changed = False
+
+    if src_file_changed:
+        with open(src_md5_file, 'w') as file_checksum:
+            file_checksum.write(src_md5)
+
+    return has_image and not src_file_changed
```
lesteve: That's slightly weird: I would think that we don't really care whether the example creates a plot or not. If its code has changed, we need to run it again, for example because its output on stdout has changed.

lesteve:
Not sure how easy this is but it would be great to be able to test this functionality.

sphinx_gallery/gen_rst.py
Titan-C: Actually this is a survivor of the original code. There is a check if there is at least one image for the script(since only the `plot_` scripts are executed, we expect an image). If the image is missing we run it again.
Now I do notice I missed a negation in front of has_image, I skipped that on the refactoring

Titan-C: What are the use cases of this image checking? I'm not sure, maybe an interrupted documentation build. Then asking for the first image might not be robust enough. Maybe it had to do with the issue of deleting images, although that happens in the `_build` directory.

Titan-C:
> Not sure how easy this is but it would be great to be able to test this functionality.

I'm open for suggestions on this. Throw random text to a file so there is always md5 mismatch?

sphinx_gallery/gen_rst.py
lesteve: > Actually this is a survivor of the original code. There is a check if there is at least one image for the script(since only the plot_ scripts are executed, we expect an image). If the image is missing we run it again.
Now I do notice I missed a negation in front of has_image, I skipped that on the refactoring

I see, it is slightly surprising but fair enough, at least it looks like the output will always be up-to-date even when no plot is produced. 

plot_* files don't necessary produce plots. It feels like looking whether the image exists is also a hacky way to figure out whether the example ran successfully. Maybe not in this PR but we should look at this to clean-up the code.

That makes me think that actually your md5 hash logic has some flaw. Imagine the example fails for example because one of the depedencies is not installed. You will still write the md5 hash, meaning that once you install the right dependencies the example won't be rerun because its content hasn't changed. Not sure how edge-casy this is and maybe we can live with it for now.

lesteve:
> I'm open for suggestions on this. Throw random text to a file so there is always md5 mismatch?

One thing is to try to structure the code in independent functions that are easily testable in a stand-alone way. Haven't put more thought into it in this particular case ...

sphinx_gallery/gen_rst.py
Titan-C: > That makes me think that actually your md5 hash logic has some flaw. Imagine the example fails for example because one of the depedencies is not installed. You will still write the md5 hash, meaning that once you install the right dependencies the example won't be rerun because its content hasn't changed. Not sure how edge-casy this is and maybe we can live with it for now.

Here is maybe another reason there is the image search

sphinx_gallery/gen_rst.py
```diff
@@ -13,18 +13,19 @@
...
 import shutil
-import traceback
-import sys
 import subprocess
+import sys
+import traceback
 import warnings
-from textwrap import dedent
-from . import glr_path_static
-from .backreferences import write_backreferences, _thumbnail_div
```
agramfort: why not keeping the local imports at the end?

it's common to start with std lib then external deps and then local files

Titan-C: Because I did an alphabetical sort. I can chage it, no string feeling for this.

Given your ordering logic,  should the imports of matplotlib, basestring and textwrap that have more complicated import instructions be also before the local files, or after all that import list?

agramfort: > On 1 janv. 2016, at 22:25, Óscar Nájera <notifications@github.com> wrote:
> 
> In sphinx_gallery/gen_rst.py:
> 
> >  import warnings
> > -from textwrap import dedent
> > -from . import glr_path_static
> > -from .backreferences import write_backreferences, _thumbnail_div
> Because I did an alphabetical sort. I can chage it, no string feeling for this.
> 
> Given your ordering logic, should the imports of matplotlib, basestring and textwrap that have more complicated import instructions be also before the local files, or after all that import list?
> 

Matplotlib is not a local file it should be Just above locals and after standard library imports

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -88,3 +88,19 @@ def test_extract_intro():
...
+        f.flush()
+        file_md5 = sg.get_md5sum(f.name)
+        # verify correct md5sum
+        assert_equal('ea8a570e9f3afc0a7c3f2a17a48b8047', file_md5)
+        # True because is a new file
+        assert_true(sg.check_md5sum_change(f.name))
+        # False because file has not changed since last check
+        assert_false(sg.check_md5sum_change(f.name))
+
+    sg.os.remove(f.name+'.md5')
```
GaelVaroquaux: I think that it would be more clear and natural to import os directly at the top of this file, rather than using the one that comes with "sg".

GaelVaroquaux:
Overall, LGTM. Thanks a lot.

It would be useful to mention this (the fact that examples are not rerun unless their content changes) somewhere in the docs.

sphinx_gallery/tests/test_gen_rst.py
agramfort: and I would put spaces around + :)

besides +1 for merge

Titan-C:
Comments addressed

GaelVaroquaux:
Still LGTM. I'll let someone else give an LGTM and merge.

agramfort:
thx @Titan-C !

GaelVaroquaux:
Sphinx-gallery is making some very nice fast-paced progress. Next release
will be nice!

Issue #65, opened 2015-10-23
Don't put examples output in sphinx-build stdout?


Titan-C:
I was wondering if it is necessary to put the stdout of the examples in the build as is done. Would it be better to hide it and instead only output those which failed.
Currently I want to debug why some nilearn examples are not building https://github.com/nilearn/nilearn/pull/592 and I notice that in stdout there is no traceback, and in the rendered html example I only get the message that the script didn't compile.

agramfort:
do you mean not show stdout in final pages? I like having them.

GaelVaroquaux:
> do you mean not show stdout in final pages? I like having them.

So do I actually. When an example is very slow, it is good to have them
to show progress.

lesteve:
> I notice that in stdout there is no traceback, and in the rendered html example I only get the message that the script didn't compile.

But the traceback should be in stderr, right ? I am pretty sure at one point you would see the traceback in the example HTML.

GaelVaroquaux:
> But the traceback should be in stderr, right ? I am pretty sure at one point
> you would see the traceback in the example HTML.

That's certainly something that we would like to have.

Titan-C:
> do you mean not show stdout in final pages? I like having them.
>So do I actually. When an example is very slow, it is good to have them
>to show progress.

stdout goes always to the final pages. I was claiming that during the build process you get all the stdout of all the examples in the same terminal and is just a mess of text. So, I think it might be nicer not to show it.
Maybe to show progress put an indicator(that might be to complex and not a priority)

> But the traceback should be in stderr, right ? I am pretty sure at one point
>  you would see the traceback in the example HTML.
> That's certainly something that we would like to have.

The traceback is printed but becomes hard to find among all the other successful stdout of the other examples. Also it might be possible to keep track of which file is failing because now the execution is done by blocks which have no link to the original file besides the stdout claiming before which file got executed(you also loose the line number correspondence)

Finaly, traceback is not printed in the rendered HTML. I'm starting to think that might be useful.

lesteve:
> Finaly, traceback is not printed in the rendered HTML. I'm starting to think that might be useful.

I seem to remember it was the case in nilearn at one point but maybe I just dreamt about it!

Titan-C:
After merging #97 closing this.
The output of a script is written to stdout during build to have the feeling of progress. Errors are written to stderr, and on the rendered example, plus a summary at the end with the traceback.

PR #77, opened 2015-12-10
[MRG+1] Traceback message in Gallery output

Titan-C:
This PR need comments. 
First there is the situation of how to deal with a broken build #45, discussion shall continue there.
Here I address, for the moment, that the traceback of a broken example should be also printed in the HTML output of the gallery following my comment on https://github.com/sphinx-gallery/sphinx-gallery/issues/65#issuecomment-151472930


agramfort:
that's fine with me

other option is to stop the doc build with a meaningful crash message

GaelVaroquaux:
> other option is to stop the doc build with a meaningful crash message

It should be controlled by an option (just like the building of the
gallery is).

lesteve:
>> other option is to stop the doc build with a meaningful crash message

> It should be controlled by an option (just like the building of the gallery is).

My 2 cents: I think the current behaviour is fine for now. The general use case is to build all the examples to find all the broken ones rather than fail on the first error.

Something I would actually find useful (not sure how easy this is): have a special image for broken examples allowing them to identify quite easily in the gallery.

sphinx_gallery/gen_rst.py
```diff
@@ -447,12 +447,14 @@ def execute_script(code_block, example_globals, image_path, fig_count,
...
-        figure_list = []
-        image_list = '%s is not compiling:' % src_file
         print(80 * '_')
-        print(image_list)
+        print('%s is not compiling:' % src_file)
         traceback.print_exc()
         print(80 * '_')
+
+        figure_list = []
+        exception_msg = indent(traceback.format_exc(), '    ')
```
lesteve: Nitpick: in order not to do traceback.print_exc() and then traceback.format_exc() which is kind of doing the same thing twice, you can do:

```python

formatted_exception = traceback.format_exc()

print(formatted_exception)

exception_msg = indent(formatted_exception, ' ' * 4)
```

lesteve:
Just tried this out and here is a snapshot:

sphinx_gallery_traceback

It would be nice to have a fixed-width font in the HTML traceback. This makes a huge difference for long tracebacks.

Titan-C:
Now it's improved.
I painted the broken image that takes over in the gallery thumbnails on failed examples. And the traceback message is not anymore in the sphinx warning box that changes font style but uses just a container.

lesteve:
> Now it's improved.
I painted the broken image that takes over in the gallery thumbnails on failed examples. And the traceback message is not anymore in the sphinx warning box that changes font style but uses just a container.

The image thing is quite nice, although we can think about tweaking the "broken" image.

I still get non fixed-width font in the warnings when generating the gallery locally though.

Titan-C:
> I still get non fixed-width font in the warnings when generating the gallery locally though.

No idea here. What does you browser tell you the CSS style is. The block shall be a sphx-glr-traceback. That one has no style for fonts. The example I'm providing gives me this

snapshot1

lesteve:
I have to say not sure whether the technical term for it is fixed-font or not but basically what I was trying to say is that I would have the same font in the traceback and in the code cell.

Titan-C:
> I would have the same font in the traceback and in the code cell.

I just learned that is syntax hightlighting for python traceback messages. Thus I'm making the traceback outputs be parsed as that. Plus an additional light red box to call attention is a traceback and that the example fails.

image

Titan-C:
As here I change some CSS I would like PR #80 to be merged first. This branch is already rebased to it, but it did not generate a branch dependency as I expected.

agramfort:
is it possible to have the build stop if an example fails? to me that should be the default behavior. It happened to me too many times to push a website update with broken examples... it happens quickly when you have dozens of examples.

Titan-C:
I added the quick fail option to break the gallery build process as soon as an example trows an Exception, added the documentation what to put in the makefile for that. But I have not made it the default behavior, as building the complete gallery and scanning for the broken examples visually from the rendered website seems more intuitive for me. You can vote on that.

doc/advanced_configuration.rst
```diff
@@ -163,3 +163,50 @@ This will download directly from PyPI our latest released code and
 save it to the current folder. This is a striped version of the
 Sphinx-Gallery module to incorporate in your project. You should also
 add it to your version control system.
+
+Build the Gallery without executing the examples
+================================================
+
+Sphinx-Gallery can parse all your examples and build the gallery
+without executing any of the scripts. This is just for speed
+visiulization processes of the gallery and the size it take your
```
agramfort: visiulization

agramfort: take -> takes

display -> to display

doc/advanced_configuration.rst
```diff
@@ -163,3 +163,50 @@ This will download directly from PyPI our latest released code and
...
+website display, or any use you can imagine for it. To achive this you
```
agramfort: achive -> achieve

doc/advanced_configuration.rst
```diff
@@ -163,3 +163,50 @@ This will download directly from PyPI our latest released code and
...
+
+Remember that for ``Makefile`` white space is significant and the indentation are tabs
+and not spaces
+
+Dealing with failing Gallery example scripts
+============================================
+
+As your project evolves some of your example scripts might stop
+executing properly. Sphinx-Gallery assist you in the discovery process
+of those bugged examples. The defaul behavior is to replace the
```
agramfort: defaul -> default

doc/advanced_configuration.rst
```diff
@@ -163,3 +163,50 @@ This will download directly from PyPI our latest released code and
...
+behavior.
+
+An extra functionality of Sphinx-Gallery is the early fail option. In
+this mode the gallery build process breaks as soon as an exeption
+occurs in the execution of the examples scripts. To activate this
+modality you need to pass a flag at the build process. It can be done
+by including in your ``Makefile``::
+
+    html-quick_fail:
+        $(SPHINXBUILD) -D quick_fail=1 -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
```
agramfort: quick_fail -> fail_on_error ?

agramfort:
besides LGTM

doc/advanced_configuration.rst
Titan-C: This one I was also thinking, early_fail, early_abort

agramfort: > This one I was also thinking, early_fail, early_abort

I find fail_on_error more explicit.

sphinx_gallery/gen_rst.py
```diff
@@ -438,21 +438,35 @@ def execute_script(code_block, example_globals, image_path, fig_count,
...
+
         print(80 * '_')
-        print(image_list)
-        traceback.print_exc()
+        print('%s is not compiling:' % src_file)
+        print(formatted_exception)
         print(80 * '_')
+
+        figure_list = []
+        image_list = codestr2rst(formatted_exception, 'pytb')
```
GaelVaroquaux: For readability, I would suggest being explicit about the keyword argument here, by writing:

<pre>
    ..., lang='pytb')
</pre>

GaelVaroquaux:
I agree with your various design choices (in particular the default).

+1 for merge. I have suggested a tiny style change.

agramfort:
it's fine to merge as is for me

thx @Titan-C

doc/Makefile
```diff
@@ -58,6 +58,11 @@ html-noplot:
 	@echo
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
 
+html-quick_fail:
```
lesteve: maybe html-quick-fail in order not to mix - and _

examples/plot_raise.py
```diff
@@ -0,0 +1,33 @@
...
+import numpy as np
+import matplotlib.pyplot as plt
+
+plt.pcolormesh(np.random.randn(100, 100))
+
+
+###############################################################################
+# Here is another error raising Block
+
+raise IOError
```
lesteve: Maybe put a more realistic error like: `plt.plot('this is not going to work')`

lesteve: What I meant is that the traceback in this case will be more interesting.

lesteve:
Is broken_stamp.svg actually used?

Slightly related is do you know why the "broken" image png appears with a black background in the gallery:
sphinx-gallery

lesteve:
My contribution for the 'quick_fail' naming: abort_on_example_error.

lesteve:
Could you rebase on master so that we make sure no extra CSS changes has slipped through in this PR?

Titan-C:
broken_stamp.svg is not used now, but is the source of for the png file. Not sure if it will become useful in the future if we want to tune the image or if #61 

The background of the image is black in the thumbnails because in the sphinx-classic css theme all body backgrounds are black, and the png image is transparent.

examples/plot_raise.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
+"""
+=============================
+Examples that fail to execute
```
lesteve: Singular seems more fitting here: Example that fails to execute

GaelVaroquaux:
Overall, LGTM. +1 for merge (ideally, please address the comments of the others).
Thanks!

agramfort:
fine with me

Titan-C:
Rebased to master and comments addressed

GaelVaroquaux:
@agramfort was good with this. I am too. Let's merge!

PR #80, opened 2015-12-29
Cleaning CSS style

### LINKED ISSUE #73: The syle of the CSS needs sanitizing
The CSS styles should be as such
<pre>
element {
   prop1: value;
   prop2: value;
}

element2 {
   prop: value;
}
</pre>

Titan-C:
This is to fix #73 
I have moved all the changes from the latest CSS version in the gallery back to the SCSS. The SCSS file is now linted with scss-lint and the resulting CSS file is exported in expanded form instead of the default nested.


Titan-C:
Changes visible in my read the docs fork http://sphinx-gallery-local.readthedocs.org/en/cssclean/
Where no difference shall be noticed

GaelVaroquaux:
I really think that the fact that you are relying on SCSS is a problem. It means that only you master the workflow to edit the CSS. As such this is a limiting factor to bringing in contributions.

Right now you are the main developer by far, so it's more beneficial that harmful. If I were you, I would train myself to work without SCSS, because it makes you more powerful.

agramfort:
+1

the size of our css file is not huge to justify the use of scss

my 2c

PR #83, opened 2016-01-03
Illustrated that links between gallery examples are possible.

emmanuelle:
Links between gallery examples are a useful feature. I added such a link to one of the ``sin_func`` gallery examples. Maybe this feature could be documented in ``advanced_configuration.rst`` as well?

### DIFF
--- a/examples/sin_func/plot_sin.py
+++ b/examples/sin_func/plot_sin.py
@@ -14,6 +14,10 @@ it corresponds mathematically to the function:
 
     x \\rightarrow \\sin(x)
 
+Note that ``sphinx-gallery`` automatically creates labels for each example from
+its filename. You can thus refer to it from any part of the documentation,
+including from other examples, as here
+
 .. seealso::
     :ref:`sphx_glr_auto_examples_sin_func_plot_sin_black_background.py` for a
     fancier plot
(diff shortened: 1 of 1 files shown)


examples/sin_func/plot_sin.py
```diff
@@ -14,6 +14,9 @@
 
     x \\rightarrow \\sin(x)
 
+.. seealso::
+    :ref:`sphx_glr_auto_examples_sin_func_plot_sin_black_background.py` for a
+    fancier plot
```
GaelVaroquaux: Maybe here you should add a text saying why this is there, for instance

"Note that sphinx-gallery automatically creates labels for each example from it's filename. You can thus refer to it from any part of the documentation, including from other examples, as here".

GaelVaroquaux:
+1 for merge although I would add the sentence above :).

I don't think that "advanced configuration" is the part of the narrative documentation where this should be added: it's not a configuration issue. That said, I agree that mentioning this in the narrative documentation would be great.

Titan-C:
I would also agree on including the same note on the readme.rst, maybe on the `The structure of the examples folder` section. Or indeed as a short note in the in the `features` section of the index.rst file

GaelVaroquaux:
I don't think that it should be in the readme: the readme should focus on
the big-picture.

agramfort:
thx @emmanuelle !

PR #84, opened 2016-01-07
STYLE: CSS tweak

GaelVaroquaux:
1. The matching on .sphx-glr-script-out+p is too general and could
   introduce space where we don't want it. I don't know why it was
   dropped in
   https://github.com/sphinx-gallery/sphinx-gallery/commit/3a8d9dcef6487f1d89434b2487c160cb37e430fa
   This seems an error too me.

2. I have made the space on the right of the output blocks disappear.
   While working on a course, I realized that it was distracting.

### DIFF
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -67,15 +67,16 @@
 }
 .sphx-glr-script-out .highlight {
   background-color: transparent;
-}
-.sphx-glr-script-out pre {
-  background-color: #fafae2;
-  border: 0;
   margin-left: 2.5em;
   margin-top: -1.4em;
+}
+.sphx-glr-script-out .highlight pre {
+  background-color: #fafae2;
+  border: 0;
   max-height: 30em;
   overflow: auto;
   padding-left: 1ex;
+  margin: 0px;
   word-break: break-word;
 }
 .sphx-glr-script-out + p {
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -78,10 +78,12 @@
   padding-left: 1ex;
   word-break: break-word;
 }
-.sphx-glr-script-out + p {
+blockquote.sphx-glr-script-out + p {
   margin-top: 1.8em;
 }
-
+blockquote.sphx-glr-script-out {
+  margin-left: 0pt;
+}
 .sphx-glr-download {
   background-color: #ffc;
   border: 1px solid #c2c22d;
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -71,14 +71,14 @@
 .sphx-glr-script-out pre {
   background-color: #fafae2;
   border: 0;
-  margin-left: 1ex;
-  margin-top: 0;
+  margin-left: 2.5em;
+  margin-top: -1.4em;
   max-height: 30em;
   overflow: auto;
   padding-left: 1ex;
   word-break: break-word;
 }
-blockquote.sphx-glr-script-out + p {
+.sphx-glr-script-out + p {
   margin-top: 1.8em;
 }
 blockquote.sphx-glr-script-out {
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -120,9 +120,7 @@ SINGLE_IMAGE = """
 
 CODE_OUTPUT = """.. rst-class:: sphx-glr-script-out
 
- **Output**:\n
-
-  ::
+ Out::
 
 {0}\n"""
 


agramfort:
fine with me

lesteve:
Not a CSS guru, @Titan-C any objections?

Titan-C:
> The matching on .sphx-glr-script-out+p is too general

I really don't understand why it is to general, as there is only a class that called .sphx-glr-script-out and the +p if for the text paragraph immediately after. I dropped blockquote because the SCSS linter suggested.

Now I'm re-testing, but I cannot see any difference in the style that your modifications make to the current state of the CSS. To me everything looks identical.
I'm checking with all the notebook styled examples that show output, with a clean build of the gallery from master in Mozilla firefox 43.0.1

Most importantly

GaelVaroquaux:
>     The matching on .sphx-glr-script-out+p is too general

> I really don't understand why it is to general, as there is only a
> class that called .sphx-glr-script-out and the +p if for the text
> paragraph immediately after.

I am thinking in terms of future evolution. I think that it is easy for
someone to modify the generated html, and create a glitch by adding
another element with .sphx-glr-script-out. Given that the "+p" is not
something people usually have in mind (ie that they need to look at
something targetting the element before). So my gut feeling is that we
need to narrow the scope of this rule.

Titan-C:
> I am thinking in terms of future evolution. I think that it is easy for someone to modify the generated html, and create a glitch by adding another element with .sphx-glr-script-out. Given that the "+p" is not something people usually have in mind (ie that they need to look at something targetting the element before). So my gut feeling is that we need to narrow the scope of this rule.

OK, I don't have enough experience to foreshadow those future glitches. We can prefix it with blockquote. Then how about the other elements of .sphx-glr-script-out, the bare, the nested .highlight and the pre? Shall we prefix them too?

Then I still miss, what is with:
> I have made the space on the right of the output blocks disappear.
>While working on a course, I realized that it was distracting.

I have not been able to identify any visual difference(or is not evident to me) between the current CSS and the one proposed.(Observation based only on the Gallery notebook styled examples, which have output and text following). Which space on the right? and that is removed with a `margin-left: 0pt;`?

GaelVaroquaux:
> Then I still miss, what is with:

>     I have made the space on the right of the output blocks disappear.
>     While working on a course, I realized that it was distracting.

> I have not been able to identify any visual difference(or is not
> evident to me) between the current CSS and the one
> proposed.(Observation based only on the Gallery notebook styled
> examples, which have output and text following). Which space on the
> right? and that is removed with a margin-left: 0pt;?

Look on the right of a (yellow) output block.

GaelVaroquaux:
OK, here is a new iteration of the CSS style that I think is better than what I had previously submitted:

output_sphx_gla

I have made the output much closer vertically.

The reason that I am working on the CSS of this part (the output), is that on a long and busy tutorial, the vertical spacing does guide the reading at all currently, and they end up hard to read.

Please review.

Titan-C:
I like this new style, looks elegant.
We might be hitting a browser compatibility issue in CSS style, not sure how relevant this is.
For me in Chrome it looks like your snapshot, in firefox the output block already starts on top of the colon. Setting `.sphx-glr-script-out pre {margin-left: 2.8em}` is enough to fixit in the default sphinx-theme.

On the read the docs theme, this comple style tweak has no effect.
spectacle tj7602

The problem is mostly irrelevant if we change the Style into 
```css
.sphx-glr-script-out .highlight {
  margin-left: 2.5em;
  margin-top: -1.4em;
}
.sphx-glr-script-out .highlight pre {
  background-color: #fafae2;
  border: 0;
  max-height: 30em;
  overflow: auto;
  padding-left: 1ex;
  word-break: break-word;
}
```
Finally I acknowledge my blindness, I fail to see the effect of this styling since the first commit
 ```css
blockquote.sphx-glr-script-out {
  margin-left: 0pt;
}

GaelVaroquaux:
@Titan-C : you reply has several points. Let my try to comment on the various:

* Firstly, '.sphx-glr-script-out pre {margin-left: 2.8em}' is fine with me, but I have the feeling that you are suggesting a better fix bellow (that I have implemented and pushed).

* Second, and most important: it's pretty much impossible for me to develop CSS changes because I cannot reproduce the read-the-doc theme. I am happy addressing problems of the read-the-doc theme, but it's not used by any project that uses sphinx-gallery. Right now, it is a strawman that prevents improve the visual experience with the other themes. If there is an easy way for a contributor to reproduce the read-the-doc theme, it would probably be useful to document it in a contributors guide.

That second point is related to your last point: on certain themes it makes a difference. Namely themes that add a margin to the right for blockquotes (there are a few that does this).

I've pushed a new version. I think that I have followed all your suggestions. Tell me if it suits you.

Titan-C:
Yes, changes are good for me. I checked in the default and read the docs theme.

>  If there is an easy way for a contributor to reproduce the read-the-doc theme, it would probably be useful to document it in a contributors guide

I use these: http://read-the-docs.readthedocs.org/en/latest/theme.html#how-do-i-use-this-locally-and-on-read-the-docs

GaelVaroquaux:
> I use these: http://read-the-docs.readthedocs.org/en/latest/theme.html#
> how-do-i-use-this-locally-and-on-read-the-docs

Looks like read the docs is a bit broken here :$.

I think that you should put information somewhere in the contributing
guide (which doesn't exists, but should be created, probably with
CONTRIBUTING.md, as in scikit-learn, that gets picked up by github). What
I would say is to test the rendering of the online documentation, one
needs to install ... and change the conf.py to ...

That said, I find it very bad practice to have in the development pattern
something that requires extra installs and that contains this amount of
magic. It's a hurdle to contributing documentation or CSS changes. We
might have to think about working around this in the future.

GaelVaroquaux:
Could we merge this before the release?

agramfort:
+1

Titan-C:
I don't understand. But this CSS change just broke in travis after the merge.

Titan-C:
@GaelVaroquaux  Since you are committing to master I'll let you continue to do it directly. so we don't clash.
My quick fix is to update https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/tests/test_gen_rst.py#L131
with the matching rst output
`code_output = 'Out::\n\n      output\n\n'`

Keep in mind the correct amount of spaces before output, github html reduces them.

GaelVaroquaux:
I've fixed the problem. Sorry.

PR #93, opened 2016-02-02
[MRG] MISC: add vertical space

GaelVaroquaux:
This commit adds vertical between the output of the script and the code
of the script, when there is only one output and it comes on top of the
code (non notebook behavior).

This is to avoid ugly-looking rendering like the following:

https://circle-artifacts.com/gh/nilearn/nilearn/1108/artifacts/0/home/ubuntu/nilearn/doc/_build/html/auto_examples/01_plotting/plot_overlay.html#sphx-glr-auto-examples-01-plotting-plot-overlay-py

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -581,7 +581,7 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
                     example_rst += code_output
                 else:
                     example_rst += code_output
-                    if code_output:
+                    if 'sphx-glr-script-out' in code_output:
                         # Add some vertical space after output
                         example_rst += "\n\n|\n\n"
                     example_rst += codestr2rst(bcontent) + '\n'
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -581,6 +581,9 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
                     example_rst += code_output
                 else:
                     example_rst += code_output
+                    if code_output:
+                        # Add some vertical space after output
+                        example_rst += "\n\n|\n\n"
                     example_rst += codestr2rst(bcontent) + '\n'
 
             else:


sphinx_gallery/gen_rst.py
```diff
@@ -581,6 +581,9 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
                     example_rst += code_output
                 else:
                     example_rst += code_output
+                    if code_output:
+                        # Add some vertical space after output
```
Titan-C: code_output is always true, even if the script produces no images, nor prints anything(pointless for an executable example maybe) code_output = '\n\n\n\n\n'

GaelVaroquaux: > code_output is always true, even if the script produces no images, nor
> prints anything(pointless for an executable example maybe) code_output
> = '\n\n\n\n\n'

Good point. I need to check that it is not that.

GaelVaroquaux: Done!

sphinx_gallery/gen_rst.py
```diff
@@ -581,6 +581,9 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
                     example_rst += code_output
                 else:
                     example_rst += code_output
+                    if not 'sphx-glr-script-out' in code_output:
```
Titan-C: I'm a bit confused.
* You want the space after the code stdout, right? (then the not is out of place)
* Or you want it after the image, when there is only the figure an no stdout?(then is ok)

sphinx_gallery/gen_rst.py
```diff
@@ -581,6 +581,9 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
...
+                        # Add some vertical space after output
+                        example_rst += "\n\n|\n\n"
                     example_rst += codestr2rst(bcontent) + '\n'
```
Titan-C: In my opinion it would be convenient to have the space always and skip the check. I was trying to get it with CCS but is just too complicated and not done.
I would put all here
```python 
example_rst += '|\n\n' + codestr2rst(bcontent) + '\n'
```

sphinx_gallery/gen_rst.py
GaelVaroquaux: >   • You want the space after the code stdout, right? (then the not is out of
>     place)

Arg. Right. Let me fix that.

GaelVaroquaux: > >   • You want the space after the code stdout, right? (then the not is out of
> >     place)

> Arg. Right. Let me fix that.

Fixed (ammended the commit)

GaelVaroquaux:
This is ready for merge: I included all the comments. Should we merge this?

GaelVaroquaux:
Thanks!

PR #98, opened 2016-02-16
plot sorted running time of examples when building the doc

agramfort:
cc @jaeilepp 

that's useful in mne-python to tracking the examples that take the most time.

@Titan-C feel free to take over to make it at your taste

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,12 @@ Change Log
 v0.1.2
 ------
 
+New features
+------------
+
+* Print aggregated and sorted list of computation times of all examples
+  in the console during the build.
+
 Bug Fixes
 .........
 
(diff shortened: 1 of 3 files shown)


agramfort:
@lesteve @Titan-C any objection?

agramfort:
@GaelVaroquaux gave its +1 IRL :)

GaelVaroquaux:
+1

Titan-C:
It looks fine. I would also ask to write about it on the Changes.rst

I'm also thinking. What if you accumulate the computation time in the same list, where the amount of code is? Maybe we can use it in the future to sort the examples by build time?

agramfort:
>
> It looks fine. I would also ask to write about it on the Changes.rst
>
 pushed. Take over if it's not the way you see it.

> I'm also thinking. What if you accumulate the computation time in the same
> list, where the amount of code is? Maybe we can use it in the future to
> sort the examples by build time?
>
it feels a bit YAGNI to me.

Titan-C:
This needs a rebase

agramfort:
done

PR #102, opened 2016-02-21
Generated by Sphinx-Gallery signature

### LINKED ISSUE #99: Add a "generated by sphinx-gallery" on the pages we generate
Make it very discreet: small and in light gray, but I think that we should have a "Generated by sphinx-gallery" at the bottom of each page, in a div that we put as a float: right, where "sphinx-gallery" would be a link to the sphinx-gallery docs.

We need more publicity.

Titan-C:
closes #99 
Adds a gray box at the right end of each html site giving publicity to sphinx-gallery


agramfort:
can you show a page to see how it looks?

Titan-C:
This is the publicity in the gallery

screenshot_20160306_191426
And within the examples
gallery_pub

A gray box close to being invisible in the lower right corner

agramfort:
I find this a bit too big I am sure we can tune this with css in each
project.

lesteve:
It feels like the Copyright thingy would be a good place to put this piece of information, doesn't it?

"Copyright 2014-2016, Óscar Nájera. Created using Sphinx 1.3.5'

Mention sphinx-gallery there with a link and possibly the sphinx-gallery version.

Not sure how easy this is.

Titan-C:
> It feels like the Copyright thingy would be a good place to put this piece of information, doesn't it?

Yes, it would be nice. I can do it for our website. But it I understood @GaelVaroquaux correctly in #99 this is to put it on users websites to make some publicity.

The issue of putting this box close to the copyright is that the rst generated by sphinx-gallery has no influence on the template footer, so I can't really put it there. I imagine I can by CSS to put it at the absolute lower, right corner. But I would never now, if it would cover some user's website design. Thus I put it at the edge of the bounding box allowed for the rst document.

Titan-C:
image

How about now I have made the box smaller

lesteve:
LGTM, we can always tweak it later if we want to improve aesthetics.

GaelVaroquaux:
It should be even smaller: the same size as the "using Sphinx" text.

It's annoying when projects use strong branding.

PR #104, opened 2016-02-29
[MRG+1] ENH: Better catch for import

Eric89GXL:
I had a problem with Mayavi throwing a `ValueError` during import when using sphinx-gallery. This makes the catch more general and thus safer.

### DIFF
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -75,7 +75,7 @@ def get_short_module_name(module_name, obj_name):
         short_name = '.'.join(parts[:i])
         try:
             exec('from %s import %s' % (short_name, obj_name))
-        except ImportError:
+        except Exception:  # libraries can throw all sorts of exceptions...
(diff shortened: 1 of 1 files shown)


GaelVaroquaux:
LGTM

Titan-C:
I agree libraries can throw all kinds of exceptions, is fine with me.

Now this is done when checking which functions where used. I guess the example should have also failed.  I wonder if one should stop all further procedures performed on file once the example fails?

Eric89GXL:
In my particular case I'm doing `make html_dev-noplot` and hit this error, so having it fail out would at this point in the code would be bad. I know the examples don't build with my current `mayavi` setup.

Titan-C:
> In my particular case I'm doing make html_dev-noplot and hit this error

In deed in this case you could not track that the example fails.

Thanks, merging.

Eric89GXL:
Thanks for the quick reviews and the excellent package

PR #97, opened 2016-02-14
[MRG] Exitcodes

Titan-C:
This expands the behavior towards failing examples.

Closes #88, as examples that fail are stopped from execution to avoid noise in output.

Closes #95. Failed examples are accumulated internally. At build finished they exit with 1.
Error messages are now in STDERR

PR #106, opened 2016-03-14
[MRG+1]: Allow unicode in code and outputs

### LINKED ISSUE #18: Unicode support

Eric89GXL:
Previously, [our] (https://github.com/mne-tools/mne-python) code had unicode in the `stdout` which gave:
```
../examples/preprocessing/plot_maxwell_filter.py is not compiling:
  File "/home/larsoner/custombuilds/sphinx-gallery/sphinx_gallery/gen_rst.py", line 474, in execute_script
    my_stdout = my_buffer.getvalue().strip().expandtabs()
  File "/usr/lib/python2.7/StringIO.py", line 271, in getvalue
    self.buf += ''.join(self.buflist)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xce in position 36: ordinal not in range(128)
```
Modified a test that fails on `master` and passes on this PR.

Works on my system (™) on Py3k and Python 2.7.

Closes #18.
Closes #19.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ install:
         if [ "$PYTHON_VERSION" == "2.7" ]; then
           conda install --yes --quiet mayavi;
           conda upgrade --yes --all;
-          conda upgrade --yes pyface;
+          pip install --upgrade pyface;
         fi;
       fi;
     - pip install -r requirements.txt
--- a/examples/plot_quantum.py
+++ b/examples/plot_quantum.py
@@ -43,8 +43,8 @@ plt.show()
 print('pass')
 
 ###############################################################################
-# (Does not work in Sphinx-Gallery)
-# ---------------------------------
+# And then:
+# ---------
 
 plt.plot(x, (1 - x) / 4)
 plt.ylabel(r'$\langle n_\uparrow n_\downarrow \rangle$')
--- a/examples/plot_quantum.py
+++ b/examples/plot_quantum.py
@@ -1,0 +1,52 @@
+# -*- coding: utf-8 -*-
+r"""
+======================
+Some Quantum Mechanics
+======================
+
+We start with a two spin system :math:`\uparrow` and :math:`\downarrow`
+
+"""
+# Author: Óscar Nájera
+
+from __future__ import division, absolute_import, print_function
+
+
+def hamiltonian(M, mu):
+    r"""Generate a single orbital isolated atom Hamiltonian in particle-hole
+    symmetry. Include chemical potential for grand Canonical calculations
+
+    .. math::
+        \mathcal{H} - \mu N = M(n_\uparrow - n_\downarrow)
+        - \mu(n_\uparrow + n_\downarrow)
+    """
+    pass
+
+###############################################################################
+# Double occupation
+# -----------------
+#
+# To find out the double occupation one uses the relation
+# (Works in Sphinx-Gallery)
+
+import matplotlib.pylab as plt
+import numpy as np
+x = np.linspace(0, 1, 20)
+plt.plot(x, (1 - x**2) / 4)
+plt.ylabel('$\\langle n_\\uparrow n_\\downarrow \\rangle$')
+plt.show()
+
+###############################################################################
+#
+# .. math:: \langle n_\uparrow n_\downarrow \rangle = \frac{2\langle V \rangle}{U}+\frac{1}{4}
+
+print('pass')
+
+###############################################################################
+# (Does not work in Sphinx-Gallery)
+# ---------------------------------
+
+plt.plot(x, (1 - x) / 4)
+plt.ylabel(r'$\langle n_\uparrow n_\downarrow \rangle$')
+plt.xlabel('U')
+plt.show()
--- a/examples/plot_quantum.py
+++ b/examples/plot_quantum.py
@@ -1,52 +1,67 @@
 # -*- coding: utf-8 -*-
 r"""
-======================
-Some Quantum Mechanics
-======================
+=================================================
+Some Quantum Mechanics, filling an atomic orbital
+=================================================
 
-We start with a two spin system :math:`\uparrow` and :math:`\downarrow`
+Considering an atomic single orbital and how to fill it by use of the
+chemical potential. This system has a four element basis, :math:`B =
+\{ \lvert \emptyset \rangle, \lvert \uparrow \rangle, \lvert
+\downarrow \rangle, \lvert \uparrow\downarrow \rangle \}`, that is the
+empty orbital, one spin up electron, one spin down electron and the
+filled orbital.
 
+The environment of the orbital is set up by an energy cost for
+occupying the orbital, that is :math:`\epsilon` and when both
+electrons meet a contact interaction corresponding to the Coulomb
+repulsion :math:`U`. Finally the chemical potential :math:`\mu` is
+what allows in the Grand canonical picture, to fill up our atomic
+orbital from a reservoir of electrons.
+
+ The the simple Hamiltonian to model this system is given by:
+
+.. math::
+   \mathcal{H} =
+        \sum_{\sigma=\uparrow,\downarrow} \epsilon c^\dagger_\sigma c_\sigma
+       + Un_\uparrow n_\downarrow - \mu \hat{N}
+
+Here :math:`c^\dagger,c` creation and annihilation operators,
+:math:`n=c^\dagger c`, and
+:math:`\hat{N}=n_\uparrow+n_\downarrow`. This Hamiltonian is diagonal
+in the basis of particle number we have chosen earlier, as the basis
+elements are also eigenvectors.
+
+.. math::
+   \mathcal{H} \lvert \emptyset \rangle &= 0 \\
+   \mathcal{H} \lvert \uparrow  \rangle &= (\epsilon - \mu) | \uparrow  \rangle \\
+   \mathcal{H} \lvert \downarrow  \rangle &= (\epsilon - \mu) | \downarrow  \rangle \\
+   \mathcal{H} \lvert \uparrow\downarrow \rangle &= (2\epsilon - 2\mu +U) \lvert \uparrow\downarrow \rangle
+
+It is easy to see, that the system will prefer to be empty if
+:math:`\mu \in [0,\epsilon)`, be single occupied if :math:`\mu \in (\epsilon, \epsilon +U)`
+and doubly occupied if :math:`\mu > \epsilon +U`.
+
+For a more rigorous treatment, the partition function has to be
+calculated and then the expected particle number can be
+found. Introducing a new variable :math:`\xi = \epsilon - \mu`, and
+:math:`\beta` corresponding to the inverse temperature of the system.
+
+.. math::
+   \mathcal{Z} &= Tr(e^{-\beta \mathcal{H}}) = 1 + 2e^{-\beta\xi} + e^{-\beta(2\xi + U)} \\
+   \langle \hat{N} \rangle &= \frac{1}{\beta} \frac{\partial}{\partial \mu} \ln \mathcal{Z}
 """
-# Author: Óscar Nájera
 
-from __future__ import division, absolute_import, print_function
-
-
-def hamiltonian(M, mu):
-    r"""Generate a single orbital isolated atom Hamiltonian in particle-hole
-    symmetry. Include chemical potential for grand Canonical calculations
-
-    .. math::
-        \mathcal{H} - \mu N = M(n_\uparrow - n_\downarrow)
-        - \mu(n_\uparrow + n_\downarrow)
-    """
-    pass
-
-###############################################################################
-# Double occupation
-# -----------------
-#
-# To find out the double occupation one uses the relation
-# (Works in Sphinx-Gallery)
+# Code source: Óscar Nájera
+# License: BSD 3 clause
 
 import matplotlib.pylab as plt
 import numpy as np
-x = np.linspace(0, 1, 20)
-plt.plot(x, (1 - x**2) / 4)
-plt.ylabel('$\\langle n_\\uparrow n_\\downarrow \\rangle$')
-plt.show()
-
-###############################################################################
-#
-# .. math:: \langle n_\uparrow n_\downarrow \rangle = \frac{2\langle V \rangle}{U}+\frac{1}{4}
-
-print('pass')
-
-###############################################################################
-# And then:
-# ---------
-
-plt.plot(x, (1 - x) / 4)
-plt.ylabel(r'$\langle n_\uparrow n_\downarrow \rangle$')
-plt.xlabel('U')
+mu = np.linspace(0, 3, 800)
+for b in [10, 20, 30]:
+    n = 2 * (np.exp(b * (mu - 1)) + np.exp(b * (2 * mu - 3))) / \
+        (1 + np.exp(b * (mu - 1)) * (2 + np.exp(b * (mu - 2))))
+    plt.plot(mu, n, label=r"$\beta={}$".format(b))
+plt.xlabel(r'$\mu$ ($\epsilon=1$, $U=1$)')
+plt.ylabel(r'$\langle N \rangle=\langle n_\uparrow \rangle+\langle n_\downarrow\rangle$')
+plt.legend(loc=0)
 plt.show()
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -12,8 +12,9 @@ example files.
 Files that generate images should start with 'plot'
 
 """
-from __future__ import (division, print_function, absolute_import,
-                        unicode_literals)
+# Don't use unicode_literals here (be explicit with u"..." instead) otherwise
+# tricky errors come up with exec(code_blocks, ...) calls
+from __future__ import division, print_function, absolute_import
 from time import time
 import ast
 import codecs
@@ -139,7 +140,8 @@ SINGLE_IMAGE = """
 """
 
 
-CODE_OUTPUT = """.. rst-class:: sphx-glr-script-out
+# This one could contain unicode
+CODE_OUTPUT = u""".. rst-class:: sphx-glr-script-out
 
  Out::
 
@@ -486,6 +488,8 @@ def execute_script(code_block, example_globals, image_path, fig_count,
         sys.stdout = my_stdout
 
         t_start = time()
+        # don't use unicode_literals at the top of this file or you get
+        # nasty errors here on Py2.7
         exec(code_block, example_globals)
         time_elapsed = time() - t_start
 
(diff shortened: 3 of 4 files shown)


Eric89GXL:
With the latest commit I get this:
```
        Adjusted coil positions by (μ ± σ): 1.2° ± 1.5° (max: 6.7°)
  File "/usr/lib/python2.7/logging/__init__.py", line 882, in emit
    stream.write(fs % msg.encode("UTF-8"))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xce in position 36: ordinal not in range(128)
Logged from file maxwell.py, line 1698

```
And the example continues to run, but I still have some issues :(

sphinx_gallery/gen_rst.py
```diff
@@ -124,7 +125,7 @@ def flush(self):
 """
 
 
-CODE_OUTPUT = """.. rst-class:: sphx-glr-script-out
+CODE_OUTPUT = u""".. rst-class:: sphx-glr-script-out
```
Titan-C: We start using in our files, so we don't miss a string
```python
from __future__ import unicode_literals
```

Titan-C:
I have still an old PR on unicode #19. What I rescue from it now is that file opening is using the python module codecs. Instead of loading the file as binary
```python
with codecs.open(filename,  'w', 'utf8') as file_content:
```

Eric89GXL:
Nice, I'll try swapping that in

sphinx_gallery/gen_rst.py
Eric89GXL: I'm not quite sure what you mean. You mean in the modules you work on that
use sphinx-gallery you've started putting in a bunch of those lines? Or
you've started putting them in the sphinx-gallery code itself?

So far we haven't needed to use them to have unicode work properly. Does it
make them work better with sphinx-gallery, or is there some other
advantage?

Titan-C:
I tested this on my personal examples in python2. It is actually a pain to get unicode to work  every time. Because it can fail when docstrings are extracted or while executing or when writing the rst file. And I'm not entirely sure how to work around this.

I also noticed that it makes a huge difference for the test case if one loads from disk the file or one gets the input from within the test script as in now.

Eric89GXL:
> I'm not entirely sure how to work around this.

I think we need to make sure `unicode` is used properly everywhere. It's a bit annoying to get right but it should be possible.

> I also noticed that it makes a huge difference for the test case if one loads from disk the file or one gets the input from within the test script as in now.

So I should add another test, then, I take it?

sphinx_gallery/gen_rst.py
Titan-C: unicode_literals takes the need away to prefix every string with `u`, makes everything unicode. My first guess was to use that instead of going through each string prefixing the `u`. But it is not that perfect, and I could not get sphinx-gallery to run my examples with unicode_literals in gen_rst.py

Eric89GXL: Did you try putting it in while on this branch, or on a previous one? It might work okay with this one since the unicode reading is a bit more unified.

Eric89GXL: (I put in the `__future__` line, removed all `u'` and `u"` and it seems to work fine)

Eric89GXL:
> I also noticed that it makes a huge difference for the test case if one loads from disk the file or one gets the input from within the test script as in now.

I don't quite get this actually. The test script writes a file to disk, and tests it. Shouldn't that cover the use case you mention? If not, can you make a small test that fails, and I can work on fixing it?

Eric89GXL:
Okay @Titan-C I switched to using `codecs` where possible. Tests pass over here an 2.7 and 3.4, and things render properly for our repo. Ready to go from my end.

If you see degenerate cases, could you try to turn them into failing tests, and either post them here or open a PR into my branch? Without seeing those bits of code, it will be hard for me to get it right.

agramfort:
one travis build is not happy

Eric89GXL:
Yeah I saw that, but it looks unrelated...?

Eric89GXL:
(Some error with Pygments)

Eric89GXL:
@Titan-C the example renders fine now, see what you think

examples/plot_quantum.py
```diff
@@ -0,0 +1,52 @@
...
+plt.show()
+
+###############################################################################
+#
+# .. math:: \langle n_\uparrow n_\downarrow \rangle = \frac{2\langle V \rangle}{U}+\frac{1}{4}
+
+print('pass')
+
+###############################################################################
+# And then:
```
Eric89GXL: @Titan-C if you know something more intelligent to put here let me know :)

agramfort:
thanks heaps @Eric89GXL !

sphinx_gallery/gen_rst.py
```diff
@@ -527,7 +531,7 @@ def execute_script(code_block, example_globals, image_path, fig_count,
 
         # Breaks build on first example error
 
-        if gallery_conf['abort_on_example_error']:
+        if gallery_conf.get('abort_on_example_error', True):
```
Titan-C: The default is false and is set up in gen_gallery.py

Eric89GXL: It's not always set -- in our tests, if I had something raise an error at an appropriate time, I got an error here for not having this property

Titan-C: > It's not always set -- in our tests

You are right. It is not set in the default dictionary, but in the gallery build configuration. Certainly we want tests to fail immediately, but the gallery to continue the build even if some examples fail as the defaults. I'll have to keep track of this in #97, there I'm writing a helper function for the tests to set defaults. 

One thing I do prefer is to have the defaults all in one place and not scattered in the code.  For now I think we can leave it like that.

Eric89GXL: Yeah I agree about keeping them in one place. I'll add a comment that it should be unified if possible.

examples/plot_quantum.py
Titan-C: something very smart from my notes
```python
# -*- coding: utf-8 -*-
r"""
=================================================
Some Quantum Mechanics, filling an atomic orbital
=================================================

Considering an atomic single orbital and how to fill it by use of the
chemical potential. This system has a four element basis, :math:`B =
\{ \lvert \emptyset \rangle, \lvert \uparrow \rangle, \lvert
\downarrow \rangle, \lvert \uparrow\downarrow \rangle \}`, that is the
empty orbital, one spin up electron, one spin down electron and the
filled orbital.

The environment of the orbital is set up by an energy cost for
occupying the orbital, that is :math:`\epsilon` and when both
electrons meet a contact interaction corresponding to the Coulomb
repulsion :math:`U`. Finally the chemical potential :math:`\mu` is
what allows in the Grand canonical picture, to fill up our atomic
orbital from a reservoir of electrons.

 The the simple Hamiltonian to model this system is given by:

.. math::
   \mathcal{H} =
        \sum_{\sigma=\uparrow,\downarrow} \epsilon c^\dagger_\sigma c_\sigma
       + Un_\uparrow n_\downarrow - \mu \hat{N}

Here :math:`c^\dagger,c` creation and annihilation operators,
:math:`n=c^\dagger c`, and
:math:`\hat{N}=n_\uparrow+n_\downarrow`. This Hamiltonian is diagonal
in the basis of particle number we have chosen earlier, as the basis
elements are also eigenvectors.

.. math::
   \mathcal{H} \lvert \emptyset \rangle &= 0 \\
   \mathcal{H} \lvert \uparrow  \rangle &= (\epsilon - \mu) | \uparrow  \rangle \\
   \mathcal{H} \lvert \downarrow  \rangle &= (\epsilon - \mu) | \downarrow  \rangle \\
   \mathcal{H} \lvert \uparrow\downarrow \rangle &= (2\epsilon - 2\mu +U) \lvert \uparrow\downarrow \rangle

It is easy to see, that the system will prefer to be empty if
:math:`\mu \in [0,\epsilon)`, be single occupied if :math:`\mu \in (\epsilon, \epsilon +U)`
and doubly occupied if :math:`\mu > \epsilon +U`.

For a more rigorous treatment, the partition function has to be calculated and then
the expected particle number can be found. Introducing a new variable
:math:`\xi = \epsilon - \mu`, and :math:`\beta` corresponding to the
inverse temperature of the system.

.. math::
   \mathcal{Z} &= Tr(e^{-\beta \mathcal{H}}) = 1 + 2e^{-\beta\xi} + e^{-\beta(2\xi + U)} \\
   \langle \hat{N} \rangle &= \frac{1}{\beta} \frac{\partial}{\partial \mu} \ln \mathcal{Z}
"""

import matplotlib.pylab as plt
import numpy as np
mu = np.linspace(0, 3, 800)
b = 10
for b in [10, 20, 30]:
    n = 2 * (np.exp(b * (mu - 1)) + np.exp(b * (2 * mu - 3))) / \
        (1 + np.exp(b * (mu - 1)) * (2 + np.exp(b * (mu - 2))))
    plt.plot(mu, n, label=r"$\beta={}$".format(b))
plt.xlabel(r'$\mu$ ($\epsilon=1$, $U=1$)')
plt.ylabel(r'$\langle N \rangle=\langle n_\uparrow \rangle+\langle n_\downarrow\rangle$')
plt.legend(loc=0)
plt.show()
```

Eric89GXL:
@Titan-C merged your PR, any other comments or does it work correctly for you now?

Titan-C:
It works for my known cases.
+1

Titan-C:
Just came to my mind. Can you put a line about this on the CHANGES.rst file, please.

sphinx_gallery/gen_rst.py
```diff
@@ -93,6 +97,18 @@ def flush(self):
         self.file2.flush()
 
 
+class MyBytesIO(BytesIO):
```
lesteve: Better name for the class ? I don't have any great suggestion I am afraid ...

sphinx_gallery/gen_rst.py
```diff
@@ -505,16 +530,16 @@ def execute_script(code_block, example_globals, image_path, fig_count,
         fig_count += 1  # raise count to avoid overwriting image
 
         # Breaks build on first example error
-
-        if gallery_conf['abort_on_example_error']:
+        # Eventually this should hopefully be unified with the default conf
+        if gallery_conf.get('abort_on_example_error', True):
```
lesteve: Is this change meant to be part of to this PR?

sphinx_gallery/gen_rst.py
```diff
@@ -505,16 +530,16 @@ def execute_script(code_block, example_globals, image_path, fig_count,
...
             raise
 
     finally:
-        os.chdir(cwd)
         sys.stdout = orig_stdout
+        os.chdir(cwd)
```
lesteve: Interesting, does interverting these two lines make any difference?

sphinx_gallery/gen_rst.py
```diff
@@ -609,7 +634,8 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
 
     time_m, time_s = divmod(time_elapsed, 60)
     example_nb.save_file()
-    with open(os.path.join(target_dir, base_image_name + '.rst'), 'w') as f:
+    with codecs.open(os.path.join(target_dir, base_image_name + '.rst'),
+                     'w', 'utf-8') as f:
```
lesteve: I would use `mode='w', encoding='utf-8'` for better readability.

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -129,24 +139,32 @@ def test_pattern_matching():
...
+                   '\n'
+                   '      Óscar output\n'
+                   '    log:Óscar\n'
+                   '    $\\langle n_\\uparrow n_\\downarrow \\rangle$\n\n'
+                   )
     # create three files in tempdir (only one matches the pattern)
     fnames = ['plot_0.py', 'plot_1.py', 'plot_2.py']
     for fname in fnames:
-        with open(os.path.join(examples_dir, fname), 'w') as f:
+        with codecs.open(os.path.join(examples_dir, fname), 'w', 'utf-8') as f:
```
lesteve: Same remark as previously: `mode='w', encoding='utf-8'`

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -129,24 +139,32 @@ def test_pattern_matching():
...
+                         'r', 'utf-8') as f:
             rst = f.read()
-            if re.search(gallery_conf['filename_pattern'],
-                         os.path.join(gallery_dir, rst_fname)):
-                assert_true(code_output in rst)
-            else:
-                assert_false(code_output in rst)
+        if re.search(gallery_conf['filename_pattern'],
+                     os.path.join(gallery_dir, rst_fname)):
+            print(code_output)
```
lesteve: I am guessing this is some debugging print statements that you meant to remove.

doc/conf.py
```diff
@@ -291,7 +291,7 @@ def setup(app):
     # Do not pop up any mayavi windows while running the
     # examples. These are very annoying since they steal the focus.
     mlab.options.offscreen = True
-except ImportError:
```
lesteve: Probably this should not be part of this PR. I seem to remember that elsewhere there was a `except Exception`since importing some libraries can raise all sorts of exceptions (that's what the comment say).

I can change this directly in master.

lesteve:
My understanding is that your fix requires the example files to be encoded in utf-8. I am wondering how badly this assumption can backfire ...

My gut feeling is that this PR is a real improvement though.

Eric89GXL:
> My understanding is that your fix requires the example files to be encoded in utf-8. I am wondering how badly this assumption can backfire ...

Well ASCII or UTF-8, yeah. Previously they had to be ASCII only, so it is an improvement even if it doesn't make it universal.

doc/conf.py
Eric89GXL: I had to have it in order to test on my system, so yes please put it in `master`, I can open another PR, or just modify it here if you don't mind having the orthogonal change here

sphinx_gallery/gen_rst.py
Eric89GXL: nah, I had deleted the `sys.stdout` line at one point and just put it back in the wrong place

sphinx_gallery/gen_rst.py
Eric89GXL: I was hitting a problem where if an example raised an error at a certain point, *this* line then raised an error during handling (because the value didn't exist) so it never got to the proper `raise` line below. Again an orthogonal change I guess but I needed it to get things working properly during testing.

sphinx_gallery/gen_rst.py
Eric89GXL: I couldn't think of a better one either, but if someone else has an idea I'm happy to change it. I used the `My` prefix because eventually it goes to a variable named `my_buffer`.

.travis.yml
```diff
@@ -39,7 +39,7 @@ install:
         if [ "$PYTHON_VERSION" == "2.7" ]; then
           conda install --yes --quiet mayavi;
           conda upgrade --yes --all;
-          conda upgrade --yes pyface;
+          pip install --upgrade pyface;
```
Eric89GXL: This is also an orthogonal change, but was necessary to make the CIs happy...

Eric89GXL:
Comments addressed

Eric89GXL:
Moved the orthogonal `Exception` change to #107

sphinx_gallery/gen_rst.py
lesteve: OK maybe something to look at some point in the future.

sphinx_gallery/gen_rst.py
```diff
@@ -93,6 +97,18 @@ def flush(self):
...
+    """Helper to deal with unicode not having buffer interface on Py2.7"""
+    def write(self, data):
+        # "unicode" objects don't have a buffer interface on Py2.7,
+        # so we need to manually convert
+        if isinstance(data, unicode):
+            data = data.encode('utf-8')
+        super(MyBytesIO, self).write(data)
+
+    def getvalue(self):
+        return super(MyBytesIO, self).getvalue().decode('utf-8')
```
lesteve: This class is actually a bit weird since it derives from BytesIO but `.getvalue` returns a string.

It seems like this is working for me:

```python

class NonUnicodeFriendlyStringIO(StringIO):
    def write(self, data):
        if not isinstance(data, unicode):
            data = data.decode('utf-8')
        StringIO.write(self, data)
```

and then later:

```python
my_buffer = NonUnicodeFriendlyStringIO()
```

sphinx_gallery/gen_rst.py
lesteve: OK something to look at in the future. It'd be great if you could provide something that reproduces the problem.

sphinx_gallery/gen_rst.py
```diff
@@ -49,9 +52,9 @@ def prefixed_lines():
         return ''.join(prefixed_lines())
 
 try:
-    from StringIO import StringIO
+    from BytesIO import BytesIO
 except ImportError:
-    from io import StringIO
+    from io import BytesIO
```
lesteve: I think this can be replaced by `from io import StringIO` if you are using the NonUnicodeFriendlyStringIO class below.

I wasn't aware of this but StringIO.StringIO doesn't have the same behaviour as io.StringIO (both exist in Python 2.7 so it's not a Python 2 vs Python 3 import difference as I naively thought). For example you can both write unicode and non unicode strings into it but then you get into trouble when you are trying to do `.getvalue`:

```python
from StringIO import StringIO
s.write('é')
s.write(u'é')
s.getvalue()
```

```
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-7-67a6f2fbafbe> in <module>()
----> 1 s.getvalue()

/volatile/le243287/miniconda3/envs/py27/lib/python2.7/StringIO.pyc in getvalue(self)
    269         _complain_ifclosed(self.closed)
    270         if self.buflist:
--> 271             self.buf += ''.join(self.buflist)
    272             self.buflist = []
    273         return self.buf

UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 0: ordinal not in range(128)
```

Eric89GXL: Yeah, the nastiness of only getting the error when doing `.getvalue()` is what drove me to use `BytesIO` (it doesn't have this problem because you explicitly set the bytes yourself) but I'll change it to `StringIO`

sphinx_gallery/gen_rst.py
```diff
@@ -505,7 +527,8 @@ def execute_script(code_block, example_globals, image_path, fig_count,
         fig_count += 1  # raise count to avoid overwriting image
 
         # Breaks build on first example error
-
+        # XXX This check can break during testing e.g. if you uncomment the
+        # `raise RuntimeError` by the `my_stdout` call, maybe use `.get()`?
```
Eric89GXL: @lesteve here you go

Eric89GXL:
Changed the class and got rid of the `.get()` with a comment that should help the next dev take a look if they want

sphinx_gallery/gen_rst.py
lesteve: Thanks for this, we should probably fix it at one point.

lesteve:
OK LGTM, merging.

lesteve:
Thanks a lot for the fix!

Eric89GXL:
Thanks for the quick reviews

agramfort:
:beers: !

PR #112, opened 2016-03-25
Exclusion regex not working

adrn:
The exclusion pattern from the documentation was not working for me, so I added an explicit example to skip and updated the regex string.

### DIFF
--- a/doc/advanced_configuration.rst
+++ b/doc/advanced_configuration.rst
@@ -41,7 +41,7 @@ in your Sphinx ``conf.py``. For example:
 
 will build all examples starting with ``plot_compute_``. The key ``filename_pattern`` accepts
 `regular expressions`_ which will be matched with the full path of the example. This is the reason
-the leading ``'/'`` is required. Users are advised to use ``os.sep`` instead of ``'/'`` if 
+the leading ``'/'`` is required. Users are advised to use ``os.sep`` instead of ``'/'`` if
 they want to be agnostic to the operating system.
 
 This option is also useful if you want to build only a subset of the examples. For example, you may
@@ -71,7 +71,7 @@ starting with ``plot_long_examples_``, you would do:
 .. code-block:: python
 
     sphinx_gallery_conf = {
-        'filename_pattern' : '/plot_(?!long_examples)'
+        'filename_pattern' : '^((?!plot_long_examples_).)*$'
     }
 
 As the patterns are parsed as `regular expressions`_, users are advised to consult the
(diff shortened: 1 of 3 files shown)


doc/advanced_configuration.rst
```diff
@@ -41,7 +41,7 @@ in your Sphinx ``conf.py``. For example:
 
 will build all examples starting with ``plot_compute_``. The key ``filename_pattern`` accepts
 `regular expressions`_ which will be matched with the full path of the example. This is the reason
-the leading ``'/'`` is required. Users are advised to use ``os.sep`` instead of ``'/'`` if 
+the leading ``'/'`` is required. Users are advised to use ``os.sep`` instead of ``'/'`` if
```
adrn: Whoops, my editor just removed the whitespace here automatically.

Titan-C:
> The exclusion pattern from the documentation was not working for me, so I added an explicit example to skip and updated the regex string

What is what you aim to do? For backwards compatibility reasons we build in the gallery examples that start with plot. So your skip_ example should be already only parsed in to HTML without running.

adrn:
Oh, right. My use case was this: I have examples that don't make plots, but where I still want the output to be displayed. So I didn't want to name them with `plot_*`. I wanted to run and display output for *most* of my examples, except those that start with a certain name.

doc/advanced_configuration.rst
```diff
@@ -71,7 +71,7 @@ starting with ``plot_long_examples_``, you would do:
 .. code-block:: python
 
     sphinx_gallery_conf = {
-        'filename_pattern' : '/plot_(?!long_examples)'
+        'filename_pattern' : '^((?!plot_long_examples_).)*$'
```
lesteve: Looking at this old PR, it would actually make sense to have a simpler example along these lines indeed, i.e. just show a regex that skips example starting with `skip`. This is a more common use case that running the examples which start with `plot_` but not the ones starting with `plot_long_examples`. On top of that the doc by saying "For example, to skip building examples starting with ``plot_long_examples_``" is slightly misleading.

I would just modify the documentation. I would neither change conf.py nor add an additional examples that is skipped (we have one already: just_code.py).

lesteve: @adrn if you want to do that and finish off this PR, that would be great!

adrn: Will try to get to this soon -- things are a bit crazy right now.

larsoner:
@adrn our config docs have changed a bit, do you want to redo this? The easiest way might be to make a new PR

larsoner:
Closing due to non-response but feel free to open a new PR if you have time

Issue #110, opened 2016-03-25
Choose thumbnail image from multiple figures in same example


adrn:
Is there a way to choose the thumbnail image figure for examples that generate multiple figures?

Titan-C:
For the moment this feature is not available/designed. It is in our plan. Possible we might implement it by introducing comment flags on the file header to specify if a particular image is to be preferred.

lesteve:
This could work I guess. You would use some kind of number inside the example source code, right, something like:

```python
sphx_glr_fig_number = 3
```

If the variable is not found it would default to 0 to match the current behaviour.

Alternatively we could have a convention that if a variable with an agreed name and of the right type (some kind of matlplotlib figure) is found in the `__main__` namespace then it is used for the gallery thubmnail . I am not convinced it is better though, a bit more complicated and intrusive in the example code.

lesteve:
> You would use some kind of number inside the example source code, right, something like:

I think part of what I was trying to say is that just using a variable in the `__main__` namespace would be simpler than to a magical comment/header that you have to parse.

PR #114, opened 2016-03-26
Allow user to choose thumbnail image

adrn:
This implements a way for the example-writer to specify which figure is displayed as the thumbnail when an example script generates multiple figures.

This is a bit of a hack and may not make sense to merge in as is, but it's at least something to start discussion. The basic idea with this implementation is that inside the docstring of a given example script, you can add a key:value pair, e.g., `thumbnail: 3` that specifies which figure in a given example to display as the thumbnail. 

Like I said: this is a hack. Right now, this uses a regexp to pick out the "thumbnail:" block. A nicer way to move forward might be to think about how you want to represent module-level, arbitrary configuration data or metadata in example files. Then, this thumbnail tag could just fit in that way. Thoughts welcome!

cc @keflavich


adrn:
This is a trial implementation to solve #110

lesteve:
Using the docstring for this seems as good as other ideas that were floated to be honest ! Some tests are failing though, haven't looked into it in more details.

We would need some tests for the extract_thumbnail_number function.

If you want to cater for the more general case (although this is probably YAGNI unless you have something in mind already). You could use something like:

```python
"""
docstring

metadata={'thumbail_figure_number': 1}
"""
```

GaelVaroquaux:
I'd rather use a comment, because if we use the docstring, the info will
be printed in the text of the example.

lesteve:
> I'd rather use a comment, because if we use the docstring, the info will be printed in the text of the example.

Fair point actually for examples that do `print(__doc__)` ...

GaelVaroquaux:
> Fair point actually for examples that do print(__doc__) ...

But also because it's rendered in the html.

lesteve:
> But also because it's rendered in the html.

I think the `thumbnail: ` line is stripped off the docstring by this PR so it should not appear in the HTML.

adrn:
Yea, the line is just deleted from of the docstring before displaying with a regular expression. If we change the key name from `thumbnail` to something like `sphinx_gallery_thumbnail_number` then I think it'd be sufficiently specific that we can do that. I'm fine with putting it in a comment too though. I'll add some tests and fix the build error. Thanks for the comments!

lesteve:
@adrn have you managed to investigate the Travis errors? We chatted with @Titan-C and think it's better to put this in a comment that can be anywhere in the file even if that means having to read the file entirely once more. We favour something like (better naming suggestions welcome):

```
# sphinx_gallery_thumbnail_number = 3
```

examples/plot_choose-thumbnail.py
```diff
@@ -0,0 +1,39 @@
+# -*- coding: utf-8 -*-
```
lesteve: If you could name this file `plot_choose_thumbnail.py` with underscores, it would be more consistent.

adrn:
Cool, working on this now. Any thoughts on 0-indexing? e.g., should the default / first plot be `sphinx_gallery_thumbnail_number = 0` or `sphinx_gallery_thumbnail_number = 1`?

adrn:
Fixed the errors and made it search in a comment instead of the docstring as you suggested. Let me know if you think this needs anything else!

sphinx_gallery/gen_rst.py
```diff
@@ -231,6 +235,22 @@ def text2string(content):
         return content
 
 
+def extract_thumbnail_number(text):
+    """ Pull out the thumbnail image number specified in the docstring. """
+
+    # check whether the user has specified a specific thumbnail image
+    pattr = re.compile("^#\s*sphinx_gallery_thumbnail_number:\s*([0-9]+)\s*$", flags=re.MULTILINE)
```
lesteve: I would use = rather than :.

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -50,7 +50,7 @@
 def test_split_code_and_text_blocks():
     """Test if a known example file gets properly split"""
 
-    blocks = sg.split_code_and_text_blocks('examples/just_code.py')
+    blocks,_ = sg.split_code_and_text_blocks('examples/just_code.py')
```
lesteve: PEP8 wants a space after the comma. You should probably configure your editor to have flake8 checks, the investment in time is worth it.

sphinx_gallery/gen_rst.py
```diff
@@ -231,6 +235,22 @@ def text2string(content):
...
+    match = pattr.search(text)
+    try:
+        thumbnail_number = int(match.groups()[0])
+    except AttributeError:
+        thumbnail_number = 1 # by default, use the first figure created
+
+    # remove the thumbnail configuration option from the document
+    text = pattr.sub("", text)
```
lesteve: I am not sure I would bother removing the thumbnail configuration from the example source code to be honest. For example I would very well imagine my self doing something like:

```python

# The next line is to pick the figure showing the negative exponential in orange
# sphinx_gallery_thumbnail_number = 2
```

in which case removing the second line is slightly confusing.

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -185,6 +185,17 @@ def test_ipy_notebook():
...
+    # which plot to show as the thumbnail image
+    for test_str in ['# sphinx_gallery_thumbnail_number: 2',
+                     '# sphinx_gallery_thumbnail_number:2',
+                     '#sphinx_gallery_thumbnail_number: 2']:
+        with tempfile.NamedTemporaryFile('w') as f:
+            f.write('\n'.join(['"Docstring"',
+                               test_str]))
+            f.flush()
+            _,cfg = sg.split_code_and_text_blocks(f.name)
+        assert cfg['thumbnail_number'] == 2
```
lesteve: The error message is slightly better with `assert_equal(cfg['thumbnail_number'], 2)` from `nose.tools`.

sphinx_gallery/gen_rst.py
adrn: Really? this looks funny to me:

```
# sphinx_gallery_thumbnail_number = 2
# Code source: Óscar Nájera
# License: BSD 3 clause
```

sphinx_gallery/tests/test_gen_rst.py
adrn: I have pep8 checks in my editor, but I find this one annoying so it's disabled. I'll fix to be consistent with this repository.

sphinx_gallery/gen_rst.py
adrn: Ok

sphinx_gallery/gen_rst.py
lesteve: Hmmm then maybe move the sphinx_gallery_thumbnail_number it somewhere else in the file ;-)

sphinx_gallery/gen_rst.py
```diff
@@ -188,7 +188,11 @@ def split_code_and_text_blocks(source_file):
         List where each element is a tuple with the label ('text' or 'code'),
         and content string of block.
     """
+    config = dict()
+
     docstring, rest_of_content = get_docstring_and_rest(source_file)
+    rest_of_content, thumbnail_number = extract_thumbnail_number(rest_of_content)
```
lesteve: You don't change rest_of_content anymore so you only need to return the thumbnail number from `extract_thumbnail_number`.

sphinx_gallery/gen_rst.py
```diff
@@ -231,6 +235,19 @@ def text2string(content):
...
+
+    # check whether the user has specified a specific thumbnail image
+    pattr = re.compile("^\s*#\s*sphinx_gallery_thumbnail_number\s*=\s*([0-9]+)\s*$", flags=re.MULTILINE)
+    match = pattr.search(text)
+    try:
+        thumbnail_number = int(match.groups()[0])
+    except AttributeError:
+        thumbnail_number = 1 # by default, use the first figure created
+
+    return thumbnail_number
```
Titan-C: I would rather write this function as `extract_example_configuration` and let it return a dictionary or named tuple. Then move all the config dictionary declaration inside here so that `split_code_and_text_blocks` remains cleaner. Also I would rename `config` to `example_config` or similar just to give it a bit more contex when reading the source code.

lesteve: Since we are only using the thumbnail_number and nobody seemed to have any suggestions of other per-example metadata that could be useful, I would be in favour of returning just the thumbnail number for YAGNI reasons.

Maybe worth investigating if it's easy to keep the thumbnail number out of `split_code_and_text_blocks` indeed.

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -185,6 +185,18 @@ def test_ipy_notebook():
...
+    # which plot to show as the thumbnail image
+    for test_str in ['# sphinx_gallery_thumbnail_number= 2',
+                     '# sphinx_gallery_thumbnail_number=2',
+                     '#sphinx_gallery_thumbnail_number = 2',
+                     '    # sphinx_gallery_thumbnail_number=2']:
+        with tempfile.NamedTemporaryFile('w') as f:
+            f.write('\n'.join(['"Docstring"',
+                               test_str]))
+            f.flush()
+            _,cfg = sg.split_code_and_text_blocks(f.name)
```
lesteve: Another PEP8 with space here.

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -77,7 +77,7 @@ def test_direct_comment_after_docstring():
                            'x, y = 1, 2',
                            '']))
         f.flush()
-        result = sg.split_code_and_text_blocks(f.name)
+        result,_ = sg.split_code_and_text_blocks(f.name)
```
lesteve: PEP8

sphinx_gallery/gen_rst.py
```diff
@@ -392,9 +409,9 @@ def scale_image(in_fname, out_fname, max_width, max_height):
                           generated images')
 
 
-def save_thumbnail(image_path, base_image_name, gallery_conf):
+def save_thumbnail(image_path, base_image_name, gallery_conf, thumb_num=1):
     """Save the thumbnail image"""
-    first_image_file = image_path.format(1)
+    first_image_file = image_path.format(thumb_num)
```
lesteve: You should rename the variable so that the thumbnail figure is not necessarily the first one.

sphinx_gallery/gen_rst.py
adrn: OK, I removed the config stuff entirely for now and moved the thumbnail stuff out of `split_code_and_text_blocks` (it was in there initially so it could remove the comment from the docstring)

adrn:
Could you restart the travis build?

lesteve:
> Could you restart the travis build?

You can do `git commit --amend` and then force-push to trigger a new build.

adrn:
It's a single button to package maintainers -- in other projects I contribute to, that is the desired way to trigger new builds...but ok, I can force-push if that is what you prefer here

Titan-C:
looks good to me
Could you please write  a line in CHANGES.rst, and a documentation section in doc/advanced_configuration.rst narrating how to use this feature and the link to the example using it.

sphinx_gallery/gen_rst.py
```diff
@@ -188,8 +188,8 @@ def split_code_and_text_blocks(source_file):
         List where each element is a tuple with the label ('text' or 'code'),
         and content string of block.
     """
-    docstring, rest_of_content = get_docstring_and_rest(source_file)
 
+    docstring, rest_of_content = get_docstring_and_rest(source_file)
```
lesteve: Avoiding unnecessary changes like this one is always appreciated.

examples/plot_choose_thumbnail.py
```diff
@@ -0,0 +1,38 @@
...
+def main():
+    x = np.linspace(-1, 2, 100)
+    y = np.exp(x)
+
+    plt.figure()
+    plt.plot(x, y)
+    plt.xlabel('$x$')
+    plt.ylabel('$\exp(x)$')
+
+    # The next line is to pick the figure showing the negative exponential in orange
```
lesteve: Maybe "The next line is to show the second figure in the gallery (plot with negative exponential in orange)"

sphinx_gallery/gen_rst.py
```diff
@@ -231,6 +231,19 @@ def text2string(content):
...
 
 
+def extract_thumbnail_number(text):
+    """ Pull out the thumbnail image number specified in the docstring. """
+
+    # check whether the user has specified a specific thumbnail image
+    pattr = re.compile("^\s*#\s*sphinx_gallery_thumbnail_number\s*=\s*([0-9]+)\s*$", flags=re.MULTILINE)
+    match = pattr.search(text)
+    try:
+        thumbnail_number = int(match.groups()[0])
```
lesteve: I would do:
```python
thumbnail_number = 1 if match is None else int(match.groups()[0])
```

I am not convinced asking for forgiveness rather than permission is better in this particular case.

sphinx_gallery/gen_rst.py
```diff
@@ -392,18 +405,18 @@ def scale_image(in_fname, out_fname, max_width, max_height):
                           generated images')
 
 
-def save_thumbnail(image_path, base_image_name, gallery_conf):
+def save_thumbnail(image_path, base_image_name, gallery_conf, thumb_num=1):
```
lesteve: The signature of the function is not great. It would be better if image_path was already the full path  of the image to be used for the thumbnail (and didn't need a .format).

Basically I would do:
```python
image_thumbnail_path = image_path.format(thumbnail_number)
```

in generate_file_rst.

image_path is not a great name, I would use something like image_path_template ... Bonus points if you fix that while you are touching this code.

adrn:
@Titan-C Added documentation and addressed comments from @lesteve

doc/advanced_configuration.rst
```diff
@@ -203,6 +203,23 @@ file. You need to add to the configuration dictionary a key called
         'default_thumb_file'     : 'path/to/thumb/file.png'}}
 
 
+choosing the thumbnail image from multiple figures
```
lesteve: Capital needed for the title

doc/advanced_configuration.rst
```diff
@@ -203,6 +203,23 @@ file. You need to add to the configuration dictionary a key called
...
+For examples that generate multiple figures, the default behavior will use
+the first figure created in each as the thumbnail image displayed in the
+gallery. To change the thumbnail image to a figure generated later in
+an example script, add a comment to the example script to specify the
+number of the figure you would like to use as the thumbnail. For example,
+to use the 2nd figure created as the thumbnail:
+
+.. code-block:: python
+
+    sphinx_gallery_thumbnail_number = 2
```
lesteve: Maybe add a # in front of this line to make it clear that it is added as a comment ?

examples/plot_choose_thumbnail.py
```diff
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 """
+.. _choose-thumbnail-example:
```
lesteve: You don't need to do that, references are created automatically for examples. Don't remember the exact reference naming though, you'll have to browse the doc and grep for something that contains `sphx_glr`

sphinx_gallery/gen_rst.py
```diff
@@ -237,10 +236,11 @@ def extract_thumbnail_number(text):
     # check whether the user has specified a specific thumbnail image
     pattr = re.compile("^\s*#\s*sphinx_gallery_thumbnail_number\s*=\s*([0-9]+)\s*$", flags=re.MULTILINE)
     match = pattr.search(text)
-    try:
+
+    if match is None: # by default, use the first figure created
```
lesteve: I you want to use an inline comment, PEP8 requires two spaces before the #.

Personally I am not a big fan of inline comments and in this particular case, I think it would make more sense to put it on a line by itself above `thumbnail_number = 1`

doc/advanced_configuration.rst
```diff
@@ -203,6 +203,23 @@ file. You need to add to the configuration dictionary a key called
...
+
+The default behavior is ``sphinx_gallery_thumbnail_number = 1``. See
```
lesteve: I would remove "The default behavior is ``sphinx_gallery_thumbnail_number = 1``" since you already mention that the default is to pick the first figure above.

adrn: Said in words and said in code. I disagree.

examples/plot_choose_thumbnail.py
lesteve: Quickly grepping, it seems like you want to remove this line and use `:ref:sphx_glr_auto_examples_plot_choose_thumbnail.py` where you need to reference this example.

lesteve:
LGTM, merging, thanks a lot for this, in particular tackling all the minor comments that quickly !

adrn:
:+1: thanks for reviewing!

PR #116, opened 2016-03-31
Fix indexing with mayavi figures.

jaeilepp:
There is an issue with indexing mayavi figures. This manifests itself in thumbnail figures not being rendered here https://circle-artifacts.com/gh/mne-tools/mne-python/621/artifacts/0/home/ubuntu/mne-python/doc/_build/html/auto_examples/index.html
See  for example ``plot all-to-all connectivity in sensor space``.
I'm not sure if this is the correct fix, but it seems to at least work around the issue.
cc @agramfort

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -334,7 +334,7 @@ def save_figures(image_path, fig_count, gallery_conf):
         e = mlab.get_engine()
         last_matplotlib_fig_num = len(figure_list)
         total_fig_num = last_matplotlib_fig_num + len(e.scenes)
-        mayavi_fig_nums = range(last_matplotlib_fig_num, total_fig_num)
+        mayavi_fig_nums = range(last_matplotlib_fig_num + 1, total_fig_num + 1)
 
         for scene, mayavi_fig_num in zip(e.scenes, mayavi_fig_nums):
             current_fig = image_path.format(mayavi_fig_num)


lesteve:
I know very little about mayavi but it looks like you are right, there is a off-by-one problem.
For example, the example "Compute point-spread functions (PSFs) for MNE/dSPM/sLORETA" shows the second image as the gallery thumbnail instead of the first one. When you have only one figure I guess this off-by-one issue cause the figure not to be rendered in the gallery.

I can reproduce the issue locally as well on the mayavi_examples/plot_3d.py example. This PR fixes it AFAICT.

Also I noticed another possible issue, it seems that for Notebook-style examples e.g. this [one](https://circle-artifacts.com/gh/mne-tools/mne-python/621/artifacts/0/home/ubuntu/mne-python/doc/_build/html/auto_examples/inverse/plot_lcmv_beamformer.html#sphx-glr-auto-examples-inverse-plot-lcmv-beamformer-py) the mayavi figure rendered in the example HTML are exactly the same. Is that expected?

agramfort:
thx @jaeilepp for looking into this.

yes @lesteve it's weird that it appears twice...

jaeilepp:
I believe the duplicate appears because of this indexing problem. The example plots a mpl figure first (index = 1) and after that, it plots a mayavi figure (index = 1 + 0) so both of the figures get a suffix 001.

agramfort:
@lesteve @Titan-C can I merge?

lesteve:
Let's merge this one. I'll double-check whether that fixes the problem with identical figures on notebook-styled examples.

lesteve:
For the record, this PR fixes the identical figure problem. It had nothing to do with Notebook-styled examples by the way. It happens when there is a mix of matplotlib and mayavi figures. The example I used to reproduce the problem:

```python

plt.plot([1, 2, 3], [1, 2, 3])

from mayavi import mlab

mlab.test_plot3d()

mlab.figure()
mlab.test_contour3d()
```

Output before PR:
bad

Output after PR:
good

lesteve:
Thanks a lot for the fix @jaeilepp!

lesteve:
It'd be great to add a test by the way, if that's not too much of a pain ...

PR #117, opened 2016-04-07
Fix to mayavi figure indexing.

jaeilepp:
I noticed that when you build tutorials like this https://circle-artifacts.com/gh/mne-tools/mne-python/645/artifacts/0/home/ubuntu/mne-python/doc/_build/html/auto_tutorials/plot_source_localization_basics.html, function ``save_figures`` gets called multiple times (notice the duplicates of the mayavi figure). Therefore the fix in https://github.com/sphinx-gallery/sphinx-gallery/pull/116 alone won't work. I tried this locally on couple of my examples and tutorials and it seems to fix the issue.

### DIFF
--- a/sphinx_gallery/tests/test_gen_rst.py
+++ b/sphinx_gallery/tests/test_gen_rst.py
@@ -184,9 +184,12 @@ def test_ipy_notebook():
 
 
 def test_save_figures():
-    """Test file naming when saving figures."""
+    """Test file naming when saving figures. Requires mayavi."""
     import matplotlib.pylab as plt
-    from mayavi import mlab
+    try:
+        from mayavi import mlab
+    except ImportError:
+        return  # Skip if mayavi is not available.
     examples_dir = tempfile.mkdtemp()
     gallery_conf = {
         'examples_dirs': examples_dir,
--- a/sphinx_gallery/tests/test_gen_rst.py
+++ b/sphinx_gallery/tests/test_gen_rst.py
@@ -15,10 +15,10 @@ import os
 import nose
 import shutil
 from nose.tools import assert_equal, assert_false, assert_true
-import matplotlib.pylab as plt
+
 import sphinx_gallery.gen_rst as sg
 from sphinx_gallery import notebook
-
+import matplotlib.pylab as plt  # Import gen_rst first to enable 'Agg' backend.
 
 CONTENT = [
     '"""'
--- a/sphinx_gallery/tests/test_gen_rst.py
+++ b/sphinx_gallery/tests/test_gen_rst.py
@@ -12,7 +12,10 @@ import json
 import tempfile
 import re
 import os
+import nose
+import shutil
 from nose.tools import assert_equal, assert_false, assert_true
+import matplotlib.pylab as plt
 import sphinx_gallery.gen_rst as sg
 from sphinx_gallery import notebook
 
@@ -185,32 +188,27 @@ def test_ipy_notebook():
 
 def test_save_figures():
     """Test file naming when saving figures. Requires mayavi."""
-    import matplotlib.pylab as plt
     try:
         from mayavi import mlab
     except ImportError:
-        return  # Skip if mayavi is not available.
+        raise nose.SkipTest('Mayavi not installed')
+    mlab.options.offscreen = True
     examples_dir = tempfile.mkdtemp()
-    gallery_conf = {
-        'examples_dirs': examples_dir,
-        'plot_gallery': True,
-        'find_mayavi_figures': True,
-    }
+
+    gallery_conf = {'find_mayavi_figures': True}
     mlab.test_plot3d()
     plt.plot(1, 1)
-    fig_list = sg.save_figures(examples_dir + '/image{0}.png', 0, gallery_conf)
+    fname_template = os.path.join(examples_dir, 'image{0}.png')
+    fig_list = sg.save_figures(fname_template, 0, gallery_conf)
     assert_equal(len(fig_list), 2)
     assert fig_list[0].endswith('image1.png')
     assert fig_list[1].endswith('image2.png')
-    for fig in fig_list:
-        os.remove(fig)
+
     mlab.test_plot3d()
     plt.plot(1, 1)
-    fig_list = sg.save_figures(examples_dir + '/image{0}.png', 2, gallery_conf)
+    fig_list = sg.save_figures(fname_template, 2, gallery_conf)
     assert_equal(len(fig_list), 2)
     assert fig_list[0].endswith('image3.png')
     assert fig_list[1].endswith('image4.png')
-    plt.close('all')
-    for fig in fig_list:
-        os.remove(fig)
-    os.rmdir(examples_dir)
+
+    shutil.rmtree(examples_dir)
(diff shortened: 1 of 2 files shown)


agramfort:
looks legit to me

thx @jaeilepp

agramfort:
wait I don't see the dupes in
https://circle-artifacts.com/gh/mne-tools/mne-python/645/artifacts/0/home/ubuntu/mne-python/doc/_build/html/auto_tutorials/plot_source_localization_basics.html
?

did you fit it already?

there is both the fsaverage and the sample brains

jaeilepp:
The first and the third figure are the same. Basically the mpl plots work fine and the mayavi figures link correctly, but the indexing is messed up. Therefore the first mpl figure gets overwritten with mayavi figure.

Titan-C:
This looks good to me. In deed one needs to include the total counting.

lesteve:
Looks good to me but same remark as last time : it would be really great to add some kind of tests for this if that's not too much of a pain.

@Titan-C by the way I was wondering how come the mayavi gallery is not included in the readthedocs ? Is it because you only can use a pip requirement file to install the dependencies ?

lesteve:
> Looks good to me but same remark as last time : it would be really great to add some kind of tests for this if that's not too much of a pain.

At the very least we could modify the plot_3d mayavi example in order to have a few cells and a mix of matplotlib/mayavi plots inside a given cell.

Titan-C:
> @Titan-C by the way I was wondering how come the mayavi gallery is not included in the readthedocs ? Is it because you only can use a pip requirement file to install the dependencies ?

This was more of a statement from @GaelVaroquaux. That we shall not put mayavi in the requirements.txt file, as having a working mayavi environment should be a responsibility of the user and not a dependency of the gallery. Also mayavi only works in python 2.7(from conda as far as I know), and the gallery has to work on python 2.6, the ubuntu 12.10 python 2.7 and python 3 and none of those can run mayavi.

lesteve:
It looks like readthedocs supports conda environments:
http://docs.readthedocs.org/en/latest/conda.html#conda-support

Maybe worth a try ?

jaeilepp:
Okay. Test added. I'm not too familiar with the code base outside ``save_figures`` so I'm just testing the naming. And it only gets tested on the environment with mayavi, but I guess that's okay.

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -181,3 +181,36 @@ def test_ipy_notebook():
 
         f.flush()
         assert_equal(json.load(f), example_nb.work_notebook)
+
+
+def test_save_figures():
+    """Test file naming when saving figures. Requires mayavi."""
+    import matplotlib.pylab as plt
```
lesteve: Move this import to the top of the file

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -181,3 +181,36 @@ def test_ipy_notebook():
...
+    try:
+        from mayavi import mlab
+    except ImportError:
+        return  # Skip if mayavi is not available.
```
lesteve: The best is to import nose at the top of the file and use `raise nose.SkipTest('Mayavi not installed')`

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -181,3 +181,36 @@ def test_ipy_notebook():
 
         f.flush()
         assert_equal(json.load(f), example_nb.work_notebook)
+
+
+def test_save_figures():
+    """Test file naming when saving figures. Requires mayavi."""
+    import matplotlib.pylab as plt
+    try:
+        from mayavi import mlab
```
lesteve: You probably want to use: `mlab.options.offscreen = True` to prevent plots mayavi plots poping up when running the tests.

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -181,3 +181,36 @@ def test_ipy_notebook():
...
+        return  # Skip if mayavi is not available.
+    examples_dir = tempfile.mkdtemp()
+    gallery_conf = {
+        'examples_dirs': examples_dir,
+        'plot_gallery': True,
+        'find_mayavi_figures': True,
+    }
+    mlab.test_plot3d()
+    plt.plot(1, 1)
+    fig_list = sg.save_figures(examples_dir + '/image{0}.png', 0, gallery_conf)
```
lesteve: Use something like:
```python
img_filename_template = os.path.join(examples_dir, 'image{0}.png')
fig_list = sg.save_figures(img_filename_template, ...)
```

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -181,3 +181,36 @@ def test_ipy_notebook():
...
+def test_save_figures():
+    """Test file naming when saving figures. Requires mayavi."""
+    import matplotlib.pylab as plt
+    try:
+        from mayavi import mlab
+    except ImportError:
+        return  # Skip if mayavi is not available.
+    examples_dir = tempfile.mkdtemp()
+    gallery_conf = {
+        'examples_dirs': examples_dir,
```
lesteve: From a quick look at the code you only need:
```python
gallery_conf = {'find_mayavi_figures': True}
```

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -181,3 +181,36 @@ def test_ipy_notebook():
...
+        'plot_gallery': True,
+        'find_mayavi_figures': True,
+    }
+    mlab.test_plot3d()
+    plt.plot(1, 1)
+    fig_list = sg.save_figures(examples_dir + '/image{0}.png', 0, gallery_conf)
+    assert_equal(len(fig_list), 2)
+    assert fig_list[0].endswith('image1.png')
+    assert fig_list[1].endswith('image2.png')
+    for fig in fig_list:
```
lesteve: Is it necessary to remove the figure files ? If it is so that the `os.rmdir` works (because the directory has to be empty) I would just use `shutil.rmtree` instead

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -181,3 +181,36 @@ def test_ipy_notebook():
...
+        os.remove(fig)
+    mlab.test_plot3d()
+    plt.plot(1, 1)
+    fig_list = sg.save_figures(examples_dir + '/image{0}.png', 2, gallery_conf)
+    assert_equal(len(fig_list), 2)
+    assert fig_list[0].endswith('image3.png')
+    assert fig_list[1].endswith('image4.png')
+    plt.close('all')
```
lesteve: Do you need this? I think since gen_rst is imported the matplotlib backend should be set to Agg and it isn't needed.

lesteve:
Your test looks neat and I have double-checked that it fails under master.

sphinx_gallery/tests/test_gen_rst.py
lesteve: For the matplotlib backend to be set to Agg correctly you probably need the matplotlib.pyplot import to be after the gen_rst one though.

lesteve:
Merging, thanks a lot for the fix + test !

GaelVaroquaux:
It would be great to have the same kind of feature to order directories with an entry in the README files.

lesteve:
:confused: hmm did you mean to comment in this issue? I am guessing you want to be able to order subsections of the gallery whereas it is done by alphabetical order at the moment. Not sure how this is related to mayavi figure indexing though.

GaelVaroquaux:
Good point. Wrong issue.

PR #118, opened 2016-04-21
Code splitter in comment header support

Titan-C:
Splitting out of the exit codes PR. This fixes the formatting of notebooks styled examples that have the structure
```
#########
# Header
#########
```


sphinx_gallery/tests/reference_parse.txt
```diff
@@ -10,13 +10,15 @@
  ('code',
   '\ndef dummy():\n    """This should not be part of a \'text\' block\'"""\n\n    ######################################\n    # Comment inside code to remain here\n    pass\n\n# this should not be part of a \'text\' block\n\n'),
  ('text',
-  '\n####################################################################\n\nMaking a line cut in sphinx\n'),
```
lesteve: The lstrip is removing the newline here. I am wondering whether it could it have an unintended consequence ... can you see any?

Titan-C: I don't see any consequence for now. Still, as a cautionary meassure I just changed the commit to make sure that all blocks end with at least one newline so that we don't have the this line break giving unwanted behavior.

lesteve:
LGTM merging, thanks !

Issue #115, opened 2016-03-28
Is it possible to use sphinx-gallery with a separate source directory?


emmanuelle:
First of all, many thanks to the developers for this great project!

I'm currently trying to use ``sphinx-gallery`` for the examples gallery of ``scikit-image`` (http://scikit-image.org/docs/dev/auto_examples/), so that ``scikit-image`` can benefit from the developments made in ``sphinx-gallery`` in the future. 

However, I'm facing some problems with relative paths (that I detail below), and I'm wondering if some developers/users have already successfully used ``sphinx-gallery`` for projects that have a separate ``source`` directory (when you answer "yes" to the following question in ``sphinx-quickstart`` 
" > Separate source and build directories (y/n) [n]: y")

I tried to reproduce the problem with a simple example, generated with ``sphinx-quickstart``. I have this doc directory with the following structure
```
.>>> tree
├── build
├── make.bat
├── Makefile
└── source
    ├── conf.py
    ├── conf.py~
    ├── examples
    │   ├── plot_adapt_rgb.py
    │   └── README.txt
    ├── index.rst
    ├── _static
    └── _templates

```
And when I run ``make html``, the following structure is generated (clipping some lines)
```
├── build
│   ├── doctrees
│   │   ├── auto_examples
│   │   │   ├── index.doctree
│   │   │   └── plot_adapt_rgb.doctree
│   │   ├── environment.pickle
│   │   └── index.doctree
│   └── html
│       ├── auto_examples
│       │   ├── index.html
│       │   └── plot_adapt_rgb.html
│       ├── _downloads
│       │   ├── plot_adapt_rgb.ipynb
│       │   └── plot_adapt_rgb.py
│       ├── genindex.html
│       ├── index.html
│       ├── objects.inv
│       ├── search.html
│       ├── searchindex.js
│       ├── _sources
│       │   ├── auto_examples
│       │   │   ├── index.txt
│       │   │   └── plot_adapt_rgb.txt
│       │   └── index.txt
│       
├── make.bat
├── Makefile
└── source
    ├── auto_examples
    │   ├── images
    │   │   ├── sphx_glr_plot_adapt_rgb_001.png
    │   │   ├── sphx_glr_plot_adapt_rgb_002.png
    │   │   └── thumb
    │   │       └── sphx_glr_plot_adapt_rgb_thumb.png
    │   ├── index.rst
    │   ├── plot_adapt_rgb_codeobj.pickle
[...]

PR #119, opened 2016-05-08
Modified path in gen_gallery_rst to solve sphinx config problem

emmanuelle:
Now it should work with the two possible configurations of sphinx (with
or without a source directory).

See #115.


lesteve:
I am going to merge this one, since @Titan-C said he didn't see any problem in the work-around in #115.

PR #122, opened 2016-06-03
ENH: make it possible to also specify abort_on_example_error in the sphinx_gallery_conf

astrofrog:
With these changes, it is possible to specify ``abort_on_example_error`` inside the normal ``sphinx_gallery_conf`` options.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,9 @@ New features
 * Print aggregated and sorted list of computation times of all examples
   in the console during the build.
 * For examples that create multiple figures, set the thumbnail image.
-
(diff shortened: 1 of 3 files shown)


lesteve:
I think this change is already in #97. @Titan-C can you double-check?

lesteve:
Actually quickly looking at it, it doesn't seem to work as expected in #97. The value in `app.builder.config.abort_on_example_error` is always going to win over `sphinx_gallery_conf`. The differences between the values in sphinx_gallery_conf and the ones in `app.builder.config`are there for historic reasons and we should try to clean this up.

Issue #123, opened 2016-06-06
unable to install on mac osx due to unicode in setup.py and README.rst


jforbess:
I am running python 3.5 on Mac OSX 10.9 and encountered this error when trying to install: 

```
Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/var/folders/ln/2s73prtj7lldhxfqjphnfnqc0000gp/T/pip-build-5az5c6_c/sphinx-gallery/setup.py", line 12, in <module>
        long_description = f.read()
      File "/Users/jforbess/Code/python/anaconda3/lib/python3.5/encodings/ascii.py", line 26, in decode
        return codecs.ascii_decode(input, self.errors)[0]
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 1399: ordinal not in range(128)

```
The unicode issues were in the long dashes used to show the project tree diagram in README.rst and the author's name in setup.py. I don't understand why mac osx has all of these errors in unicode/ascii decoding but I have seen them in a few different places recently. I solved this for myself by removing that text, but it would be helpful for others if this could be solved upstream.

lesteve:
Thanks a lot for the bug report, I am guessing somehow your locale is set to ascii so that `open` uses ascii by default.

We should still fix this in sphinx-gallery by using `codecs.open` with `encoding='utf-8'` as we do in other places. If you have the bandwith to open a PR, it would be more than welcome!

PR #124, opened 2016-06-07
Specify utf-8 encoding in file open

jforbess:
Fixing issue #123 to explicitly open README.rst as utf-8.


lesteve:
Thanks for your PR, as you can see Travis has some failures for Python 2.7. The reason is that `open` does not have a `encoding` keyword in Python 2, which is why I recommended to use codecs.open in the associated issue.

Another remark: it is good practice to create a branch (i.e. not use master) when you do a PR.

jforbess:
Sorry, I'm clearly a newbie. I will address next week. 

Sent on the move.

lesteve:
> Sorry, I'm clearly a newbie. I will address next week.

No worries, just ping us when you think you are done and Travis is green.

PR #130, opened 2016-07-10
UTF-8 in setup.py

### LINKED ISSUE #123: unable to install on mac osx due to unicode in setup.py and README.rst

Titan-C:
This should fix #123 and close #124 

@jforbess is this working for you?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,10 +5,11 @@
 Installer Sphinx extension for gallery generator
 """
 
+import codecs
 from setuptools import setup, find_packages
 import sphinx_gallery
 
-with open('README.rst') as f:
+with codecs.open('README.rst', mode='r', encoding='utf-8') as f:
     long_description = f.read()
 
 


jforbess:
Yes, thanks. Sorry I didn't follow up on such an easy fix. Summer.

PR #131, opened 2016-07-10
Include a reference for each gallery subsection

### LINKED ISSUE #127: Section labels should be inserted in the index.rst of the gallery
Right now, it is not possible to point to the different sections of the gallery using the :ref:`label` syntax of sphinx.

Sphinx-gallery should insert section labels in the index.rst. It should probably prefix the names with sphx_glr, as it does in the examples.

Titan-C:
When describing a gallery subsection or the main gallery one writes
in the README.txt file which supports RestructuredText syntax and one
should put references for the headings.

But because one can forget an the seems to be no conflict in Sphinx if a
header has to references Sphinx-Gallery prepends an automatically
generated label for each subsection

Fix #127

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -284,9 +284,12 @@ def save_thumbnail(image_path_template, src_file, file_conf, gallery_conf):
 def generate_dir_rst(src_dir, target_dir, gallery_conf, seen_backrefs):
     """Generate the gallery reStructuredText for an example directory"""
 
+    fhindex = """\n\n.. _sphx_glr_{0}:\n\n""".format(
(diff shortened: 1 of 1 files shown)


lesteve:
Could we have this feature included in the doc with a reference that points to a subsection of the gallery examples.

GaelVaroquaux:
I am +1 for merge.

@Titan-C : there is a trivial conflict here. Would you mind fixing it, and we merge?

Titan-C:
do not merge, the proposed new reference is on full path on disk. Not usefull

Titan-C:
should be fixed now

GaelVaroquaux:
Gr8. Merging!

Issue #128, opened 2016-06-12
Create a zip file with all examples as Python scripts and notebooks


GaelVaroquaux:
It would be useful to have a zip file accessible from the top of the gallery containing all examples as Python scripts and notebooks.

This zip file should be inserted in the index.rst using the download sphinx command (http://www.sphinx-doc.org/en/stable/markup/inline.html#referencing-downloadable-files)

banilo:
+1 That would be very helpful IMHO, not only for workshops, but also individual users.

Titan-C:
closed with #132

PR #132, opened 2016-07-10
[MRG] Script downloads

Titan-C:
To answer #128 I propose this to download all example files per section in python and Jupyter notebooks

I'm still unsure of the presentation. If the download should be in the yellow boxes we use in the files or a simpler plain text. Also if it would be at the start of the example section or at the end.

spectacle d10639

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@ git master
 New features
 ''''''''''''
 * Enhanced CSS for download buttons
+* Download buttons at the end of each section in the gallery to
+  download all scripts or Jupyter notebooks together in a zip file.
 
 Bug Fixes
 '''''''''
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,10 @@ New features
 ''''''''''''
 * Enhanced CSS for download buttons
 * Download buttons at the end of each section in the gallery to
-  download all scripts or Jupyter notebooks together in a zip file.
+  download all python scripts or Jupyter notebooks together in a zip
+  file. New config variable `download_section_examples` to toggle this
+  effect. Activated by default
+
 
 Bug Fixes
 '''''''''
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,10 +7,10 @@ git master
 New features
 ''''''''''''
 * Enhanced CSS for download buttons
-* Download buttons at the end of each section in the gallery to
-  download all python scripts or Jupyter notebooks together in a zip
-  file. New config variable `download_section_examples` to toggle this
-  effect. Activated by default
+* Download buttons at the end of the gallery to download all python
+  scripts or Jupyter notebooks together in a zip file. New config
+  variable `download_all_examples` to toggle this effect. Activated by
+  default
 
 
 Bug Fixes
--- a/sphinx_gallery/downloads.py
+++ b/sphinx_gallery/downloads.py
@@ -37,9 +37,9 @@ def python_zip(file_list, gallery_path, extension='.py'):
 
     Parameters
     ----------
-    file_list : list of str
+    file_list : list of strings
         Holds all the file names to be included in zip file
-    gallery_path : str
+    gallery_path : string
         path to where the zipfile is stored
     extension : str
         '.py' or '.ipynb' In order to deal with downloads of python
@@ -48,9 +48,10 @@ def python_zip(file_list, gallery_path, extension='.py'):
         variable while generating the zip file
     Returns
     -------
-    str : zip file name, written as `target_dir_{python,jupyter}.zip`
+    zipname : string
+        zip file name, written as `target_dir_{python,jupyter}.zip`
         depending on the extension
-"""
+    """
     zipname = gallery_path.replace(os.path.sep, '_')
     zipname += '_python' if extension == '.py' else '_jupyter'
     zipname = os.path.join(gallery_path, zipname + '.zip')
@@ -65,7 +66,17 @@ def python_zip(file_list, gallery_path, extension='.py'):
 
 
 def list_downloadable_sources(target_dir):
-    """Returns a list of python source files is target_dir"""
+    """Returns a list of python source files is target_dir
+
+    Parameters
+    ----------
+    target_dir : string
+        path to the directory where python source file are
+    Returns
+    -------
+    list
+        list of paths to all Python source files in `target_dir`
+    """
     return [os.path.join(target_dir, fname)
             for fname in os.listdir(target_dir)
             if fname.endswith('.py')]
@@ -78,12 +89,12 @@ def generate_zipfiles(gallery_dir):
 
     Parameters
     ----------
-    gallery_dir : str
+    gallery_dir : string
         path of the gallery to collect downloadable sources
 
     Return
     ------
-    download_rst: str
+    download_rst: string
         RestructuredText to include download buttons to the generated files
     """
 
--- a/sphinx_gallery/downloads.py
+++ b/sphinx_gallery/downloads.py
@@ -12,15 +12,14 @@ from __future__ import absolute_import, division, print_function
 import os
 import zipfile
 
-CODE_DOWNLOAD = """**Total running time of the script:**
-({0:.0f} minutes {1:.3f} seconds)\n\n
+CODE_DOWNLOAD = """
 \n.. container:: sphx-glr-download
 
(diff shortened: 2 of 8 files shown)


Titan-C:
I also need a test for this. Any ideas?

Titan-C:
Since #138 download button are so nice I'll put the same buttons at the end of each section to download the complete zip file of example scripts.
I still need a test

sphinx_gallery/gen_rst.py
```diff
@@ -444,6 +454,37 @@ def save_thumbnail(image_path_template, src_file, gallery_conf):
...
+        variable while generating the zip file
+    Returns
+    -------
+    void : zip file is written as `target_dir_{python,jupyter}.zip`
+    depending on the extension
+"""
+    zipname = '_python' if extension == '.py' else '_jupyter'
+    zipname = target_dir + zipname + '.zip'
+
+    zipf = zipfile.ZipFile(zipname, mode='w')
```
GaelVaroquaux: I suspect that, to work reliably on all OSes, it's useful to use "mode='wb'".

GaelVaroquaux: I would use a with statement: "with ZipFile(...) as zipf:". The benefit is that if there is an exception, the file is closed. The drawback is that it only works in Python 2.7 and later.

sphinx_gallery/gen_rst.py
```diff
@@ -444,6 +454,37 @@ def save_thumbnail(image_path_template, src_file, gallery_conf):
...
+    extension : str
+        '.py' or '.ipynb' In order to deal with downloads of python
+        sources and jupyter notebooks the file extension from files in
+        file_list will be removed and replace with the value of this
+        variable while generating the zip file
+    Returns
+    -------
+    void : zip file is written as `target_dir_{python,jupyter}.zip`
+    depending on the extension
+"""
```
GaelVaroquaux: Style: I think that this should be indented.

sphinx_gallery/gen_rst.py
Titan-C: > I would use a with statement: "with ZipFile(...) as zipf:". The benefit is that if there is an exception, the file is closed. The drawback is that it only works in Python 2.7 and later.

I was aware of the context manager for this. But Sphinx-Gallery still
supports python 2.6 so I didn't use it. I remember I keep that support
since it was for your Neurodebian systems. Should I drop support for
Python 2.6? @lesteve suggested to drop support on Python 2.6 when NumPy
does it, which shall be anytime soon as it is announced for their 1.12
release.

GaelVaroquaux: > I was aware of the context manager for this. But Sphinx-Gallery still supports
> python 2.6 so I didn't use it.

Sounds reasonnable.

> I remember I keep that support since it was for your Neurodebian
> systems. Should I drop support for Python 2.6? @lesteve suggested to
> drop support on Python 2.6 when NumPy does it, which shall be anytime
> soon as it is announced for their 1.12 release.

Well, let's wait a bit after the 1.12 release, and drop it after that.

Titan-C: > I suspect that, to work reliably on all OSes, it's useful to use "mode='wb'".

I did this change and now travis fails. The zipfile does not take binary in the mode options.

GaelVaroquaux: >     I suspect that, to work reliably on all OSes, it's useful to use "mode='wb'".

> I did this change and now travis fails. The zipfile does not take binary in the mode options.

OK. My bad. Sorry

Titan-C:
I refactored this a bit just to work with the test. Not a great test but is something.

GaelVaroquaux:
Maybe there should be an option in the conf.py to disable this. Maybe some people are not going to want it.

GaelVaroquaux:
Also, I am curious: why did you choose to do it at the level of the section, rather than at the level of the whole gallery?

GaelVaroquaux:
A design question: do people think that we should be putting the filename in the download button or not? The benefit is that it makes it easier for people to find that file in their download folder, in case it is crowded. The drawback is that it makes the download button bigger, and a bit perturbing on the page.

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -260,6 +261,22 @@ def test_save_figures():
...
+
+def test_zip_notebooks():
+    """Test generated zipfiles are not corrupt"""
+    gallery_conf = build_test_configuration(examples_dir='examples')
+    examples = [fname
+                for fname in sorted(os.listdir(gallery_conf['examples_dir']))
+                if fname.endswith('.py')]
+    sg.python_zip(examples, gallery_conf['examples_dir'],
+                  gallery_conf['gallery_dir'])
+    zipfilename = os.path.join(gallery_conf['gallery_dir'], "_python.zip")
```
GaelVaroquaux: I just realized: are the tests creating files on the disk? If so, they should remove them after, in a try/finally setting.

Ideally, these files should be created in a temporary folder, as sometimes the build folder is not writable.

Titan-C:
> A design question: do people think that we should be putting the filename in the download button or not? The benefit is that it makes it easier for people to find that file in their download folder, in case it is crowded. The drawback is that it makes the download button bigger, and a bit perturbing on the page.

The download button has the filename without the file extension. I'll add the file extension.

sphinx_gallery/tests/test_gen_rst.py
Titan-C: for the test they are saved in a temporary folder. The function `build_test_configuration` takes care making this folder and setting other variables.

Titan-C:
Now the gallery section name is now correctly set in the zipfile download file name.

sphinx_gallery/tests/test_gen_rst.py
GaelVaroquaux: > for the test they are saved in a temporary folder. The function build_test_configuration takes care making this folder and setting other variables.

OK, thanks!

GaelVaroquaux:
Aren't you worried that it makes too big a button on the gallery page? One per section is frequent.

Titan-C:
> Aren't you worried that it makes too big a button on the gallery page? One per section is frequent.

No. The buttons are quite thin, they are the same ones of the individual examples. The example thumbnail takes 5 more times vertical space.  And for big packages is negligible, nilearn has 5 section that isn't much compared to the number of examples. Same for scikit-learn

Titan-C:
With the latest commit it is now optional to have this download buttons for each section. The default is nevertheless to have this option activated. How to disable is in the documentation.

lesteve:
> With the latest commit it is now optional to have this download buttons for each section. The default is nevertheless to have this option activated. How to disable is in the documentation.

Personally I'd be in favour of only one button to download all the examples. The only use case I can see for this "download all examples" button is for people who are in tutorials, want only the examples and not the package source. In this case I don't really see why it would make sense to have a per-section download option.

GaelVaroquaux:
> Personally I'd be in favour of only one button to download all the
> examples. The only use case I can see for this "download all examples"
> button is for people who are in tutorials, want only the examples and
> not the package source. In this case I don't really see why it would
> make sense to have a per-section download option.

That's how I feel too. I don't really understand the usecase for
downloading only one section.

Titan-C:
As the common agreement is to have download files for the complete gallery I updated this PR for that behaviour.

CHANGES.rst
```diff
@@ -7,6 +7,11 @@ git master
 New features
 ''''''''''''
 * Enhanced CSS for download buttons
+* Download buttons at the end of each section in the gallery to
```
GaelVaroquaux: Is this description still accurate?

sphinx_gallery/downloads.py
```diff
@@ -0,0 +1,114 @@
...
+# Author: Óscar Nájera
+# License: 3-clause BSD
+
+from __future__ import absolute_import, division, print_function
+
+import os
+import zipfile
+
+CODE_DOWNLOAD = """**Total running time of the script:**
+({0:.0f} minutes {1:.3f} seconds)\n\n
```
GaelVaroquaux: I find it quite surprising to have the running time of the script part of the download template.

Titan-C: Indeed this is a bad use of the variable content

Titan-C:
Comments addressed

Issue #120, opened 2016-05-18
Sphinx warns about not readable image file (although the image is found for html build)


emmanuelle:
I'm currently trying to port the doc of scikit-image to sphinx-gallery (see https://github.com/scikit-image/scikit-image/pull/2078). 

During the html build of the doc, sphinx issues the following warning (treated as error by CI):
```
 /home/emma/travail/projects/scikit-image/doc/source/user_guide/numpy_images.rst:None: 
WARNING: image file not readable: user_guide/../../_images/sphx_glr_plot_camera_numpy_001.png
```

``sphx_glr_plot_camera_numpy_001.png`` is generated by one of the examples of the gallery, and included in ``numpy_images.rst`` as 

```
.. image:: ../../_images/sphx_glr_plot_camera_numpy_001.png
    :width: 45%
    :target: ../auto_examples/numpy_operations/plot_camera_numpy.html
```

From the warning, it seems that the path is wrong, but the html page generated from the build displays correctly the image (and the link to the gallery example works too). Therefore, I don't understand the origin of the warning (and maybe I wouldn't mind it so much if didn't make our Travis CI fail :-)). 

This path issue might be another facet of the problem of having a separated ``source`` directory, as also illustrated in #115, but I'm not sure whether that point matters here.

lesteve:
As far I can see from the nilearn doc, we use things like:

```
.. image:: ../auto_examples/01_plotting/images/sphx_glr_plot_atlas_001.png
   :target: ../auto_examples/01_plotting/plot_atlas.html
```

Can you try to follow a pattern like this and see whether that fixes the warning?

I am not going to claim I understand how sphinx works in detail, but at one point I understood that the _images folder was something that sphinx was creating automatically by looking at all the images referenced to and copying them there. It could well be that your image is referenced correctly in the example HTML causing it to end up in _images. This would kind of explain why the image is not found at sphinx generation time but shows up correctly in the HTML.

emmanuelle:
Thank you so much for helping me dive into sphinx's black magics :-) ! It worked, warning fixed, made my day!

lesteve:
Glad to hear that! Closing this issue then.

GaelVaroquaux:
> Glad to hear that! Closing this issue then.

That's closing an issue a bit fast, IMHO. The problem is that we had a
fairly senior Python developer not managing to use sphinx-gallery right
after giving it a good try. If she is not managing by her own, it means
[...]

Issue #50, opened 2015-08-05
Relative paths of example directories not relative to doc root


florian-wagner:
I have asked something similar in the closed issue #40, but I still think there is something wrong with the paths. Shouldn't they be relative to the documentation source (i.e. the location of conf.py) rather than relative to the location where `sphinx-build` is called?

``` bash
cd ~/tmp
git clone https://github.com/sphinx-gallery/sphinx-gallery
sphinx-build -q sphinx-gallery/doc html_output # does not work
cd sphinx-gallery/doc && sphinx-build -q . html_output # works fine
```

The output of the first `sphinx-build` call is:

``` bash
WARNING: 'default' html theme has been renamed to 'classic'. Please change your html_theme setting either to the new 'alabaster' default theme, or to 'classic' to keep using the old default.

Exception occurred:
  File "/usr/lib/python3.4/os.py", line 237, in makedirs
    mkdir(name, mode)
PermissionError: [Errno 13] Permission denied: '../../../examples'
```

lesteve:
> Shouldn't they be relative to the documentation source (i.e. the location of conf.py) rather than relative to the location where sphinx-build is called?

I think the convention we are following is the latter, the main reason for that is that we are always building the documentation from the doc folder as far as I know. Do you see something especially wrong with that ?

florian-wagner:
sphinx itself does allow to build the documentation everywhere and I think sphinx-gallery should not break this functionality. 

It is very useful to separate the source from compiled files including the docs. Always building the documentation in the source tree itself is especially annoying when calling the status of your version control, as a lot of untracked files are generated.

lesteve:
Fair points. To be brutally honest, I don't see anything happening in near future on this front, just because these issues are not important enough for our use cases. Of course PRs are more than welcome!

FWIW, we are generally building documenation using the Makefile in doc rather than directly issuing sphinx commands.

Titan-C:
This is fixed in #190

Issue #40, opened 2015-06-29
Support for absolute paths in 'examples_dir' and 'gallery_dir'


florian-wagner:
Is it possible to allow absolute paths to be set for `examples_dir` and `gallery_dir`? Or in other words: Is it necessary to transform the input variables to paths relative to app.build.srcdir? If not, this would enable a bit more flexibility when using cmake.

Titan-C:
Yes they have to be relative to app.build.src because this then becomes
the absolute root for the website and sphinx relies on this local root to
resolve all link addresses

florian-wagner:
Okay, thanks a lot. Then I have to rearrange a bit on my side.

PR #133, opened 2016-07-23
[NO MRG - WIP] Improved use of relative paths

Titan-C:
This aims to fix in a broader sense issues #40 #50 #115 #119 #120 

The issue is that, sphinx-gallery (sphinx way of work) relies on relative paths within the documentation to resolve correctly the back reference links and thumbnail images from arbitrary location in the documentation source.

If one is to build the documentation not inside the doc source using the `make html` command one runs into problems. PR #119 addresses this issue by changing directory during build. But this solution neglects the link embedding in the final html.

This should not be merged before #97

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,6 +7,6 @@ with-doctest=1
 doctest-extension=rst
 
 [build_sphinx]
-source-dir = docs/
-build-dir  = docs/_build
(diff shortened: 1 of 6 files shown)


Titan-C:
Closing in favor of #190

PR #135, opened 2016-07-27
stdout info on file build and time

### LINKED ISSUE #134: Time elapsed should be printed for each example rather than for each block
Probably due to some changes in #97, it seems like "Time elapsed" is written once per block. Also it doesn't seem like the filename of the example being run is printed anymore which I think is useful.

```
 - time elapsed : 0.24 sec
 - time elapsed : 0.27 sec
 - time elapsed : 0.05 sec
 - time elapsed : 0.091 sec
 - time elapsed : 0.085 sec
```

Summary: the filename of the example + time elapsed should be printed once per example.

Titan-C:
fix #134

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -555,7 +555,6 @@ def execute_code_block(code_block, example_globals,
         os.chdir(cwd)
         sys.stdout = orig_stdout
 
-    print(" - time elapsed : %.2g sec" % time_elapsed)
     code_output = u"\n{0}\n\n{1}\n\n".format(images_rst, stdout)
     block_vars['fig_count'] += fig_num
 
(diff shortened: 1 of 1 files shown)


sphinx_gallery/gen_rst.py
```diff
@@ -659,4 +659,6 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
         example_rst += SPHX_GLR_SIG
         f.write(example_rst)
 
+    print("{0} build in : {1:.2g} sec\n".format(src_file, time_elapsed))
```
lesteve: I would use `"{0} ran in {1:.2g} seconds"`

lesteve:
LGTM besides the minor comment, feel free to merge when you have tackled it.

PR #139, opened 2016-08-23
[MRG] Notebook inline math

Titan-C:
If inline math appears multiple times in the same line it would be joined in the same block. The non-greedy regex fixes that.
If inline math spans over multiple lines, it would not be parsed. The re.DOTALL fixes that

- [x] Extend test to cover this.

For discussion. The example `plot_quantum.py` uses heavily latex. I noticed the previous bugs from it. It also uses a multiline math that to be rendered correctly in the jupyter notebook it requires to be wrapped with:
```latex
\begin{align}
\end{align}
```
Should I make this the default for the rst syntax:
```rst
.. math: some latex
```

### DIFF
--- a/sphinx_gallery/notebook.py
+++ b/sphinx_gallery/notebook.py
@@ -55,7 +55,8 @@ def rst2md(text):
 
     math_eq = re.compile(r'^\.\. math::((?:.+)?(?:\n+^  .+)*)', flags=re.M)
     text = re.sub(math_eq,
-                  lambda match: r'$${0}$$'.format(match.group(1).strip()),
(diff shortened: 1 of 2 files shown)


Titan-C:
I'll use the align environment for latex equations. Test is included

GaelVaroquaux:
How does single line math block render with the align environment ? No rendering error?

GaelVaroquaux:
Aside from that question,  it looks good to me. Good job with the tests.

Titan-C:
>How does single line math block render with the align environment ? No rendering error?

For a single line of math \begin{equation} and \begin{align} are completely equivalent. I learned that here.

PR #140, opened 2016-08-30
Seaborn enviroment clean up

Titan-C:
If seaborn is loaded in one example, examples afterward keep using its plotting style. This patch clears its presence. I'm still looking for different way to sandbox the python execution.

I'm also getting errors in python3 on the seaborn script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ addons:
     packages:
     - python-numpy
     - python-matplotlib
+    - python-scipy
     - python-pip
     - python-coverage
     - xvfb # For mayavi headless
@@ -32,17 +33,24 @@ before_install:
       fi
 
 install:
+    # force no mkl because mayavi requires old version of numpy
+    # which then crashes with pandas and seaborn
     - if [ "$DISTRIB" == "conda" ]; then
-        conda create --yes -n testenv python=$PYTHON_VERSION pip numpy
-        setuptools matplotlib pillow sphinx nose coverage;
+        conda create --yes -n testenv python=$PYTHON_VERSION pip nomkl numpy
+        setuptools matplotlib pillow sphinx nose coverage seaborn;
         source activate testenv;
         if [ "$PYTHON_VERSION" == "2.7" ]; then
-          conda install --yes --quiet mayavi;
+          conda install --yes mayavi;
           conda upgrade --yes --all;
           pip install --upgrade pyface;
         fi;
       fi;
     - pip install -r requirements.txt
+    # This will build pandas as a dependency in ubuntu 12.04
+    - if [ "$DISTRIB" == "ubuntu" ]; then pip install seaborn; fi;
+
+    # Seaborn does not support python 2.6 thus is not available in conda
+    - if [ "$PYTHON_VERSION" == "2.6" ]; then pip install seaborn; fi;
     - python setup.py install
 
 # To test the mayavi environmen follow the instructions at
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,18 @@ Change Log
 git master
 ----------
 
+New features
+''''''''''''
+* Enhanced CSS for download buttons
+
+Bug Fixes
+'''''''''
+* When seaborn is imported in a example the plot style preferences are
+  tranfered to plots executed aftewards. The CI is set up such that
+  users can follow how to get the compatible versions of
+  mayavi-pandas-seaborn and nomkl in a conda environment to have all
+  the features available.
+
 v0.1.3
 ------
 
--- a/examples/plot_seaborn.py
+++ b/examples/plot_seaborn.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+r"""
+Using Seaborn
+=============
+
+A example from their website
+
+"""
+# Author: Michael Waskom
+# License: BSD 3 clause
+
+from __future__ import division, absolute_import, print_function
+
+import seaborn as sns
+import matplotlib.pyplot as plt
+sns.set(style="white")
+
+df = sns.load_dataset("iris")
+
+g = sns.PairGrid(df, diag_sharey=False)
+g.map_lower(sns.kdeplot, cmap="Blues_d")
+g.map_upper(plt.scatter)
+g.map_diag(sns.kdeplot, lw=3)
--- a/examples/plot_seaborn.py
+++ b/examples/plot_seaborn.py
@@ -1,23 +1,29 @@
 # -*- coding: utf-8 -*-
 r"""
-Using Seaborn
-=============
+Seaborn example
+===============
 
-A example from their website
-
+Preview the capture of seaborn styles in plots
 """
 # Author: Michael Waskom
 # License: BSD 3 clause
 
 from __future__ import division, absolute_import, print_function
 
+
+import numpy as np
 import seaborn as sns
-import matplotlib.pyplot as plt
-sns.set(style="white")
 
-df = sns.load_dataset("iris")
+# Enforce the use of default set style
+#sns.set(style="darkgrid", palette="Set2")
 
-g = sns.PairGrid(df, diag_sharey=False)
-g.map_lower(sns.kdeplot, cmap="Blues_d")
-g.map_upper(plt.scatter)
-g.map_diag(sns.kdeplot, lw=3)
+# Create a noisy periodic dataset
+sines = []
+rs = np.random.RandomState(8)
+for _ in range(15):
+    x = np.linspace(0, 30 / 2, 30)
+    y = np.sin(x) + rs.normal(0, 1.5) + rs.normal(0, .3, 30)
+    sines.append(y)
+
+# Plot the average over replicates with bootstrap resamples
+sns.tsplot(sines, err_style="boot_traces", n_boot=500)
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -642,6 +642,14 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
             example_rst += text2string(bcontent) + '\n'
             example_nb.add_markdown_cell(text2string(bcontent))
 
+    # Unload seaborn after all examples run
+    if 'seaborn' in sys.modules:
+        sys.modules['seaborn'].reset_defaults()
+        sys.modules['seaborn'].reset_orig()
+
+    # Reset Matplotlib to default
+    plt.rcdefaults()
+
     # Writes md5 checksum if example has build correctly
     # not failed and was initially meant to run(no-plot shall not cache md5sum)
     if block_vars['execute_script']:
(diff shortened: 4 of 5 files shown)


GaelVaroquaux:
LGTM. Thanks a lot. Let's wait for Travis to finish and then merge.

Titan-C:
Yes the unload was wrong. I have decided to avoid the deletion of the module in favor of resetting to the default plots. Something relevant to is that the reset shall be done after the complete script is executed and not the block. Since later on notebook styled examples using seaborn will appear.

I still have the issue installing seaborn in the ubuntu environment. It times out.

GaelVaroquaux:
Reseting to the default plots won't work because if you have a second example that uses seaborn, it will not load seaborn from scratch, hence seaborn's defaults won't get installed. You don't see that problem with only one seaborn example, you need two.

What's wrong with the code that I had written? It seemed to work, and my understand of Python's import logic tells me that it should.

.travis.yml
```diff
@@ -6,8 +6,6 @@ addons:
     - python-numpy
     - python-matplotlib
     - python-scipy
-    - python-pandas
-    - python-seaborn
```
GaelVaroquaux: seaborn is indeed not in the standard Ubuntu (actually it's in the 16.04), but I am surprised that we can't get pandas.

I think that the strategy would be pandas via ubuntu. It has compiled code, so it's easier that way, and seaborn via pip.

Titan-C:
>Reseting to the default plots won't work because if you have a second example that uses seaborn, it will not load seaborn from scratch, hence seaborn's defaults won't get installed. You don't see that problem with only one seaborn example, you need two.

Locally I did put more examples. But since my examples come from the seaborn gallery they explicitly load a style, so I did not notice.

> What's wrong with the code that I had written? It seemed to work, and my understand of Python's import logic tells me that it should.

It does not work in python 3, since you delete an entry form the sys.modules
```python-traceback
    for module in sys.modules.keys():
RuntimeError: dictionary changed size during iteration
```

.travis.yml
Titan-C: So which versions shall we support. On the first try pandas was indeed installed from ubuntu sources and since seaborn was not there pip did installed it. But the build fails because the version of pandas is ubuntu is so old that it loocks for attributes in the pandas dataframe that are not implemented yet.

```python-traceback
Unexpected failing examples:
../examples/plot_seaborn.py failed leaving traceback:
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/sphinx_gallery-0.1.3.post1-py2.7.egg/sphinx_gallery/gen_rst.py", line 524, in execute_code_block
    exec(code_block, example_globals)
  File "<string>", line 10, in <module>
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/seaborn/utils.py", line 428, in load_dataset
    if df.iloc[-1].isnull().all():
  File "/usr/lib/pymodules/python2.7/pandas/core/frame.py", line 1437, in __getattr__
    (type(self).__name__, name))
AttributeError: 'DataFrame' object has no attribute 'iloc'
```

That is the reason I installed pandas and seaborn from pip, but it fails again probably now the matplotlib version is to old
```python-traceback
Unexpected failing examples:
../examples/plot_seaborn.py failed leaving traceback:
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/sphinx_gallery-0.1.3.post1-py2.7.egg/sphinx_gallery/gen_rst.py", line 524, in execute_code_block
    exec(code_block, example_globals)
  File "<string>", line 15, in <module>
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/seaborn/axisgrid.py", line 1322, in map_diag
    diag_ax = ax._make_twin_axes(sharex=ax, frameon=False)
AttributeError: 'AxesSubplot' object has no attribute '_make_twin_axes'
```

GaelVaroquaux:
> It does not work in python 3, since you delete an entry form the sys.modules

>     for module in sys.modules.keys():
> RuntimeError: dictionary changed size during iteration

OK, the fix is easy:

for module in list(sys.modules.keys())

.travis.yml
GaelVaroquaux: How about installing an old version of seaborn? You could do "pip install seaborn==0.5"

Titan-C:
Seaborn is not available for python 2.6 in conda. In pip it kind of  gets installed and runs the particular example that is present

For the ubuntu install even with seaborn 0.5 it fails

Titan-C:
It works and travis validates. It was a configuration pain.
Few thoughts:
* There was a big collision with mayavi-conda-mkl-pandas-seaborn. Mayavi in conda requires very specific versions of numpy and scipy and pyface. There is a forced update of pyface to make sphinx-gallery work with mayavi. If on top we install pandas/seaborn there are more problems as all versions are not entirely compatible, either because they link to the wrong BLAS(mkl/openblas) or numexpr does not get loaded. I had also to scan which examples of seaborn could run in this setup and I test on a forced nomkl setup.
* The ubuntu version is 12.04 and neither seaborn nor pandas are there. When they are installed through pip, the ubuntu matplotlib 1.1.1 is just to old and some plots seaborn can't be generated, I had to look for compatible ones. I think getting a more current version of matplotlib should be the user responsibility.
* Seaborn does not support python 2.6, we do for the moment. Python 2.6 is installed by conda and does not provide seaborn. With pip I can manage to get it installed although not  everything will work. We will drop support for python 2.6 in the near future so is a minor concern.

GaelVaroquaux:
>   • There was a big collision with mayavi-conda-mkl-pandas-seaborn.

Yes. Tight dependencies are a pain :$.

>   • The ubuntu version is 12.04 and neither seaborn nor pandas are there.

We should upgrade our travis instance to use 14:04. I think that it's
easy.

>    We will drop support for python 2.6 in the near future so is a minor concern.

+1. For this sklearn needs to drop 2.6, but I am in favor of this. Ping
@lesteve, @ogrisel, @amueller.

GaelVaroquaux:
+1 for merge. Good job!

amueller:
I'm ok with dropping 2.6 though I don't understand why we need to drop it because of this. Does sphinxgallery depend on seaborne? Or is it only used in examples?

GaelVaroquaux:
> I'm ok with dropping 2.6 though I don't understand why we need to drop it
> because of this. Does sphinxgallery depend on seaborne? Or is it only used in
> examples?

It's used in the examples because seaborn needs special handling (it
hacks the rcParams on import).

Setting up the continuous integration was a bit hard :)

amueller:
But we won't run the CI for sphinx gallery, so sklearn doesn't care, right?

GaelVaroquaux:
> But we won't run the CI for sphinx gallery, so sklearn doesn't care, right?

No, it's on the side of sphinx-gallery: we would like to stop supporting
Python 2.6, partly because it's a pain to run the CI on it.

amueller:
Sorry I didn't have coffee yet and I'm slow

> +1. For this sklearn needs to drop 2.6, but I am in favor of this. Ping

This sound like you propose sklearn drops 2.6, maybe I misunderstood. If that's what you propose, I didn't understand yet why proposed it. We have no CI problems with 2.6 for sklearn, right?

GaelVaroquaux:
I think that I am not explaining things well. Sphinx_gallery would like
to drop support for 2.6. But scikit-learn still support 2.6. I thought
that it would be good for the doc of scikit-learn to be buildable on all
versions that it supports. Hence, I thought that it was a reason for
sphinx_gallery not to drop 2.6. But maybe it's not a big deal if
scikit-learn's docs doesn't build on 2.6.

amueller:
Ok I got it. Well I don't mind if docs don't build on 2.6. I think people with 2.6 (if there are any) would rather have the docs not built than not being able to use sklearn. I would rather not drop 2.6 for 0.18 because 0.18 is such a big release and moves the model_selection module.
We can announce that we drop 2.6 for 0.19 in the 0.18 release?

lesteve:
> I'm still looking for different way to sandbox the python execution.

Just a thought: could we not run each example in a separate process?

It's not written anywhere that sphinx-gallery supports Python 2.6 and I could find at least one `.format` statement with `{}` that hints that it is not supported.

```
❯ git grep -C3 '{}' | grep -C3 format
sphinx_gallery/gen_rst.py-        raise ValueError(
sphinx_gallery/gen_rst.py-            "Example docstring should have a header for the example title "
sphinx_gallery/gen_rst.py-            "and at least a paragraph explaining what the example is about. "
sphinx_gallery/gen_rst.py:            "Please check the example file:\n {}\n".format(filename))
sphinx_gallery/gen_rst.py-
sphinx_gallery/gen_rst.py-    return first_paragraph
sphinx_gallery/gen_rst.py-
```

lesteve:
Just to be clear I don't think we should bother about Python 2.6 for sphinx-gallery.

GaelVaroquaux:
> Just to be clear I don't think we should bother about Python 2.6 for sphinx-gallery.

Let's stop bothering about it, and move our CI away from it. We can also
move to 14:04 on travis. All these will make our life easier!

Titan-C:
>It's not written anywhere that sphinx-gallery supports Python 2.6 and I could find at least one .format statement with {} that hints that it is not supported.

I missed this one. This is not having enough tests that will catch this.

Regarding what Sphinx-Gallery supports is probably not a well defined issue. Most of the time I have more trouble supporting the ubuntu 12.04 environment(very old matplotlib) than the conda python 2.6(formatting strings). And the core of Sphinx-Gallery has little inconvenience supporting this. Of course moving away of these would make my life easier.

The problem comes when dealing with the optional supported modules: that is mayavi and now seaborn. And in this PR the biggest pain was to support the environment where both of this where included.  Mayavi can only run in a conda environment on python 2.7 and we would not bother on supporting more options, because at least @agramfort and I thought as this is an extra feature I will certainly come to the user responsibility to have all the dependencies well installed(reason why Sphinx-Gallery setup does not pull any dependencies by itself).

Should we have the same mindset for seaborn? and support it on updated environments only?

Titan-C:
If there are not other suggestions on this PR code. I'll clean up the git commit history before merging.

GaelVaroquaux:
> I would rather not drop 2.6 for 0.18 because 0.18 is such a big
> release and moves the model_selection module.

+1 on that.

I am on the same page as you with regards to building docs!

> We can announce that we drop 2.6 for 0.19 in the 0.18 release?

Thanks for asking the question on the ML.

Looks like we have the opinion on the question. Great!

GaelVaroquaux:
:+1: for merge on my side.

Titan-C:
Rebased and ready for merge. I reordered the commits and somehow Github reorders commits by time in the website but still recognizes that the last commit is the changes log and sends that one to travis.

GaelVaroquaux:
Maybe you can do a "git commit --amend" and force push, to trigger travis on the last commit.

Titan-C:
`git commit --amend --date="$(date -R)"` did the job

GaelVaroquaux:
And everything passes beautifully! Great job. Merging.

PR #142, opened 2016-08-30
ENH: Improve markdown conversion

### LINKED ISSUE #141: RST directives remain in ipynb
The IPython ipynb downloads don't strip all irrelevant stuff e.g. links:

http://martinos.org/mne/dev/_downloads/plot_background_filtering.ipynb

Looks like this:

screenshot from 2016-08-30 10-01-29

Should we remove lines starting with `.. `?

Eric89GXL:
Closes #141.

1. Removes some of the offending residual restructuredtext directives:
    screenshot from 2016-08-30 13-04-42

2. Adds warning/note boxes:
    screenshot from 2016-08-30 13-05-00

3. Adds image support:
    screenshot from 2016-08-30 13-05-31

4. Corrects problem with greedy regex for inline math.

Ready for review/merge from my end.

### DIFF
--- a/sphinx_gallery/notebook.py
+++ b/sphinx_gallery/notebook.py
@@ -11,10 +11,12 @@ Class that holds the Jupyter notebook information
 # License: 3-clause BSD
 
 from __future__ import division, absolute_import, print_function
+from functools import partial
 import json
 import os
 import re
 import sys
+
 
 def ipy_notebook_skeleton():
     """Returns a dictionary with the elements of a Jupyter notebook"""
@@ -46,6 +48,14 @@ def ipy_notebook_skeleton():
     return notebook_skeleton
 
 
+def directive_fun(match, directive):
+    """Helper to fill in directives"""
+    directive_to_alert = dict(note="info", warning="danger")
+    return ('<div class="alert alert-{0}"><h4>{1}</h4><p>{2}</p></div>'
+            ''.format(directive_to_alert[directive], directive.capitalize(),
+                      match.group(1).strip()))
+
+
 def rst2md(text):
     """Converts the RST text from the examples docstrigs and comments
     into markdown text for the Jupyter notebooks"""
@@ -60,6 +70,29 @@ def rst2md(text):
                   text)
     inline_math = re.compile(r':math:`(.+?)`', re.DOTALL)
     text = re.sub(inline_math, r'$\1$', text)
+
+    directives = ('warning', 'note')
+    for directive in directives:
+        directive_re = re.compile(r'^\.\. %s::((?:.+)?(?:\n+^  .+)*)'
+                                  % directive, flags=re.M)
+        text = re.sub(directive_re,
+                      partial(directive_fun, directive=directive), text)
+
+    links = re.compile(r'^ *\.\. _.*:.*$\n', flags=re.M)
+    text = re.sub(links, '', text)
+
+    refs = re.compile(r':ref:`')
+    text = re.sub(refs, '`', text)
+
+    contents = re.compile(r'^\s*\.\. contents::.*$(\n +:\S+: *$)*\n',
+                          flags=re.M)
+    text = re.sub(contents, '', text)
+
+    images = re.compile(
+        r'^\.\. image::(.*$)(\n *:alt:(.*$))?[\n +:\S+:.*$]*', flags=re.M)
+    text = re.sub(
+        images, lambda match: '![{1}]({0})\n'.format(
+            match.group(1).strip(), (match.group(3) or '').strip()), text)
 
     return text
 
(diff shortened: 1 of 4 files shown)


GaelVaroquaux:
Beautiful!

There are now conflicts, and I suspect that it's due to #139 being merged. Maybe this PR could actually be refactored to weave in well with the code introduced in #139.

Titan-C:
It looks great, thank you.

I recently merged a fix for the rendered math too, which covers some extra cases too. You'll have to rebase.

Finally, would it be possible to have the content of the test also displayed in the gallery? the one on the tutorials folder that covers more the notebook styled examples.  That way we can make people aware that we support this.

sphinx_gallery/notebook.py
```diff
@@ -46,6 +48,14 @@ def ipy_notebook_skeleton():
     return notebook_skeleton
 
 
+def directive_fun(match, directive):
+    """Helper to fill in directives"""
+    directive_to_alert = dict(note="info", warning="danger")
+    return ('<div class="alert alert-{0}"><h4>{1}</h4><p>{2}</p></div>'
+            ''.format(directive_to_alert[directive], directive.capitalize(),
```
lesteve: I don't think you need the `''`, you could do something like:

```py
def directive_fun(match, directive):
    """Helper to fill in directives"""
    directive_to_alert = dict(note="info", warning="danger")
    return ('<div class="alert alert-{0}"><h4>{1}</h4><p>{2}</p></div>'
            .format(directive_to_alert[directive], directive.capitalize(),
                    match.group(1).strip()))
```

sphinx_gallery/notebook.py
```diff
@@ -57,9 +67,33 @@ def rst2md(text):
...
+
+    refs = re.compile(r':ref:`')
+    text = re.sub(refs, '`', text)
+
+    contents = re.compile(r'^\s*\.\. contents::.*$(\n +:\S+: *$)*\n',
+                          flags=re.M)
+    text = re.sub(contents, '', text)
+
+    images = re.compile(
+        r'^\.\. image::(.*$)(\n *:alt:(.*$))?[\n +:\S+:.*$]*', flags=re.M)
```
lesteve: Why the square brackets, did you mean to use parentheses instead in `[\n +:\S+:.*$]*`?

lesteve:
I am a bit worried about the maintainability of ReST to markdown conversion via regexps going forward. I don't know whether there is a cleaner alternative. I think it's fine as long as we don't build high expectations about the markdown support in the notebook.

GaelVaroquaux:
> I am a bit worried about the maintainability of ReST to markdown
> conversion via regexps going forward. I don't know whether there is a
> cleaner alternative. I think it's fine as long as we don't build high
> expectations about the markdown support in the notebook.

For now, I think that we should do this. In the longer term, maybe we
should use docutils.examples.internals: the first argument is a document
tree that we can walk (eg with the "traverse" method).

lesteve:
FWIW I found out that there is some markdown support in sphinx. Maybe using markdown in the examples would make sense since we generally do not use advanced ReST functionalities. That was just a thought and talking about the long term here obviously.

Quoting from https://github.com/sphinx-doc/sphinx/issues/825#issuecomment-180197293:

> Actually, sphinx can build Markdown document by using:

> pip install commonmark recommonmark

> conf.py

> ```
source_parsers = {
    '.md': 'recommonmark.parser.CommonMarkParser',
}
source_suffix = ['.rst', '.md']
```

> It's still limited, but usable.

Eric89GXL:
> I recently merged a fix for the rendered math too, which covers some extra cases too. You'll have to rebase.

Yeah that must be the conflict, rebase should be straightforward (I hope!).

> Maybe using markdown in the examples would make sense since we generally do not use advanced ReST functionalities. 

Maybe, but now in MNE-Python we have a lot of tutorials in ReST syntax, so changing them would be :( But maybe going forward it would make sense if someone wants to try it out and make nice examples.

Eric89GXL:
Rebased and comments addressed, including updating an example:

screenshot from 2016-08-31 10-10-58

GaelVaroquaux:
>     Maybe using markdown in the examples would make sense since we
>     generally do not use advanced ReST functionalities.

I am not excited in having a mixture of markup language in a
documentation: cognitive overhead.

GaelVaroquaux:
> >     Maybe using markdown in the examples would make sense since we
> >     generally do not use advanced ReST functionalities.

By the way: nilearn is using the :func: and :ref: directives quite a lot.

Titan-C:
+1 on this

> > Maybe using markdown in the examples would make sense since we  generally do not use advanced ReST functionalities.

> I am not excited in having a mixture of markup language in a
documentation: cognitive overhead.

I also like markdown better, but that is because is a simple syntax to follow. This simplicity comes with the fact that is not standardized there are many flavors of markdown. Markdown simplicity comes with the issue that it can't correctly handle many interconnected documents, and that would be the use of :ref: directives and similar it is just good for a single document.

Few days ago I found this document on why not using markdown for documentation
http://ericholscher.com/blog/2016/mar/15/dont-use-markdown-for-technical-docs/

PR #144, opened 2016-09-15
[MRG] CSS: improve rendering of download buttons

GaelVaroquaux:
This PR:

* Fixes the overflow visible in the download buttons (as discussed in https://github.com/scikit-learn/scikit-learn/pull/7354#issuecomment-246689940)

* Centers the buttons correctly

* Make the sphinx-gallery signature slightly less visible (let's not annoy downstream users)

Before, in sphinx-gallery's documentation:

image

After:

image

Before in scipy-lectures
image

After:
image

### DIFF
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -111,6 +111,7 @@ div.sphx-glr-footer {
 div.sphx-glr-download {
   display: inline-block;
   margin: 1em auto 1ex 2ex;
+  vertical-align: middle;
 }
 
 div.sphx-glr-download a {
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -140,6 +140,12 @@ div.sphx-glr-download a {
     }
 }
 
+@media (min-width: 40em) {
+    div.sphx-glr-download a {
+	min-width: 16em;
+    }
+}
+
 
 div.sphx-glr-download code.download {
   display: inline-block;
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -128,6 +128,11 @@ div.sphx-glr-download a {
   text-align: center;
 }
 
+/* The last child of a download button is the file name */
+div.sphx-glr-download a span:last-child {
+    font-size: smaller;
+}
+
 @media (min-width: 20em) {
     div.sphx-glr-download a {
 	min-width: 10em;
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -130,6 +130,7 @@ div.sphx-glr-download a {
 
 div.sphx-glr-download code.download {
   display: inline-block;
+  white-space: normal;
 }
 
 div.sphx-glr-download a:hover {
@@ -150,18 +151,15 @@ ul.sphx-glr-horizontal img {
   height: auto !important;
 }
 
-p.sphx-glr-signature {
-  text-align: right;
-  margin-left: auto;
-  display: table;
-}
-
 p.sphx-glr-signature a.reference.external {
   -moz-border-radius: 5px;
   -webkit-border-radius: 5px;
   border-radius: 5px;
   padding: 3px;
   font-size: 75%;
+  text-align: right;
+  margin-left: auto;
+  display: table;
 }
 
 a.sphx-glr-code-links:hover{
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -103,6 +103,11 @@ thumbnail with its default link Background color */
 blockquote.sphx-glr-script-out {
   margin-left: 0pt;
 }
+
+div.sphx-glr-footer {
+    text-align: center;
+}
+
 div.sphx-glr-download {
   display: inline-block;
   margin: 1em auto 1ex 2ex;
@@ -115,11 +120,13 @@ div.sphx-glr-download a {
   border: 1px solid #c2c22d;
   color: #000;
   display: inline-block;
+  /* Not valid in old browser, hence we keep the line above to override */
+  display: table-caption;
   font-weight: bold;
-  max-width: 45ex;
   padding: 1ex;
   text-align: center;
 }
+
 
 div.sphx-glr-download code.download {
   display: inline-block;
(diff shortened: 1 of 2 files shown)


Titan-C:
I would say some buttons become really small. 
gal

I would put a min-width instead of completely deleting the max-width. Otherwise the button look too small and we might loose visual uniformity.

sphinx_gallery/_static/gallery.css
```diff
@@ -143,8 +150,13 @@ ul.sphx-glr-horizontal img {
   height: auto !important;
 }
 
+p.sphx-glr-signature {
+  text-align: right;
+  margin-left: auto;
+  display: table;
+}
+
```
Titan-C: You can put 
margin-left: auto;
display: table;

in the next class and you get the same effect without having this separate class

sphinx_gallery/_static/gallery.css
```diff
@@ -115,12 +120,14 @@ div.sphx-glr-download a {
...
   font-weight: bold;
-  max-width: 45ex;
   padding: 1ex;
   text-align: center;
 }
 
+
 div.sphx-glr-download code.download {
   display: inline-block;
 }
```
Titan-C: To solve the conflict for scikit-learn with the boostrap.min.css we need to add an extra `white-space: normal;` in this class. Or shall that be kept only in scikit-learn css files?

GaelVaroquaux:
> To solve the conflict for scikit-learn with the boostrap.min.css we
> need to add an extra white-space: normal; in this class. Or shall that
> be kept only in scikit-learn css files?

This sounds something reasonnable. And other people might be using
bootstrap. I'll add it to the PR.

GaelVaroquaux:
I've addressed all the comments, in particular the one on the width of the buttons. It now renders as such:
image

Titan-C:
I tested this PR on top of my new PR #147 that allows to change themes on the fly. It looks good and uniform on the classic theme and also the read the docs theme.

But one last nitpick. In the nature theme which is at the base for nilearn and scikit-learn you get this quite often
image

The jupyter notebook is usually 3 lines height and the Python source is 2. The boxes are unequal. Should we care?  adding `min-height: 3em;` might be enough `div.sphx-glr-download a` to give some extra uniformity.

GaelVaroquaux:
Hum, my fix doesn't fix what you had noticed above, as it fixes only for the download in the gallery. Could you please give me the steps to reproduce? I haven't managed.

Titan-C:
```bash
git checkout -b css_download_multitheme master
git pull https://github.com/Titan-C/sphinx-gallery.git sphinx-themes
git pull https://github.com/GaelVaroquaux/sphinx-gallery.git css_download
cd doc
make html theme=nature
```
This will, in a new branch, load my changes, then yours. On that branch you can build the docs under any theme, nature is the case of my snapshot. Test over the actual examples, `just_code.py` in the snapshot

lesteve:
The PR with themes has been merged into master so you just need to rebase on master to be able to test with the nature theme.

GaelVaroquaux:
> The PR with themes has been merged into master so you just need to rebase on master to be able to test with the nature theme.

Yes, I got it, and I can reproduce. I just haven't had time. Maybe this
morning.

GaelVaroquaux:
I've pushed a change that does vertical alignment:

image

GaelVaroquaux:
I've just pushed one extra set of changes that means that if the screen is wide enough, the buttons will be wider and they won't be 3 lines-long. On my screen:

image

Titan-C:
Great! 
But again a new Nitpick. I see in the read the docs theme
buttons

When the file name is just too long, and because this boxes are `code` html tags. The read the docs puts `overflow-x: auto;` to them. Should we add to default `div.sphx-glr-download code.download` an extra `overflow-x: hidden;`? or put that into our theme overlay?

GaelVaroquaux:
Reproduced and fixed. I also removed the inner bevel of the buttons created by the RTD theme:
image

Titan-C:
I see you are cutting the notebook extension into `ipyn` and `b`

GaelVaroquaux:
> I see you are cutting the notebook extension into ipyn and b

Yes, because the word is too long, the browser needs to cut it.

Titan-C:
I was thinking it we need to have the file name in the button at all? The original reason the file names where there is to declare the download link. But now we have buttons. So maybe to keep visual uniformity  just have download buttons with out writing the filename on it. That will make it easier too, to setup the CSS styles.
It is then the user's browser which will uncover the filename upon clicking.

GaelVaroquaux:
The reason to have the filename is that people download the file, it then
goes in a huge directory with many many files downloaded (eg ~/Downloads
under Ubuntu) and people don't know which one it is. Of course, they can
find this out by looking at the "download" list of their browser, but it
makes a case for struggles and errors.

lesteve:
I know very little about CSS but it seems very hard to have the download buttons render well in many different themes (and I am guessing it depends a lot of the number of characters in the filename ...). I feel it is turning a little bit into bikeshedding.

I also feel we should prioritise the one on readthedocs (since this is our doc) and maybe the default sphinx theme. For other themes I am afraid people will have to tweak their CSS if they want to achieve perfection.

Having said that I am fine merging this as is and revisit later if it turns out to be too visually annoying.

GaelVaroquaux:
OK, you're pushing my web skills to their limits!

Here is an improvement: I am using the fact that the last element of the download button is the file name to make it smaller: screenshots, first with the nature theme, and then with the rtd theme

image

image

image

image

GaelVaroquaux:
I only partly agree with @lesteve. In a sense, yes it's going to be impossible to get everything perfect for every theme. However, people tend not to be very good at CSS, so the more work we do, the more they will love us. In addition, we need to have flashy selling points, and if using sphinx-gallery makes a gallery look awesome with no work, people will feel more driven to adopt us.

To illustrate that the current CSS works quite well with yet another theme, here is how it looks in the scipy-lectures:

image

image

Titan-C:
> Of course, they can find this out by looking at the "download" list of their browser, but it makes a case for struggles and errors.

Usually the browser handles that very well listing your last download.

The smaller fonts work well enough for me. 

+1

GaelVaroquaux:
>     Of course, they can find this out by looking at the "download" list of their browser, but it makes a case for struggles and errors.

> Usually the browser handles that very well listing your last download.

My comment is not a theoretical one. It's an experimental observation
based on teaching many classes to non technical people (eg master
students in psychology) who don't know well how to use a computer.

lesteve:
@Titan-C gave its +1, so let's merge this one!

A for effort on this one @GaelVaroquaux !

PR #145, opened 2016-09-15
[MRG] CSS: add a CSS class to links in code

GaelVaroquaux:
The goal is to enable downstream to style the doc-resolv links in code examples specially.

### DIFF
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -151,3 +151,25 @@ p.sphx-glr-signature a.reference.external {
   padding: 3px;
   font-size: 75%;
 }
+
+a.sphx-glr-code-links:hover{
+    text-decoration: none;
+}
+
+a.sphx-glr-code-links[title]:hover:before{
+    background: rgba(0,0,0,.8);
+    border-radius: 5px;
+    color: white;
+    content: attr(title);
+    padding: 5px 15px;
+    position: absolute;
+    z-index: 98;
+    width: 15.5em;
+    word-break: normal;
+    display: inline-block;
+    text-align: center;
+    text-indent: 0;
+    margin-left: -5em;
+    margin-top: 1.2em;
+}
+
--- a/sphinx_gallery/docs_resolv.py
+++ b/sphinx_gallery/docs_resolv.py
@@ -345,7 +345,8 @@ def _embed_code_links(app, gallery_conf, gallery_dir):
                                                     gallery_dir))
 
     # patterns for replacement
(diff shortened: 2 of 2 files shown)


GaelVaroquaux:
To make this functionality more discoverable, I've added a title to the link, that hints to what the link points to, and CSS code so that the title appears with no delay when the user hovers:
image

Titan-C:
this looks very nice

lesteve:
Very nice indeed!

@Titan-C I don't remember does readthedocs build on each PR or only on each commit into master? It would be nice to see the generated sphinx-gallery doc on PRs ...

lesteve:
Merged thanks.

lesteve:
Hmmm actually I should have tested this a bit more I get two popups for some reason (both locally and on readthedocs). Here is a snapshot:

sphinx-gallery

GaelVaroquaux:
> Hmmm actually I should have tested this a bit more I get two popups for some
> reason (both locally and on readthedocs).

OK. I'll fix this ASAP.

I actually really hate readthedocs because it makes it so hard to test
things without merging.

lesteve:
> I actually really hate readthedocs because it makes it so hard to test things without merging.

In this case I get this locally too though. In this case I'll blame myself for merging while blindly trusting someone ;-).

lesteve:
Having said that it'd be great if we could use the readthedocs theme so that it looks the same locally and on readthedocs ... no idea whether it's difficult or not.

GaelVaroquaux:
>   That's a bummer, because "title" is used by readers for blind people.

Apparently that's not always the case.
https://silktide.com/i-thought-title-text-improved-accessibility-i-was-wrong/

I've pushed a fix on
https://github.com/sphinx-gallery/sphinx-gallery/pull/146

Titan-C:
> I don't remember does readthedocs build on each PR or only on each commit into master? It would be nice to see the generated sphinx-gallery doc on PRs ...

readthedocs only builds after is pushed into master and for every commit to which version tag is put.

I'm also thinking in having the docs build in CircleCI as you do in nilearn/scikit-learn

Titan-C:
> it'd be great if we could use the readthedocs theme so that it looks the same locally and on readthedocs ... no idea whether it's difficult or not.

I do that manually from time to time. I can set it up as part of our configuration. It will of course require you have the readthedocs theme installed(is in conda and pip)

GaelVaroquaux:
> I do that manually from time to time. I can set it up as part of our
> configuration. It will of course require you have the readthedocs theme
> installed(is in conda and pip)

Would it be possible to have it as a different make target? I actually
think that it's useful to build under different themes, to avoid having
choices that are too specific to a given theme.

PR #147, opened 2016-09-18
Sphinx themes

Titan-C:
In response to the discussion in https://github.com/sphinx-gallery/sphinx-gallery/pull/145#issuecomment-247575808

Here I propose to change the html theme passing an extra variable to the make target. In this case on can build locally the read the docs theme by using

```bash
make html theme=rtd
```

Also all sphinx themes become available. Thus to change to nature theme is as simple as
```bash
make html theme=nature
```

If one does not pass the `theme=value` instruction the `classic` default is taken.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,17 @@ Change Log
 git master
 ----------
 
+Developer changes
+'''''''''''''''''
+* Support on the fly theme change for local builds of the
+  Sphinx-Gallery docs. Passing to the make target the variable `theme`
+  builds the docs with the new theme. All sphinx themes are availible
+  plus read the docs online theme under the value `rtd` as shown in this
+  usage example.::
+
+    $ make html theme=rtd
+
+
 v0.1.4
 ------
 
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -4,6 +4,17 @@ Change Log
 git master
 ----------
 
+Developer changes
+'''''''''''''''''
+* Support on the fly theme change for local builds of the
+  Sphinx-Gallery docs. Passing to the make target the variable `theme`
+  builds the docs with the new theme. All sphinx themes are availible
```
lesteve: available

GaelVaroquaux:
I just saw with @titan-C (met him on the bus), and here is another
solution to the same problem: how about replacing "Jupyter notebook
files" by "Jupyter notebooks"? That should both fix the problem of
vertical alignment, and make the buttons smaller (or rather avoid them
getting bigger) which I think is a good thing.

lesteve:
@GaelVaroquaux I guessed you meant to comment on #144 rather than this one.

GaelVaroquaux:
Indeed :$

lesteve:
Good stuff, let's merge this one!

PR #148, opened 2016-09-24
[MRG] Notebook CLI

Titan-C:
This is a command line utility to expose sphinx gallery notebook generator.
I manually tested it on bash and zsh.

calling

``` bash
sphx_ex2nb.py *.py
```

will convert all python files into notebooks


Titan-C:
@GaelVaroquaux Any comments on this one?

lesteve:
Just curious, are you using the sphx_ex2nb.py yourself?

setup.py
```diff
@@ -30,7 +30,7 @@
     packages=find_packages(),
     package_data={'sphinx_gallery': ['_static/gallery.css', '_static/no_image.png',
                                      '_static/broken_example.png']},
-    scripts=['bin/copy_sphinxgallery.sh'],
+    scripts=['bin/copy_sphinxgallery.sh', 'bin/sphx_ex2nb.py'],
```
lesteve: Maybe something less cryptic like sphx_glr_example_to_notebook? Not 100% convinced so better suggestions welcome! My point is that trying to shorten names as much as possible does not make too much sense since you have tab-completion these days ...

bin/sphx_ex2nb.py
```diff
@@ -0,0 +1,41 @@
...
+                    help='Input Python file script to convert. '
+                    'Supports multiple files and shell wildcards'
+                    ' (e.g. *.py)')
+
+
+def main():
+    args = parser.parse_args()
+
+    for src_file in args.file:
+        file_name = os.path.basename(src_file)
```
lesteve: For consistency I think this should be `src_filename` and `filename`. I tend to use `filename` for file names and `file` for object files (i.e. as returned from `open`) to make it easier to differentiate between the two.

sphinx_gallery/notebook.py
```diff
@@ -118,6 +126,8 @@ def __init__(self, file_name, target_dir):
         self.write_file = os.path.join(target_dir, self.file_name)
         self.work_notebook = ipy_notebook_skeleton()
         self.add_code_cell("%matplotlib inline")
+        self.fill_notebook(script_blocks)
+        self.save_file()
```
lesteve: It's generally considered best practice to have minimal logic in the constructor. In this case I would rather have an API like:

```py

nb = Notebook(source_filename)
nb.to_notebook(target_filename)
```

The parsing of the blocks from `source_filename` could happen in the constructor.

Titan-C:
Comments addressed. Notebook generation is simpler now as well. And I have included test for the command line utility

.travis.yml
```diff
@@ -68,6 +68,7 @@ before_script:
 script:
     - if [ "$DISTRIB" == "ubuntu" ]; then python setup.py nosetests; fi
     - if [ "$DISTRIB" == "conda" ]; then nosetests; fi
+    - nosetests bin # to test jupyter notebook generator CLI
```
lesteve: not very conventional ... maybe it would be great if the scripts in bin, was something super simple like:

```py
from sphinx_gallery.notebook import main

if __name__ == '__main__':
    main()
```

This way you can still test the parser in sphinx_gallery/tests.

Titan-C: I put this line because I could not get nosetest to scan the `bin` folder. Pytest did not have issues on that. But indeed I'll move everything to the notebook parsing file.

Titan-C:
Rebased to master and comments addressed.  Cleaned the commit history too

sphinx_gallery/tests/test_notebook.py
```diff
@@ -66,3 +75,39 @@ def test_convert():
...
+            assert_equal(json.load(fname), example_nb)
+
+###############################################################################
+# Notebook shell utility
+
+
+class CommandLineTest(TestCase):
+    """Test the Sphinx-Gallery python to Jupyter notebook converter CLI"""
+
+    def test_with_empty_args(self):
```
lesteve: why still the class methods rather than functions?

lesteve: I push a change in your branch because apart from this it is good for merging I think.

lesteve:
Merging this one, thanks!

Titan-C:
Thanks for the changes. I'm not sure if you got the comment about using the class methods for test. My idea behind it is that that is how unittest does in the standard library tests for exceptions. And since we plan to move to pytest. I did not wanted to include more use to nosetest.

lesteve:
Oops sorry, I missed your reply somehow. In order not to have camelCase in our tests I would either the same method as in scikit-learn so that we can have `from sphinx_gallery.testing import assert_raises_regex` or it seems like pytest has its own way of checking exception message (from http://doc.pytest.org/en/latest/assert.html):

```py
def test_match():
    with pytest.raises(ValueError) as excinfo:
        myfunc()
    excinfo.match(r'.* 123 .*')
```

lesteve:
> Cleaned the commit history too

Don't worry about that too much, to be honest I tend to use the "Squash and merge" button these days and edit the commit message if they were too many commits and/or too many useless commit messages.

PR #151, opened 2016-10-09
[MRG] Documentation Auto API

### LINKED ISSUE #126: BUG: Spamming of warnings about `.exclude`
Currently my build gets hundreds of these lines:

```
/home/larsoner/custombuilds/mne-python/doc/generated/mne.write_labels_to_annot.rst:8: SEVERE: Problems with "include" directive path:
InputError: [Errno 2] No such file or directory: 'generated/mne.write_labels_to_annot.examples'.
```

It makes it difficult to find the actual useful Sphinx warnings.

I think it comes from:

```
doc/_templates/module.rst:   .. include:: {{fullname}}.{{item}}.examples
```

Is there a way to have the template only include the `.examples` file if it exists, or by default filter out any warnings in `sphinx` that occur of this form?

Titan-C:
This PR improves the documentation on getting Sphinx gallery backreferences into auto documented API. Closes #126  

There are some incompatible changes. The backreferences are disabled by default. The user must activate them manually by specifying the directory where to save them.


GaelVaroquaux:
I am a bit worried about those backward incompatible changes. We now have a sizeable number of packages who use us, and we are going to break their documentation build the minute we release. 

Is there any way to have a depreciation cycle? Could we have step by step instructions to migrate from the old way to the new way?

Titan-C:
> Is there any way to have a depreciation cycle? Could we have step by step instructions to migrate from the old way to the new way?

I would also like to make a deprecation cycle. Maybe I can leave the old default value and put a warning that I'll now require the variable to be configured. The migration is just set up the value, the complete goal of this to make the user more aware of this configuration as it is error prone see https://github.com/sphinx-gallery/sphinx-gallery/issues/126#issuecomment-237981328, many places have to be configured coherently. 

This PR is mostly a documentation change to deal with the confusion of setting up the sphinx-autodoc and sphinx-autosummary to work with sphinx gallery backreferences, read #126. 

My approach to it was to force the user to become conscious of this process, as I don't see a way to automate this process, see https://github.com/sphinx-gallery/sphinx-gallery/issues/126#issuecomment-238181976.

lesteve:
Some conflicts to fix here.

Titan-C:
Rebased to master. New docs on the backreferences configuration. Proof reading required.

Titan-C:
Can I get proof readers on this one? @lesteve @Eric89GXL 
This is a documentation change, and behavior change of the gallery to deal with the issues from #126 about configuring the backreferences to be displayed automatically generated APIs

doc/advanced_configuration.rst
```diff
@@ -114,56 +114,65 @@ dictionary within your Sphinx ``conf.py`` file :
...
+=======================================
 
-Sphinx-Gallery also enables you, when documenting your modules, to
-reference to the examples that use that particular class or
-function. For example if we are documenting the numpy.linspace
-function its possible to embed a small gallery of examples using it
-like this:
+Sphinx-Gallery enables you, when documenting your modules, to
+reference to the examples that use a particular function. For example
+if we are documenting the ``numpy.exp`` function its possible to embed
```
Eric89GXL:
```
:func:`numpy.exp`
```
This should work with sphinx crossref

Eric89GXL:
I can confirm that this change gets everything working for MNE:
```
-    'mod_example_dir': 'generated',
+    'backreferences_dir': 'generated',
```
And the changes make sense to me.

Eric89GXL:
Also, having both config values (old and new) I get the same error:
```
Exception occurred:
  File "/home/larsoner/custombuilds/sphinx-gallery/sphinx_gallery/gen_gallery.py", line 99, in parse_config
    + backreferences_warning)
ValueError: Old configuration for backreferences detected 
using the configuration variable `mod_example_dir`
```
If both are present, I think the new one should be used and no warning issued. That way I can just update MNE to have redundant old and new keys, it will work for people with old and new sphinx_gallery installs, and somewhere down the line I'll clean up the `conf.py`.

lesteve:
> And the log basically just has the ValueError traceback. Is this intentional? I expected a warning, but successful build.

I seem to remember a discussion about this but I can't find it. IMHO warnings are easily ignored especially in the middle of a noisy output like the one sphinx + sphinx-gallery produces. In principle, I would rather have an error saying that something is very likely configured wrong somewhere and please read the doc or remove this line. What I suspect is that nobody has this setting right, but (lucky) people using exactly the same setup as scikit-learn and @Eric89GXL who asked about the warnings he was getting and thanks to @Titan-C managed to set everything correctly.

Having said that, if we think it is better to warn for a few releases and then have an error I am fine with this as well.

Most importantly what I think is important is that the warning/error message clearly states:
* the feature that is in play here: mostly having mini galleries in the reference API. A link to how it looks in sphinx-gallery is needed because I don't think explaining it with words is good enough.
* if you don't care about this feature, do this (I am guessing removing one line in conf.py is enough?).
* If you care about the feature and want a quick fix, try to replace 'mod_example_dir' by 'backreferences_dir'
* If the quick fix does not produce the mini-galleries then go and read very carefully this section of the doc.

> If both are present, I think the new one should be used and no warning issued. That way I can just update MNE to have redundant old and new keys, it will work for people with old and new sphinx_gallery installs, and somewhere down the line I'll clean up the conf.py.

From this I am guessing that you don't have sphinx-gallery bundled in externals (the latter is what most projects do AFAIK). If that is the case would you consider requiring a recent enough version of sphinx_gallery in your conf.py (once sphinx-gallery 0.1.8 is released) ?

Eric89GXL:
> From this I am guessing that you don't have sphinx-gallery bundled in externals (the latter is what most projects do AFAIK). If that is the case would you consider requiring a recent enough version of sphinx_gallery in your conf.py (once sphinx-gallery 0.1.8 is released) ?

We could do that, sure. Allowing the old value to seems like a reasonable low-cost way to maintain better backward compatibility/transition with this change, but it probably (hopefully) won't affect too many people anyway.

Titan-C:
reviewed and rebased

sphinx_gallery/gen_gallery.py
```diff
@@ -67,13 +67,9 @@ def clean_gallery_out(build_dir):
...
-def generate_gallery_rst(app):
-    """Generate the Main examples gallery reStructuredText
-
-    Start the sphinx-gallery configuration and recursively scan the examples
-    directories in order to populate the examples gallery
-    """
-    print('Generating gallery')
+def parse_config(app):
+    """Process the Sphinx Gallery configuration"""
+    # TODO: Test this behavior.
```
Titan-C: This goes for the future

Titan-C:
The backreferences configurations will now only issue warnings. Old and new configurations should work.

Eric89GXL:
Now with both config entries present, I get a warning about using the new config value, which we can live with for a while to get people migrated to the newer version.

LGTM, thanks @Titan-C

sphinx_gallery/gen_gallery.py
lesteve: @Titan-C sorry I don't have time to review this PR in detail so feel free to merge it. 

It does not have to happen in this PR but tests for the both old and new configuration are a must. It is very easy to break this kind of things without realizing and the impact on the user that encounters the regression is very bad IMO.

PR #152, opened 2016-10-09
Update travis to drop ubuntu 12 test

Titan-C:
while working on #151, I hit again the problem of tests not passing on ubuntu. Problem is the seaborn support and requirement for pandas cython for pandas.
I have created a more descriptive test matrix, which loads ubuntu 12 and 14. Ubuntu 14 does not have seaborn either but the pandas version is updated enough to work with seaborn.
After few tests I can't get ubuntu 12 to work without upgrading all python packages with pip.

Travis still claims ubuntu 14 is a test feature and builds on it take longer. That is the reason of the container matrix. All test on conda are ubuntu 12 containers and for ubuntu 14 specific packages the environment is different.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,25 +3,33 @@ sudo: false
 addons:
   apt:
     packages:
-    - python-numpy
-    - python-matplotlib
-    - python-scipy
-    - python-pip
-    - python-coverage
     - xvfb # For mayavi headless
 
 virtualenv:
         system_site_packages: true
 
-env:
-    matrix:
-        - DISTRIB="ubuntu" PYTHON_VERSION="2.7"
-        # This environment tests the oldest supported anaconda env
-        - DISTRIB="conda" PYTHON_VERSION="2.6"
-        # This environment tests is for mayavi
-        - DISTRIB="conda" PYTHON_VERSION="2.7"
-        # This environment tests the newest supported anaconda env
-        - DISTRIB="conda" PYTHON_VERSION="3.4"
+matrix:
+  include:
+    - os: linux
+      dist: trusty
+      sudo: required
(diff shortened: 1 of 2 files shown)


lesteve:
Sorry I am not following, what is the point of testing two versions of Ubuntu?

lesteve:
Also these days you can get pandas wheels. The reason why you are not getting them is probably because Ubuntu comes with `pip` that is too old. You can try doing `pip install -U pip` to bring it up-to-date (possibly with sudo in your case). Afterwards you should be able to do `pip install seaborn` without any problem.

GaelVaroquaux:
You should drop support for Ubuntu 12, and for Python 2.6.

Titan-C:
> Sorry I am not following, what is the point of testing two versions of Ubuntu?

Travis is still ubuntu 12 and that is the stable and fast environment to
test, ubuntu 14 is still beta and takes longer.

The tests in on ubuntu 12 is there just because I was testing if I could
still get it to work. But I'll drop it entirely.

> Also these days you can get pandas wheels. The reason why you are not
> getting them is probably because Ubuntu comes with pip that is too
> old. You can try doing pip install -U pip to bring it up-to-date
> (possibly with sudo in your case). Afterwards you should be able to do
> pip install seaborn without any problem. 

Yes, as I wrote earlier. I can not get everything to run on the native
ubuntu environment without updating everything on pip. If everything is
new on the virtual environment then the test shall not be called ubuntu.

> You should drop support for Ubuntu 12, and for Python 2.6.

I'll drop both then.

Titan-C:
Since master is failing with the latest update this PR is a bit urgent. 
I rebased this PR. It drops support for Ubuntu 12 and Python 2.6. Includes testing on Ubuntu 14 packages.

PR #157, opened 2016-10-26
Warn the user to set the Agg backend

### LINKED ISSUE #153: Document that the matplotlib backend will be set to Agg
While trying to use sphinx-gallery to document my mplcursors (https://mplcursors.readthedocs.io/en/stable/) package, I got the following (well-known) warning:

```
UserWarning:  This call to matplotlib.use() has no effect
because the backend has already been chosen;
matplotlib.use() must be called *before* pylab, matplotlib.pyplot,
or matplotlib.backends is imported for the first time.
```

Turns out that in my `conf.py`, I already import `mplcursors/__init__.py` (in order to get the version set by `versioneer` out of it), and that this triggers an import of `matplotlib.pyplot` (which I can actually get rid of, but that's irrelevant here).  A simple solution was simply to set the backend to `Agg` before importing `mplcursors`.

Documenting this point could perhaps save some head-scratching for others too.

Titan-C:
closes #153 

This includes an additional warning message when loading the Agg backend if matplotlib has loaded any other backend before.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -55,23 +55,25 @@ except ImportError:
 
 from io import StringIO
 
-mpl_backend_msg = """
-
-Sphinx-Gallery relies on the matplotlib Agg backend to render figures
-and write them to files. You are currently using the {} backend.
-Sphinx-Gallery will terminate the build now, because changing backends
-is not well supported by matplotlib. We advise you to move
-sphinx_gallery imports before any matplotlib-dependent import. Moving
-sphinx_gallery imports at the top of your conf.py file should fix this
-issue.\n\n"""
-
 try:
     # make sure that the Agg backend is set before importing any
     # matplotlib
     import matplotlib
-    matplotlib.use('Agg')
-    if matplotlib.get_backend() != 'agg':
-        raise ValueError(mpl_backend_msg.format(matplotlib.get_backend()))
+    matplotlib.use('agg')
+    matplotlib_backend = matplotlib.get_backend()
+
+    if matplotlib_backend != 'agg':
+        mpl_backend_msg = (
+            "Sphinx-Gallery relies on the matplotlib 'agg' backend to "
+            "render figures and write them to files. You are "
+            "currently using the {} backend. Sphinx-Gallery will "
+            "terminate the build now, because changing backends is "
+            "not well supported by matplotlib. We advise you to move "
+            "sphinx_gallery imports before any matplotlib-dependent "
+            "import. Moving sphinx_gallery imports at the top of "
+            "your conf.py file should fix this issue")
+
+        raise ValueError(mpl_backend_msg.format(matplotlib_backend))
 
     import matplotlib.pyplot as plt
 except ImportError:
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -55,11 +55,24 @@ except ImportError:
 
 from io import StringIO
 
+mpl_backend_msg = """
+
+Sphinx-Gallery relies on the matplotlib Agg backend to render figures
+and write them to files. You are currently using the {} backend.
+Sphinx-Gallery will terminate the build now, because changing backends
+is not well supported by matplotlib. We advise you to move
+sphinx_gallery imports before any matplotlib-dependent import. Moving
+sphinx_gallery imports at the top of your conf.py file should fix this
+issue.\n\n"""
+
 try:
     # make sure that the Agg backend is set before importing any
     # matplotlib
     import matplotlib
     matplotlib.use('Agg')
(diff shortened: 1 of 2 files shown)


lesteve:
Why not an error, i.e. is there a use case where keeping on going with a backend other than Agg is useful? I haven't tried but for an interactive backend I would expect the build to stop at the first `plt.show()` ...

Titan-C:
> Why not an error, i.e. is there a use case where keeping on going with a backend other than Agg is useful? I haven't tried but for an interactive backend I would expect the build to stop at the first plt.show()

I thought of a warning because the only reason to use Agg is to avoid the pop up. The use case is none. Yes interactive backend will pop the figures on `plt.show()` and the user will have to close them by hand.
I'll make it a Value error

GaelVaroquaux:
> Why not an error, i.e. is there a use case where keeping on going with a
> backend other than Agg is useful?

SVG would work.

> I haven't tried but for an interactive backend I would expect the build
> to stop at the first plt.show() ...

I think that we monkey-patch matplotlib to avoid this.

The real reason that we set the backend to agg is to avoid using an X
server on headless servers.

Unless proven that other backends don't work, I am +1 for warning rather
than error.

Titan-C:
>> Why not an error, i.e. is there a use case where keeping on going with a
 backend other than Agg is useful?

> SVG would work.

I actually gave it a try. For now writing images in a different format than PNG is not really supported. Main immediate problem is thumbnails(we use pillow to resize them), I actually hacked it to just pass the svg as they don't really need to be resized. With PDF, web browsers can't display pdf images only as embed or opening them individual. Thus in both cases it does not look nice. 

But my new question is Agg can generate PNG and SVG and EPS and PDF(And at some point we could support all those formats), why bother selecting the other non interactive backends?

>> I haven't tried but for an interactive backend I would expect the build
 to stop at the first plt.show() ...

> I think that we monkey-patch matplotlib to avoid this.

Monkey-patching seem unnecessary hard. I tried loading matplotib in interactive mode. And although it works well in the Python/IPython interpreter to continue execution after plt.show() is called. It doesn't when running sphinx-build and the figures open in my display, it is not until I close all of them that the build can continue with the next example. But most important (except the seaborn example) once I close the image there are no figure objects to save, thus no figure gets saved and we have no gallery.

> Unless proven that other backends don't work, I am +1 for warning rather
than error.

I thought of the warning as a better alternative than raising the error, because I have no idea of the users use cases. I already updated the PR to give the error some hours ago, going back is not a major trouble. The question I now repeat is why use other backend if Agg outputs to all other formats? Well Cairo does too, but that one is more complicated to install as a dependency. So it a user needs that one he for sure knows why.

I believe the main point of this PR and issue #153 is to tell the user that he is unintentionally loading pylab or pyplot before Sphinx-Gallery can set the backend to AGG. So that he can change his configuration.

@anntzer any comment on this?

anntzer:
Clearly, interactive backends do not work (because they pop up windows); I think just erroring out clearly with them should be enough.
Whether you want to bother supporting other noninteractive backends is up to you...

sphinx_gallery/gen_rst.py
```diff
@@ -55,11 +55,25 @@ def prefixed_lines():
...
+is not well supported by matplotlib. We advise you to review your
+Sphinx setup to use the Agg backend.\n\n"""
+
 try:
     # make sure that the Agg backend is set before importing any
     # matplotlib
     import matplotlib
+    # Check if we've already set up a backend
+    current_backend = matplotlib.rcParams['backend']
+    if 'matplotlib.backends' in sys.modules and current_backend.lower() != 'agg':
```
lesteve: I think this is simpler:

```py

matplotlib.use('Agg')

if matplotlib.get_backend() != 'Agg':
    raise ValueError(...)
```

sphinx_gallery/gen_rst.py
```diff
@@ -55,11 +55,25 @@ def prefixed_lines():
 
 from io import StringIO
 
+mpl_backend_msg = """
+
+Sphinx-Gallery relies on the matplotlib Agg backend to render figures
+and write them to files. You are currently using the {} backend.
+Sphinx-Gallery will terminate the build now, because changing backends
+is not well supported by matplotlib. We advise you to review your
+Sphinx setup to use the Agg backend.\n\n"""
```
lesteve: The recommendation should be "Move the sphinx_gallery imports at the top of your conf.py file" rather than "review your configuration", Maybe a more nuanced thing like "sphinx_gallery imports should be before any matplotlib-dependent import. Moving the sphinx_gallery imports at the top of your conf.py file should fix this issue" or something better phrased.

lesteve:
@Titan-C I pushed some minor changes into your branch (github allows maintainers to do that). I'll wait for the CI to come back green and merge this one.

lesteve:
OK merging this one, thanks @Titan-C.

PR #161, opened 2016-11-08
MRG: Dont close figures until end of script

Eric89GXL:
Sometimes it is necessary to have figures that are not closed within a given code-block. This modifies the gallery behavior to leave the figures open until the script is done. This is useful e.g. when creating a figure in one code block, and querying some parameter from it later. Our particular use case came from creating a Mayavi figure in one code block, and needing to get the viewport transformation from it in another code-block.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -335,11 +335,23 @@ def save_figures(image_path, fig_count, gallery_conf):
     """
     figure_list = []
 
+    # We don't actualy *close* figures until the entire script is done
+    # executing (in clean_modules) so that users have access to the figures
+    # in case they need to query properties (e.g., for Mayavi scene, the
+    # view transformation) in different code blocks.
+    # Instead we monkey-patch _sg_captured to the figure/scene.
+
+    # In the future we could even add an option to the scripts to allow
+    # recapturing within a code-block like (# sg:recapture1) could
+    # recapture figure number 1 even if it's already captured.
+
     fig_managers = matplotlib._pylab_helpers.Gcf.get_all_fig_managers()
     for fig_mngr in fig_managers:
         # Set the fig_num figure as the current figure as we can't
         # save a figure that's not the current figure.
         fig = plt.figure(fig_mngr.num)
+        if hasattr(fig, '_sg_captured'):
+            continue
         kwargs = {}
         to_rgba = matplotlib.colors.colorConverter.to_rgba
         for attr in ['facecolor', 'edgecolor']:
@@ -351,6 +363,7 @@ def save_figures(image_path, fig_count, gallery_conf):
         current_fig = image_path.format(fig_count + fig_mngr.num)
         fig.savefig(current_fig, **kwargs)
         figure_list.append(current_fig)
+        fig._sg_captured = True
 
     if gallery_conf.get('find_mayavi_figures', False):
         from mayavi import mlab
@@ -360,12 +373,14 @@ def save_figures(image_path, fig_count, gallery_conf):
         mayavi_fig_nums = range(last_matplotlib_fig_num + 1, total_fig_num + 1)
 
         for scene, mayavi_fig_num in zip(e.scenes, mayavi_fig_nums):
+            if hasattr(scene, '_sg_captured'):
+                continue
             current_fig = image_path.format(mayavi_fig_num)
             mlab.savefig(current_fig, figure=scene)
+            scene._sg_captured = True
             # make sure the image is not too large
             scale_image(current_fig, current_fig, 850, 999)
             figure_list.append(current_fig)
-        mlab.close(all=True)
 
     # Depending on whether we have one or more figures, we're using a
     # horizontal list or a single rst call to 'image'.
@@ -513,7 +528,6 @@ def execute_code_block(code_block, example_globals,
     if not block_vars['execute_script']:
         return stdout, time_elapsed
 
-    plt.close('all')
     cwd = os.getcwd()
     # Redirect output to stdout and
     orig_stdout = sys.stdout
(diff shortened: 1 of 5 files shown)


Eric89GXL:
An alternative solution would be to do everything that is needed for a given figure within a single code block, but that's not as flexible as this solution.

sphinx_gallery/gen_rst.py
```diff
@@ -351,6 +363,7 @@ def save_figures(image_path, fig_count, gallery_conf):
         current_fig = image_path.format(fig_count + fig_mngr.num)
         fig.savefig(current_fig, **kwargs)
         figure_list.append(current_fig)
+        fig._sg_captured = True
```
lesteve: Rather than this monkey-patching I think it would be slightly better to have an additional argument in save_figures, which is a set that remembers the already captured figures.

Alternatively we could have a matplotlib index and a mayavi index keeping track of the number of the last plot saved.

lesteve:
It would be great to add tests, preferentially both with matplotlib and mayavi. Not a matplotlib expert, I could not find something that goes wrong because of closing all the figures at the end of a block, suggestions welcome.

sphinx_gallery/gen_rst.py
Eric89GXL: Yeah I could do that. I don't think indexing will work:
```
>> import matplotlib.pyplot as plt
>>> fig = plt.figure()
>>> fig.number
1
>>> plt.close('all')
>>> fig = plt.figure()
>>> fig.number
1
```
Instead I could maintain a list and check with `is`. The problem with that approach is that it might lead to memory issues, because the figures can't be garbage-collected even if they are closed by the user. So I'm actually starting to think that the monkey patching, although ugly, might be better.

Eric89GXL:
I'm not sure about whether it causes problems with `matplotlib` but it definitely does for `mayavi`. And if we're going to do it for one, we might as well do it for both to be safe.

I'll look into the Travis failure and try to add tests soon.

Eric89GXL:
Okay, tests added. I put in a bunch of fixes for Windows-related building because I'm on my Windows dev machine currently. It would be worth adding AppVeyor testing at some point and cleaning up Windows support properly.

agramfort:
travis is not happy :(

Eric89GXL:
All better

Titan-C:
I'll will start with a consistency check of how we want to present output, and how we should understand the notebook styled examples or code and outputs interleaved with prose. 
Let's put this simple script to test

```python
# -*- coding: utf-8 -*-
"""
The sin function with a phase shift
===================================

"""

# Code source: Óscar Nájera
# License: BSD 3 clause

import numpy as np

x = np.linspace(0, 2 * np.pi, 100)
y = np.sin(x)
plt.plot(x, y, label=r'$\sin(x)$')

# now put an angle
y = np.sin(x + np.pi / 4)
plt.plot(x, y, lw=2, label=r'$\sin(x+\pi/4)$')

plt.legend(loc=0)
plt.xlabel('$x$')
plt.ylabel('$\sin(x)$')
plt.show()
```

If we run this directly from the command line we get a window to pop up with the 2 curves in the same figure. If I run this in the gallery at master I get 2 figures, each one holds 1 curve.
2016-11-14-220135_404x978_scrot

And the rendered Jupyter notebook produces a similar output, as indeed we have defined 2 blocks.
2016-11-14-220638_398x882_scrot

Running this PR I get a not consistent result, at least in my linux PC the second block does not show any output, and that I assume is totally unwanted. I would have expected the second figure to be a composition of the first and second plot instructions. Showing 2 curves in the figure.
2016-11-14-221010_403x857_scrot

Now to the core of my discussion. I do see the intention in a tutorial to show how every new line of code adds to the drawn figure and so to show each code block as an intermediate step. But I feel in a notebook styled example we should restrict ourselves to plot individually what each block is declaring.  Especially since the examples are targeted for education purposes, that once the students try out the jupyter notebook file they shall get the same output. To me the easiest way to accomplish this is to just close all figures after saving them for each block.

Second thought, is that jupyter notebooks do allow to recover the plotted figure from the previous cell and compose on top of that, but the syntax is quite involved and I would be even more against to support that for our simple set of examples when just rewriting the plot instructions is so simple.

Third thought is that if we want to support accumulating the figures. I would rather first include a flag to tell sphinx-gallery to deviate from the behavior of making each block individual and stop closing figures after each block. Then make the author of the script be responsible for which figures are open(alive), and which he wants to close. In that case sphinx-gallery shall save and show for each block all active figures. I think this would also avoid the monkey patching of the figure object, simplify the numbering/naming of saved figures as is just a counter. And as put in the current PR comment, recapturing a picture in this case is trivial.

lesteve:
I do agree there is some tension between seeing the example as a Jupyter notebook and seing it as a script, for example when you do multiple `plt.plot` in different cells, it is not clear to me what the behaviour should be in this case.

In @Eric89GXL's case though, I think that we are not in such a tricky case. IIUC he plots two different figures in two different cells. The problem is that closing the first figure (let's call it `fig`) make the variable `fig` not usable any more (to be honest it could well be considered as a mayavi bug, but that may be a very naive interpretation of the problem given that I know close to nothing about mayavi). The second plot tries to use `fig`, not to plot on top of the first figure but to reuse its "viewport transformation" and that's where the problem happens.

I would be in favour of not closing the figures since that seems to be a useful work-around in this case, unless we can spot a real blocker in the fine case of trying to plot figures in different cells.

I would leave plotting on top of an existing figure and recapturing as undefined behaviour for now.

Titan-C:
> I would be in favour of not closing the figures since that seems to be a useful work-around in this case, unless we can spot a real blocker in the fine case of trying to plot figures in different cells.

But for mayavi or for matplotlib?

lesteve:
> But for mayavi or for matplotlib?

I guess for both, be it only for consistency reasons.

Eric89GXL:
Closing for now until consensus

lesteve:
I am +1 on closing the figures only at the end of the script.

@Eric89GXL what is your current work-around this in MNE-python?

Eric89GXL:
In the example where we needed to modify an existing (mayavi) figure, we instead recreated it in the next block and modified it.

lesteve:
/'//> In the example where we needed to modify an existing (mayavi) figure, we instead recreated it in the next block and modified it.

So you have some duplication (the lines to create the figure) in your example?

Eric89GXL:
Yeah, that's what we ended up going with I think

PR #166, opened 2016-11-16
MRG: Add __file__

Eric89GXL:
In one of my scripts I want to use `__name__` but in `master` it is not available.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -640,6 +640,8 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
         # Examples may contain if __name__ == '__main__' guards
         # for in example scikit-learn if the example uses multiprocessing
         '__name__': '__main__',
+        # And they might use the __file__ parameter
+        '__file__': example_file,
     }
 
     # A simple example has two blocks: one for the
--- a/sphinx_gallery/tests/test_gen_rst.py
+++ b/sphinx_gallery/tests/test_gen_rst.py
@@ -37,6 +37,8 @@ CONTENT = [
     '"""',
     '',
     '# and now comes the module code',
+    'assert __name__ == "__main__"',
+    'assert ".py" in __file__',
     'import logging',
     'import sys',
     'x, y = 1, 2',


Eric89GXL:
... and by `__name__` I mean `__file__`

Eric89GXL:
Travis is happy, ready for review/merge from my end

lesteve:
LGTM, merging thanks!

lesteve:
@Eric89GXL actually thinking about it more, `example_file` is something like `auto_examples/plot_seaborn.py`, is that what you expected? I would have thought that it should point to the example i.e. either `examples/plot_seaborn.py` or maybe even better its absolute path.

Eric89GXL:
Yeah pointing to the actual original python file would be best

lesteve:
> Yeah pointing to the actual original python file would be best

PR welcome ;-) !

GaelVaroquaux:
I just thought about something: it seems that __name__ is defined in an IPython notebook, but not __file__.

I would hence be in favor of not supporting __file__. It's going to confuse users that run things in IPython notebooks.

Eric89GXL:
That's true. However, any example that a given project has that uses `__file__` already has this problem, and thus we aren't making anything worse by supporting it, except maybe because of the IPython-notebook download button. Maybe we should disable that for any example file that uses `__file__`?

GaelVaroquaux:
> That's true. However, any example that a given project has that uses __file__
> already has this problem, and thus we aren't making anything worse by
> supporting it, except maybe because of the IPython-notebook download button.

The problem is that users are strongly demanding the IPython notebook
feature. These users expect examples to work in IPython notebooks.

> Maybe we should disable that for any example file that uses __file__?

Why do examples need the file? IMHO they should be rewritten. People will
copy-paste them in IPython notebooks and run into problems.

By supporting __file__ in sphinx-gallery you are just hidding this
problem.

The core problem is that Python now has several execution models and that
they are not compatible. It's a problem created by notebook, and not by
sphinx-gallery.

Eric89GXL:
Fair enough. Feel free to roll back the commit in `master`

GaelVaroquaux:
> Fair enough. Feel free to roll back the commit in master

Will you be able to adapt your example?

Eric89GXL:
Yeah, after some tweaking I got it working using a different tactic than `__file__`. I don't know if it will work in all cases, but at least it did here.

GaelVaroquaux:
I reverted, but stopped short of documenting this behavior, because if
it's a major blocker for many people we will support __file__.

lesteve:
Good catch @GaelVaroquaux !

PR #171, opened 2016-11-22
[MRG+1] Use plt.get_fignums rather than a private helper

lesteve:
This is available since matplotlib 0.98: http://matplotlib.org/api/api_changes.html#changes-for-0-98-x


GaelVaroquaux:
+1 to merge.

Issue #165, opened 2016-11-15
When README.txt does not finish with a new line: Thumbnail doesn't work; raw html is displayed


trbedwards:
I'm having issues trying to get Sphinx gallery to render thumbnails correctly.

When I build the documentation and open the gallery index.html on my web browser, some of the thumbnails are non-clickable. For those thumbnails that are non-clickable, I get this raw html displayed: `<div class=”sphx-glr-thumbcontainer” tooltip=”This script prints ‘hello world’ “>`. The problem is shown in the attached screenshot.

<img width="441" alt="sphinx_gallery_screenshot" src="https://cloud.githubusercontent.com/assets/1222553/20309730/0054168e-ab41-11e6-841f-0977f3ce2af4.PNG">

To reproduce this problem I created a simple directory structure like so:

```
example_sphinx_project/
  doc/
  examples/
    README.txt
    hello_script.py
    goodbye_script.py
  pysrc/
    mypackage/
      __init__.py
      mod1.py
      mod2.py
```

I ran `sphinx_quickstart` in `doc/`, using all the default options. In `index.rst` I added `Examples <auto_examples/index.rst>` to the toctree. In `conf.py` I added the following 
```
sphinx_gallery_conf = {
  'examples_dirs': '../examples',
  'gallery_dirs': 'auto_examples'
} 
```

Yet my thumbnails don't work properly. 

I've had this problem occur with sphinx_gallery 1.5.0 using sphinx 1.4.8 on Windows and sphinx 1.4.6 on Linux. 

Thanks

lesteve:
Thanks for the precise description! A good way to make it easier to reproduce would be to attach a zip of your `example_sphinx_project` to the issue.

trbedwards:
Ah sorry about that, for some reason I thought that .zip files weren't allowed on github! Here it is attached 
[example_sphinx_project.zip](https://github.com/sphinx-gallery/sphinx-gallery/files/592387/example_sphinx_project.zip)

lesteve:
It is a reasonably recent feature (maybe a few months ago?)

trbedwards:
I had a closer look html, and I noticed something off about the double quotes. They appear to be non-ascii. So I had a look at the html source. Here is the div for the non-working bit (hello_world.py):
```
[...]

Issue #164, opened 2016-11-15
Gallery directory without README.txt is skipped from execution but nevertheless generate an index


Titan-C:
Directories without a README.txt are not considered for the gallery and skipped from parsing or execution. This is the case when we want to have directories dedicated to hold data files that shall not be transformed into a gallery but are nevertheless needed nearby.

If the user forgets to put the readme.txt file in the main examples folder, a warning is printed and the build continues. The generated website is a page with the download buttons of the examples.

My expected behavior would be, since is the case of the main gallery directories, to stop the build immediately.

lesteve:
Yeah "error with user-friendly message" a lot greater than "warning" IMHO.

PR #175, opened 2016-11-27
Fix readme bugs

Titan-C:
This are simple fixes to issues #164 and #165


sphinx_gallery/gen_gallery.py
```diff
@@ -20,6 +20,12 @@
 from .docs_resolv import embed_code_links
 from .downloads import generate_zipfiles
 
+try:
+    FileNotFoundError
+except NameError:
+    # Python2
+    FileNotFoundError = IOError
```
lesteve: I forgot to tell you when I saw that in a different PR but this is a slightly WTF. Just raise IOError this way you don't have to do a Python2/3 dance.

lesteve:
One PR one issue is a really good habit, have I said that already ;-)?

PR #177, opened 2016-11-27
[MRG] Build gallery when example has SyntaxError

Titan-C:
As reported in #126 if an example has a SyntaxError gallery build
breaks. This provides the necessary changes to continue build and
raise an error at the end on the summary of errors

### DIFF
--- a/sphinx_gallery/py_source_parser.py
+++ b/sphinx_gallery/py_source_parser.py
@@ -11,6 +11,13 @@ import ast
 import re
 from textwrap import dedent
 
+SYNTAX_ERROR_DOCSTRING = """
+SyntaxError
+===========
+
+Example script with invalid Python syntax
+"""
+
 
 def get_docstring_and_rest(filename):
     """Separate `filename` content between docstring and the rest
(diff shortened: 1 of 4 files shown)


sphinx_gallery/gen_rst.py
```diff
@@ -186,6 +186,8 @@ def extract_intro(filename):
     """ Extract the first paragraph of module-level docstring. max:95 char"""
 
     docstring, _ = get_docstring_and_rest(filename)
```
lesteve: My original idea was (untested so maybe there is a problem with it), just change `get_docstring_and_rest` to return `"", full_content_of_the_file including docstring`. Then you can run later the example normally and you don't have more code changes to do. Maybe we require an non empty docstring in which case just return something like "A SyntaxError happened while parsing the example" as the docstring.

Titan-C:
Rebased and comments addressed.

lesteve:
Very nice, merging, thanks !

sphinx_gallery/py_source_parser.py
```diff
@@ -27,12 +34,16 @@ def get_docstring_and_rest(filename):
     # can't use codecs.open(filename, 'r', 'utf-8') here b/c ast doesn't
     # seem to work with unicode strings in Python2.7
     # "SyntaxError: encoding declaration in Unicode string"
-    with open(filename, 'rb') as f:
-        content = f.read()
+    with open(filename, 'rb') as fid:
```
lesteve: Just curious, why `fid`, i.e. what does it stand for?

Titan-C: `fid` is a name I constantly encounter, so I just took it as convention. The reason to change is that my linter is always complaining about single letter variables

lesteve: I have never seen `fid` for a file I have to say ...  maybe it is more used in a C context where `fopen` returns the file identifier.

GaelVaroquaux:
Usually the term in the Python world is "file handle".

PR #137, opened 2016-08-05
WIP: Fix link error to sphinx file

Eric89GXL:
Using the bleeding-edge (updated today) version of Sphinx, it looks like files that were originally:
```
_build/html/generated/mne.pick_types.rst.html
```
Are now put in:
```
_build/html/generated/mne.pick_types.html
```
The fix here just strips the `.rst` before appending the `.html`, but will probably only work on latest `sphinx`. Is this a bug that should be fixed by `sphinx-gallery`, or did Sphinx do something they shouldn't have done by changing the filename?

Issue #162, opened 2016-11-15
Add AppVeyor to test sphinx-gallery on Windows


lesteve:
From https://github.com/sphinx-gallery/sphinx-gallery/pull/161#issuecomment-260251762, looks like there are some problems.

lesteve:
The best would be to start from the nilearn appveyor.yml which is a lot simpler that the scikit-learn one.

PR #178, opened 2016-12-09
MRG: Error on sphinx 1.5

Eric89GXL:
Starting over on #137.


lesteve:
We know sphinx-gallery tests are not passing on Windows ... we have an issue https://github.com/sphinx-gallery/sphinx-gallery/issues/162 and an ongoing PR. Maybe do not spend too much time on getting the tests to pass on Windows.

I have heard of the astropy CI stuff but never looked into it in detail, what do they give us exactly?

Eric89GXL:
I removed the AppVeyor stuff, don't worry. PR forthcoming. CI-helpers simplify Anaconda installation, I'll show you shortly :)

In the meantime this should be good to go.

lesteve:
Hmm I do not get any errors on your PR with sphinx 1.5 but I get plenty of "HTTPError 404" printout like this:
```
Embedding documentation hyperlinks in examples..
        processing: plot_function_identifier.html
        processing: just_code.html
        processing: index.html
        processing: plot_strings.html
        processing: plot_gallery_version.html
The following error has occurred:

<HTTPError 404: 'Not Found'>
The following error has occurred:

<HTTPError 404: 'Not Found'>
The following error has occurred:

<HTTPError 404: 'Not Found'>
The following error has occurred:

<HTTPError 404: 'Not Found'>
        processing: plot_choose_thumbnail.html
The following error has occurred:

<HTTPError 404: 'Not Found'>
The following error has occurred:

<HTTPError 404: 'Not Found'>
        processing: plot_colors.html
The following error has occurred:

<HTTPError 404: 'Not Found'>
The following error has occurred:

<HTTPError 404: 'Not Found'>
The following error has occurred:

<HTTPError 404: 'Not Found'>
        processing: plot_raise.html
The following error has occurred:

<HTTPError 404: 'Not Found'>
        processing: plot_exp.html
The following error has occurred:

<HTTPError 404: 'Not Found'>
The following error has occurred:

<HTTPError 404: 'Not Found'>
        processing: plot_quantum.html
The following error has occurred:
```

You can see them in the Travis log actually:
https://travis-ci.org/sphinx-gallery/sphinx-gallery/jobs/182623353#L735

I am wondering where they are coming from ...

Eric89GXL:
I see those, too. I've been using the equivalent of this branch for a while and haven't noticed any broken links or anything... Does it only happen for you on this branch? Maybe I need to dig into Sphinx to figure out where it's coming from and have it throw a more explicit error.

lesteve:
Does it only happen for you on this branch?

I guess this is the only branch I have which works with sphinx 1.5 ;-) so it is hard to know whether sphinx 1.5 or this PR is the culprit for these noisy HTTPError messages.

Eric89GXL:
Looks like it's coming from `docs_resolv.py` around 374:
```
                    try:
                        link = doc_resolvers[this_module].resolve(cobj,
                                                                  full_fname)
                    except (HTTPError, URLError) as e:
                        print("The following error has occurred:\n")
                        print(repr(e))
                        continue
```

Eric89GXL:
More explicit error generation in latest commit:
```
Error in /home/travis/build/sphinx-gallery/sphinx-gallery/doc/_build/html/tutorials/plot_notebook.html for module numpy ({'module_short': 'numpy', 'name': 'linspace', 'module': 'numpy'}):
```

Eric89GXL:
... so the underlying problem is probably in `SphinxDocLinkResolver` somewhere.

.travis.yml
```diff
@@ -45,7 +45,7 @@ install:
     # force no mkl because mayavi requires old version of numpy
     # which then crashes with pandas and seaborn
     - if [ "$DISTRIB" == "conda" ]; then
-        conda create --yes -n testenv python=$PYTHON_VERSION pip nomkl numpy
+        conda create --yes -n testenv python=$PYTHON_VERSION pip nomkl numpy scipy
```
GaelVaroquaux: I am puzzled: why was adding scipy necessary for this fix?

Eric89GXL: There were some `seaborn` errors without it I thought

Eric89GXL:
The problem seems to be that SphinxDocLinkResolver tries to get the link to e.g. http://docs.scipy.org/doc/numpy-1.9.1/reference/generated/numpy.html but it doesn't exist. Now it gives a more informative error:

https://travis-ci.org/sphinx-gallery/sphinx-gallery/jobs/182643937#L1139

In the meantime, this gets Travis passing and SG at least more usable. I'm not sure how to fix the underlying SphinxDocLinkResolver issue. Looks like that's code by @Titan-C?

GaelVaroquaux:
@Eric89GXL : I do note that this gets travis passing, and I didn't see anything that looked wrong to me, so overall I am :+1: for the merge.

But still, I don't get why we need scipy on the CI. Maybe you answered this question, but I didn't understand the answer, then.

lesteve:
> But still, I don't get why we need scipy on the CI. Maybe you answered this question, but I didn't understand the answer, then.

We install seaborn so getting pandas and scipy automatically via the dependency resolver anyway.

GaelVaroquaux:
> We install seaborn so getting pandas and scipy automatically via the
> dependency resolver anyway.

OK. Good point.

Still puzzled but it's not really important.

+1 from my side.

Eric89GXL:
I added a new commit to removes the SciPy change, and it works fine. I thought it threw an error at one point but I must have misread it!

Eric89GXL:
(Also FWIW `python-scipy` is installed in the Ubuntu distro version, so adding it to the conda version is a little bit more explicit.) Ready for merge from my end anyway. Should I open an issue about the unresolved links?

sphinx_gallery/docs_resolv.py
```diff
@@ -232,8 +233,15 @@ def _get_link(self, cobj):
                 fname_idx = value[cobj['name']][0]
 
         if fname_idx is not None:
-            fname = self._searchindex['filenames'][fname_idx] + '.html'
-
+            fname = self._searchindex['filenames'][fname_idx]
+            try:
+                import sphinx
+            except Exception:
```
lesteve: Why would the sphinx import fail?

Eric89GXL: It looked like nowhere in the package did `sphinx` actually get imported, so I didn't want to add a top-level import. But since this is `sphinx-gallery` I guess it's okay, eh? :)

sphinx_gallery/docs_resolv.py
```diff
@@ -232,8 +233,15 @@ def _get_link(self, cobj):
...
+                fname = os.path.splitext(fname)[0]
+            else:
+                if LooseVersion(sphinx.__version__) >= LooseVersion('1.5'):
```
lesteve: Bonus points if you add a comment to clarify why the work-around for 1.5 is needed.

Eric89GXL:
@lesteve done

Eric89GXL:
Oh wait, I realized the problem with the link resolution. We will have an issue with building docs with 1.5+ but linking to older doc versions. I can push a fix.

Eric89GXL:
I don't love this fix, but it should work. Eventually once more pages start using 1.5+ we'll end up with a lot of errors being shown for links that couldn't be grabbed even though they work on the second iteration of the `for fname in fnames` loop...

GaelVaroquaux:
Please open an issue for the unresolved links (or check that there is not one already). They have been annoying me too.

OK, I was about to merge this PR, but it seems that you're on something, so I'll hold back a wee bit.

sphinx_gallery/docs_resolv.py
```diff
@@ -4,12 +4,14 @@
...
 # Documentation link resolver objects
 from __future__ import print_function
+from distutils.version import LooseVersion
 import gzip
 import os
 import posixpath
 import re
 import shelve
 import sys
+import sphinx
```
GaelVaroquaux: Are the sphinx import and the LooseVesion import still useful?

sphinx_gallery/docs_resolv.py
```diff
@@ -232,22 +234,34 @@ def _get_link(self, cobj):
...
+                        html = get_data(link, self.gallery_dir)
+                        self._page_cache[link] = html
+                except Exception as e:
+                    pass
+                else:
+                    break
             else:
-                html = get_data(link, self.gallery_dir)
-                self._page_cache[link] = html
+                raise e
```
lesteve: Would `raise` not be work to raise the last caught exception? My experience is that `raise e` creates a misleading traceback especially on Python 2.7 since you lose the provenance of the original exception.

sphinx_gallery/docs_resolv.py
```diff
@@ -375,8 +389,12 @@ def _embed_code_links(app, gallery_conf, gallery_dir):
...
                                                                   full_fname)
                     except (HTTPError, URLError) as e:
-                        print("The following error has occurred:\n")
-                        print(repr(e))
+                        if isinstance(e, HTTPError):
+                            extra = e.code
+                        else:
+                            extra = e.reason
+                        print("\t\tError resolving %s.%s: %r (%s)"
+                              % (cobj['module'], cobj['name'], e, extra))
```
lesteve: For another PR, but it does seem that it would be useful to have the URL that we tried to access, rather than the module ...

Eric89GXL: Yeah I agree, but that will require a good bit of refactoring because the URL is a function or two deeper than this. And really it should be refactored to use the proper local filename based on sphinx version, and then check the two possible remote names, rather than trying both in both cases as it does now. But I think that will again require a lot of refactoring to work.

Eric89GXL:
Unnecessary imports removed.

sphinx_gallery/docs_resolv.py
```diff
@@ -232,22 +232,34 @@ def _get_link(self, cobj):
...
+
+                    if hasattr(link, 'decode'):
+                        link = link.decode('utf-8', 'replace')
+
+                    if link in self._page_cache:
+                        html = self._page_cache[link]
+                    else:
+                        html = get_data(link, self.gallery_dir)
+                        self._page_cache[link] = html
+                except Exception:
```
GaelVaroquaux: Do you really want to catch _all_ the exceptions? Can't we be slightly more specific?

Sorry for making a new comment at each round, I am seeing these aspects only gradually.

Eric89GXL: It's okay, this bit is new :)

Yes I can catch just HTTPError and URLError

lesteve:
I think the fix is fine, you could tidy it up by putting the for loop in a function but we can do that later.

+1 for merging, thanks a lot @Eric89GXL.

I wish I would have paid more attention when #137 was created, I have to admit I have no recollection of it at all ...

lesteve:
> I think the fix is fine, you could tidy it up by putting the for loop in a function but we can do that later.

And the warnings about not being able to access the URL would make sense inside this function as well, since at this point you know the two URLs you tried to access (rather than catching the exception later and having only one URL).

Eric89GXL:
It's alright, I think Anaconda just rolled out 1.5 so I doubt many people have been hit by it. I also probably should have followed up once the issue didn't magically disappear like I was hoping it would...

Exception made less broad, Travis builds look good (green and logs have no complaints).

lesteve:
OK I am going to merge this one, thanks a lot !

PR #185, opened 2016-12-13
[MRG] Drop styling of codelinks tooltip use title

Titan-C:
Following the recommendation done on https://github.com/scikit-learn/scikit-learn/pull/7986#issuecomment-265322148, this PR simplifies the tooltips for code links

Currently the tooltip appears immediately after hovering the link and with message "Link to". It is visually uniform in firefox and chromium.
before

This changes uses title and lets the browser handle this popup which appears after a few seconds of hovering over the link. It also changes "link to"->"View"
Thus in firefox it is similar to the previous one but without the like break
firefox
And in chromium is a different style
chromium

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ Change Log
 git master
 ----------
 
+New features
+''''''''''''
+
+* Drop styling in codelinks tooltip. Replaced for title attribute which is managed by the browser.
+
 v0.1.7
 ------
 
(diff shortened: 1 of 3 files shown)


lesteve:
I tested this locally and this works fine. Visually I think this is fine and this simplifies the CSS, so +1 from me.

@GaelVaroquaux do you have an opinion on this one?

lesteve:
Hmmm this needs a rebase too.

Titan-C:
Rebased

lesteve:
Maybe an entry in the changelog?

Titan-C:
Entry in changelog was added

Any comments on this @GaelVaroquaux ? We drop visual uniformity across browsers in favor of simpler CSS. True reason is moving from tooltip to title and letting the browser handle this. Suggestion from @jnothman

PR #189, opened 2016-12-21
cached conda gives problems in new Circle CI Builds

Titan-C:
Shamelessly copy from MNE-python Circle CI script not to install conda again

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -4,8 +4,6 @@ machine:
     PATH: "/home/ubuntu/miniconda/envs/circleenv/bin:/home/ubuntu/miniconda/bin:$PATH"
 
 dependencies:
-  cache_directories:
-    - "~/miniconda"
   pre:
     # Disable pyenv (no cleaner way provided by CircleCI as it prepends pyenv version to PATH)
     - rm -rf ~/.pyenv


lesteve:
What problem does this solve?

Titan-C:
> What problem does this solve?

Build on the same branch or PR start using the cache and then fail when conda tries to install. See

https://circleci.com/gh/Titan-C/sphinx-gallery/11

lesteve:
I see, they are caching the miniconda install. Not sure I am a big fan of the idea, I'd rather remove ~/miniconda from the cache.

Titan-C:
> I see, they are caching the miniconda install. Not sure I am a big fan of the idea, I'd rather remove ~/miniconda from the cache.

I see the cache is because one downloads a lot. Comparing timings for us downloading conda with mayavi and mkl + install takes a little more than a minute. For MNE-Python they need 3 min to recover their cache which includes more than the conda install. I'm fine with dropping the cache

Titan-C:
It's a bit weird, but building without saving cache breaks the build because cache was already there. I'm rebuilding from CircleCi to clean everything

lesteve:
Let's merge this one thanks!

PR #191, opened 2017-01-03
[MRG] Remove nonfunctional text2string function

naoyak:
`text2string()` uses `ast.literal_eval()` but `ast` is never imported in the file. So the try-except block is catching a `NameError`?

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -83,7 +83,7 @@ from .downloads import CODE_DOWNLOAD
 from .py_source_parser import (get_docstring_and_rest,
                                split_code_and_text_blocks)
 
-from .notebook import jupyter_notebook, text2string, save_notebook
+from .notebook import jupyter_notebook, save_notebook
 
 try:
     basestring
@@ -589,7 +589,7 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
                 example_rst += codestr2rst(bcontent) + '\n'
 
         else:
-            example_rst += text2string(bcontent) + '\n'
+            example_rst += bcontent + '\n\n'
 
     clean_modules()
 
--- a/sphinx_gallery/notebook.py
+++ b/sphinx_gallery/notebook.py
@@ -19,14 +19,6 @@ import sys
 from .py_source_parser import split_code_and_text_blocks
 
 
-def text2string(content):
-    """Returns a string without the extra triple quotes"""
-    try:
-        return ast.literal_eval(content) + '\n'
-    except Exception:
-        return content + '\n'
-
-
 def jupyter_notebook_skeleton():
     """Returns a dictionary with the elements of a Jupyter notebook"""
     py_version = sys.version_info
@@ -170,7 +162,7 @@ def fill_notebook(work_notebook, script_blocks):
         if blabel == 'code':
             add_code_cell(work_notebook, bcontent)
         else:
-            add_markdown_cell(work_notebook, text2string(bcontent))
+            add_markdown_cell(work_notebook, bcontent + '\n')
 
 
 def save_notebook(work_notebook, write_file):


Titan-C:
Good catch. I don't seem to put enough attention to my linter. I'm actually thinking this function is a residue of earlier prototype of the notebook styled examples, when the proposed syntax was to use the free triple quoted python strings to denote the blocks of prose. I think we don't even need this function anymore.

Although it might be early to say we don't need this, I'm unsure about our users use cases. But #148 introduces this bug and was merged not long ago, some users have not even moved to a version with this change.

@naoyak could you give it a try and remove text2string from our code? and just replace its call for adding a newline to the strings. And if you can imagine a way to put a new test to this would be amazing.

naoyak:
As far as I can tell that line throws an exception every time (due to `ast` not being available), so in all cases it does `return content + '\n'`.

I was actually looking at this for scikit-learn/scikit-learn#8143. The built docs are here: https://7986-843222-gh.circle-artifacts.com/0/home/ubuntu/scikit-learn/doc/_build/html/stable/index.html

@Titan-C If you want I can try your suggestion on that PR and see if the sklearn docs still build correctly. How does that sound?

Titan-C:
> @Titan-C If you want I can try your suggestion on that PR and see if the sklearn docs still build correctly. How does that sound?

That look good for me. Scikit-learn is at a version pre PR#148, so if you find something crashing there or not crashing will be good. And we can clean the code overhere.

naoyak:
@Titan-C fyi here's the diff I'm testing out: https://github.com/scikit-learn/scikit-learn/pull/8143/commits/678a2761170a5a08fce82120414cb210e594af29
sklearn docs should be up at http://scikit-learn.org/circle?8021 in a bit

naoyak:
So, building another set of scikit-learn docs after rebasing: http://scikit-learn.org/circle?8257

@Titan-C let me know if there is a specific type of potential glitch I should be looking out for (commit is https://github.com/scikit-learn/scikit-learn/pull/8143/commits/d5c11b009aa96ef18861b5e36b3be4db2c563a6c)

lesteve:
@naoyak could you add a test for text2string?

Titan-C:
I'm actually in favour of dismissing tex2string. It does not serve any purpose.

naoyak:
As far as I can tell removing text2string doesn't do anything (indeed it seems to have been doing nothing all along due to the missing import). I can revise this PR.

naoyak:
I removed `text2string` and substituted references to it.

lesteve:
> I'm actually in favour of dismissing tex2string. It does not serve any purpose.

I thought at one point it would strip docstring from triple back-quotes around it. If this does not have any use, I don't mind removing it though ...

sphinx_gallery/gen_rst.py
```diff
@@ -589,7 +589,7 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
                 example_rst += codestr2rst(bcontent) + '\n'
 
         else:
-            example_rst += text2string(bcontent) + '\n'
+            example_rst += '\n\n'
```
Titan-C: Here you forget to keep the `bcontent`

Titan-C:
> I thought at one point it would strip docstring from triple back-quotes around it. If this does not have any use, I don't mind removing it though

Yes when I used the tokenizer to extract the docstring and the notebook's text/prose where also triple quoted strings I required this eval to remove the quotes. But docstring are now extracted with ast, and the text/prose blocks are recognized by the regular expressions.

sphinx_gallery/gen_rst.py
naoyak: Doh, sorry. fixed

lesteve:
> Yes when I used the tokenizer to extract the docstring and the notebook's text/prose where also triple quoted strings I required this eval to remove the quotes. But docstring are now extracted with ast, and the text/prose blocks are recognized by the regular expressions.

OK makes sense, thanks for the details.

Titan-C:
Merged. Thank you @naoyak !

Issue #192, opened 2017-01-04
Lack of `install_requires`


dopplershift:
I know `setup.py` notes that not having `install_requires` is a deliberate decision, but is it possible to revisit that decision?

Right now `pip install sphinx-gallery` results in an incomplete install--I still need to manually install sphinx, matplotlib, and pillow to have a functional install. While sphinx and matplotlib are a given for my project, I need to now add `pillow` as a manual install step to my documentation CI builds--and it will break any time dependencies update in the future. Not to mention this becomes a much bigger headache should there be any version-specific needs in the future.

dopplershift:
Documenting this fact prominently would be a nice start.

Putting things in my own `requirements.txt` doesn't solve my core issue here: that I *manually* must do the job of a package manager and stay up to date on what dependencies must be installed to have sphinx-gallery function properly. I've never before encountered a Python package that does this and doesn't put things in `install_requires`. (`install_requires` is considered best practice here: https://packaging.python.org/distributing/#install-requires)

It feels like wrong software engineering practice to me, since not only do I have to repeat information that exists and is used elsewhere, but I'm also codifying in the configuration for my own project information that is completely unrelated.

I don't know the problems that led to this solution, so at the very least I'd suggest better commenting/documenting the actual problem rather than saying "install_requires breaks stuff" (the pull request that made this change literally says nothing). Ideally, I'd revisit the problem and find a solution that doesn't require doing something so far from normal in the Python ecosystem--but obviously that's not my call.

dopplershift:
To be clear, I have no problem using a `requirements.txt`. My problem is that mine would have to be something like:
```
numpy
scipy
matplotlib
sphinx
sphinx-gallery
# The following are needed to make sphinx-gallery work
pillow
 # outdated
joblib
```
So now I have to maintain `requirements.txt` not just for the things I want in the environment, but I have to make sure I'm up-to-date on what changes you guys make, like no-longer needing joblib,  or adding new dependencies. It feels really wrong to be putting in that file things that I don't directly want, but are unlisted requirements, from pip's perspective, for sphinx-gallery.

lesteve:
For the record, I looked at the doc and it turns out that you can include sphinx_gallery requirements in your own requirements file like this:

```
# your requirements go here
numpy
scipy

# sphinx-gallery requirements
-r https://raw.githubusercontent.com/sphinx-gallery/sphinx-gallery/master/requirements.txt
```

[...]

PR #193, opened 2017-01-05
README states that dependencies are on the user's responsibility

Titan-C:
This is to address issue #192


lesteve:
LGTM, merging.

PR #187, opened 2016-12-16
ENH: Cleaner output

Eric89GXL:
This does two things:

1. Suppresses output when doing `plot_gallery=False` saying it took 0 seconds to run a bunch of examples that didn't run. It's not very informative, and it makes the sphinx output very long, thus making things like checking for good sphink markup, bad links, etc. harder (what `plot_gallery=False` is mostly used for, I assume).

2. Uses `app.status_iterator` for the last stage of processing where code is embedded. Instead of having tens or hundreds of lines, now the output looks like the rest of sphinx, e.g.:
    ```
    auto_examples[100%] plot_left_cerebellum_volume_source.html
    auto_tutorials[100%] plot_ica_from_raw.html
    ```

PR #198, opened 2017-01-17
[MRG] Fix :orphan: sphinx-directive multiple calls

Titan-C:
PR #187 introduced the use of the `orphan` sphinx directive to avoid the warnings of the gallery examples not being in any toctree. This directive has to be called once per rst file and not before every section, otherwise `:orphan:` is presented in the html file. This PR fixes that behaviour


lesteve:
The CircleCI [artifacts](https://51-26191147-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/doc/_build/html/auto_examples/index.html) look good, so merging this one, thanks a lot!

PR #207, opened 2017-03-08
Don't error when input directory is the same as the output directory

ngoldbaum:
In my project, I'd like to have:

```
sphinx_gallery_conf = {
    'examples_dirs': 'cookbook',
    'gallery_dirs': 'cookbook',
}
```

This almost works right now except for the error raised by the block modified in this pull request. Since the error is harmless in this situation, let's just catch the exception and ignore it.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -546,7 +546,10 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
 
     src_file = os.path.normpath(os.path.join(src_dir, fname))
     example_file = os.path.join(target_dir, fname)
-    shutil.copyfile(src_file, example_file)
+    try:
+        shutil.copyfile(src_file, example_file)
+    except shutil.SameFileError:
+        pass
     script_blocks = split_code_and_text_blocks(src_file)
     amount_of_code = sum([len(bcontent)
                           for blabel, bcontent in script_blocks


Titan-C:
I personally think is not a good idea to have the source dir as the gallery dir. When the example scripts are executed they can generate or download or manipulated data and I prefer to have that isolated from the data generated for the gallery, rst files, images, checksums, and code objects. That way you can clean the gallery generated files and folders safely.
Including your PR would imply that this option is save to use. I rather leave it for the users to personally do this modification in their codebase if they really need it.

ngoldbaum:
> That way you can clean the gallery generated files and folders safely.

Presumably you can also use your version control tool to do that

> I rather leave it for the users to personally do this modification in their codebase if they really need it.

That's too bad, I'd prefer not to vender `sphinx-gallery` but if that's the approach I'll need to take so be it, although it will make me less likely to upstream features I'd like to add.

lesteve:
> I personally think is not a good idea to have the source dir as the gallery dir.

I have to agree. @ngoldbaum do you have a strong reason for this?

GaelVaroquaux:
> I have to agree. @ngoldbaum do you have a strong reason for this?

It's pretty bad practice to mix source and output.

To give you an example of what might go wrong: trying to flush the
generate documentation, I once removed all the files in Mayavi's SVN (and
because I did it in a script, I pushed to master without realizing).

GaelVaroquaux:
> > I have to agree. @ngoldbaum do you have a strong reason for this?

> It's pretty bad practice to mix source and output.

Another reason: it's going to make it easy for changes in sphinx-gallery
to break user's code, as sphinx-gallery assumes that it fully controls
the output directory.

ngoldbaum:
I'm trying to switch from our own custom cookbook sphinx extension to use sphinx gallery since I think sphinx gallery looks nice. Our current sphinx extension writes the output to the same directory as the input and we have not had any trouble with that. I'd like to move to sphinx gallery with a minimum of churn, adding this helps me accomplish that.

Given the amount of pushback I've gotten for as simple a change as this as well as the other issues I've commented on and opened yesterday I think I'm going to vendor sphinx gallery and add the features I need to the vendored version. It's clear it will simply take too much emotional effort to upstream changes I need. Thanks for creating this project, but please keep in mind that lecturing new contributors like in this thread comes across as very hostile, especially on text over the internet.

Please feel free to close this issue if you do not want to take the patch.

GaelVaroquaux:
> Thanks for creating this project, but please keep in mind that lecturing new
> contributors like in this thread comes across as very hostile, especially on
> text over the internet.

Nathan, I've done a lot of open source for a lot of years, and this
process is normal and to be expected. Libraries cannot grow for many
years and stay healthy without push back on every feature that gets
added. Elsewhere they become a hodge-podge of features with no guiding
principle.

I understand where you are coming from: adapting an existing codebase.
However, using comon tools requires comon conventions. The cost is that
we have to adapt to others. The benefit is that we share the code.

I am sorry that you are taking this as hostile. You should see the amount
of crap that I have sometimes recieved as an open source maintainer,
being insulted for not solving other people's problem.

Issue #199, opened 2017-01-26
Rendering HTML in examples


rth:
Since sphinx-gallery displays the example output on a web page, I was wondering if it was possible to render directly some HTML  produced by the example script.

For instance, in jupyter notebooks, one can render some HTML with,
```py
from IPython.display import display, HTML
display(HTML('<span style="background-color:red;"> test</span>'))
```
so if the example `.py` script contained a `display(HTML(...))` call, is it possible to just embed the corresponding HTML code into the example page? I could not find anything about this topic in the docs.

If that not currently supported, what would be the advised approach to  add this functionality (either through `display(HTML(...))` or some other mechanism)?  Thank you.

Titan-C:
> I was wondering if it was possible to render directly some HTML  produced by the example script.

I would rather be against getting this feature. We capture stdout and
save the plots, and would rather stay within those boundaries. The idea
is that you can run the scripts directly from a python interpreter and
not meet inconveniences.

> For instance, in jupyter notebooks, one can render some HTML

Yes jupyter notebooks are amazing but that is a different ecosystem than
us. Starting to support Jupyter notebooks features would imply we put
their ecosystem in our dependencies, that is not going to happen.


> If that not currently supported, what would be the advised approach to  add this functionality (either through `display(HTML(...))` or some other mechanism)?  Thank you.

I encourage you to get creative and hack into our code. My approach
would be to save a file on disk with your desired output and then use
the support we give to rst syntax in the code annotations to do a
`literalinclude` of that file and render it as html.

This of course will not work directly with our current code, because
code is executed where the source file is but the output rst file is
saved in your documentation folder where Sphinx can find it to parse it
into html. So you will need to work your way around to give the correct
relative paths for the include on the simple side, or implement a way to
transport desired output files into the sphinx documentation folder.

rth:
@Titan-C Thanks a lot for this exhaustive response and sorry for my very late reply..

Following your explanations, I managed to make it work. Assuming that,
 * examples are under `examples/`
 * the docs are under `doc/`
 * that the docs are built with `make html` from under `docs`,

in order to render HTML contained in the `html_data` variable, adding the following lines at the end of the example script, seems to work,

```py
tmp_dir = os.path.join('..', 'doc', 'examples')
if os.path.exists(tmp_dir):
    # building the docs with sphinx-gallery 
    with open(os.path.join(tmp_dir, 'out.html'), 'wt') as fh:
        fh.write(html_data)
[...]

Issue #150, opened 2016-10-07
matplotlib animations support


dfroger:
That would be nice to support GIF images, like in this example: http://eli.thegreenplace.net/2016/drawing-animated-gifs-with-matplotlib/

lesteve:
I am not sure how easy it is for us to do. My instinct is why not but it is not that crucial, in general you can select a few static plots that are as clear as an animated gif.

dfroger:
I'll try to implement it for [matplotlib animation](http://matplotlib.org/api/animation_api.html) in general (.git or .mp4). I'll keep you informed.

lesteve:
One of the technical hurdle I can see: after running the example we get the list of figures opened by matplotlib and save each of them (look at save_figures in gen_rst.py for more details). It seems like for matplotlib animations you need to call .save on the matplotlib.Animation object instead.

Titan-C:
I haven't used animation. But if is a different object, you should be able to write a new function(as we do for mayavi figures) an make it part of the chain.
Other thing that is not a priority I guess, but might be good to keep in mind is our support for latex/pdf export, see #61.

dfroger:
@lesteve  Yes, I think I will add an `if` case like [the one for mayavi](https://github.com/sphinx-gallery/sphinx-gallery/blob/4e460513197101594ff6c7eee2f77b25b709c1fa/sphinx_gallery/gen_rst.py#L340), in which I will call `.save` of `matplotlib.Animation`.

@Titan-C  ok, I'll keep this in mind.

I can't work on it right now, but I will spend some time on it this week, so it may progress slowly :)

NelleV:
In general, it would be super useful for matplotlib to be able to display animation, so we would greatly appreciate this addition to `sphinx-gallery`.

lesteve:
@NelleV here is the code we use to find all the matplotlib figures that have been generated by the example and save each of them:
https://github.com/sphinx-gallery/sphinx-gallery/blob/47f7149b6a0ecbb548de5bf482d42543126431bd/sphinx_gallery/gen_rst.py#L323

I am not a matplotlib expert at all, so here is a naive question: is there a way from the figure object to know that a matplotlib.Animation object was used to plot on this figure?

NelleV:
There is an animation object.
I'll check the code out and see whether we can extend it to movies.

t-makaro:
[...]

Issue #25, opened 2015-04-10
Multiprocesor support?


Titan-C:
I was wondering if one could generate the examples in parallel processes?
I now python has a feature for sending processes to individual cpu. I can not be directly used right now because currently we scan recursively the example folder instead of generating a list of example files to treat first. It will also need to keep track of the back references.

GaelVaroquaux:
Hi,

I think that this is not high priority. I would much more like to focus
on getting IPython notebook output.

lesteve:
Agreed. FWIW I have some hacky lines of bash using GNU parallel to run the examples in parallel and make sure that they all work. For some reason I haven't fully investigated, the speed-up I got running the nilearn examples was 2x (i.e. roughly 10 minutes instead of 20 minutes).

All I am trying to say is that such a speed-up is not going to change your life that much.

GaelVaroquaux:
> the speed-up I got running the nilearn examples was 2x (i.e. roughly 10
> minutes instead of 20 minutes).

I/O?

> All I am trying to say is that such a speed-up is not going to change
> your life that much.

Well, for the wider ecosystem anyhow IPython notebook support would be a
much bigger benefit.

Eric89GXL:
FYI you can now run your own examples using `make html_dev-pattern`, so some lines of `bash` can be used to customize it for each repo. It's not automated, but it works e.g. for CircleCI.

ogrisel:
Now that notebook exports have been implemented I think we should bump up the priority of this one. Building the scikit-learn documentation takes ages and lazy reviewers tend to not check the rendering and the integration of example generated figures because of that. I have 12 cores on my workstation and only one is working at the moment...

ogrisel:
@lesteve this could probably be implemented with joblib.Parallel quite easily instead of using gnu parallel. One just need to make sure that the example building function does not return large python objects back to the main process but instead directly writes the output of the execution to disk. (e.g. images and joblib cached stuff).

lesteve:
> Now that notebook exports have been implemented I think we should bump up the priority of this one.

I wasn't aware that generating notebooks would take a lot of time ... are you confident that's the culprit here ?

I guess using joblib.Parallel would alleviate the problem seen in #57 (seaborn style set in one example was kept for all the other examples) since you could run each example in a separate process as mentioned in https://github.com/sphinx-gallery/sphinx-gallery/pull/140#issuecomment-243461269.

NelleV:
[...]

PR #208, opened 2017-03-08
Add the ability to optionally include images in a gallery that were saved to disk

### LINKED ISSUE #206: Images saved outside of pyplot are not picked up in the gallery
It looks like `sphinx-gallery` only picks up on an image if it is saved via pyplot. I'm working with a library that uses matplotlib for plotting but *does not* use pyplot, instead wrapping matplotlib's object oriented interface.

It would be nice if in addition to intercepting images saved via pyplot there was also a scan for `.png` files in the directory that executes a given script. That way tools that generate images independent of matplotlib.pyplot will still be able to include images in a gallery.

If this is an agreeable feature to add I'd be happy to take a shot at adding it. Also if my description isn't clear I can take a shot at making a minimal reproducible example.

ngoldbaum:
This resolves #206. I'll add documentation and a test for this feature and the new configuration option if this approach is copacetic.

This change makes it so that we can include images in a gallery for code that does not save image via pyplot or mayavi. It's enabled via a new configuration option `find_saved_figures`, which defaults to `False`.

If it's `True`, we inspect the directory containing the example script (i.e. the directory where the script was being run) for png images. If we find any, we move them to the output directory for the gallery and give them names following the template used for pyplot figures.

I've had to modify the API of the internal `save_figures` function slightly. As far as I can tell this function only has consumers inside of `sphinx-gallery` so it should be ok to change the API, but please let me know if that's incorrect.

There are a couple things that could be improved:

* Search for png images that already exist before running the script so we can find the list of images created *just* by the script
* Search for images that aren't in png format or somehow allow users to specify the image format.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -17,6 +17,7 @@ Files that generate images should start with 'plot'
 from __future__ import division, print_function, absolute_import
 from time import time
 import codecs
+import glob
 import hashlib
 import os
 import re
@@ -229,18 +230,19 @@ def md5sum_is_current(src_file):
     return False
 
 
-def save_figures(image_path, fig_count, gallery_conf):
+def save_figures(image_path, execution_path, fig_count, gallery_conf):
     """Save all open matplotlib figures of the example code-block
 
     Parameters
     ----------
     image_path : str
         Path where plots are saved (format string which accepts figure number)
+    execution_path : str
+        Path where the example script was executed.
     fig_count : int
         Previous figure number count. Figure number add from this number
     gallery_conf : dict
         Contains the configuration of Sphinx-Gallery
-
     Returns
     -------
     images_rst : str
@@ -280,6 +282,14 @@ def save_figures(image_path, fig_count, gallery_conf):
             scale_image(current_fig, current_fig, 850, 999)
             figure_list.append(current_fig)
         mlab.close(all=True)
+
+    if gallery_conf.get('find_saved_figures', False):
+        png_paths = glob.glob(os.sep.join([execution_path, '*.png']))
+        cur_count = len(figure_list)
+        for fig_num, path in enumerate(png_paths, start=1):
+            current_fig = image_path.format(fig_count + fig_num + cur_count)
+            shutil.move(path, current_fig)
+            figure_list.append(current_fig)
 
     return figure_rst(figure_list, gallery_conf['src_dir'])
 
@@ -482,8 +492,11 @@ def execute_code_block(code_block, example_globals,
         if my_stdout:
             stdout = CODE_OUTPUT.format(indent(my_stdout, u' ' * 4))
         os.chdir(cwd)
-        images_rst, fig_num = save_figures(block_vars['image_path'],
-                                           block_vars['fig_count'], gallery_conf)
+        images_rst, fig_num = save_figures(
+            block_vars['image_path'],
+            os.path.dirname(src_file),
+            block_vars['fig_count'],
+            gallery_conf)
 
     except Exception:
         formatted_exception = traceback.format_exc()
--- a/sphinx_gallery/tests/test_gen_rst.py
+++ b/sphinx_gallery/tests/test_gen_rst.py
@@ -243,14 +243,14 @@ def test_save_figures():
     mlab.test_plot3d()
     plt.plot(1, 1)
     fname_template = os.path.join(gallery_conf['gallery_dir'], 'image{0}.png')
-    image_rst, fig_num = sg.save_figures(fname_template, 0, gallery_conf)
+    image_rst, fig_num = sg.save_figures(fname_template, '', 0, gallery_conf)
     assert_equal(fig_num, 2)
     assert '/image1.png' in image_rst
     assert '/image2.png' in image_rst
 
     mlab.test_plot3d()
     plt.plot(1, 1)
-    image_rst, fig_num = sg.save_figures(fname_template, 2, gallery_conf)
+    image_rst, fig_num = sg.save_figures(fname_template, '', 2, gallery_conf)
     assert_equal(fig_num, 2)
     assert '/image2.png' not in image_rst
     assert '/image3.png' in image_rst


lesteve:
I haven't tested this but I think this will be a problem with the notebook-style examples where you have multiple cells and not all figures are saved in the same cell. I believe some figures will be repeated in the HTML.

Also the order of saved figures will not be related with the order of the example execution which may be confusing if you have multiple figures.

Rather than trying to treat similarly saved png and matplotlib figures, maybe we could have an additional "Saved Files" section, which shows saved files by name. This way it is more clear where the image comes from in the rendered HTML. I am thinking aloud here and I don't have a good solution for what you want to do right now.

lesteve:
> I haven't tested this but I think this will be a problem with the notebook-style examples where you have multiple cells and not all figures are saved in the same cell. I believe some figures will be repeated in the HTML.

Actually I just quickly tested it and this looks fine, I am guessing because you are moving images out of the way as soon as you discover them.

> Also the order of saved figures will not be related with the order of the example execution which may be confusing if you have multiple figures.

I found an example of this in your gallery actually:
http://yt-project.org/docs/dev/cookbook/simple_plots.html#showing-and-hiding-axis-labels-and-colorbars
In the snippet you first have no_axes_sliceplot and then no_axes_no_colorbar_sliceplot but in the images it is the other way around.

I think we should sort the result of `glob.glob` using `os.path.getctime` or something similar.

> Rather than trying to treat similarly saved png and matplotlib figures, maybe we could have an additional "Saved Files" section, which shows saved files by name. This way it is more clear where the image comes from in the rendered HTML.

I thought a bit more about this and I stick to my proposal. So basically on top of the two things we already have as the output of a cell:
* Interactive figures
* Standard Output
we would have "Saved images". It would be good to display the filename reasonably prominently to make it easier to track which image is which file and understand where it comes from in the example.

By the way this feature could be a way of supporting matplotlib animations #150 by saving to a .gif in the example.

@Titan-C @GaelVaroquaux any opinions?

GaelVaroquaux:
I am quite uncomfortable with that, because it is moving us further away
from the notebook model. We are already not really notebooks, and things
behave a bit different in the notebook and in the Python examples. This
will increase the gap. I don't like it.

lesteve:
> I am quite uncomfortable with that, because it is moving us further away from the notebook model

Yeah I thought the same too actually. It should be possible to add some kind of formatting in the rendered HTML to make it clear that those are saved image files. Differentiating between interactive figures and saved images is actually the point of adding another section in the output.

GaelVaroquaux:
Still, it's pulling the project away from notebooks, while everybody else
is going onto notebooks.

I haven't fully understood the compeling reason for saving to disk.

lesteve:
> I haven't fully understood the compeling reason for saving to disk.

The main point is that it makes sphinx-gallery usable by packages that saves images to disk (as seems to be the case of [yt](http://yt-project.org)) and do not have figures e.g. `matplotlib.pyplot` does.

One thing to keep in mind also is that the changes needed to make it work are fairly small. It is not a conclusive argument by itself but still certainly something to take into account.

GaelVaroquaux:
> The main point is that it makes sphinx-gallery usable by packages that saves
> images to disk (as seems to be the case of yt) and do not have figures e.g.
> matplotlib.pyplot does.

yt doesn't use matplotlib? I thought that it did (sorry, I am behind the
great firewall of China, and google doesn't work, and I am not very good
with baidu :$).

For non matplotlib-based plotting, the vision was to have backends that
register with sphinx-gallery in the conf.py. I thought that we had done
that with Mayavi, but looking at the codebase, we haven't.

I would much prefer that solution: it avoids modifying sphinx-gallery
each time a new situation comes. We could have a callback that is called
after the execution of each cell, with the locals and globals passed in
and a few other meta-data (such as the file name executed), and it would
return the restructured code to be included.

ngoldbaum:
yt does use matplotlib but not pyplot.

There is no global list of figures creat d by a yt plotting session (since we're not using matplotlib) so the approach used by sphinx-gallery for mayavi and pyplot simply won't work, there's no way to figure after a script has run which figures it produced except by inspecting a folder for imag s saved to disk. If I were using a plugin here the plugin would just be looking for images saved to disk.

What if I were trying to document a library like Pillow which saves images to disk completely independent of matplotlib and also does not have a global registry of figures? I'd need something exactly like what's in this pull request. This is a sufficiently general desire that I think it would be very helpful to include in the library itself.

GaelVaroquaux:
Based on @choldgraf 's feedback in #206, I think that we should implement this functionality.

A bit more work is needed on this PR, mainly:

* The option is the conf.py should not be a True/False, but a regex to decide which files are matched. That way it would be possible to include other files than pngs, but also to be more restrictive than capturing all the files

* There needs to be narrative documentation added, so that users discover this functionality

* There should be a bit of formatting / CSS work so that the images appear as different from the ones produced inline. Something like a light title saying "Files produced" and maybe a box with a light border.

Titan-C:
I though I replied to this, but since I'm out on a school with other
things in mind I just have not. I actually like the idea of capturing
other output objects, which we now limit to stdout, matplotlib & mayavi
plots.

In such scheme I agree with @lesteve to give this new output its own
output block, not mix it as if they were the standard matplotlib pop-up
plots.

What I don't like of this implementation is that the image capturing is
just to generic: just a glob to `*.png`. This could give unwanted
behavior. I image for example in the scikit-image project they certainly
load `png` or `jpg` images for further processing. Letting this glob
just capture everything in the local folder and move it to the gallery
images output would just break their gallery, they would run out of
their locally available images. Yes this feature is disabled by default,
but I just fear the generality of the procedure to break things. Second
concern for the future as Multiprocessor support is becoming a requested
feature PR #25. Race conditions will become a real issue on this
implementation. 2 parallel threads will glob the same files on disk, then
rendered htmls will have mixed images and then one thread will move the file
first making the other thread raise a FilenotfoundError.

On top of @gaelvaroquaux suggestions. The conf.py option shall be
`capture_yt_save_figures`, it needs to be indeed a regex. But ignorant
of the yt code, is it possible to include an internal variable that
names written files such that it is specific they are output files and
from which python script they are generated? see concerns above.

In the case we want to make this very general an capture other arbitrary
objects other than png, like gif animations issue #150. Then I would go
for making a per file sphinx-gallery comment that signal how to capture
a specific file. And then force the user to use the rst comments to load
the captured file, see my comments on issue #199, weather is an image,
video, html, text file for download, numpy data file, etc. and how we
want to display it? as downloadable file? an rendered image.

alexhuth:
I don't have any comments implementation, but I just want to chime in and add that this would be a very helpful feature for pycortex. A lot of our plotting is actually done using webGL in a browser, and we really want to make a pretty gallery to show off the pretty outputs. Being able to insert screenshots as thumbnails would be a great help!

lesteve:
> Being able to insert screenshots as thumbnails would be a great help!

Are the screenshots generated by the example or are they done manually ?

alexhuth:
I'm hoping to do the screenshots manually so they could include the GUI. 

Currently it's possible to automatically screenshot just the webGL canvas (with no GUI). One workaround that I hadn't considered before would be to load the automatic screenshot and then display it using matplotlib in order to generate a thumbnail for the gallery. That would be possible, but *ugly* (and would end up with an example showing the user how to do something stupid that they would never want to do).

lesteve:
> I'm hoping to do the screenshots manually so they could include the GUI.

This is what I thought. Your use case is slightly different than the yt case then. One of the point of sphinx-gallery is to make sure that the script and its outputs are in sync by running the example. I think the goal of this PR is to allow capturing saved images generated by the example.

Note that notebook-style cells allow you to write any rst as comments so you can include images if you have a screenshot somewhere. It's just that they will not be used for the thumbnail in the gallery.

tacaswell:
attn @choldgraf

choldgraf:
what's the status of this PR? it sounds like there's still debate as to whether to implement this feature in the first place... @ngoldbaum or @lesteve any thoughts on this? I think we should decide and get this PR either merged or closed :-)

ngoldbaum:
I don't have any time to spend on this. If it's causing noise to have this open, please close it.

Issue #209, opened 2017-03-11
CSS class of images is a bit generic


choldgraf:
It seems like the rendered images in the HTML have a class of `align-center`. It might be helpful for some projects to set a class that is a little bit more sphinx-gallery specific (e.g. `sg-inline`). That way projects could write CSS rules for the output. Any reason not to implement something like this at least as an option in the config file? e.g. `{'img_class': 'something'}`

Titan-C:
`align-center` is sphinx default class for images. But there is no reason in my opinion to not allow images to have a specific CSS class. In PR #210, I have added  the classes `sphx-glr-img` and `sphx-glr-li-img`, which use our standard namespace.

choldgraf:
Cool! Thanks for the quick turnaround!

larsoner:
@choldgraf this can be closed

choldgraf:
ah yes, #210 is now merged!

PR #210, opened 2017-03-11
Gallery output images have own css class

Titan-C:
In response to #209

### DIFF
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -207,6 +207,16 @@ ul.sphx-glr-horizontal img {
   height: auto !important;
 }
 
+.sphx-glr-single-img {
+  margin: auto;
+  display: block;
+}
+
+.sphx-glr-multi-img {
+  max-width: 42%;
+  height: auto;
+}
+
 p.sphx-glr-signature a.reference.external {
   -moz-border-radius: 5px;
   -webkit-border-radius: 5px;
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -210,11 +210,19 @@ ul.sphx-glr-horizontal img {
 .sphx-glr-single-img {
   margin: auto;
   display: block;
+  max-width: 100%;
+  transition: transform .2s;
 }
 
 .sphx-glr-multi-img {
   max-width: 42%;
   height: auto;
+  transition: transform .2s;
+}
+
+.sphx-glr-single-img:hover, .sphx-glr-multi-img:hover {
+  transform: scale(1.5);
+  z-index: 200;
 }
 
 p.sphx-glr-signature a.reference.external {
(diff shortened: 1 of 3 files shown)


Titan-C:
A little survey on preferences: Given that the gallery output images now get an own CSS class. Should we keep Sphinx standard styling directives `:center:` and `:scale:` or just move all that to CSS?

lesteve:
`sphx-glr-li-img` does not seem like a great name. Naming suggestion: `sphx-glr-single-img` and `sphx-glr-img`. Better names more than welcome.

I am fine moving things to the CSS.

Titan-C:
I moved image styles to CSS. Changed to new names `sphx-glr-single-img` and `sphx-glr-multi-img`. In the case of multilpe images, they won't be always rescaled to 42% of their size, but have a max-width of 42%. So depending of the availible space they will look bigger or smaller. This makes the output responsive, which looks good. The only feature I start to miss is that when using sphinx's `:scale:` the image becomes clickable to see the original full size.

lesteve:
> The only feature I start to miss is that when using sphinx's :scale: the image becomes clickable to see the original full size.

I think you could add a `target:` in the `image` directive for both single and multi images, this way they both would be clickable.

Other than this, LGTM.

choldgraf:
if I it's not too complex, I agree with @lesteve about adding a target, though I don't think it should be blocking on this PR if it *is* complicated

larsoner:
@Titan-C I just hit a use case where this is necessary, as I need to add `max-width: 100%` to the gallery image CSS to avoid the image spilling over. It seems like this needs a doc update, and possibly the `:target:` update (though I'm not sure how easy or difficult this would be). Any time for this? Or could I take over?

Titan-C:
@larsoner feel free to take over. Although I think we should merge this directly. `:target:` update is very hard to get working because it takes an url, whereas `image` takes the path of the file and then moves the image to the `_images` directory. The correct way to handle that would be extending the `image` directive in Sphinx, which I'm against because is too complicated to deal with and give maintenance. My alternative would be to solve in the same fashion as I did for the gallery thumbnails only with CSS, maybe by hovering over the image.

Titan-C:
I'm proposing a CSS zoom on the images.

larsoner:
I just tried it and for galleries with large images it's difficult to avoid mousing over the image and hitting the zoom behavior, which in turn covers up the narrative. This makes interaction a bit clunky, so -1 on that for me. (Aethestically I also find the changes too distracting / invasive to make it a default.) But otherwise the CSS change of `max-width: 100%` works well for me.

larsoner:
@lesteve are you okay with this one as is? Seems like a pretty straightforward CSS flexibility / extension now.

+1 for merge from me, fixes PySurfer doc build https://github.com/nipy/PySurfer/issues/225 (though the new floating buttons also break it :) )

choldgraf:
+1 on merge - always better to use classes than to hard-code css styles IMO

choldgraf:
:tada:

PR #212, opened 2017-03-14
__file__ global variable in script execution

### LINKED ISSUE #211: NameError: name '__file__' is not defined
I just upgraded my ancient wordcloud sphinx setup to use sphinx-gallery. However, it seems like ``__file__`` is missing from the globals. Is that on purpose?
I want to load a file that's in the same directory as the example. Is there a better way to do that?

Titan-C:
Fix #211 
@amueller does this work for you?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,13 @@ Change Log
 git master
 ----------
 
+
+Documentation
+'''''''''''''''
+
+* Frequently Asked Questions added to Documentation. Why `__file__` is
+  not defined?
+
 v0.1.10
 -------
 
@@ -36,7 +43,6 @@ Developer changes
 
 * Move testing to py.test
 * Include link to github repository in documentation
-
 
 v0.1.8
 ------
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -1,0 +1,25 @@
+Frequently Asked Questions
+==========================
+
+.. contents:: **Contents**
+    :local:
+    :depth: 1
+
+
+Why is `__file__` not defined? What can I use?
+----------------------------------------------
+
+The global `__file__` variable defined by Python when running scripts
+is not defined on Jupyter notebooks. Since Sphinx-Gallery supports
+notebook styled examples and also exports to Jupyter notebooks we
+agreed on keeping this variable out of scope when executing the
+example scripts.
+
+Instead of `__file__` use :func:`os.getcwd` to get the directory where
+the file is located. Sphinx-Gallery executes the examples scripts in
+their source directory.
+
+
+.. seealso::
+    `Github PR #166 <https://github.com/sphinx-gallery/sphinx-gallery/pull/166>`_
+    `Github PR #212 <https://github.com/sphinx-gallery/sphinx-gallery/pull/212>`_
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -68,6 +68,7 @@ Contents:
 
    getting_started
    advanced_configuration
+   faq
    syntax
    utils
    reference
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -579,6 +579,7 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
         # Examples may contain if __name__ == '__main__' guards
         # for in example scikit-learn if the example uses multiprocessing
         '__name__': '__main__',
+        # Don't ever support __file__: Issues #166 #212
     }
 
     # A simple example has two blocks: one for the


amueller:
yeah that's the fix I did locally, thanks :)

lesteve:
There was already a PR like this and it was reverted because `__file__` does not exist in a Jupyter notebook:
https://github.com/sphinx-gallery/sphinx-gallery/pull/166

amueller:
I'm ok with not supporting it as long as there is an alternative for my use-case ;)
I want to load a file that's stored with the example. How do I get the path?

GaelVaroquaux:
> I want to load a file that's stored with the example. How do I get the path?

Maybe the right solution is that we CWD to the path where your example is
before running it (and CWD back after). That way, you can use os.cwd.

G

amueller:
but that doesn't work when just running the example script if you execute it from another folder, right?

GaelVaroquaux:
> but that doesn't work when just running the example script if you execute it from another folder, right?

Yes, but we can't satisfy both this and the notebook model. In a sense
the notebook model is more limited and hence frustrating when you're just
trying to do scripts. But there is value to the users to provide
notebook.

amueller:
It seems the best solution would be for me to check if ``__file__`` exists, and if so, use it, and otherwise assume we cwd'ed?

amueller:
though if I rely on a local file, how is the notebook download gonna be helpful? Neither of the downloads will work by themselves, right?

amueller:
But I'm generally not sold on the downloads, you can just copy and paste the script, which is usually much easier than opening a downloaded notebook. You can't open a notebook from the browser, you need to navigate to the download folder and start a server there....

GaelVaroquaux:
Well, it's complicated.

One the one hand, you are right that from a user interaction perspective,
the download is completaly broken because there is no easy way of opening
notebooks files. I think that this is a problem from the IPython side,
and I hope that nteract will solve it. It does pose real problems when we
teach.

On the other hand, notebooks and downloads enable us to do very didactic
example (eg
http://nilearn.github.io/auto_examples/plot_decoding_tutorial.html#sphx-glr-auto-examples-plot-decoding-tutorial-py).
For teaching, they are awesome. And the teaching (whether it is actual
teaching, or just someone running the code at home) is more smooth than
with copy pasting.

GaelVaroquaux:
> It seems the best solution would be for me to check if __file__ exists, and if so, use it, and otherwise assume we cwd'ed?

Why do you mind that much assuming that the user is in the right
directory? In the Mayavi examples, long before the notebook, this is what
we would do, and it is a simplifying assumption that works very well.

amueller:
I can assume that if I gain something from it ;) And I mind it because it will mean many more open issues on my issue tracker.

As I said above, the downloads actually don't include the file that I'm trying to load, so if I'm using ``__file__`` the notebook will be invariably broken - and the ``.py`` download too. For my use-case I probably should just remove the download buttons as there is no way to download a working example.

So there are two possible solutions: You can make sphinx-gallery cwd and I can add code to check if ``__file__`` exists and otherwise assume it's in the right directory.
Or you can add ``__file__`` to the globals.
Both solutions result in running the script locally to work, but the downloads not working. Only adding ``__file__`` makes the examples slightly easier.
I'm ok with either solution but I don't see a strong argument with the notebook perspective.

GaelVaroquaux:
I'd rather avoid adding __file__, because it means that people will be
creating broken notebooks without realizing it.

That said, maybe it would be good to raise the issue of the
presence/absence of __file__ with the IPython people.

amueller:
Alright. But the notebooks they create will probably be still as broken, as mine will be, even without relying on ``__file__``.

GaelVaroquaux:
> Alright. But the notebooks they create will probably be still as broken, as mine will be, even without relying on __file__.

You mean because it assumes that the data is there and it's not?

amueller:
yeah

Titan-C:
I completely forgot about the previous PR and why `__file__` is not
included. I should add a comment on that before including global
variables in gen_rst.py

On the other hand calling `os.getcwd()` must always work because
examples are run in the directory where the original source file is
located. That is to ensure required data is available, and that saved
data(for later use in cache) remains there and not in the gallery
created directory in the `doc` folder (which can be always be wiped).

>> Alright. But the notebooks they create will probably be still as broken, as mine will be, even without relying on __file__.
>
> You mean because it assumes that the data is there and it's not?
> yeah

There is not much Sphinx-Gallery can do here, if the user does not have
the data.

Titan-C:
I'm fixing this issue with documentation that `os.getcwd` should be used

amueller:
Ok so if I want my example to be able to run from the command line (from another folder) I need to do
```python
path = os.path.basename(__file__)  if "__file__" in locals() else os.getcwd()
```

Issue #113, opened 2016-03-26
Some external functions / classes don't resolve into links


adrn:
For example, in [this example](http://sphinx-gallery.readthedocs.org/en/latest/auto_examples/plot_exp.html#sphx-glr-auto-examples-plot-exp-py) in the sphinx-gallery gallery, `np.linspace` does not become a link, but all other numpy functions do.

lesteve:
Yeah this limitation is kind of known but I don't think anyone has had the bandwith to look into it in more details ... Insights or PR welcome :-) !

dfroger:
I'm trying to fix this issue too. Here is where I am.

For [the example](http://sphinx-gallery.readthedocs.io/en/latest/auto_examples/plot_exp.html#sphx-glr-auto-examples-plot-exp-py), the code in  [docs_resolv.py](https://github.com/sphinx-gallery/sphinx-gallery/blob/829dc34c1f2b40fc3a4c9738e882892c9c162d8f/sphinx_gallery/docs_resolv.py#L389-L404) behaves like this snippet:

    import re

    names = ['<span>np</span>', '<span>plt</span>', '<span>pd</span>']

    expr = re.compile(r'(?<!\.)\b' +  '|'.join(re.escape(name) for name in names))

    lines = ['<span>x</span> <span>np</span> <span>x</span>',
             '<span>x</span> <span>plt</span> <span>x</span>',
             '<span>x</span> <span>pd</span> <span>x</span>']

    for line in lines:
        print(expr.search(line) is not None)

which prints:

    False
    True
    True

In the `regexp`, word boundary `\b` apply only to first `name` of the `names` list. As `\b` does not matches `<` (first character in `name`), value `name` is not replaced to add `<a href`

The regular expression seems to want to work on something like `np.linspace`, where `(?<!\.)` prevent matching `xxx.np.linspace`, and `\b` to prevent matching `xxxnp.linspace`.

However, the regular expression is not used on something like `np.linspace`, but on:

    <span class="n">np</span><span class="o">.</span><span class="n">linspace</span>

lesteve:
@dfroger this seems like a good lead, sorry not to have answered sooner. Have you had the chance to get to the bottom of this?

lesteve:
[...]

PR #214, opened 2017-03-17
[MRG] Fix embedded missing link

lesteve:
Fixes part of https://github.com/sphinx-gallery/sphinx-gallery/issues/113. The regular expression was not matching for the first (i.e. also the longest) name.

I have to say I don't understand why the regex `'r(?<!\.)\b` was here ... but I double-check that on sphinx-gallery examples, it does add a link for the longest name.

For example this adds a link to `np.linspace` in this [example](http://sphinx-gallery.readthedocs.io/en/stable/auto_examples/plot_exp.html#sphx-glr-auto-examples-plot-exp-py) and a link to `np.random.RandomState` in this [example](http://sphinx-gallery.readthedocs.io/en/latest/auto_examples/plot_seaborn.html#sphx-glr-auto-examples-plot-seaborn-py).

Obviously adding a test would be good, but I have already spent too much time on this to my liking ...

### DIFF
--- a/sphinx_gallery/docs_resolv.py
+++ b/sphinx_gallery/docs_resolv.py
@@ -414,9 +414,8 @@ def _embed_code_links(app, gallery_conf, gallery_dir):
 
             # ensure greediness
             names = sorted(str_repl, key=len, reverse=True)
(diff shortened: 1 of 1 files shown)


Titan-C:
Thank you! Merged

PR #217, opened 2017-04-01
Move testing to py.test

### LINKED ISSUE #163: Move to py.test
Scikit-learn and joblib are in the process of moving to py.test since nose is not maintained any more. sphinx-gallery is the kind of small project that would be great as a test bed for the py.test move.

Titan-C:
Closes #163


.travis.yml
```diff
@@ -65,8 +65,7 @@ before_script:
   - sleep 3 # give xvfb some time to start
 
 script:
-    - if [ "$DISTRIB" == "ubuntu 14" ]; then python setup.py nosetests; fi
-    - if [ "$DISTRIB" == "conda" ]; then nosetests; fi
+    - python setup.py test
```
lesteve: Is there an advantage compared to using something like pytest sphinx_gallery? I think we dropped this in joblib because python setup.py test is just not that common in my experience. Happy to be proven wrong.

Titan-C:
> Is there an advantage compared to using something like pytest sphinx_gallery? I think we dropped this in joblib because python setup.py test is just not that common in my experience. Happy to be proven wrong.

I'm not aware of usage statistics. The idea behind this is setuptools
`setup.cfg` and that setuptools can use the dependencies needed for
testing without installing them. If pytest is already installed there is
no advantage. pytest does read `setup.cfg` independent of setuptools, so
you can do a direct call like

```bash
pytest
```

And it will work with all the custom configuration.

lesteve:
Why did you move the tests outside of the sphinx_gallery folder? Having tests inside the package is something that is completely standard.

Titan-C:
> Why did you move the tests outside of the sphinx_gallery folder? Having tests inside the package is something that is completely standard.

I don't remember anymore. I was experimenting with pytest and I needed
to try it for it to work on whatever I was testing. I did not keep track
to revert it back. I'll check that again later.

Titan-C:
> Why did you move the tests outside of the sphinx_gallery folder? Having tests inside the package is something that is completely standard.

I put the test back inside.

Titan-C:
mentions of nose are out

lesteve:
A very common expectation is to be able to run `pytest sphinx_gallery` to run the tests. At the moment this fails. One of the reason is that there is no `__init__.py` in sphinx_gallery/tests. There is another reason that I have no time to look into.

Also you have a non-optional dependency on matplotlib. I am not sure how big of a deal this is.

lesteve:
A very common expectation is to be able to run `pytest sphinx_gallery` to run the tests. At the moment this fails. One of the reason is that there is no `__init__.py` in sphinx_gallery/tests. There is another reason that I have no time to look into.

Also you have a non-optional dependency on matplotlib. I am not sure how big of a deal this is.

Titan-C:
> A very common expectation is to be able to run `pytest sphinx_gallery` to run the tests. At the moment this fails. One of the reason is that there is no `__init__.py` in sphinx_gallery/tests. There is another reason that I have no time to look into.

Having `__init__.py` is irrelevant for pytest scan. If you run `pytest`
only the test run well. The problem with `pytest sphinx_gallery` is that
pytest will in that case `cd sphinx_gallery` and in that case some
filepaths fail.

Now that sphinx-gallery can be called from arbitrary directories with
the proper `conf.py` the test files would probably be configured to also
allow calls from pytest from arbitrary locations. It is extra work,
which represented no gain. I will nevertheless end up implementing that
as a consequence of PR #225, where I plan to, in the same way as Sphinx,
use pytest to test itself create test build locations, and then all
this temporary build folders would end up being also a test case of
execution of the gallery build from arbitrary locations.

> Also you have a non-optional dependency on matplotlib. I am not sure how big of a deal this is.

Yes, that comes because I deleted all mentions of nose in the code. All
the nose hacks for making it not to compete with the sphinx-build, and
for segregating the test scan from the execution.

On a simple point of view: Matplotlib is a critical dependency of Sphinx
gallery, if it is not installed you should get the error at runtime if
not installed. For know we don't have in the test scripts any use of it
but with PR #225 those shall come into play.

On a broader point of view: We have users that embed Sphinx-Gallery in
their code base. And for some particular choice of their own
Matplotlib may not available in their test environment. But they still
pass their test scan over the doc/sphinxext/ folder and even continue to
use nose. Then removing all the nose hacks and having matplotlib as not
optional may annoy them.

I personally think this should not represent a big deal. Scikit-learn
(source all those hacks) already skips the test scan and the linter over
the sphinx-extensions.

lesteve:
Please make `pytest sphinx_gallery` work. This is non negociable, people will expect this will run the tests.

> The problem with `pytest sphinx_gallery` is that pytest will in that case `cd sphinx_gallery` and in that case some filepaths fail.

Nope this is not what is happening.

Quickly investigating (although I shouldn't have because I have plenty of other things on my plate trust me on this), the arguments of pytest find their way to jupyter_to_python_cli, probably sys.argv is used when args is None (passing args=[] may be an option). `sphinx_gallery` is thus passed to jupyter_to_python_cli as the python file to convert, hence the `IsADirectoryError: [Errno 21] Is a directory: 'sphinx_gallery'`.

Titan-C:
Thank you for your time @lesteve, you were right on the source of the error for `pytest sphinx_gallery`. I just pushed a tested fix to master. It works now.

lesteve:
FYI I pushed a change in master to add `__init__.py` into `sphinx_gallery/tests`.

PR #233, opened 2017-05-04
[WIP] adding folder sorting

choldgraf:
This is a riff off of @lesteve 's suggestion. Basically it adds an extra configuration option that takes a dictionary of `example_dir: list_of_folder_names` pairs.

It will sort folders first with the explicit ordering of `list_of_folder_names` for that `example_dir`, and if any other folders are not explicitly listed in the explicit sort, it will append them to the end of the sort.

LMK if this is generally a good approach and then I can add tests + make changes.


choldgraf:
First mentioned in #37

sphinx_gallery/gen_gallery.py
```diff
@@ -29,6 +29,7 @@
 DEFAULT_GALLERY_CONF = {
     'filename_pattern': re.escape(os.sep) + 'plot',
     'examples_dirs': os.path.join('..', 'examples'),
+    'examples_dirs_order': None,
```
lesteve: I would have a example_dirs_sort_key or something, which would be used like this in conf.py:

```
'example_dirs_sort_key': ExplicitOrderKey(['folder', 'another_folder'])
```

Not by that default, we should still sort alphabetically, but I am guessing this is what `key=None` does.

sphinx_gallery/gen_gallery.py
```diff
@@ -301,3 +313,22 @@ def setup(app):
 
     app.connect('build-finished', sumarize_failing_examples)
     app.connect('build-finished', embed_code_links)
+
+
+class ExplicitOrderKey(object):
```
lesteve: We could put this in `sphinx_gallery/sorting.py` or something like this.

sphinx_gallery/gen_gallery.py
```diff
@@ -301,3 +313,22 @@ def setup(app):
...
+    """Explicitly order a list of strings."""
+    def __init__(self, ordered_list):
+        if not isinstance(ordered_list, list):
+            raise ValueError('ordered_list must be a list')
+        self.ordered_list = ordered_list
+        self.unordered_items = []
+
+    def __call__(self, item):
+        if item in self.ordered_list:
```
lesteve: Unless you have a specific use case in mind, I would be in favour of requiring the user to specify all the folders and raise an exception if `item` is not in `self.ordered_list`. IIRC `os.listdir` is not deterministic so this could introduce ordering changes in the gallery that are confusing to the user.

choldgraf:
Good points - here's what it looks like in my `conf.py`:

```
sphinx_gallery_conf = {
    ...
    'examples_dirs': ['examples']
    'examples_dirs_order': {'examples': ['test_c', 'test_b']}
```
So the way I had originally done it, you'd give a dictionary where keys would be a folder in which keys are folders in `examples_dirs` and values correspond to names of folders within. We could just make it a simple list of folder names that are relative to the project root, though that could be a lotta text.

re: ordering, I added in support for non-explicit folders in case there are many folders in there and you just care that one or two show up before everything else. I think I may be in an atypical usecase as I'm trying to get all of matplotlib's docs into sphinx gallery. Either way, I think it'd be easy to sort any remaining folders alphabetically so they're consistent. WDYT?

sphinx_gallery/gen_gallery.py
choldgraf: Yep - that's what key=None does, so by default the behavior won't change.

sphinx_gallery/gen_gallery.py
choldgraf: I'll give it a shot and we can see how it looks there

choldgraf:
see latest push for addressing the above comments!

choldgraf:
Also added a test to make sure the explicit ordering works as intended.

choldgraf:
LMK if there's anything else I should do (e.g. perhaps we should discuss the `conf.py` variable names / values?

choldgraf:
ping  :-)

sphinx_gallery/gen_gallery.py
Titan-C: `None` feels very strange to me, like giving the option to be completely random. It also hides the type we expect in this field. I would set as default the empty dictionary.

sphinx_gallery/gen_gallery.py
```diff
@@ -180,11 +182,22 @@ def generate_gallery_rst(app):
...
+        if gallery_conf['examples_dirs_order'] is not None:
+            base_examples_dir = os.path.basename(examples_dir)
+            ordering = gallery_conf['examples_dirs_order']
+            if not isinstance(ordering, dict):
+                raise ValueError('If not None, examples_dirs_order'
+                                 ' must be a dictionary of folder:'
+                                 ' ordering pairs')
+            if base_examples_dir in ordering.keys():
+                key = ExplicitOrderKey(ordering[base_examples_dir])
+
```
Titan-C: Once our default is the empty dictionary we don't need the first check. Directly start checking the user has input a dictionary the config or raise the ValueError. The message of the error is unclear to me, we need to make this more comprehensible.

sphinx_gallery/gen_gallery.py
```diff
@@ -180,11 +182,22 @@ def generate_gallery_rst(app):
 
         computation_times += this_computation_times
 
+        key = None
+        if gallery_conf['examples_dirs_order'] is not None:
+            base_examples_dir = os.path.basename(examples_dir)
```
Titan-C: This will be very prone to failure. This is because `examples_dir` is an absolute path, but with `os.path.basename` you just get the last folder name. Nothing guaranties you that it will match with the user folder path. I think we should move this code block inside `_prepare_sphx_glr_dirs`, that one has the user's original paths, and we can make it return the absolute paths to work on + the sorting key function

Titan-C:
I went various times over the code and changed my mind many times as well and I can not yet decide. Thus I just write my thoughts, please criticize them.
### The configuration:
* I first thought the dictionary to be the best option for holding the order of the galleries. In that case our config would be:
```python
sphinx_gallery_conf = {
     # the list of strings holding the many main galleries
    'examples_dirs': ['../examples']
   # for each gallery a key with the same string as in the previous list.
    'examples_dirs_order': {'../examples': ['test_c', 'test_b']} 
```
But here you have to repeat some entries. And repeating is error prone(at least for me)
* Then I thought in nested lists
```python
sphinx_gallery_conf={
    # we have everything in the same variable. Gallery dirs are inside a list.
    # If a gallery has an explicit sorting, the entry is a tuple with the gallery 
    # path and a list of subfolders
    'examples_dirs' : [('../examples',  ['test_c', 'test_b']),
                       '../tutorials'] # the second gallery has no explicit sorting
```
### The backend
* I think we should move all this work to the function `_prepare_sphx_glr_dirs`. That one shall output the example dirs, gallery dirs, already in absolute paths and sorting keys for each gallery subsection
* I'm in favour, if explicit ordering is given, that all folders have to be explicitly given. But in that case why have a sorting key? Just use the given list. On the other hand if folders are not specified in the list, but they have the `README.txt` should we raise an error or silently skip them?(And in this case having a function to listdir and sort is needed again). +1 for sorting key.

sphinx_gallery/gen_gallery.py
choldgraf: cool - I think this is a good point, let me see if I can work something out per your suggestions that's more stable

lesteve:
As far as I am concerned a sort key is the best API I can think of. Small changes to the code (just add a `key` argument to the `sort` call), fully generic and I could not find use cases where a sort key did not cut it.

Also I want to keep within-folder and between-folder ordering separate as I commented somewhere else. This PR is about between-folder ordering, which I think is the most common thing that people want to do to order the gallery, let's keep it this way.

choldgraf:
> On the other hand if folders are not specified in the list, but they have the README.txt should we raise an error or silently skip them?

Skip them as in don't add them to the gallery at all? In this case I would say why not just sort them alphabetically :-)  but if you guys don't want to support that usecase I think throwing an error is better.

lesteve:
> In this case I would say why not just sort them alphabetically :-) but if you guys don't want to support that usecase I think throwing an error is better.

I would be in favour of raising an exception if a folder is not present in the `ExplicitOrderKey` instance with an informative error message. It is better to nudge the user towards fully specifying the folders he cares about. In general I don't expect the list of the folders to be that long, so it should not be too painful to maintain.

The good thing if we have the sort key as API is that the user can implement the sorting logic the way he wants to, so your use case to specify only a few folders and then sorting the rest alphabetically could be catered for (as long as your are willing to write the sort key logic yourself).

Longer term, you could imagine having a class (or a parameter in `ExplicitOrderKey`) to allow missing items. In my mind this is more useful for within folder ordering where you only care about a few examples being shown first (e.g. the very introductory examples) and don't really care about the ordering of the others.

choldgraf:
> I would be in favour of raising an exception if a folder is not present in the ExplicitOrderKey instance with an informative error message.

Sounds like you two are in agreement on this point, so I can go with this for the current PR. I think sorting w/ a list that doesn't cover all the folders is only useful for people w/ large numbers of folders in the examples folder (e.g., matplotlib :P )

> so your use case to specify only a few folders and then sorting the rest alphabetically could be catered for (as long as your are willing to write the sort key logic yourself).

isn't this use-case already implemented in this PR? ;-)

choldgraf:
or are you saying that you could give a sort key in the `conf.py` file itself?

lesteve:
> or are you saying that you could give a sort key in the conf.py file itself?

Yep that is the idea, sorry if I was not clear enough.

choldgraf:
I see - that's an interesting idea. Fine with me. How would users use `conf.py` for explicit sorting then? They'd just always have to write their own key? In that case, maybe it'd be useful to either put "recipes" for common keys in the docs, or have some pre-defined keys in `sorting.py` that users could import in their `conf.py`.

lesteve:
Probably what I have in mind is a bit scattered over a few issues/PRs but basically, we would provide a few basic sorting keys for convenience e.g. the `ExplicitOrderKey` (by the way the name can probably be improved) and leave the user implement more complicated things if he wants to.

Putting the sort keys in `sorting.py` or `utils.py` (or a better name) would be fine with me.

choldgraf:
yeah that's what I'm imagining too then. So in `conf.py` you could have something like:

```
from sphinx_gallery.sorting import ExplicitOrderKey

key = ExplicitOrderKey(['folder1', 'folder2'])

sphinx_gallery_config_dict  = {
      ...
      folder_sort_key: {'gallery_folder': key}
}
```
?

lesteve:
Something along these lines. I was thinking of something like a global `folder_sort_key`, e.g.:

```
sphinx_gallery_config_dict = {
    ...
    folder_sort_key = ExplicitOrderKey(['folder', 'another_folder'])
}
```

Do you see anything wrong with that? You could argue that if you have multiple top folders of examples (e.g. in sphinx_gallery we have `examples` and `tutorials` but most usages of sphinx-gallery I know of only have one) the ExplicitOrderKey is going to mix both folders. I don't think this is a huge issue, and I'd rather go for simplicity here, unless there is a very good reason where you would want a per top-folder sort key. Even then you can always write your own sort key ;-).

choldgraf:
ah I see - I was imagining people could do `{'folder_1': key1, 'folder_2': key2}`. I think you're right that we are getting into the weeds of potentially rare use cases :)

How about in this PR we just let people give a key which takes paths relative to the documentation root, rather than relative to the gallery folder. That way they can be explicit. You could then easily do something like:

```
gallery_1_sort = ['/gallery/folder_1', '/gallery/folder_2']
gallery_2_sort = ['/gallery2/folder_1', '/gallery2/folder_2']
sphinx_gallery_config_dict = {
    ...
    folder_sort_key = ExplicitOrderKey(gallery_1_sort + gallery_2_sort)
}
```

lesteve:
AFAIK each top-folder of examples generate a separate gallery so I don't think this is a problem. Basically imagine you have a folder structure like this:
```
tutorials
|---- tuto_folder
|---- tuto_another_folder
examples
|---- ex_folder
|---- ex_another_folder
```

Then in my mind, having:

```
folder_sort_key = ExplicitOrderKey(['tuto_folder', 'tuto_another_folder', 'ex_folder', 'ex_another_folder'])
```
would work.

About which paths (relative to the root project, relative to conf.py, absolute) gets passed to the `sort` call I have to say that @Titan-C probably has better insights about this than me.

choldgraf:
Yeah but what if you had

```
tutorials
|---- folder_1
|---- tuto_another_folder
examples
|---- folder_1
|---- ex_another_folder
```
And you wanted folder_1 to be in a different position in each folder? I am trying to think of annoying edge-cases that we'd have to engineer around in the future, not trying to be a jerk :-)

Titan-C:
I have some of the exotic use case. I make galleries according to
topics. So I have a gallery on lattice simulations, other in quantum
Monte Carlo methods, other in perturbation theory, but all use the same
base library so they belong to the same package. And of course each
gallery has a certain amount of subsections which are subprojects in each
topic. I never really cared about sorting, because it is not aimed
beyond me having a visual overview of what script produced which result.

That is the reason I ended up with the idea of having nested lists to
specify the various galleries and the subsection sorting within each
one.

I would agree on directly exposing the sort key to the user, in that way
a config would look to me

```python
from sphinx_gallery.sorting import ExplicitOrder
sphinx_gallery_conf = {
    "examples_dirs": [("../pert_theory", len),
                      ("../Monte_Carlo", ExplicitOrder(["2D Ising", "3D Ising",
                                                        "2D Random Field Ising",
                                                        "2D Potts"])),
                      "../dimer_lattice"],
    "gallery_dirs": ["pert_theory", "Monte_Carlo", "dimer_lattice"]}
```

This way my first gallery is sorted by the length of the folder name,
the second is explicit and the last is our standard alphabetically. In
that way the user can also put his sort key at will. And if the sort key
function is exposed the we can have explicit sort which raises an
exception `ExplicitOrderStrict` if a folder is missing and one that
silently continues `ExplicitOrderLeaders`. Other naming accepted.


> About which paths (relative to the root project, relative to conf.py, absolute)
> gets passed to the `sort` call I have to say that @Titan-C probably has better
> insights about this than me.

This is something to be careful. "examples_dirs" takes a path relative
to conf.py. On execution `os.listdir` receives the absolute path of the
folder and outputs a list with the names of the directories in the path,
without the leading path. Then it is only this names that get
sorted. That is the reason I wanted to "glue" the sorting-key to the
example path definition and not making it a dictionary lookup.

lesteve:
Yeah I thought about that. That could happen, I guess. 

A possible solution: rather than folders you would need to pass more complete paths to the `sort` call (maybe it is already the case I am not sure). The `ExplicitOrderKey` would use these more complete paths as well.

choldgraf:
So it sounds like we have two proposed solutions: 1 that uses nesting (similar to the current implementation) and one that just supports a single key. Any of you feel strongly about one of those two options?

lesteve:
> So it sounds like we have two proposed solutions: 1 that uses nesting (similar to the current implementation) and one that just supports a single key. Any of you feel strongly about one of those two options?

Let's keep this one simple please and go for the single key that does only between-folder ordering.

Titan-C:
>> So it sounds like we have two proposed solutions: 1 that uses nesting (similar to the current implementation) and one that just supports a single key. Any of you feel strongly about one of those two options?
>
> Let's keep this one simple please and go for the single key that does only between-folder ordering.

This PR is only concerned about between-folder ordering. My only concern
is that in the case of multiple galleries having a single key specified
apart from the galleries path definition could give some problems.

lesteve:
> My only concern is that in the case of multiple galleries having a single key specified apart from the galleries path definition could give some problems.

My take on it is that we should go for the simplest use case first. Also if the paths that are passed to the `sort` function are complete enough (i.e. contains tutorials or examples in my example above) you could still decide to implement the logic the way you want in your key function so in practice you could have a different sorting-logic on a per top-folder level.

choldgraf:
+1 to simplest use case first. If our key just takes a list of paths to folders, then users could write whatever function they want to create such a list, and then pass it to our key, right?

choldgraf:
See latest push, which requires all folders to be explicitly ordered, and the folder position is related given relative to `conf.py`, similar to how you'd specify `examples_dirs`. So your configuration could look like this:

```
   ...
    'examples_dirs': '../examples',
    'gallery_folder_order': ['../examples/test_c', '../examples/test_b',
                                          '../examples/test_d', '../examples/test_a']
   ...
```

and if you didn't specify one of the folders in there, it would throw a ValueError.

Looks good?

sphinx_gallery/gen_gallery.py
```diff
@@ -14,11 +14,13 @@
 import copy
 import re
 import os
+import os.path as op
```
Titan-C: I know this is a common convention for some. We have seen it inside some PR for sphinx-gallery. Nevertheless we prefer to be explicit and write `os.path` every time.

sphinx_gallery/gen_gallery.py
```diff
@@ -181,17 +195,23 @@ def generate_gallery_rst(app):
...
-                                                                        seen_backrefs)
-                fhindex.write(this_fhindex)
-                computation_times += this_computation_times
+
+        examples_folders = [op.join(examples_relative_dir, idir)
+                            for idir in os.listdir(examples_dir)
+                            if op.isdir(op.join(examples_relative_dir, idir))]
+        for path_directory in sorted(examples_folders, key=key):
+            directory = path_directory.split(examples_relative_dir)
+            directory = directory[-1].strip('/')
```
Titan-C: This is an unnecessary complication

Titan-C:
@choldgraf thanks for keeping on with this

> So your configuration could look like this:
```python
    'examples_dirs': '../examples',
    'gallery_folder_order': ['../examples/test_c', '../examples/test_b',
                             '../examples/test_d', '../examples/test_a']
```
I feed this makes it very complicated for the setup and for the code.  I really prefer to keep `examples_dirs` relative to `conf.py` and the `gallery_folder_order` the directory list for a direct comparison to the output to `os.listdir`. That is to say relative path to the example folder. I really want to push for a dictionary of a nested list to look like
```python
"examples_dirs" : "../examples",
"gallery_subsection_order": {"../examples": ["test_a", "test_z", "test"]}
```
or
```python
"examples_dirs":[("../examples",  ["test_a", "test_z", "test"])]
```

A final thought. I know CI is not complaining, but this PR fails to build our documentation correctly. The back references minigallery has only the seaborn example(no Idea why) and the gallery which has a subsection does not show that subsection. That misses the point of this PR which is to order the subsections. Paths are not handled correctly. It is just a too complicated implementation.

choldgraf:
> I really want to push for a dictionary of a nested list to look like

This is something closer to the first version of this PR, which used nested dictionaries. I thought you guys were pushing back on that and preferred a single list, but maybe I mis-interpreted that. It would be good to have agreement before I spend the time implementing some other structure.

> A final thought. I know CI is not complaining, but this PR fails to build our documentation correctly. 

This is a good point - it's hard to tell what exactly is being tested in travis. It sounds like there is some functionality that isn't covered in the tests. Good point about checking the sphinx gallery docs as an eye-test. I'll do that in the future.

choldgraf:
Can we get some consensus on whether to do dictionary w/ list vs. one list of paths? I'd rather not spend the time implementing something if it'll get reverted later on. Just wanna know which direction to move towards... @Titan-C @lesteve

Titan-C:
@choldgraf. Thank you for keeping up with this. I get the feeling this is converting into bikeshedding. Thus I decided to try out my own implementation, which turned out to be way more complex than I expected but it is now done as a proof on concept in PR #234. I have based it on your work and some of my ideas too. I have certainly encountered many problems that weren't evident as we were discussing the issue and given how you wrote your code, now I understand, since in my first iteration I thought some of it was unnecessarily complicated. This feature request is a little more complex than I first imagined.

Here are my new ideas:
We are dealing only with the subsection ordering. I think that given how simple it is (at least in my mind) to rename the folders and keep the alphabetic folder sorting, we can allow in this PR or mine PR #234 to have a sort key for each big gallery. (This is mostly to deal with my use case and config of comment https://github.com/sphinx-gallery/sphinx-gallery/pull/233#issuecomment-300277287). Making a sortkey for each gallery explicitly also avoids the issue of comment https://github.com/sphinx-gallery/sphinx-gallery/pull/233#issuecomment-300251107. The individual sortkey also allows the user to try a different implementation, for a particular gallery without interfering with other galleries he might have, even if most of our current users only have one gallery.

choldgraf:
@Titan-C hey - sounds good to me. I was working on this this morning as well actually...I'll push the latest version now and we can compare. TBH I am happy with whatever implementation, so long as it's not super confusing for users to implement it. Wanna take a look and then decide which PR to go with? I mostly would like this to be merged sometime before the next Matplotlib 2.1 release (which, given history, will probably be in 2020 so maybe it's not such a big deal 😆 )

choldgraf:
latest push has these changes...works with nested dictionary sorting. It's a little bit hacky but maybe not too bad

lesteve:
I am afraid I have to insist on the single global sort key option rather than cramming the sort key into the `example_dirs` key. As I commented elsewhere, as long as the paths are complete enough (i.e. don't lose the top-folder name) you would be able to implement a separate-logic through a custom sort key.

There may be some issues with paths that I haven't had (and probably won't have before a while) the time to fully understand but I tend to think of it as a different issue. If I can give my gut feeling: there may be discrepancies at the moment but I reckon that all paths should be relative to conf.py.

choldgraf:
I think the question should be: what does the user side of things look like? In conf.py, do they:

1. create sorting keys for each root folder
1. create one sorting key for all root folders
1. supply a nested dictionary of root folder: lists of folder names pairs
1. supply a list of folder names relative to `conf.py`

Answering this question is most important and then the details of the implementation can evolve over time as long as the user-facing API doesn't change.

Also on this note, that's twice now that I've spent a bunch of time implementing something that is then asked to revert. I'm not going to make and more pushes to this PR until we have a group consensus on this.

lesteve:
I am in favor of `1.` as is probably clear from my other posts above.

Pros:
* simplicity
* genericity: covers all use cases that I can think of as long as "complete enough" paths are passed into the `sort` call.

Cons:
* maybe "complete enough" paths is the hard part I am not sure.

About other proposals I feel that cramming sorting logic into `example_dirs` is less than great. If we implement within-folder sorting are we going to add the within-folder sort key in `example_dirs` as well? I feel having a separate key for between-folder ordering and a separate one for within-folder ordering is the right way to go.

> Also on this note, that's twice now that I've spent a bunch of time implementing something that is then asked to revert. I'm not going to make and more pushes to this PR until we have a group consensus on this.

I do get how frustrating this is and I am deeply sorry about that.

choldgraf:
The main challenge I ran into when implementing 1 is that it entails changing the machinery of the `gen_rst` function, which currently iterates just through folder names within each root folder, it doesn't iterate through the full folder paths. I guess another con is that it requires users to utilize a sorting key, no? I don't think most people use them so it'd be a barrier for some devs.

@Titan-C thoughts?

Titan-C:
> About other proposals I feel that cramming sorting logic into example_dirs is less than great. If we implement within-folder sorting are we going to add the within-folder sort key in example_dirs as well?

It's true, I'm certainly putting too many things together. I would be fine having one sortkey for everything. And input as a dictionary of example path relative to `conf.py` and a list of folders names having the order as is now, the sortkey is not exposed to the user.

choldgraf:
I tend to lean against asking users to create a sorting key, as my guess is that most of them don't know how to do this. Perhaps they can optionally create one themselves, but could also supply a dictionary of `root_folder: list_of_nested_folders` pairs. WDYT?

Titan-C:
Indeed, I don't think exposing the sortkey to the user is a good
idea. The user should only supply a dictionary with the listed
folders. The sortkey is for internal use, to be able to raise exceptions
and log errors, to have a better control of the input rather than directly
iterating over the list of folders provided by the user.

choldgraf:
@Titan-C so, is this different from the current push? If so, lemme know what you'd like me to modify :-)

lesteve:
> I tend to lean against asking users to create a sorting key, as my guess is that most of them don't know how to do this.

Hmm I tried to argue why having a sortkey is a good idea but it looks like I failed so far :disappointed:  ... I think having an example in the doc covering the most common case i.e. using ExplicitOrderKey would be good enough to alleviate concerns that most user do not know how to do this.

GaelVaroquaux:
> I tend to lean against asking users to create a sorting key, as my guess is
> that most of them don't know how to do this. Perhaps they can optionally create
> one themselves,

Optional sorting key is also my feeling.

choldgraf:
I think that you make some good points @lesteve , just to go point-by-point so that we're on the same page:

The 'pros' you note are:

> simplicity

I'd argue that sorting keys are the most simple from the standpoint of this PR, but the least simple from the standpoint of the users. I've been doing dev in python for 4 years now, and this PR was the first time I had heard of a sorting key, and I had to spend some time trying to figure out just how they work. My concern is we are requiring users to learn a new (relatively idiosyncratic) python concept just so that they can sort folders.

> genericity: covers all use cases that I can think of as long as "complete enough" paths are passed into the sort call.

I agree with this one, though can we imagine other use-cases here? What other kind of sorting rule would a person want other than "I want to choose the order of my folders manually"? I'm a little bit worried that we are going to make something more confusing for users by trying to accommodate a use-case that it's unclear people would really want.

Either way, here's the actionable stuff:

If we want to use a sorting key, I think we should do:

* Tell users to give a dictionary of `gallery_dir: sorting_key` pairs and tell them that the sorting key will handle the ordering of folders within each gallery directory.
* Provide users `ExplicitSortKey` with the code itself, and also include a bit in the docs for how to do this in case they wanna roll their own.

Would that work? It's sort of a generalization of the current PR so shouldn't be too tough to implement.

@Titan-C curious to hear what you think about this as well!

Titan-C:
This has gone very far into opinions for a matter I would have solved by renaming my folders with a number prefix, zero python implementation.

If we want to have some custom order, then a sortkey, I fully agree, is the best option. I would have gone to one sortkey per gallery, I have a proof of concept in PR#234, but for the sake of consensus with @lesteve, let's make one for all galleries.

For the sake of consensus too: Let's expose the sortkey to the user. Only to make him aware that it is being used. That means I take back that the users should only supply a dictionary with the listed
folders.

In concluding point of view. I'm happy enough if the user config looks like:

```python
from sphinx_gallery.sorting import ExplicitOrder
sphinx_gallery_conf = {
    "examples_dirs": ["../examples", "../tutorials", "../references"]
    "gallery_subsection_order": ExplicitOrder({"../examples":["first","second","third"],
                                               "../references":['test_z','test_x', 'test_y']})}
```

Then we write in the docs that we use a sortkey to provide order to gallery subsections. That our provided `ExplicitOrder` sortkey takes as input a dictionary where the key must match the `examples_dir` items and values are a list of folder names. This key raises an exception if a folder
is missing on the list. Finally the user can provide his personal global sortkey, for all galleries if he ever dares to implement one.

choldgraf:
fwiw my preference would be to do the above thing, but have people supply `gallery_dir: key` pairs, rather than having a key that takes a dictionary of `gallery_dir: list_of_folder` pairs. It'd be simpler to implement since we wouldn't need to change any of the machinery of how we iterate through folders, and would be more extensible if people want to do different things with different folders. That said I don't feel super super strongly about it, so if everybody agrees on a different approach I can do that...

Titan-C:
After a face to face discussion with @lesteve we have agreed on what we want. I have prepared a working implementation in PR #234.
Details of naming variables are still to be agreed on. It has been so painful to get this working, that I would kindly appreciate you comments on it too.

choldgraf:
I am about to head off for a few days so I'll have spotty internet, but I'd say that if you two have met F2F about this, I trust your judgment about the API design and am willing to close this PR in lieu of your proposal. My general 2 cents are that things should be as simple as possible for the user, and to require minimal extra knowledge in order to implement this feature on the user end!

Otherwise I'll take a look when I get back around Thursday or Friday!

PR #234, opened 2017-05-08
[MRG] Other gallery subsection sorter

Titan-C:
A competing implementation of #233 to deal with #37. In this case the gallery sections are sorted using a sortkey given by the user. This sortkey can be individually configured for each main gallery. For the case of Sphinx-Gallery's main examples I have reorganized it to have to subsections that now with this implementation would appear in a different order than their folder names would dictate.
The configuration uses the concept of nested list. There is a main list of strings in "examples_dirs' to define where galleries examples are located, but that string can be turned into a tuple, containing the gallery directory and the sortkey. This PR provides the sortkey of explicit ordering, which raises an exception is a subgallery folder is missing from the list, but does not raise errors if folders which don't generate a gallery are not in the list.

Closes #233


doc/advanced_configuration.rst
```diff
@@ -77,6 +77,39 @@ starting with ``plot_long_examples_``, you would do:
 As the patterns are parsed as `regular expressions`_, users are advised to consult the
 `regular expressions`_ module for more details.
 
+.. _sub_gallery_order:
+
+Sorting the subsections inside a gallery
+========================================
+
+Gallery subsections are sorted by default alphabetically by their folder
+name, and as such you can always organize them by changing you folder
```
choldgraf: your folder

doc/advanced_configuration.rst
```diff
@@ -77,6 +77,39 @@ starting with ``plot_long_examples_``, you would do:
...
+names. An alternative option that Sphinx-Gallery offers is to use a sortkey
```
choldgraf: remove "that Sphinx-Gallery offers"

doc/advanced_configuration.rst
```diff
@@ -77,6 +77,39 @@ starting with ``plot_long_examples_``, you would do:
...
+to organize those subsections. We provide an explicit order sortkey where
+you have to define the order of all subfolders in your galleries.
+
+.. code-block:: python
+
+    from sphinx_gallery.sorting import ExplicitOrderStrict
+    sphinx_gallery_conf = {
+        'examples_dirs': ['../examples','../tutorials'],
+        'subsection_order': ExplicitOrderStrict(['../examples/sin_func',
+                                                '../examples/no_output',
```
choldgraf: need an extra space to line up the indent w/in the list, no?

doc/advanced_configuration.rst
```diff
@@ -77,6 +77,39 @@ starting with ``plot_long_examples_``, you would do:
...
+                                                '../tutorials/seaborn']),
+    }
+
+Here we build 2 main galleries `examples` and `tutorials`, each of them
+with subsections. To specify their order explicitly in the gallery we
+import :class:`sphinx_gallery.sorting.ExplicitOrderStrict` and initialize
+it with the list of all subfolders with their paths relative to `conf.py`
+in the order you prefer them to appear. You can use Linux paths, and if you
+documentation is build in a Windows system, paths will be transformed to
```
choldgraf: built

doc/advanced_configuration.rst
```diff
@@ -77,6 +77,39 @@ starting with ``plot_long_examples_``, you would do:
...
+work accordingly, the converse does not hold.
+
+If you so desire you can implement your own sorting key. It will be
+provided the relative paths to `conf.py` of each sub gallery folder.
```
choldgraf: Maybe a link to more information about sorting keys / how to build them?

doc/conf.py
```diff
@@ -299,6 +300,7 @@ def setup(app):
     'matplotlib': ('http://matplotlib.org/', None),
 }
 
+from sphinx_gallery.sorting import ExplicitOrderStrict
```
choldgraf: What benefit do we get from calling it `Strict` at the end? Why not just call it `ExplicitOrder` and then upon initialization give people the option of passing `strict=True` and defining what this does in the docstring?

sphinx_gallery/sorting.py
```diff
@@ -0,0 +1,40 @@
...
+import types
+
+
+class ExplicitOrderStrict(object):
+    """Sorting key for all galleries sub sections
+
+    Parameters
+    ----------
+    ordered_list : list, tuple, types.GeneratorType
+        Hold the paths of each gallery subfolder"""
```
choldgraf: I usually put the ending """ on another line but just my preference

sphinx_gallery/sorting.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def __init__(self, ordered_list):
```
choldgraf: so here we could have `explicit=True` on by default

sphinx_gallery/sorting.py
```diff
@@ -0,0 +1,40 @@
...
+                             "tuple or Generator, which hold"
+                             "the paths of each gallery subfolder")
+
+        self.ordered_list = list(os.path.normpath(path)
+                                 for path in ordered_list)
+
+    def __call__(self, item):
+        if item in self.ordered_list:
+            return self.ordered_list.index(item)
+        else:
```
choldgraf: and here could be an `elif self.strict is False: return np.inf` or something like that.

Though if we don't want to support that behavior it's fine with me, just don't think we need to call it `ExplicitOrderStrict` in that case since it is *always* strict

doc/conf.py
```diff
@@ -325,6 +327,10 @@ def setup(app):
         'numpy': 'http://docs.scipy.org/doc/numpy-1.9.1'},
     'examples_dirs': examples_dirs,
     'gallery_dirs': gallery_dirs,
+    'subsection_order': ExplicitOrderStrict(['../examples/sin_func',
```
QuLogic: It's not exactly clear to me from this PR how this `ExplicitOrderStrict` class is of benefit to a user over just specifying a list?

lesteve: I would be fine with allowing a list for simplifying the user's life. We should still allow a sort key to cover the general case. There were long discussions on this in related issues and PRs.

sphinx_gallery/sorting.py
Titan-C: Ok. Since it is always strict I'll drop it from the name.

doc/advanced_configuration.rst
```diff
@@ -77,6 +77,39 @@ starting with ``plot_long_examples_``, you would do:
...
+                                                 '../tutorials/seaborn']),
+    }
+
+Here we build 2 main galleries `examples` and `tutorials`, each of them
+with subsections. To specify their order explicitly in the gallery we
+import :class:`sphinx_gallery.sorting.ExplicitOrder` and initialize
+it with the list of all subfolders with their paths relative to `conf.py`
+in the order you prefer them to appear. You can use Linux paths, and if you
+documentation is built in a Windows system, paths will be transformed to
+work accordingly, the converse does not hold.
```
choldgraf: haha - you mean we don't support `C://` paths on linux?!?! Dealbreaker. ;-)

doc/advanced_configuration.rst
```diff
@@ -77,6 +77,39 @@ starting with ``plot_long_examples_``, you would do:
...
+names. An alternative option is to use a sortkey to organize those
+subsections. We provide an explicit order sortkey where you have to define
+the order of all subfolders in your galleries.
+
+.. code-block:: python
+
+    from sphinx_gallery.sorting import ExplicitOrder
+    sphinx_gallery_conf = {
+        'examples_dirs': ['../examples','../tutorials'],
+        'subsection_order': ExplicitOrder(['../examples/sin_func',
```
choldgraf: Could we add one sentence just pointing out the fact that even though there are multiple examples_dirs, there is only one `ExplicitOrder` key given to subsection_order? I know that people could infer this from looking at the folder names in the sorting key, but I think it might confuse some folks who assume you'd need to give two `ExplicitOrder` keys, one for each gallery folder.

choldgraf:
I had 1 very short comment, other than that LGTM

doc/advanced_configuration.rst
Titan-C: Yes. File-system absolute paths from windows will not work (and I guess in linux it won't work eithr). We require always absolute paths to the location of the conf.py

doc/advanced_configuration.rst
Titan-C: Sure. Better to be redundant

doc/advanced_configuration.rst
```diff
@@ -94,17 +94,20 @@ the order of all subfolders in your galleries.
...
 with subsections. To specify their order explicitly in the gallery we
-import :class:`sphinx_gallery.sorting.ExplicitOrder` and initialize
-it with the list of all subfolders with their paths relative to `conf.py`
-in the order you prefer them to appear. You can use Linux paths, and if you
-documentation is built in a Windows system, paths will be transformed to
-work accordingly, the converse does not hold.
+import :class:`sphinx_gallery.sorting.ExplicitOrder` and initialize it with
+the list of all subfolders with their paths relative to `conf.py` in the
+order you prefer them to appear. Keep in mind that we use a single sort key
+for all the galleries that are build, thus we include the prefix of each
```
choldgraf: *built

(sorry for nitpicking)

doc/advanced_configuration.rst
```diff
@@ -94,17 +94,20 @@ the order of all subfolders in your galleries.
...
+gallery in the corresponding subsection folders. One does not define a
+sortkey per gallery. You can use Linux paths, and if you documentation is
```
choldgraf: *if your documentation

doc/advanced_configuration.rst
Titan-C: Don't worry. This one is hard for me, I realize I have made that same mistake on this PR earlier on.

Titan-C:
Should be fine now. Thanks @choldgraf, @QuLogic, @lesteve for the review on this and the heated discussion about how implement ordering for the subsections. I'm merging now.

choldgraf:
Cool! 🎉 🎉 🎉 🎉

lesteve:
Great stuff!

PR #160, opened 2016-11-07
[MRG] Large codebase refactoring

Titan-C:
This is inspired by the way PR #148 handles the creation of the jupyter notebook, and requires that PR to be merged first. The idea is writing the rst file shall be completely independent of the execution of the script. This PR splits the script execution from the preparation of the RST file

PR #190, opened 2016-12-21
[MRG] Better path handling

Titan-C:
This replaces #133 
As brought back to life by #186, Sphinx-Gallery does not handle the cases when the documentation structure is in a different place than the location of the makefile and the conf.py files.

This PR aims to fix that and preceding issues stated in  #40 #50 #115 #119 #120

For test completeness I know build in circle CI the documentation in various location, with different commands a different themes.

As can be seen from the build logs(https://circleci.com/gh/Titan-C/sphinx-gallery/12?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). The examples are run only on the first build, later builds can take advantage of the cached files as they are referenced in the same locations and the output to the described location. For review. This are the 3 builds

* alabaster theme: https://12-26191147-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/alabaster_html/index.html
* classic theme: https://12-26191147-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/doc_build/html/tutorials/index.html
* read the docs theme: https://12-26191147-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/doc/_build/html/advanced_configuration.html#establishing-local-references-to-examples

Since it is possible I have missed cases where links/images/references/etc are not well resolved, the links above point to different places of the documentation I have checked, but please help me explore. And even better tell me if we can elaborate a way to test on this.

There is an issue with circle ci and the cached conda installation that is addressed in #189

PR #239, opened 2017-05-14
[MRG] Separate rst notebook generation from execution of the script

Titan-C:
This is a refactoring idea from my old PR #160, that was too large to review and accept. I aim to separate the task of every function to make them more testable at some point. The second goal and reason of this PR is to make the generator of the rst notebook(What Sphinx uses to do the html) to work very similar to the jupyter notebook generator. And later at some point have them use the same input

### Executive summary

Refactor for independent rst file construction. Function
``sphinx_gallery.gen_rst.generate_file_rst`` does not anymore compose the
rst file while it is executing each block of the source code. Currently
executing the example script ``execute_script`` is an independent
function and returns structured in a list the rst representation of the
output of each source block. ``generate_file_rst`` calls for execution of
the script when needed, then from the rst output it composes an rst
document which includes the prose, code & output of the example which is
the directly saved to file including the annotations of binder badges,
download buttons and timing statistics.


sphinx_gallery/gen_rst.py
```diff
@@ -244,6 +244,7 @@ def save_figures(image_path, fig_count, gallery_conf):
         current_fig = image_path.format(fig_count + fig_num)
         fig.savefig(current_fig, **kwargs)
         figure_list.append(current_fig)
+    plt.close('all')
```
lesteve: Note to myself: this is moved from execute_code_block.

sphinx_gallery/gen_rst.py
```diff
@@ -260,7 +261,7 @@ def save_figures(image_path, fig_count, gallery_conf):
             figure_list.append(current_fig)
         mlab.close(all=True)
 
-    return figure_rst(figure_list, gallery_conf['src_dir'])
+    return figure_rst(figure_list, gallery_conf['src_dir']), len(figure_list)
```
lesteve: This definitely feels nice to simplify the return type.

sphinx_gallery/gen_rst.py
```diff
@@ -604,58 +600,113 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
...
+
+    save_thumbnail(image_path_template, src_file, file_conf, gallery_conf)
+
+    example_nb = jupyter_notebook(script_blocks)
+    save_notebook(example_nb, example_file.replace('.py', '.ipynb'))
+
+    if block_vars['execute_script']:
+        logger.debug("%s ran in : %.2g seconds\n", src_file, time_elapsed)
+        # Writes md5 checksum if example has build correctly
+        # not failed and was initially meant to run(no-plot shall not cache
```
lesteve: missing space between run and "(no-plot"

sphinx_gallery/gen_rst.py
```diff
@@ -604,58 +600,113 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
...
-        if time_elapsed >= gallery_conf["min_reported_time"]:
-            example_rst += ("**Total running time of the script:**"
-                            " ({0: .0f} minutes {1: .3f} seconds)\n\n".format(
-                                time_m, time_s))
-        example_rst += CODE_DOWNLOAD.format(fname,
-                                            fname.replace('.py', '.ipynb'))
-        example_rst += SPHX_GLR_SIG
-        f.write(example_rst)
+    write_fname : str
+        Filename with full path where to save the rST file
```
lesteve: I think this is either rst (as in the extension) or reST (short name for reSrucuturedText). I'd rather use the extension, less change to be confused with REST. This may happen in other places in this PR, please check.

sphinx_gallery/gen_rst.py
lesteve: missing space between run and (no-plot. Bonus points if you improve the wording of the comment because it is a bit hard to follow (saying many things at once).

sphinx_gallery/gen_rst.py
```diff
@@ -604,58 +600,113 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
...
+        # Writes md5 checksum if example has build correctly
+        # not failed and was initially meant to run(no-plot shall not cache
+        # md5sum)
+        with open(example_file + '.md5', 'w') as file_checksum:
+            file_checksum.write(get_md5sum(example_file))
+
+    return intro, time_elapsed
+
+
+def rst_notebook_cell(script_blocks, output_blocks, file_conf, gallery_conf):
```
lesteve: I think the name is not great. For me cell is similar to block. Here you are generating almost the full example rst unless I am missing something. Also notebook does not seem to be related either. In my mind, notebook only pertains to when you have multiple cells (see `is_example_notebook_like` below).

sphinx_gallery/gen_rst.py
```diff
@@ -604,58 +600,113 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
...
 
-    sys.argv = argv_orig
-    clean_modules()
 
-    # Writes md5 checksum if example has build correctly
-    # not failed and was initially meant to run(no-plot shall not cache md5sum)
-    if block_vars['execute_script']:
-        with open(example_file + '.md5', 'w') as file_checksum:
-            file_checksum.write(get_md5sum(example_file))
+def save_rst_notebook(example_rst, write_file, time_elapsed, gallery_conf):
```
lesteve: Not sure notebook applies here, see my other comment. I'd rather use "example" or something better.

sphinx_gallery/gen_rst.py
```diff
@@ -604,58 +600,113 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
...
-                     mode='w', encoding='utf-8') as f:
-        if time_elapsed >= gallery_conf["min_reported_time"]:
-            example_rst += ("**Total running time of the script:**"
-                            " ({0: .0f} minutes {1: .3f} seconds)\n\n".format(
-                                time_m, time_s))
-        example_rst += CODE_DOWNLOAD.format(fname,
-                                            fname.replace('.py', '.ipynb'))
-        example_rst += SPHX_GLR_SIG
-        f.write(example_rst)
+    write_fname : str
```
lesteve: Maybe output_file. Also docstring and parameter name are not matching.

choldgraf:
I had not noticed this before, but think this is a great effort

choldgraf:
I had not noticed this before, but think this is a great effort

lesteve:
Seems like this needs a rebase.

sphinx_gallery/gen_rst.py
Titan-C: I'm not sure to get the point of this comment.

choldgraf:
feel free to ping when travis is happy!

Titan-C:
@choldgraf, @lesteve  this is ready for a new review. It's just a large refactor of how we generate the rst files into smaller independent functions.

sphinx_gallery/gen_rst.py
lesteve: Just a compliment, maybe you are not used to that in a review ;-)

lesteve:
Reviewing refactors like this when the code moves around is a bit tricky. Is there a way you can help making it more palatable ? Maybe with an executive summary listing the main changes or with inline comments in github around the main changes ?

choldgraf:
agree w/ @lesteve that a list of major changes would be helpful, this is a big-ish PR so we should iterate on it quickly and get it merged before it becomes stale again :-)

Titan-C:
I just wrote the executive summary in the changes file. It is really only making smaller functions of what was `generate_file_rst` by splitting the interleaved composition of the rst document and the execution of the example script into 2 separate processes.
I'm having trouble with travis now, unrelated to the PR something in matplotlib is not working correctly anymore.

lesteve:
Thanks a lot for the executive summary. For the record I was more asking for one in the github PR to help reviewers rather than in CHANGES.rst.

choldgraf:
I took the liberty of editing the top-level comment to paste @Titan-C 's executive summary

re: this PR, what's left to do? Travis is unhappy right now, but other than this is there any code that needs to be added? I've found that refactors/modularizing steps are particularly difficult to pick back up if the PR is left dangling...

Titan-C:
Travis should be happy now that #357 is merged I'll push again to restart the build.

lesteve:
IMO we should release a version, review and (hopefully) merge this PR soon after the release.

larsoner:
Sounds like a good plan to me, will give us time to work out any kinks this might introduce.

larsoner:
Is this good to go? IIUC it's holding up #313, which should provide a clear path for 3 or so open issues

choldgraf:
It sounds like we are in general +1 on merging this as soon as the conflicts are fixed, since we've now had a new release since the latest comment! @Titan-C WDYT?

Titan-C:
This updates are hard after some time passes. Test pass locally. Let's see all CI

Titan-C:
@larsoner I'm not really familiar with the new CircleCI v2. What is this merge check that is failing? Should I just make a rebuild?

larsoner:
It tries to merge with upstream master in a custom command ( not 2.0 specific) and the merge commit seems to have broken that. I suspect that making a trivial commit to master then restarting the build might fix it.

Alternatively you could create a commit tag like `[circle nomerge]` then change the code to respect it using `git log` I can do this if it's not clear how. We do a variance in of this in MNE here for example

https://github.com/mne-tools/mne-python/blob/master/.circleci/config.yml#L75

choldgraf:
I'm +1 on doing what we gotta do to get this merged :-)  boo stale PRs!!

larsoner:
Build docs:

https://256-26191147-gh.circle-artifacts.com/0/rtd_html/index.html

choldgraf:
I'm gonna merge this tomorrow morning unless @Titan-C wants to either merge first, or suggest we wait for some reason!

Titan-C:
I just noticed the runtime of scripts is broken and the binder badge too. Git missed to give me a conflict on those. I'll check that now.

choldgraf:
hmmmm, you mean that when you click the binder link it doesn't direct you to the right file path, right?

Titan-C:
> hmmmm, you mean that when you click the binder link it doesn't direct you to the right file path, right?

Yes, because instead of having autoexamples in the path it has ../

My problem is that I could not reproduce it locally. Neither in my
python 2 nor python 3 environments.

choldgraf:
that's strange...could it be a dependency version difference?

Titan-C:
Ok, I managed to reproduce the problem locally. It is because we don't build the docs in CircleCI within the `doc` folder. I somehow have reintroduced a bug that I already solved in #190

Titan-C:
This should be fixed now. Fix was easy, I wanted to write an extra test.

Titan-C:
Merged. Happy this is finally in!

PR #246, opened 2017-05-25
BUG: Fix reading with non-utf-8 default encoding

dopplershift:
`open()` defaults to text mode, invoking the default system encoding. For
files written in utf-8, this results in decde errors on systems where
the default encoding is something else. This is especially problematic
on windows where the default is cp1252.

While I found this problem trying to build docs on Windows, I can reproduce on my mac with:
```sh
LANG=en_US.ascii make clean html
```

Everything seems to work fine just by opening the file in binary mode so that no implicit encoding is assumed.

### DIFF
--- a/sphinx_gallery/docs_resolv.py
+++ b/sphinx_gallery/docs_resolv.py
@@ -6,6 +6,7 @@ Link resolver objects
 =====================
 """
 from __future__ import print_function
+import codecs
 import gzip
 import os
 import posixpath
@@ -42,14 +43,14 @@ def _get_data(url):
             encoding = resp.headers.get('content-encoding', 'plain')
         data = resp.read()
         if encoding == 'plain':
-            pass
+            data = data.decode('utf-8')
         elif encoding == 'gzip':
             data = StringIO(data)
-            data = gzip.GzipFile(fileobj=data).read()
+            data = gzip.GzipFile(fileobj=data).read().decode('utf-8')
         else:
             raise RuntimeError('unknown encoding')
     else:
-        with open(url, 'r') as fid:
+        with codecs.open(url, mode='r', encoding='utf-8') as fid:
             data = fid.read()
 
     return data
@@ -250,9 +251,6 @@ class SphinxDocLinkResolver(object):
                     else:
                         link = posixpath.join(self.doc_url, fname)
 
-                    if hasattr(link, 'decode'):
-                        link = link.decode('utf-8', 'replace')
-
                     if link in self._page_cache:
                         html = self._page_cache[link]
                     else:
@@ -271,14 +269,8 @@ class SphinxDocLinkResolver(object):
                 for mod in self.extra_modules_test:
                     comb_names.append(mod + '.' + cobj['name'])
             url = False
-            if hasattr(html, 'decode'):
-                # Decode bytes under Python 3
-                html = html.decode('utf-8', 'replace')
 
             for comb_name in comb_names:
-                if hasattr(comb_name, 'decode'):
-                    # Decode bytes under Python 3
-                    comb_name = comb_name.decode('utf-8', 'replace')
                 if comb_name in html:
                     url = link + u'#' + comb_name
             link = url


dopplershift:
So looks like this broke a test that explicitly expects `get_data()` to return text. Would it be better to keep the `open()` in text mode and instead explicitly pass `encoding='utf-8'` to `open()`?

Titan-C:
In this case we are opening a file generated by Sphinx, which nowadays should be utf-8 regardless of python version. Thus we can use the same solution we have put in place for other files and open with:

with codecs.open(url, mode='r', encoding='utf-8') as fid:

That should pass the tests.

Could you please add a line to `appveyor.yml` to test for this bug.

dopplershift:
I've made the codecs change. Not sure what exactly you're looking for me to add to appveyor specifically.

Titan-C:
You have a typo my friend. It's codecs not codes.

Appveyor is our test for windows, and this issue did not show up. I
wanted to be able to make a test where I see this behavior show up,
build the gallery inside a windows encoding that reproduces your bug.

dopplershift:
Oops. 🐑 

I'm well acquainted with AppVeyor, was just confused that you mentioned it specifically rather than adding a test to the suite. Either way, I can say:

1. I'm not clear enough on what the problem is exactly to write a test without more guidance
2. Currently, your test suite passes fine on my mac with `LANG=en_US.ascii`; this is not the case when I build my docs, so it seems like the relevant lines aren't executed by any tests at the moment

Also, it looks like we have new test failures that resemble my original problem...

lesteve:
I haven't looked into it in detail, but it looks like your change broke the tests on Python 2 on all platforms.

pllim:
Ops, I didn't see this before opening #254. I don't think Python 2 accepts encoding.

sphinx_gallery/docs_resolv.py
```diff
@@ -49,7 +50,7 @@ def _get_data(url):
         else:
             raise RuntimeError('unknown encoding')
     else:
-        with open(url, 'r') as fid:
+        with codecs.open(url, mode='r', encoding='utf-8') as fid:
             data = fid.read()
```
Eric89GXL: Can you try `open(url, 'rb') as fid` and `fid.read.decode('utf-8')` instead? Maybe it will fix Travis

dopplershift:
Let's see what happens...

dopplershift:
That would be a no.

Eric89GXL:
Can you try locally? It's possible that the data are used later with a `'w'` that needs to be `'wb'` with `.encode('utf-8')`, but you need to check the sphinx error log to find out.

dopplershift:
I'm having trouble understanding WTH is going on:

```pytb
  File "/Users/rmay/miniconda3/envs/py27/lib/python2.7/site-packages/sphinx/setup_command.py", line 192, in run
    app.build(force_all=self.all_files)
  File "/Users/rmay/miniconda3/envs/py27/lib/python2.7/site-packages/sphinx/application.py", line 357, in build
    self.emit('build-finished', None)
  File "/Users/rmay/miniconda3/envs/py27/lib/python2.7/site-packages/sphinx/application.py", line 489, in emit
    return self.events.emit(event, self, *args)
  File "/Users/rmay/miniconda3/envs/py27/lib/python2.7/site-packages/sphinx/events.py", line 79, in emit
    results.append(callback(*args))
  File "/Users/rmay/repos/sphinx-gallery/sphinx_gallery/docs_resolv.py", line 464, in embed_code_links
    _embed_code_links(app, gallery_conf, gallery_dir)
  File "/Users/rmay/repos/sphinx-gallery/sphinx_gallery/docs_resolv.py", line 398, in _embed_code_links
    full_fname)
  File "/Users/rmay/repos/sphinx-gallery/sphinx_gallery/docs_resolv.py", line 315, in resolve
    link = self._get_link(cobj)
  File "/Users/rmay/repos/sphinx-gallery/sphinx_gallery/docs_resolv.py", line 277, in _get_link
    html = html.decode('utf-8', 'replace')
  File "/Users/rmay/miniconda3/envs/py27/lib/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xb6' in position 3042: ordinal not in range(128)
```

It's decoding as UTF-8, which somehow triggers an ascii problem--is that because on Python 2.7 it's actually an ascii str with a bad character but doesn't know it yet?

Titan-C:
I'm not sure which was the original source of the problem. The build fails now because because one is trying to decode twice. Once from get_data and again in line 277. Certainly the test
```python
if hasattr(html,'decode')
```
is not good to test if one has a bytes type for py3, as that is also a property for unicode and str in python 2

dopplershift:
Last one seemed to do it finally, so I've rebased and cleaned up. I got it working by making sure to `decode('utf-8')` the results coming from `urlopen` and removing all the places checking for a `decode` attribute. Tests pass (I assume they will again) and I can confirm these changes fix my original problem.

Eric89GXL:
LGTM

sphinx_gallery/docs_resolv.py
```diff
@@ -42,15 +42,15 @@ def _get_data(url):
...
         elif encoding == 'gzip':
             data = StringIO(data)
-            data = gzip.GzipFile(fileobj=data).read()
+            data = gzip.GzipFile(fileobj=data).read().decode('utf-8')
         else:
             raise RuntimeError('unknown encoding')
     else:
-        with open(url, 'r') as fid:
-            data = fid.read()
+        with open(url, mode='rb') as fid:
```
lesteve: It would be better to use `codecs.open(..., encoding='utf-8')` to be consistent with what we are doing elsewhere in the code.

dopplershift:
Done. For the record, this is only the second use of `codecs.open()` in the library itself (outside of 4 in tests and 1 in `setup.py`)--raw uses of `open()` are much more common. I agree though that `codecs` is a better solution.

Titan-C:
LGTM. Merging. Thanks @dopplershift

Titan-C:
I just noticed that this breaks the docs
https://readthedocs.org/projects/sphinx-gallery/builds/5622558/
I'm not sure why it happens as travis does build the docs in python 3.5, and there was no problem there.

dopplershift:
I'll check if I can reproduce with Python 3.5 and Sphinx 1.5.3 locally.

dopplershift:
Well that's a no....onto trying readthedocs docker image

dopplershift:
Well then. Can't even reproduce using readthedocs' Docker image. Will keep digging, though thoughts on where are welcome.

dopplershift:
Can somebody with proper permissions try cleaning the build environment on readthedocs.org and try rebuilding? The *only* thing that makes sense at this point is that there's a stale `searchindex.db` file lurking around.

If that *is* the problem, we could try to add some handling around:
```python
data = search_index[url]
```
But I'm not sure just how long you expect `searchindex.dab` to be persisted.

Titan-C:
Thank you for taking the time to look on this. I just reset the docs build after a full wipe. It works on current master.
https://readthedocs.org/projects/sphinx-gallery/builds/5627339/

PR #249, opened 2017-05-29
Escape tooltips for any HTML special characters.

QuLogic:
Fixes #327.


Titan-C:
This looks good to me. Thank you!

PR #250, opened 2017-05-30
Cleanup console output during build

### LINKED ISSUE #230: Console output is very verbose
For every example, sphinx-gallery outputs:
 * One line stating it's going to run the example.
 * One line stating it ran the example and how long it took.
 * One blank line.
 * The example and its runtime in a summary after all the examples are run.

That's at least 4 times as many lines as necessary. Plus there are 4 lines for every directory that's skipped.

This seems a bit too verbose. It would be much nicer if it followed sphinx's example and printed filenames on top of each other. When warnings come up, they get their own _single_ line. The runtime summary is probably not really necessary by default, but once the above things are cut down, it wouldn't be as much of a problem.

QuLogic:
This PR tries to reduce the console output during a build, as well as pass a lot of status message through the Sphinx logging API. This gives the immediate benefit of matching styling with Sphinx easily. I tried to add colours to various items to match Sphinx appearance, but I'm open to suggestions as to the logging level of the messages.

Before:
[asciicast](https://asciinema.org/a/122701)
After:
[asciicast](https://asciinema.org/a/122702)

Since sphinx-gallery's documentation prints a lot, it might not totally clear, so here's a Matplotlib doc build with the patched branch:
[asciicast](https://asciinema.org/a/122706)

Fixes #230.

### DIFF
--- a/sphinx_gallery/docs_resolv.py
+++ b/sphinx_gallery/docs_resolv.py
@@ -30,6 +30,9 @@ from io import StringIO
 from . import sphinx_compatibility
 
 
+logger = sphinx_compatibility.getLogger('sphinx-gallery')
+
+
 def _get_data(url):
     """Helper function to get data over http or from a local file"""
     if url.startswith('http://'):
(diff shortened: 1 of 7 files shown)


lesteve:
This looks quite good. Have you tested this with an old sphinx version e.g. 1.5 or even 1.4?

QuLogic:
I tried with 1.4.8, but I'm not sure how old a sphinx is supported.

Titan-C:
> I tried with 1.4.8, but I'm not sure how old a sphinx is supported.

Sphinx-1.4.8 is on ubuntu16.10. On travis our oldest supported platform is the ubuntu14.04 and there we enforce sphinx-1.4.9(not 1.2.2 as in the distribution) and that is because the huge problems Sphinx-1.5 caused(and story repeats now with Sphinx-1.6), so we test on a version just before that. Up until that point things didn't change so drastically in Sphinx. My point of view would be not to try to support even older version, and only add a new entry into travis a test for Sphinx-1.5 on a ubuntu14 machine which is the oldest we support.

Titan-C:
I will reference to #65, where conclusion was to have an indicator of progress through stdout, and that error messages should be easy to find.
This PR is great at reducing the verbose output we have, I really like it, but looking at the travis output and the video of the output I'm concerned that you loose the information of which example is producing the text output during the build. If someone ever reads the build log that would make it very confusing. On the other hand since #97 we repeat the traceback of the failing examples which is really important and worth reading.

I would really like to keep information of which example gave the output, instead of it being text polluting the logs, without any context
Second, unsure how easy it would be to implement, I'm in favour our reducing output so I would also like to send all this output to the Sphinx-verbose logger(or log) and require the user to be explicit if he wants to see all this output during build.

QuLogic:
I'm not sure it would be possible to keep outputs together without capturing output from the examples. And I'm thinking that perhaps if that were done, the output should go in the HTML (also?).

QuLogic:
Oh, I didn't realize that there's already stdout capture and passed through a tee to the console. I just took out the tee and logged it as verbose; then it can be seen by setting Sphinx logging to verbose.

Titan-C:
Thank you for working on this. Merging

PR #252, opened 2017-06-15
Fix for when sys.argv is used in examples

### LINKED ISSUE #251: Problem with examples using sys.argv e.g. through argparse.ArgumentParser
Add this in `examples/plot_sys_argv.py`:
```py
"""
Title

explanation
"""

import argparse

parser = argparse.ArgumentParser()
parser.parse_args()
```

```
Executing file /home/lesteve/dev/sphinx-gallery/examples/plot_sys_argv.py
usage: sphinx-build [-h]
sphinx-build: error: unrecognized arguments: -b html -d _build/doctrees . _build/html
Makefile:76: recipe for target 'html' failed
make: *** [html] Error 2
```

The error is very confusing, what happens is that `sys.argv` are from the `sphinx-build` command and are passed into the example. I had a quick fix somewhere I will try to find and open a PR.

FYI it happens in scikit-learn where some examples are using `sys.argv`.

lesteve:
Fix #251.

The fix is not the nicest one (maybe changing `sys.argv` inside a `try` clause and changing it back in `except` would make it a bit better) but it works ...

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,13 @@ New features
   `#233 <https://github.com/sphinx-gallery/sphinx-gallery/pull/233>`_ and
   `#234 <https://github.com/sphinx-gallery/sphinx-gallery/pull/234>`_
 
+Bug Fixes
+'''''''''
+
+* Reset ``sys.argv`` before running each example. See
+  `#252 <https://github.com/sphinx-gallery/sphinx-gallery/pull/252>`_
+  for more details.
+
 v0.1.11
 -------
 
--- a/examples/plot_sys_argv.py
+++ b/examples/plot_sys_argv.py
@@ -1,0 +1,14 @@
+"""
+Using ``sys.argv`` in examples
+
+Note that your example will be run by sphinx-gallery without arguments.
+"""
+
+import argparse
+import sys
+
+parser = argparse.ArgumentParser(description='Toy parser')
+parser.add_argument('--option', default='default',
+                    help='a dummy optional argument')
+print('sys.argv:', sys.argv)
+print('parsed args:', parser.parse_args())
--- a/examples/plot_sys_argv.py
+++ b/examples/plot_sys_argv.py
@@ -1,5 +1,6 @@
 """
 Using ``sys.argv`` in examples
+==============================
 
 Note that your example will be run by sphinx-gallery without arguments.
 """
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -568,7 +568,9 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
 
     argv_orig = sys.argv[:]
     if block_vars['execute_script']:
-        print('Executing file %s' % src_file)
+        # We want to run the example without arguments. See
+        # https://github.com/sphinx-gallery/sphinx-gallery/pull/252
+        # for more details.
         sys.argv[0] = src_file
         sys.argv[1:] = []
 
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -565,6 +565,13 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
     time_elapsed = 0
     block_vars = {'execute_script': execute_script, 'fig_count': 0,
                   'image_path': image_path_template, 'src_file': src_file}
+
+    argv_orig = sys.argv[:]
+    if block_vars['execute_script']:
+        print('Executing file %s' % src_file)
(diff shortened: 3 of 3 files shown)


lesteve:
ping @Titan-C @choldgraf.

CHANGES.rst
```diff
@@ -4,6 +4,7 @@ Change Log
 git master
 ----------
 
+<<<<<<< 5eb83a7c61a79656f24e61f5c90c631082265359
```
Titan-C: Left over from rebase?

examples/plot_sys_argv.py
```diff
@@ -0,0 +1,14 @@
+"""
+Using ``sys.argv`` in examples
```
Titan-C: This needs to be an RST heading

sphinx_gallery/gen_rst.py
```diff
@@ -565,6 +565,13 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
     time_elapsed = 0
     block_vars = {'execute_script': execute_script, 'fig_count': 0,
                   'image_path': image_path_template, 'src_file': src_file}
+
+    argv_orig = sys.argv[:]
+    if block_vars['execute_script']:
+        print('Executing file %s' % src_file)
```
Titan-C: We currently use the Sphinx logger not the print statement. The execution of the file is stated in `generate_dir_rst` and not here anymore.

sphinx_gallery/gen_rst.py
```diff
@@ -565,6 +565,13 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
...
+        sys.argv[0] = src_file
```
Titan-C: Is this necessary? `sys.argv[0]` corresponds to the executing binary, most probably it would be `python` as example scripts are not executable on their own (there is no shebang). I wouldn't like to have this hint to the `src_file`, because people might start abusing it. We already don't provide the `__file__` variable because of incompatibility with the Jupyter notebooks.

lesteve: Otherwise `sys.argv[0]` is `sphinx-build` leading to a very confusing error message like the one shown in https://github.com/sphinx-gallery/sphinx-gallery/issues/251#issue-236174916.

CHANGES.rst
lesteve: fixed.

examples/plot_sys_argv.py
lesteve: fixed

sphinx_gallery/gen_rst.py
lesteve: Removed the `print` statement. I think that is what you were aiming at.

Titan-C:
Thanks, merging

lucyleeow:
@lesteve could you explain this a bit more? My understanding is that you cannot use any sys args in your examples ``.py`` files as they are [all set to `[]`](https://github.com/sphinx-gallery/sphinx-gallery/blob/4bcd358d540603cad78a3e69c027006853438262/sphinx_gallery/gen_rst.py#L649). But this doesn't correspond to the [example](https://sphinx-gallery.github.io/auto_examples/plot_sys_argv.html) in the documentation where it seems a sys arg is used?

lesteve:
> My understanding is that you cannot use any sys args in your examples .py files as they are set to [].

You are right, basically what happens is the equivalent of `python your-example.py` (i.e. no arguments are provided to the example script).

> But this doesn't correspond to the example in the documentation where it seems a sys arg is used?

It is consistent with this example, `parsed args: Namespace(option='default')` means this is the default value of the parameters so the same thing as what you get when doing `python your-example.py`. In other words:

```
❯ python examples/plot_sys_argv.py
sys.argv: ['examples/plot_sys_argv.py']
parsed args: Namespace(option='default')
```

lucyleeow:
Ah I see, thank you for explaining. You mentioned:

> FYI it happens in scikit-learn where some examples are using sys.argv.

And I noticed that there are scikit-learn examples that do use `sys.argv` (e.g. [here](https://github.com/scikit-learn/scikit-learn/blob/82845abac64a2de9ef851d9838ab857d74dea8fd/doc/tutorial/text_analytics/skeletons/exercise_01_language_train_model.py#L23)). Is this because they are using different configurations or...?

lesteve:
> Is this because they are using different configurations or...?

I think that the example you mention is not run by sphinx-gallery. AFAIK scikit-learn sphinx-gallery config only use the `examples` folder and your file somewhere in `doc/tutorial`.

lucyleeow:
Ah okay - that makes sense - thanks for clarifying.

PR #254, opened 2017-06-20
Enforce UTF8 encoding because that is not default on Windows

### LINKED ISSUE #253: UnicodeDecodeError on Windows when building doc
Windows 7, Anaconda, Python 3.5.3, Astropy 2.0.dev19112, Sphinx 1.5.6

untitled

pllim:
Fix #253 

Also see astropy/sphinx-automodapi#29

UPDATE: Close #246

### DIFF
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -11,7 +11,8 @@ Parses example file code in order to keep track of used functions
 from __future__ import print_function
 import ast
 import os
-
+import platform
+import sys
 
 # Try Python 2 first, otherwise load from Python 3
 try:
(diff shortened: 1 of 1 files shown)


lesteve:
I think #246 had the right approach of always trying to use utf-8. The only thing is that I never had the time to understand why the tests were failing ...

pllim:
If situation is dire enough, maybe this can be a "quick fix" until the other one has a chance to be investigated. This patch works for my use case and I have no motivation to investigate further.

sphinx_gallery/backreferences.py
```diff
@@ -130,7 +131,15 @@ def identify_names(code):
...
-    example_code_obj = identify_names(open(example_file).read())
+
+    # https://github.com/sphinx-gallery/sphinx-gallery/issues/253
+    if (platform.system().lower().startswith('win') and
+            sys.version_info[0] >= 3):
+        open_kwargs = {'encoding': 'utf8'}
+    else:
+        open_kwargs = {}
+
+    example_code_obj = identify_names(open(example_file, **open_kwargs).read())
```
lesteve: This is not going to work on Python 2.

Eric89GXL: I'm running Python2 and it seems to work

Eric89GXL: ... oh wait, I thought this was on my PR, never mind!

pllim: @lesteve , what do you mean? The tests passed.

Titan-C:
I get the feeling this PR and the corresponding issue #253 are not related to the problem being addressed in PR #246

lesteve:
> I get the feeling this PR and the corresponding issue #253 are not related to the problem being addressed in PR #246

They are related in the sense that we should be using `codecs.open(..., encoding='utf-8')` instead of `open` everywhere in our source code otherwise the default platform encoding could be used, which is generally not what we want.

Titan-C:
Indeed. I wasn't specific enough either. This PR is targeting other place in the codebase which is not what PR #246 is covering. We still need to fix it here too.

PR #255, opened 2017-06-20
Fix status_iterator deprecation warning

pllim:
Fix this warning with Sphinx 1.6:
```
.../sphinx/application.py:445: RemovedInSphinx17Warning:
    app.status_iterator() is now deprecated. Use sphinx.util.status_iterator() instead.
  RemovedInSphinx17Warning)
```

c/c @bsipocz

### DIFF
--- a/sphinx_gallery/docs_resolv.py
+++ b/sphinx_gallery/docs_resolv.py
@@ -15,6 +15,11 @@ import sys
 
 from sphinx.util.console import fuchsia
 
+try:
+    from sphinx.util import status_iterator
+except ImportError:
+    status_iterator = None
+
 # Try Python 2 first, otherwise load from Python 3
 try:
     import cPickle as pickle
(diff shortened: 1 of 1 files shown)


sphinx_gallery/docs_resolv.py
```diff
@@ -370,9 +375,14 @@ def _embed_code_links(app, gallery_conf, gallery_dir):
     flat = [[dirpath, filename]
             for dirpath, _, filenames in os.walk(html_gallery_dir)
             for filename in filenames]
-    iterator = app.status_iterator(
-        flat, os.path.basename(html_gallery_dir), colorfunc=fuchsia,
-        length=len(flat), stringify_func=lambda x: os.path.basename(x[1]))
+    if status_iterator is None:
```
lesteve: Thanks for your PR. For slightly better readability, I would do something like:

```py
try:
    from sphinx.util import status_iterator
except ImportError:
    # sphinx < 1.6
    status_iterator = app.status_iterator

iterator = status_iterator(...)
```

Titan-C:
I actually prefer the solution from PR #250

pllim:
Ah, I was unaware of the fix being implemented already on #250.

PR #257, opened 2017-06-28
Enforce UTF8 encoding in backreferences

pllim:
Fix #253
Another attempt of #254

Also see astropy/sphinx-automodapi#29

PR #225, opened 2017-04-16
[MRG] Testing configuration

Titan-C:
Taking in to advantage the movement to pytest in PR #217, I'm looking at how Sphinx tests itself using pytest and plan to produce some test builds to check sphinx-gallery builds the gallery with the correct configuration.

This PR depends on PR #217 and shall not be merged before that.

PR #259, opened 2017-06-29
Revert "Enforce UTF8 encoding in backreferences"

Titan-C:
Reverts sphinx-gallery/sphinx-gallery#257.
In python 2 the codeblocks are not being properly recognized to embed the documentation links.

### DIFF
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -11,7 +11,7 @@ Parses example file code in order to keep track of used functions
 from __future__ import print_function
 import ast
 import os
-import codecs
+
 
 # Try Python 2 first, otherwise load from Python 3
 try:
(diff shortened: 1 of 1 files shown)


lesteve:
Can we have a test for this so that we don't break it again?

Titan-C:
> Can we have a test for this so that we don't break it again?

One day in the future. When I can run pytest over the built galleries. PR #225 is a step towards that goal where I start running Sphinx within pytest over some test templates to check for specific output

lesteve:
> One day in the future. When I can run pytest over the built galleries. PR #225 is a step towards that goal where I start running Sphinx within pytest over some test templates to check for specific output

I was hoping that this is possible to test this in isolation (rather than generating a full gallery, if I have understood the motivation behind #225 correctly) but maybe I am too naive ...

I am guessing this is possible to fix `identify_names` so that it can handle correctly unicode in Python 2.

PR #263, opened 2017-06-30
Save and use original line numbers where possible

### LINKED ISSUE #248: Confusing line numbers in traceback
When building e.g. the s-g docs, one gets the traceback
```
/home/antony/src/extern/sphinx-gallery/examples/plot_raise.py failed leaving traceback:
  File "/usr/lib/python3.6/site-packages/sphinx_gallery/gen_rst.py", line 472, in execute_code_block
    exec(code_block, example_globals)
  File "<string>", line 2, in <module>
NameError: name 'iae' is not defined
```
but the error actually occurs at line 23 of the .py file (I realize this is probably the line number *within the cell*, but this can make troubleshooting a bit trickier).

QuLogic:
This PR saves line numbers from the original source file, then uses that information with `ast` to ensure that tracebacks use the correct lines and filenames, which fixes #248.

Additionally, it adds an option `line_numbers` that can be enabled globally or per-file as a `sphinx_gallery_line_numbers` comment which allows for numbering of source code lines. As an example, this is enabled locally on the `plot_raise.py` file.

Some other refactorings are included to make these changes easier.


QuLogic:
Rebased to fix conflicts.

sphinx_gallery/gen_rst.py
```diff
@@ -135,36 +138,25 @@ def write(self, data):
...
-        # by default, use the first figure created
-        thumbnail_number = 1
-    else:
-        thumbnail_number = int(match.groups()[0])
-
-    return thumbnail_number
-
-
-def extract_intro(filename):
+def extract_intro(filename, docstring):
```
Titan-C: I feel this is a good opportunity to change the logic of how Sphinx-Gallery works. This function fails after the example has been built if the docstring is not compliant, and crashes the entire gallery build process.
Then one has to change the file to update the docstring and rebuild the file.
I would like to put the check of a compliant docstring in the `get_docstring_and_rest` function and raise the exception there.

sphinx_gallery/gen_rst.py
```diff
@@ -521,13 +541,13 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
...
-    script_blocks = split_code_and_text_blocks(src_file)
+    file_conf, script_blocks = split_code_and_text_blocks(src_file)
     amount_of_code = sum([len(bcontent)
-                          for blabel, bcontent in script_blocks
+                          for blabel, bcontent, lineno in script_blocks
                           if blabel == 'code'])
 
     if md5sum_is_current(example_file):
-        return amount_of_code, 0
+        return get_docstring_and_rest(example_file)[0], amount_of_code, 0
```
Titan-C: return script_blocks[0][1], like in the end of the function.
Even better we can return the introduction instead of the docstring, here and at the end. Then we can use the introduction directly in `generate_dir_rst`

sphinx_gallery/gen_rst.py
```diff
@@ -438,46 +464,38 @@ def execute_code_block(src_file, code_block, example_globals,
...
         t_start = time()
         # don't use unicode_literals at the top of this file or you get
         # nasty errors here on Py2.7
-        exec(code_block, example_globals)
+        exec(compile(code_ast, src_file, 'exec'), example_globals)
         time_elapsed = time() - t_start
 
+    except Exception:
+        sys.stdout = orig_stdout
+        images_rst = handle_exception(sys.exc_info(), src_file, block_vars,
```
Titan-C: This is a wrong logic/naming of sphinx-gallery. I've been trying to get rid of this in PR #239, the traceback is not an image_rst, but a formatted output. Since later on we format stdout for the successful output, I believe `handle_exceptions` should save to that variable too to keep consistency.

sphinx_gallery/py_source_parser.py
```diff
@@ -68,32 +69,55 @@ def split_code_and_text_blocks(source_file):
...
+        name = match.group(1)
+        value = match.group(2)
+        try:
+            value = ast.literal_eval(value)
+        except (SyntaxError, ValueError):
+            logger.warning(
+                'Sphinx-gallery option %s was passed invalid value %s',
+                name, value)
+        else:
+            file_conf[name] = value
```
Titan-C: We can also extend the test that checks for the thumbnail number with the line number config

QuLogic: I'm not sure there's much point to testing the line number config as it's pretty much the same to parse as thumbnail number.

QuLogic: Fixed.

sphinx_gallery/gen_rst.py
QuLogic: Renamed.

sphinx_gallery/gen_rst.py
QuLogic: Done.

sphinx_gallery/gen_rst.py
QuLogic: I don't quite understand what difference this would make.

QuLogic:
Rebased due to conflicts.

Titan-C:
Thanks, merging

PR #264, opened 2017-06-30
Correctly re-raise errors in doc resolver.

### LINKED ISSUE #213: No active exception to raise
Due to docs.scipy.org being down/slow (I think) I'm seeing this message:
```pytb
Exception occurred:
  File "/Users/rmay/miniconda3/envs/py36/lib/python3.6/site-packages/sphinx_gallery/docs_resolv.py", line 264, in _get_link
    raise
RuntimeError: No active exception to reraise
```

That traces back to this block:
```python
            for fname in fnames:
                try:
                    if self._is_windows:
                        fname = fname.replace('/', '\\')
                        link = os.path.join(self.doc_url, fname)
                    else:
                        link = posixpath.join(self.doc_url, fname)

                    if hasattr(link, 'decode'):
                        link = link.decode('utf-8', 'replace')

                    if link in self._page_cache:
                        html = self._page_cache[link]
                    else:
                        html = get_data(link, self.gallery_dir)
                        self._page_cache[link] = html
                except (HTTPError, URLError, IOError):
                    pass
                else:
                    break
            else:
                raise
```
`raise` is being called from the `else` clause from the `for` loop, so clearly there's no exception to be raised at that point. I have no idea what the code should be doing, but (as the interpreter indicates) there's no way that code can operate correctly.

QuLogic:
Fixes #213.

### DIFF
--- a/sphinx_gallery/docs_resolv.py
+++ b/sphinx_gallery/docs_resolv.py
@@ -246,7 +246,7 @@ class SphinxDocLinkResolver(object):
             # built with < 1.5 or >= 1.5 regardless of what we're currently
             # building with, so let's just check both :(
             fnames = [fname + '.html', os.path.splitext(fname)[0] + '.html']
-            for fname in fnames:
+            for i, fname in enumerate(fnames):
                 try:
                     if self._is_windows:
                         fname = fname.replace('/', '\\')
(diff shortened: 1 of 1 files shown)


lesteve:
LGTM, merging, thanks a lot!

PR #265, opened 2017-07-09
Fix py37 changes

tacaswell:
python/cpython#46 or bpo-29463 added docstring as a property to the module class that comes out of AST (so the free strings no longer appear an the first element in the parsed files).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,7 @@ matrix:
     - os: linux
       env: DISTRIB="conda" PYTHON_VERSION="3.6"
     - os: linux
+      dist: trusty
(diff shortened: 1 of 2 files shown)


tacaswell:
Will collide with #263

QuLogic:
#263 is merged now.

tacaswell:
The 2.7 failure is download related, not sure why nightly is not working....

Eric89GXL:
LGTM

choldgraf:
this looks good to me as well - thanks @tacaswell for helping put out this tiny-ish fire before it became a big-ish fire :-)

PR #244, opened 2017-05-22
auto-generate binder badges

choldgraf:
This is a prototype for auto-adding binder badges to sphinx-gallery examples. It utilizes a super-beta Binder API that is definitely not public yet, but this is a proof-of-concept. You can see a live version at the sphinx-gallery here:

http://predictablynoisy.com/sphinx-gallery/auto_examples/index.html

pretty cool!

Lemme know people's thoughts, comments, etc. I don't expect this to go live for a little while anyway because we need to finish the beta version of the new binder backend first.

Some ideas / questions:

* Right now the Binder badge is put at the end above the buttons...that work?
* The badge is linking from the `mybinder.org` svg. Do we want that image?
* It's linking to the ipynb file in `_downloads`...so not much opportunity to move between ipynb files within the same binder instance. I think that's fine for this PR but maybe in the future there's a more elegant solution
* All of the relevant configuation options are done with `binder_<option>`. Alternatively we could have a dictionary, e.g. `{binder_conf: {'org': 'choldgraf', 'repo': 'sphinx-gallery', 'reqs': '../requirements.txt'}}`

PR #268, opened 2017-07-15
Push the docs from Circle CI into github

Titan-C:
Read the docs has built our docs for a long time, and it has some limitations. For example we can not build the examples gallery for mayavi. And with PR #244, we need to have the jupyter notebooks on a github repository to use binder.

This PR builds the documentation on CircleCI and if succesful it pushed the results to github pages.
Now the build address is 
https://sphinx-gallery.github.io
For the moment it will only build master branch. 
I image we can slowly move towards this website and drop readthedocs.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -36,6 +36,5 @@ deployment:
   staging:
     branch:
       - master
-      - push_doc
     commands:
         - ./docs_deploy.sh rtd_html
--- a/circle.yml
+++ b/circle.yml
@@ -38,5 +38,4 @@ deployment:
       - master
       - push_doc
     commands:
-        - touch rtd_html/.nojekyll
         - ./docs_deploy.sh rtd_html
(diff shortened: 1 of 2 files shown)


GaelVaroquaux:
+1 for moving to github-hosted docs.

choldgraf:
Another option would be to use a tool such as `doctr`:

https://github.com/drdoctr/doctr

You can run sphinx as a part of the build process, and then configure doctr to push to `gh-pages` if the build completes. Sounds like this works as well too, just throwing it out there :-)

Titan-C:
I had no idea of `doctr`.  Nevertheless there are some advantages in doing the push from Circle CI, namely that I can see the build.

Titan-C:
I'll merge this, to also launch a build with the latest release of the gallery.

choldgraf:
+1 - there's definitely a benefit to using circleCI to build the docs

choldgraf:
(and doctr is new but useful! worth keeping an eye on it in the future.)

PR #272, opened 2017-07-29
Back-reference optimizations

QuLogic:
Finding back-references involves downloading a file for every possibly used object in order to do two things:
 1. Check for the correct extension
 2. Check for the correct anchor

All this downloading makes back-referencing slow, so instead:
 1. Download the index page, parse out the `DOCUMENTATION_OPTIONS` and find the `FILE_SUFFIX` from there. If it's not defined, it's a _really_ old Sphinx so assume `.rst.html`.
 2. Use the same method of finding the anchor as the JavaScript search code.

With these two changes, we can stop downloading all those files and just download 2 per back-reference URL. Additionally, use Sphinx's built-in parser to cut down on the parsing code.


sphinx_gallery/docs_resolv.py
```diff
@@ -216,66 +163,46 @@ def __init__(self, doc_url, gallery_dir, searchindex='searchindex.js',
...
 
     def _get_link(self, cobj):
         """Get a valid link, False if not found"""
 
-        fname_idx = None
-        full_name = cobj['module_short'] + '.' + cobj['name']
-        if full_name in self._searchindex['objects']:
-            value = self._searchindex['objects'][full_name]
-            if isinstance(value, dict):
-                value = value[next(iter(value.keys()))]
```
QuLogic: Note that this is completely wrong. For example, in `plot_gallery_version`, `np.random.random` is linked to [`np.random.lognormal`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.lognormal.html#numpy.random). I believe this is also in part due to #273.

Titan-C: I just tested this PR. but in this case it does not even recognize `np.random.random` I'll see with #273

sphinx_gallery/docs_resolv.py
```diff
@@ -75,102 +78,51 @@ def get_data(url, gallery_dir):
...
+        if value[0] in '"\'':
+            value = value[1:-1]
+        elif value == 'false':
+            value = False
+        elif value == 'true':
+            value = True
+        else:
+            value = int(value)
+
+        docopts[key] = value
```
Titan-C: Can you provide a test for this function. Or at least move this parse to an independent function an test that part at least.

choldgraf: +1 for tests

choldgraf:
this sounds like a great addition, thanks @QuLogic ... I don't have a ton of experience w/ intersphinx code but I agree that it'd be good to test any new parsing functionality built in here.

sphinx_gallery/docs_resolv.py
```diff
@@ -77,6 +78,53 @@ def get_data(url, gallery_dir):
...
```
QuLogic: This parsing is a separate function?

Anyway, test added.

sphinx_gallery/docs_resolv.py
```diff
@@ -30,6 +30,9 @@
 
 from io import StringIO
 
+from sphinx.search import js_index
+from sphinx.util import jsdump
```
QuLogic: Removed.

sphinx_gallery/docs_resolv.py
QuLogic: Definitely should be because of #273.

choldgraf:
just a ping that #273 is merged now!

Titan-C:
This works well on my machine. I relaunched Circle CI, but it did not make the build on top of master to include #273, so I could not check on the other versions.
+1 after a rebase, just to check.

Eric89GXL:
On my OSX machine, on current `master` it gets stuck on the embedding step. With this PR things at least don't hang, so +1 for merge from my end!

QuLogic:
Rebased and `np.random.random` is now pointing to the correct function [here](https://159-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/doc/_build/html/auto_examples/plot_gallery_version.html#sphx-glr-auto-examples-plot-gallery-version-py).

Titan-C:
Merged. Thanks @QuLogic

PR #273, opened 2017-07-29
Ensure shortened module path points to expected object.

QuLogic:
For example, `matplotlib.pyplot.figure` can be shortened to `matplotlib.figure`, but the former is a function (that is used a lot) while the latter is a module (which is generally used much less).

### DIFF
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -76,12 +76,23 @@ class NameFinder(ast.NodeVisitor):
 
 def get_short_module_name(module_name, obj_name):
     """ Get the shortest possible module name """
+    scope = {}
+    try:
+        # Find out what the real object is supposed to be.
+        exec('from %s import %s' % (module_name, obj_name), scope, scope)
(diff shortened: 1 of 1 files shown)


QuLogic:
This might be a bit overcautious if something like a top-level shortcut like `from a import foo` exists when the actual object is `a.b.c.d.foo` and `a.b.foo` exists as a different object (or similar.)

sphinx_gallery/backreferences.py
```diff
@@ -76,12 +76,21 @@ def get_mapping(self):
 
 def get_short_module_name(module_name, obj_name):
     """ Get the shortest possible module name """
+    scope = {}
+    try:
+        exec('from %s import %s' % (module_name, obj_name), scope, scope)
```
choldgraf: could you add a comment to highlight what/why this is being called here?

QuLogic: Added.

Titan-C:
This works well enough for me. I'm merging this.

PR #278, opened 2017-07-31
[WIP] Expose examples sortkey

Titan-C:
This is my proposal for #275 
Here I directly expose a sortkey for ordering the examples within each subsection. 
Default is for the moment case sensitive alphabetic ordering, which uses the `None` keyword. Functions for amount of code or file size are also provided.
Needs the documentation. Feel free to take over this PR.

PR #279, opened 2017-07-31
MRG: Update docs

Eric89GXL:
Closes #276.
Closes #277.

~~WIP because building freezes on my machine at `Embedding documentation hyperlinks in examples ...` (40%) so I want to see what CircleCI gives, and I still want/need to address #277.~~

PR #281, opened 2017-08-02
MRG: Allow sorting subsection files

### LINKED ISSUE #275: ENH: Allow setting example order
I recently migrated PySurfer over to SG, and want to be able to set the example file order. There currently is no way to do that, correct? The workarounds would be to either build subdirectories, or re-title the examples e.g. "01. Basic visualization" etc. to get them in the correct order, which is not so satisfactory.

I propose adding an `example_order` dict to `conf.py` to complement the existing `subsection_order`. The design would be nested dicts:
```
'example_order': {'examples': ['plot_basic.py', 'plot_advanced.py']}
```
So the first-level dict would be key-value pair for `subsection: order`. If a subdirectory isn't listed, use default (title-alphabetical) sorting. If one is listed, it must list all SG-relevant files in the directory or an error is thrown. This seems the most conservative API-wise and should be fully backward compatible.

Eric89GXL:
Closes #275.
Closes #278. (Take over of this PR.)

This adds sort keys for sorting within subsections. I don't need explicit reordering currently (filename-based is good enough for now) so I didn't bother implementing it for now. But it at least sets up the framework for it later.

~~WIP for now because I'm not updating the doc until #279 is merged and I can rebase to avoid conflicts.~~

### DIFF
--- a/doc/advanced_configuration.rst
+++ b/doc/advanced_configuration.rst
@@ -18,6 +18,7 @@ file:
 - ``examples_dirs`` and ``gallery_dirs`` (:ref:`multiple_galleries_config`)
 - ``filename_pattern`` (:ref:`build_pattern`)
 - ``subsection_order`` (:ref:`sub_gallery_order`)
+- ``within_subsection_order`` (:ref:`within_gallery_order`)
 - ``reference_url`` (:ref:`link_to_documentation`)
 - ``backreferences_dir`` and ``doc_module`` (:ref:`references_to_examples`)
 - ``default_thumb_file`` (:ref:`custom_default_thumb`)
@@ -140,6 +141,29 @@ the converse does not hold.
 
 If you so desire you can implement your own sorting key. It will be
 provided the relative paths to `conf.py` of each sub gallery folder.
+
+
+.. _within_gallery_order:
+
+Sorting gallery examples
+========================
+
+Within a given gallery (sub)section, the example files by default are ordered
+by the amount of code, i.e. :class:`sphinx_gallery.sorting.AmountOfCodeSortKey`
+as::
+
+    from sphinx_gallery.sorting import AmountOfCodeSortKey
+    sphinx_gallery_conf = {
+        ...
+        'within_subsection_order': AmountOfCodeSortKey,
+    }
+
+However, other options exist that can be instantiated for use with
+``within_subsection_order``:
+
+- :class:`sphinx_gallery.sorting.FileSizeSortKey` to sort by file size.
+- :class:`sphinx_gallery.sorting.FileNameSortKey` to sort by file name.
+- :class:`sphinx_gallery.sorting.ExampleTitleSortKey` to sort by example title.
 
 
 .. _link_to_documentation:
--- a/doc/advanced_configuration.rst
+++ b/doc/advanced_configuration.rst
@@ -148,19 +148,23 @@ provided the relative paths to `conf.py` of each sub gallery folder.
 Sorting gallery examples
 ========================
 
-Within a given gallery (sub)section, the example files by default are ordered
-by the amount of code, i.e. :class:`sphinx_gallery.sorting.AmountOfCodeSortKey`
-as::
+Within a given gallery (sub)section, the example files are ordered by
+using the standard :func:`sorted` function with the ``key`` argument by default
+set to
+:class:`NumberOfCodeLinesSortKey(src_dir) <sphinx_gallery.sorting.NumberOfCodeLinesSortKey>`,
+which sorts the files based on the number of code lines::
 
-    from sphinx_gallery.sorting import AmountOfCodeSortKey
+    from sphinx_gallery.sorting import NumberOfCodeLinesSortKey
     sphinx_gallery_conf = {
         ...
-        'within_subsection_order': AmountOfCodeSortKey,
+        'within_subsection_order': NumberOfCodeLinesSortKey,
     }
 
-However, other options exist that can be instantiated for use with
+However, multiple convenience classes are provided for use with
 ``within_subsection_order``:
 
+- :class:`sphinx_gallery.sorting.NumberOfCodeLinesSortKey` (default) to sort by
+  the number of code lines.
 - :class:`sphinx_gallery.sorting.FileSizeSortKey` to sort by file size.
 - :class:`sphinx_gallery.sorting.FileNameSortKey` to sort by file name.
 - :class:`sphinx_gallery.sorting.ExampleTitleSortKey` to sort by example title.
--- a/doc/advanced_configuration.rst
+++ b/doc/advanced_configuration.rst
@@ -160,7 +160,7 @@ which sorts the files based on the number of code lines::
         'within_subsection_order': NumberOfCodeLinesSortKey,
     }
 
-However, multiple convenience classes are provided for use with
+In addition, multiple convenience classes are provided for use with
 ``within_subsection_order``:
 
 - :class:`sphinx_gallery.sorting.NumberOfCodeLinesSortKey` (default) to sort by
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -375,8 +375,10 @@ def generate_dir_rst(src_dir, target_dir, gallery_conf, seen_backrefs):
 
     if not os.path.exists(target_dir):
         os.makedirs(target_dir)
-    sorted_listdir = [fname for fname in sorted(os.listdir(src_dir))
-                      if fname.endswith('.py')]
+    listdir = [fname for fname in os.listdir(src_dir)
+               if fname.endswith('.py')]
+    sorted_listdir = sorted(
+        listdir, key=gallery_conf['within_subsection_order'](src_dir))
     entries_text = []
     computation_times = []
     build_target_dir = os.path.relpath(target_dir, gallery_conf['src_dir'])
@@ -385,29 +387,25 @@ def generate_dir_rst(src_dir, target_dir, gallery_conf, seen_backrefs):
         'Generating gallery for %s ' % build_target_dir,
         length=len(sorted_listdir))
     for fname in iterator:
-        intro, amount_of_code, time_elapsed = generate_file_rst(
+        intro, time_elapsed = generate_file_rst(
             fname,
             target_dir,
             src_dir,
             gallery_conf)
         computation_times.append((time_elapsed, fname))
-        new_fname = os.path.join(src_dir, fname)
         this_entry = _thumbnail_div(build_target_dir, fname, intro) + """
 
 .. toctree::
    :hidden:
 
    /%s\n""" % os.path.join(build_target_dir, fname[:-3]).replace(os.sep, '/')
-        entries_text.append((amount_of_code, this_entry))
+        entries_text.append(this_entry)
 
         if gallery_conf['backreferences_dir']:
             write_backreferences(seen_backrefs, gallery_conf,
                                  target_dir, fname, intro)
 
-    # sort to have the smallest entries in the beginning
-    entries_text.sort()
-
-    for _, entry_text in entries_text:
+    for entry_text in entries_text:
         fhindex += entry_text
 
     # clear at the end of the section
@@ -531,8 +529,6 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
     -------
     intro: str
         The introduction of the example
-    amount_of_code : int
-        character count of the corresponding python script in file
     time_elapsed : float
         seconds required to run the script
     """
@@ -541,13 +537,10 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
     example_file = os.path.join(target_dir, fname)
     shutil.copyfile(src_file, example_file)
     file_conf, script_blocks = split_code_and_text_blocks(src_file)
-    amount_of_code = sum([len(bcontent)
-                          for blabel, bcontent, lineno in script_blocks
-                          if blabel == 'code'])
     intro = extract_intro(fname, script_blocks[0][1])
 
     if md5sum_is_current(example_file):
-        return intro, amount_of_code, 0
+        return intro, 0
 
     image_dir = os.path.join(target_dir, 'images')
     if not os.path.exists(image_dir):
@@ -646,4 +639,4 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
     if block_vars['execute_script']:
         logger.debug("%s ran in : %.2g seconds", src_file, time_elapsed)
 
-    return intro, amount_of_code, time_elapsed
+    return intro, time_elapsed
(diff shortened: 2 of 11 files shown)


Eric89GXL:
Rebased, [docs added](https://169-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/rtd_html/advanced_configuration.html#within-gallery-order), and CIs are happy. Ready for review/merge from my end.

doc/advanced_configuration.rst
```diff
@@ -142,6 +143,29 @@ If you so desire you can implement your own sorting key. It will be
...
+.. _within_gallery_order:
+
+Sorting gallery examples
+========================
+
+Within a given gallery (sub)section, the example files by default are ordered
+by the amount of code, i.e. :class:`sphinx_gallery.sorting.AmountOfCodeSortKey`
+as::
+
+    from sphinx_gallery.sorting import AmountOfCodeSortKey
```
choldgraf: Why not `NumberOfLines` sortkey? I know that's kind of a bikeshed but I think it's more explicit

doc/advanced_configuration.rst
```diff
@@ -142,6 +143,29 @@ If you so desire you can implement your own sorting key. It will be
 provided the relative paths to `conf.py` of each sub gallery folder.
 
 
+.. _within_gallery_order:
+
+Sorting gallery examples
+========================
+
+Within a given gallery (sub)section, the example files by default are ordered
```
choldgraf: Do we have any short description in the docs that just explains that ordering in general is handled by "sorting keys" which are a way of defining a custom sorting behavior with the `sorted` function? Maybe that's too much info but I was confused the first time I saw sort keys being used.

doc/advanced_configuration.rst
```diff
@@ -142,6 +143,29 @@ If you so desire you can implement your own sorting key. It will be
...
+    sphinx_gallery_conf = {
+        ...
+        'within_subsection_order': AmountOfCodeSortKey,
+    }
+
+However, other options exist that can be instantiated for use with
+``within_subsection_order``:
+
+- :class:`sphinx_gallery.sorting.FileSizeSortKey` to sort by file size.
+- :class:`sphinx_gallery.sorting.FileNameSortKey` to sort by file name.
```
choldgraf: just a thought on `ExplicitFilenameSortKey` (which I know you aren't implementing here): I think it would get really confusing if we implemented this, since right now doing explicit sorting for folders *requres* that all folders are listed. We couldn't do the same thing for explicit filenames because there would be too many IMO

sphinx_gallery/gen_rst.py
```diff
@@ -375,8 +375,10 @@ def generate_dir_rst(src_dir, target_dir, gallery_conf, seen_backrefs):
 
     if not os.path.exists(target_dir):
         os.makedirs(target_dir)
-    sorted_listdir = [fname for fname in sorted(os.listdir(src_dir))
-                      if fname.endswith('.py')]
+    listdir = [fname for fname in os.listdir(src_dir)
+               if fname.endswith('.py')]
+    sorted_listdir = sorted(
+        listdir, key=gallery_conf['within_subsection_order'](src_dir))
```
choldgraf: I think we should have a check somewhere that sees if one of the allowed keys are given, and if not (e.g., if somebody use some other wonky object) we'd throw an error that says "this must be one of XXX sortkeys"

sphinx_gallery/gen_rst.py
```diff
@@ -385,29 +387,25 @@ def generate_dir_rst(src_dir, target_dir, gallery_conf, seen_backrefs):
         'Generating gallery for %s ' % build_target_dir,
         length=len(sorted_listdir))
     for fname in iterator:
-        intro, amount_of_code, time_elapsed = generate_file_rst(
+        intro, time_elapsed = generate_file_rst(
```
choldgraf: I like that this is being factored out

sphinx_gallery/sorting.py
```diff
@@ -46,3 +48,71 @@ def __call__(self, item):
...
+
+    Parameters
+    ----------
+    src_dir : str
+        The source directory.
+    """
+
+    def __call__(self, filename):
+        src_file = os.path.normpath(os.path.join(self.src_dir, filename))
+        file_conf, script_blocks = split_code_and_text_blocks(src_file)
```
choldgraf: ah I see that it isn't strictly `NumberOfLines` per-se, since you're only counting the code lines. I think it's fine to keep it `AmountOfCode` in that case, but maybe clarify in the docs that this will count only code lines and not all the lines in the `.py` files

sphinx_gallery/sorting.py
```diff
@@ -46,3 +48,71 @@ def __call__(self, item):
...
+class ExampleTitleSortKey(_SortKey):
+    """Sort examples in src_dir by example title.
+
+    Parameters
+    ----------
+    src_dir : str
+        The source directory.
+    """
+
+    def __call__(self, filename):
```
choldgraf: I think this will break for many people's sphinx-gallery setups, as there are a few projects (ahem, MNE-python being one of them) that put sphinx tags just above their titles. I feel like I opened an issue about this a while back but the dev team wasn't keen on supporting that use-case...

sphinx_gallery/tests/test_gen_gallery.py
```diff
@@ -162,3 +164,69 @@ def test_config_backreferences(config_app):
...
+    'examples_dirs': 'src',
+    'gallery_dirs': 'ex',
+    'within_subsection_order': FileSizeSortKey,
+}""")
+def test_example_sorting_filesize(config_app):
+    """Test sorting of examples by filesize."""
+    _check_order(config_app, 'filesize')
+
+
+@pytest.mark.conf_file(content="""
```
choldgraf: These tests feel like magic to me

doc/advanced_configuration.rst
Eric89GXL: It's not just number of lines but more number of code characters IIUC

sphinx_gallery/gen_rst.py
Eric89GXL: But people in principle can use their own

sphinx_gallery/sorting.py
Eric89GXL: Ah yes I should probably parse the title however it's found and parsed elsewhere (and refactor to DRY)

sphinx_gallery/tests/test_gen_gallery.py
Eric89GXL: The conf stuff? Agreed but I copy pasted (then mostly simplified) from earlier code so I assumed thus was best practice for the repo

doc/advanced_configuration.rst
Eric89GXL: Sorry, number of code lines rather (which you note below)

doc/advanced_configuration.rst
Eric89GXL: Yeah we'll have to think harder about that one, but I don't think that the current API will make the problem harder necessarily

sphinx_gallery/sorting.py
Eric89GXL: Actually this already uses `split_code_and_text_blocks` which calls `get_docstring_and_rest`, and `script_blocks[0]` is the `docstring`, which is already cleaned. I even added this to the code:
```
        assert script_blocks[0][1].strip().split('\n')[0][0] == '='
```
And it passed, so we should be okay here.

Eric89GXL:
Okay @choldgraf I changed the name to `NumberOfCodeLinesSortKey`, added a brief bit to get people comfortable with `sorted` / `key`, and added some junk comment lines at the top of one the `plot_1.py` test file to ensure the title-ordering doesn't care about code comments. Updated rendered doc is here:

https://170-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/rtd_html/advanced_configuration.html#sorting-gallery-examples

sphinx_gallery/tests/testconfs/src/plot_1.py
```diff
@@ -1,7 +1,9 @@
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
 """
```
choldgraf: no no - it's not encoding lines, it's sphinx references. People put stuff like

```
.. _my_ref:

My title
--------
```

which you're technically not needed to do since SG auto generates references, but [people do it anyway](https://github.com/mne-tools/mne-python/blob/master/examples/io/plot_read_events.py) :-)

doc/advanced_configuration.rst
```diff
@@ -148,19 +148,23 @@ provided the relative paths to `conf.py` of each sub gallery folder.
...
-    from sphinx_gallery.sorting import AmountOfCodeSortKey
+    from sphinx_gallery.sorting import NumberOfCodeLinesSortKey
     sphinx_gallery_conf = {
         ...
-        'within_subsection_order': AmountOfCodeSortKey,
+        'within_subsection_order': NumberOfCodeLinesSortKey,
     }
 
-However, other options exist that can be instantiated for use with
+However, multiple convenience classes are provided for use with
```
choldgraf: If you're going to make edits can you make this "In addition" instead of "However"?

If there aren't any more commits needed then I can edit and then merge (but see my other comment)

sphinx_gallery/tests/testconfs/src/plot_1.py
Eric89GXL: Ahh okay, I'll probably have to do something smarter then, I'll look into it more.

sphinx_gallery/gen_rst.py
```diff
@@ -153,12 +153,15 @@ def codestr2rst(codestr, lang='python', lineno=None):
     return code_directive + indented_block
 
 
-def extract_intro(filename, docstring):
+def extract_intro_title(filename, docstring):
```
choldgraf: this naming makes me thing that it extracts *just* the title of the introduction. could we make it `extract_intro_and_title`?

sphinx_gallery/gen_rst.py
```diff
@@ -153,12 +153,15 @@ def codestr2rst(codestr, lang='python', lineno=None):
...
     """ Extract the first paragraph of module-level docstring. max:95 char"""
 
     # lstrip is just in case docstring has a '\n\n' at the beginning
     paragraphs = docstring.lstrip().split('\n\n')
+    paragraphs = [p for p in paragraphs if not p.startswith('..')]
```
choldgraf: I think this is going to fix some other bugs in the process 🎉

sphinx_gallery/gen_rst.py
```diff
@@ -153,12 +153,15 @@ def codestr2rst(codestr, lang='python', lineno=None):
...
     if len(paragraphs) > 1:
+        title = paragraphs[0].split('\n')
+        title = ' '.join(t for t in title if t[0] not in ['=', '-', '^', '~'])
```
choldgraf: Maybe this could be a list constant somewhere? `RST_TITLE_CHARACTERS`?

sphinx_gallery/tests/test_gen_gallery.py
```diff
@@ -172,6 +172,7 @@ def _check_order(config_app, key):
     regex = '.*:%s=(.):.*' % key
     with codecs.open(index_fname, 'r', 'utf-8') as fid:
         for line in fid:
+            print(line)
```
choldgraf: just making sure this is intentional and not for debugging

choldgraf:
In general this looks great - a few small comments from me then LGTM

Eric89GXL:
@choldgraf done, I made the check for RST title chars better based on their official description and renamed the function.

choldgraf:
nice - so it sounds like this will work if you either put `===` *just* after, or *both* before and after the title, yeah?

LGTM...I'll give it 24 hrs and merge if nobody else brings anything up. That OK w/ you?

Eric89GXL:
> nice - so it sounds like this will work if you either put === just after, or both before and after the title, yeah?

Yeah and RST supports both so we should be good there.

Ready for merge from my end.

choldgraf:
🎉 🎉 🎉

lesteve:
Sorry I completely missed this one. I think this is great stuff. Quickly looking at the diff, I have the following comments though:

I think we should have an instance (rather than a class) as the value associated with the `within_subsection_order` to be consistent with `subsection_order`. Another strong reason is that some people may want to control the order explicity (and I am hoping `ExplicitOrder` may be used for this).

Another consistency problem, it would be good to have a consistent naming for sorting keys (i.e. finishing with `Order` or `SortKey`).

choldgraf:
> I think we should have an instance (rather than a class) as the value

I think this is a good point. No harm in having people give `NumberOfLinesSortKey()` and I agree it keeps things more standard. Would we need to change the code at all or would the sorting behavior stay the same regardless of whether its an instance or a class?

> it would be good to have a consistent naming for sorting keys 

+1, either is fine with me but maybe I have a small preference for `Order`. So we do `ExplicitOrder`, `NumberOfLinesOrder`, `AmountOfCodeOrder` etc. @Eric89GXL do you have thoughts?

Eric89GXL:
> I think we should have an instance (rather than a class) as the value

What eventually gets called is e.g. `instance = FileNameSortKey(src_dir); instance(filename)`. We could pass an instance, but then `__call__` would need to take `src_dir` as an arg, and return something (a function?) that would then operate on `fname`. Is this preferred?

This is also why the subsection ordering and within-subsection ordering need to be kept separate -- the former operates on effectively just `ExplictOrder(order).__call__(fname)` and the other on what would need to be `ExplicitExampleOrder(order)(src_dir)(fname)`. So I'm not sure `ExplicitOrder` can be used as a key using the existing interface.

I used the name `SortKey` to make it clear that it will be used as a ``key`` argument to the `sorted` function. Until just now, I actually didn't know if `ExplictOrder` needed to return ints (i.e., the index in the order), or if it also used as a sort key (i.e., just needed to provide anything that could be ordered by `sort`) when I looked deeper into the code. I'm fine with `*Order`, too, but we should document in the code that it's really just used as a sort key under the hood so the next dev knows immediately upon looking at the class.

lesteve:
> This is also why the subsection ordering and within-subsection ordering need to be kept separate -- the former operates on effectively just `ExplictOrder(order).__call__(fname)` and the other on what would need to be `ExplicitExampleOrder(order)(src_dir)(fname)`. So I'm not sure `ExplicitOrder` can be used as a key using the existing interface.

The hope when we were discussing the design was that they would be exactly the same. Is there a good reason the `within_subsection_order` sort key not be called on a list of the fully qualified filenames inside each `src_dir`?

Personally I am fine with SortKey because I know what a sort key is. The argument was that a significant chunk of our users may not have heard of a sort key. In the end I think that's why `Order` was used.

Eric89GXL:
> The hope when we were discussing the design was that they would be exactly the same. Is there a good reason the within_subsection_order sort key not be called on a list of the fully qualified filenames inside each src_dir?

It is eventually called that way -- any class for `within_subsection_order` will effectively operate as `Class(my_order)(src_dir)(fname)`. Any class for  `subsection_order`, by contrast, will effectively need to operate as `Class(my_order)(fname)`. How should these be unified?

Issue #277, opened 2017-07-30
ENH: Add thumbnail size setter


Eric89GXL:
It looks like the thumbnail dimensions are static:

https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/gen_rst.py#L357

I'm happy to add it as a `conf.py` key (`thumbnail_size` tuple?) if people agree it would be useful.

Titan-C:
Thumbnails are of fixed size and rather big(compared to the gallery thumbnail size) as on scikit-learn they were used on other places in their docs(the carousel for example).

I'm not sure about adding this extra setter. But if we go for it, I might say start clustering it with other options for the thumbnail, like the default image. Something like
```python
'gallery_thumbnails':{'default': 'no_image.png', 'size': (200, 140),'broken':'broken_example.png'}
```
Of course that would break backwards compatibility. As we have the default image as a separate key. But in this way we could expose those defaults to the config rather than setting them up inside the codebase. Then this default will also be more discoverable and help with #276

choldgraf:
It sounds like @Titan-C is suggesting that using CSS to resize images would be a decent solution, is that right?

Titan-C:
We already resize with CSS. My guess is @Eric89GXL wants to set a new smaller size on the Thumbnail, so save on space as the thumbnail images is larger than the saved size. Otherwise, if it is on the aesthetics of the thumbnail, yes changing CSS at will is the easiest and can be done on the user side.

Eric89GXL:
I actually just want larger images in the gallery section. If this can
already be done with CSS, then I'll just make a documentation PR update.

choldgraf:
were you able to figure this out w/ CSS? (and if so can we close this?)

Eric89GXL:
Will auto-close on merge

Eric89GXL:
CSS isn't sufficient, I have opened a different bug report about that (#282). Closing this one to move discussion there.

PR #283, opened 2017-08-03
MRG: Add thumbnail_size option

### LINKED ISSUE #282: BUG: pillarboxes are added to square images when scaling thumbnails
I couldn't figure out why I couldn't get my square-figure output to be size 250x250, then I looked at the actual thumbnail image, and it looked like this (ignore the improperly captured static image, this is a Mayavi problem that I need to track down), which I have put in a quoted bit of text so you can see the offset:

> sphx_glr_plot_fmri_activation_thumb

This gets created from this original image, which has no such offset:

> sphx_glr_plot_fmri_activation_001

In other words, SG adds white pillars on either side of the otherwise square image. It seems like this shouldn't be necessary and that CSS should be able to take care of whatever purpose this would serve (e.g., centering the element) instead. Can the rescaling routines be changed to fix this?

Eric89GXL:
Closes #282.

WIP because #279 needs to be merged first, then I can properly document how to get different thumbnail sizes.

In #277 @Titan-C you suggested this:

> I might say start clustering it with other options for the thumbnail, like the default image... Of course that would break backwards compatibility.

Here I've decided to avoid the step of combining keys. It seems like it could create a bit of deprecation pain for little gain in terms of readability.

I also added the "show" `Makefile` option because I've found it helpful previously, but I can split that off or remove it if need be.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -198,3 +198,6 @@ pseudoxml:
 	$(SPHINXBUILD) -b pseudoxml $(ALLSPHINXOPTS) $(BUILDDIR)/pseudoxml
 	@echo
 	@echo "Build finished. The pseudo-XML files are in $(BUILDDIR)/pseudoxml."
+
+show:
+	@python -c "import webbrowser; webbrowser.open_new_tab('file://$(PWD)/$(BUILDDIR)/html/index.html')"
--- a/doc/advanced_configuration.rst
+++ b/doc/advanced_configuration.rst
@@ -21,6 +21,7 @@ file:
 - ``reference_url`` (:ref:`link_to_documentation`)
 - ``backreferences_dir`` and ``doc_module`` (:ref:`references_to_examples`)
 - ``default_thumb_file`` (:ref:`custom_default_thumb`)
+- ``thumbnail_size`` (:ref:`setting_thumbnail_size`)
 - ``line_numbers`` (:ref:`adding_line_numbers`)
 - ``download_section_examples`` (:ref:`disable_joint_download`)
 - ``plot_gallery`` (:ref:`without_execution`)
@@ -55,6 +56,7 @@ For this you use in your Sphinx ``conf.py`` file a list of directories in
 the sphinx configuration dictionary::
 
     sphinx_gallery_conf = {
+        ...
         'examples_dirs': ['../examples', '../tutorials'],
         'gallery_dirs': ['auto_examples', 'tutorials'],
     }
@@ -72,6 +74,7 @@ if this naming convention does not suit your project, you can modify this patter
 in your Sphinx ``conf.py``. For example::
 
     sphinx_gallery_conf = {
+        ...
         'filename_pattern': '/plot_compute_',
     }
 
@@ -85,6 +88,7 @@ want to build only one example so that you can link it in the documentation. In 
 you would do::
 
     sphinx_gallery_conf = {
+        ...
         'filename_pattern': 'plot_awesome_example.py',
     }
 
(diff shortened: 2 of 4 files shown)


Eric89GXL:
I have now tested this and it works well to get PySurfer's images scaled nicely. It also makes the scaling slightly nicer for the BROKEN and default images by unifying the scaling routines instead of using different sizes. (This will have some presumably small negative impact on page load times.)

CIs are happy. Ready for review/merge from my end.

doc/Makefile
```diff
@@ -198,3 +198,6 @@ pseudoxml:
 	$(SPHINXBUILD) -b pseudoxml $(ALLSPHINXOPTS) $(BUILDDIR)/pseudoxml
 	@echo
 	@echo "Build finished. The pseudo-XML files are in $(BUILDDIR)/pseudoxml."
+
+show:
+	@python -c "import webbrowser; webbrowser.open_new_tab('file://$(PWD)/$(BUILDDIR)/html/index.html')"
```
choldgraf: I'm +1 for this...my fingers will appreciate it in the long run

doc/advanced_configuration.rst
```diff
@@ -85,6 +88,7 @@ want to build only one example so that you can link it in the documentation. In
 you would do::
 
     sphinx_gallery_conf = {
+        ...
```
choldgraf: good idea for these ellipses

doc/advanced_configuration.rst
```diff
@@ -478,24 +493,36 @@ the example script.
...
+maintaining the original aspect ratio. This size can be controlled with the
+``thumbnail_size`` entry as, e.g.::
+
+    sphinx_gallery_conf = {
+        ...
+        'thumbnail_size': (250, 250),
+    }
+
+The gallery uses various CSS classes to display these thumbnails, which
+default to maximum 160x160px. To change this, e.g. to display the images
```
choldgraf: why do you type the sizes in regular font here, but use code font above?

sphinx_gallery/gen_rst.py
```diff
@@ -313,8 +313,12 @@ def scale_image(in_fname, out_fname, max_width, max_height):
...
     height_sc = int(round(scale * height_in))
 
-    # resize the image
-    img.thumbnail((width_sc, height_sc), Image.ANTIALIAS)
+    # resize the image using resize; if using .thumbnail and the image is
+    # already smaller than max_width, max_height, then this won't scale up
+    # at all (maybe could be an option someday...)
+    img = img.resize((width_sc, height_sc), Image.BICUBIC)
+    # img.thumbnail((width_sc, height_sc), Image.BICUBIC)
+    # width_sc, height_sc = img.size  # necessary if using thumbnail
```
choldgraf: what are these commented-out lines for? future dev?

doc/advanced_configuration.rst
Eric89GXL: Python vs CSS (approximately)

Eric89GXL: Python vs CSS (approximately)

sphinx_gallery/gen_rst.py
Eric89GXL: Using the thumbnail (non-upsampling) method if we ever want it (as suggested by the preceding comments)

Eric89GXL:
Yes I tested these code suggestions on restoring the PySurfer examples back to 250x250px

Eric89GXL:
Before/after would be:

hinx-gallery.readthedocs.io/en/latest/auto_examples/index.html

https://168-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/rtd_html/auto_examples/index.html

Though it's not terribly interesting because SG uses the defaults.

choldgraf:
ewww @Eric89GXL you're using matplotlib < 2.0?!?! That jet is killing me :-)

LGTM

Issue #221, opened 2017-04-10
Allow option to not display total running time in the rendered docs


anntzer:
For non-computationally-heavy projects, the running times will usually all be in the order of milliseconds, and thus mostly irrelevant noise.

Titan-C:
We can do this but I still like to see that the example runs fast. And that trying it out won't take much time. Do you want to deactivate this measure for the complete gallery of only for the quick examples?
@lesteve , @GaelVaroquaux any opinions?

anntzer:
A configurable threshold would be fine too.  Usually if one is writing examples that take milliseconds to run that should be pretty obvious.  Notifying for the opposite (unexpectedly slow examples) may be more relevant.

GaelVaroquaux:
I think that a switch to globally turn off could be an option. Another
solution would be to give a specific CSS class, and CSS could then be
used to hide. But many users would have a hard time figuring this out, so
the switch is probably better (both would be useful, for theming
purposes).

anntzer:
Essentially closed by #290.

PR #290, opened 2017-08-31
Add threshold below which computation time is not reported.

anntzer:
See #221.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,12 @@ Change Log
 git master
 ----------
 
+New features
+''''''''''''
+
+* Added ``min_reported_time`` configurable.  For examples that run faster than
+  that threshold (in seconds), the execution time is not reported.
+
 v0.1.12
 -------
 
--- a/doc/advanced_configuration.rst
+++ b/doc/advanced_configuration.rst
@@ -29,6 +29,7 @@ file:
 - ``find_mayavi_figures`` (:ref:`find_mayavi`)
 - ``abort_on_example_error`` (:ref:`abort_on_first`)
 - ``expected_failing_examples`` (:ref:`dont_fail_exit`)
+- ``min_reported_time`` (:ref:`min_reported_time`)
 
 Some options can also be set or overridden on a file-by-file basis:
 
@@ -552,6 +553,20 @@ in your own site's CSS file:
     .sphx-glr-thumbcontainer a.internal {
         padding: 270px 10px 0 !important;
     }
+
+
+.. _min_reported_time:
+
+Minimal reported time
+=====================
+
+By default, Sphinx-gallery logs and embeds in the html output the time it took
+to run each script.  If the majority of your examples runs quickly, you may not
+need this information.
+
+The ``min_reported_time`` configuration can be set to a number of seconds.  The
+duration of scripts that ran faster than that amount will not be logged nor
+embedded in the html output.
 
 
 .. _regular expressions: https://docs.python.org/2/library/re.html
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -45,6 +45,7 @@ DEFAULT_GALLERY_CONF = {
     'failing_examples': {},
     'expected_failing_examples': set(),
     'thumbnail_size': (400, 280),  # Default CSS does 0.4 scaling (160, 112)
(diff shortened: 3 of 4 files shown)


sphinx_gallery/gen_rst.py
```diff
@@ -638,9 +638,10 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
     save_notebook(example_nb, example_file.replace('.py', '.ipynb'))
     with codecs.open(os.path.join(target_dir, base_image_name + '.rst'),
                      mode='w', encoding='utf-8') as f:
-        example_rst += "**Total running time of the script:**" \
-                       " ({0: .0f} minutes {1: .3f} seconds)\n\n".format(
-                           time_m, time_s)
+        if time_elapsed >= gallery_conf["min_reported_time"]:
```
choldgraf: is there any situation in which not forcing this to be a float will bite us?

doc/advanced_configuration.rst
```diff
@@ -554,4 +555,18 @@ in your own site's CSS file:
     }
 
 
+.. _min_reported_time:
```
choldgraf: mind if I clean up this wording a small amount before merging? happy to do it myself as a quick commit-then-merge

anntzer: Feel free to force push to my branch.

sphinx_gallery/gen_rst.py
anntzer: If somebody sets the min computation time to not a float then that will crash but what can you do?...  Obviously an explicit type check could be added, feel free to push one.

larsoner: I'd assume YAGNI and rely on the Python error message to be explicit enough

choldgraf: ah I was just thinking to add a `float(gallery_conf["min_reported_time"])` in there as opposed to doing a check

larsoner: sure, that might give better errors in any case

anntzer: That (so support strings when we want a float) I think would be unpythonically wrong.

larsoner: If we eventually add it as a command-line option I think it would come in as `str`, though

choldgraf: ok I'm gonna wrap it in float for now and we can make it error later if we want

CHANGES.rst
```diff
@@ -4,6 +4,11 @@ Change Log
 git master
 ----------
 
+New features
+''''''''''''
+
+* Added ``min_reported_time`` configurable.
```
lesteve: You will need to shortly explain what this does.

doc/advanced_configuration.rst
```diff
@@ -554,4 +555,18 @@ in your own site's CSS file:
...
+
+Minimal reported time
+=====================
+
+By default, Sphinx-gallery logs and embeds in the html output the time it took
+to run each script.  If the majority of your examples runs quickly, you may not
+need this information.
+
+The ``min_reported_time`` confirguration can be set to a number of seconds.
```
lesteve: typo in configuration

sphinx_gallery/gen_gallery.py
```diff
@@ -247,11 +248,12 @@ def generate_gallery_rst(app):
 
             fhindex.write(SPHX_GLR_SIG)
 
-    if gallery_conf['plot_gallery']:
+    if gallery_conf['plot_gallery'] and gallery_conf['min_reported_time']:
```
lesteve: I don't understand why you are adding `and gallery_conf['min_reported_time']`. The vaiue of min_reported_time is 0 so this does not log by default, right?

CHANGES.rst
anntzer: done

doc/advanced_configuration.rst
anntzer: fixed

sphinx_gallery/gen_gallery.py
anntzer: Good catch, probably leftover from earlier design.

choldgraf:
Is this ready to go @anntzer ?

anntzer:
sure.

PR #292, opened 2017-09-10
Report version to sphinx.

anntzer:
Allows one to require a version from sphinx, see
http://www.sphinx-doc.org/en/stable/config.html#confval-needs_extensions
http://www.sphinx-doc.org/en/stable/extdev/index.html#extension-metadata

Not doing `from . import __version__` as this would also set the version
of the submodule.

### DIFF
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -16,8 +16,7 @@ import copy
 import re
 import os
 
-from . import glr_path_static
-from . import sphinx_compatibility
+from . import sphinx_compatibility, glr_path_static, __version__ as _sg_version
 from .gen_rst import generate_dir_rst, SPHX_GLR_SIG
 from .docs_resolv import embed_code_links
 from .downloads import generate_zipfiles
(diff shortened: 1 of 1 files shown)


larsoner:
> Not doing from . import __version__ as this would also set the version of the submodule.

It's probably better to do `from . import __version__ as _sg_version` or so to avoid this problem while still using relative imports

anntzer:
sure, fixed

Titan-C:
It looks good.
Travis on py2 ubuntu failed. It is certainly unrelated to this. I refreshed the build, but we might need to pin seaborn or pandas version in this build until it passes.

choldgraf:
looks like 2.7 failed again...what should be done to get the fix in this PR?

Titan-C:
> looks like 2.7 failed again...what should be done to get the fix in this PR?

I noticed that the 2.7 on ubuntu uses seaborn 0.8.1, whereas the conda
installation has 0.8. Or something like that, I'm writing out of
memory. We could pin the seaborn version in the travis file.

GaelVaroquaux:
The failure is unrelated this PR is useful and good to go. Given that it has had approval above, I am merging.

PR #293, opened 2017-09-21
MRG: Fix gzip/BytesIO error

larsoner:
Over at MNE we are getting:
```
embedding documentation hyperlinks...

Exception occurred:
  File "/home/ubuntu/mne-python/sphinx-gallery/sphinx_gallery/docs_resolv.py", line 50, in _get_data
    data = StringIO(data)
TypeError: initial_value must be unicode or None, not str
```
I have a fix but I want to see if I can expose it first (TDD!).

### DIFF
--- a/sphinx_gallery/docs_resolv.py
+++ b/sphinx_gallery/docs_resolv.py
@@ -28,7 +28,7 @@ except ImportError:
     import urllib.parse as urllib_parse
     from urllib.error import HTTPError, URLError
 
-from io import StringIO
+from io import BytesIO
 
 from sphinx.search import js_index
 
@@ -47,7 +47,7 @@ def _get_data(url):
         if encoding == 'plain':
             data = data.decode('utf-8')
         elif encoding == 'gzip':
-            data = StringIO(data)
+            data = BytesIO(data)
             data = gzip.GzipFile(fileobj=data).read().decode('utf-8')
         else:
             raise RuntimeError('unknown encoding')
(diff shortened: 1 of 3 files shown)


larsoner:
Okay CIs are happy other than the standard Seaborn failure.

This one is currently breaking all of our CI builds, so it's a high priority from my end. Ready for review/merge.

sphinx_gallery/docs_resolv.py
```diff
@@ -47,7 +47,7 @@ def _get_data(url):
         if encoding == 'plain':
             data = data.decode('utf-8')
         elif encoding == 'gzip':
-            data = StringIO(data)
+            data = BytesIO(data)
```
choldgraf: Is the reason this doesn't work a python 2 vs. 3 thing?

choldgraf:
LGTM...travis problem doesn't seem related to this PR

sphinx_gallery/docs_resolv.py
larsoner: I originally thought so, but then the first commit (true TDD!) broke *every* CI, and on 3k the error is:
```
  File "/home/travis/miniconda2/envs/testenv/lib/python3.6/site-packages/sphinx_gallery-0.1.12-py3.6.egg/sphinx_gallery/docs_resolv.py", line 50, in _get_data
    data = StringIO(data)
TypeError: initial_value must be str or None, not bytes
```

choldgraf: weird!

choldgraf:
I'll merge in a couple hours in case somebody else has thoughts

sphinx_gallery/docs_resolv.py
lesteve: I think the data you get over the network is supposed to be bytes in Python 3, so BytesIO makes a lot of sense to me.

lesteve:
I was not aware of the seaborn failure, I think this is because we get pandas via apt-get and it is quite ancient i.e. 0.13.1.

larsoner:
> I was not aware of the seaborn failure, I think this is because we get pandas via apt-get and it is quite ancient i.e. 0.13.1.

This was discussed a bit [last week](https://github.com/sphinx-gallery/sphinx-gallery/pull/292#issuecomment-328571839). I agree a newer `pandas` could help fix it.

In the meantime is this good to go from your end @lesteve? It's blocking SG functionality currently :(

lesteve:
Yes this is good to go, thanks a lot, merging!

PR #294, opened 2017-09-21
[MRG] Deactivate virtualenv provided by Travis

lesteve:
This should fix the Travis build failure seen in https://github.com/sphinx-gallery/sphinx-gallery/pull/292#issuecomment-328479686.

Travis provides numpy in its default virtual env and I am guessing that they upgraded the version recently. In any case we want to have control over the numpy we are using in this case the numpy from apt-get

I did some minor clean-ups in .travis.yml while I was at it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,7 @@ virtualenv:
 matrix:
   include:
     - os: linux
-      dist: trusty
-      sudo: required
-      env: DISTRIB="ubuntu 14" PYTHON_VERSION="2.7"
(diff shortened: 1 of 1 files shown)


Titan-C:
I'm surprised about Travis giving a numpy version in their virtual env which is more updated that the OS provided. In general Travis has very outdated versions of everything.
That though aside. This seems to solve our problems well enough, I'm merging this.

lesteve:
Back in the days (when numpy wheels were not available for Linux) Travis did that for projects that had a dependency on numpy and were doing `pip install numpy` which would build numpy from source and take 5 minutes or so.

PR #296, opened 2017-10-02
MRG+1: Use intersphinx

larsoner:
Closes #287.

Simplifies link resolution by using `intersphinx`.

~~I'm not sure the best way to test this. I was doing it locally by having a small `testbuild` directory with a `conf.py`, `index.rst`, `examples/plot_numpy_scipy.py`. I could turn this into a test if it seems useful. Would the best way be to directly include the files, copy them to a tempdir, call `Sphinx(...)`, and check the output links to ensure they contain links to numpy/scipy/matplotlib docs, as they should?~~

PR #298, opened 2017-10-03
[MRG] ENH: only directive for downloads

### LINKED ISSUE #297: Download button and 'generated by sphinx-gallery' should be only:: html
When generating a PDF with a project using sphinx-gallery, the examples appear with 
"""
Download Python source code:
Download Jupyter notebook:
Generated by Sphinx-Gallery
"""
Which is very strange. We should put the corresponding rst code in "only" directives so that it appears only on html output

GaelVaroquaux:
Fixes #297


larsoner:
Do we generate any PDFs in the tests / on CircleCI? Perhaps we should. We could maybe add it to the `conf.py` for #296 so we can easily inspect the output, but we'd have to be careful about LaTeX requirements.

GaelVaroquaux:
Adding LateX on Circle will increase the build times a lot. Are we sure that we want to do that?

larsoner:
> Are we sure that we want to do that?

I'll open a PR so we can see how much time/pain it adds :)

larsoner:
When I rebase this on my `intersphinx`, do `make latexpdf`, and view it, I still see the:
```
Download Python source code:
Download Jupyter notebook:
plot_numpy_scipy.py
plot_numpy_scipy.ipynb
Gallery generated by Sphinx-Gallery
```
I don't understand how this is possible given the code. I assume you tested locally and it works for you?

larsoner:
Oh actually it looks like a `make clean` fixes it. +1 for merge from me

choldgraf:
TIL about the `only` directive. neat!

GaelVaroquaux:
> Oh actually it looks like a make clean fixes it. +1 for merge from me

Thanks for checking!

PR #303, opened 2017-10-09
[MRG+1] FIX: Update packages

larsoner:
Let's see if CircleCI is happy with this.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -6,6 +6,7 @@ machine:
 dependencies:
   pre:
     # Disable pyenv (no cleaner way provided by CircleCI as it prepends pyenv version to PATH)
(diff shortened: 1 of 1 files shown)


larsoner:
CircleCI is happy:

https://circleci.com/gh/sphinx-gallery/sphinx-gallery/225#artifacts/containers/0

choldgraf:
Is there a concern that this will randomly break at some point in the future if a package gets updated?

GaelVaroquaux:
If it breaks when a package is updated, it is quite a good thing: it will make explicit a breakage that would have happened on users computer.

+1 for merge

lesteve:
LGTM, merging, thanks a lot!

PR #304, opened 2017-10-09
FIX: Update docs

### LINKED ISSUE #302: Having `plot_gallery` as a bool in conf.py raises warning
Having

```python
plot_gallery = True
```

in `conf.py` results in (seem independent of sphinx version)

```
16:00 $ python make.py html
Running Sphinx v1.5.2
Initializing GitHub plugin

Warning, treated as error:
WARNING: The config value `plot_gallery' has type `bool', defaults to `str'.

Building HTML failed.

```

This fails the build if you have sphinx set to treat warnings as errors.

Replacing with

```python
plot_gallery = 'True'
```

works as expected

larsoner:
I put this in 1.12 but let me know if it should go in `git master`

Closes #302

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,9 @@ New features
 * Use intersphinx for all function reference resolution `#296 <https://github.com/sphinx-gallery/sphinx-gallery/pull/296>`_
 * Sphinx only directive for downloads `#298 <https://github.com/sphinx-gallery/sphinx-gallery/pull/298>`_
 * Allow sorting subsection files `#281 <https://github.com/sphinx-gallery/sphinx-gallery/pull/281>`_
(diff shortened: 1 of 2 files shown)


GaelVaroquaux:
I think that this should go in master.

Aside from that, +1 for merge

lesteve:
> I put this in 1.12 but let me know if it should go in git master

There was a typo in the section it should have been v0.1.13 (rather than v0.1.12) which I fixed in master. This change is in v0.1.13 section as shown by https://github.com/sphinx-gallery/sphinx-gallery/commit/ea60d182bafe8cfee6eba851b46a2d6b8402933b.

I pushed an update to the doc while I was at it.

lesteve:
Merging, thanks a lot @larsoner!

PR #305, opened 2017-10-15
[MRG] FIX import local modules in examples

NelleV:
This patch allows to import a module placed in the same folder as the examples when executing the example by adding the current directory to the sys.path.

I used a deepcopy to restore the sys.path to the original state (before running the example).

This patch relates to a problem we are having on https://github.com/matplotlib/matplotlib/pull/9402
I am honestly not entirely convince that sphinx-gallery *should* support this, as I think that good examples are self-contained examples.

### DIFF
--- a/examples/local_module.py
+++ b/examples/local_module.py
@@ -1,0 +1,5 @@
+"""
+A trivial local module to provide a value for plot_exp.py.
+"""
+
+N = 100
--- a/examples/plot_exp.py
+++ b/examples/plot_exp.py
@@ -14,9 +14,13 @@ stacking multiple images.
 import numpy as np
 import matplotlib.pyplot as plt
 
+# You can use modules local to the example being run, here
+# we just use a trivial NumPy wrapper
+from local_module import N  # = 100
+
 
 def main():
-    x = np.linspace(-1, 2, 100)
+    x = np.linspace(-1, 2, N)
     y = np.exp(x)
 
     plt.figure()
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -14,6 +14,7 @@ Files that generate images should start with 'plot'.
 # tricky errors come up with exec(code_blocks, ...) calls
 from __future__ import division, print_function, absolute_import
 from time import time
+import copy
 import ast
 import codecs
 import hashlib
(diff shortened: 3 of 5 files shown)


choldgraf:
can anyone think of a way in which this could bite us in the long-run?

larsoner:
The way I have solved this previously is by doing `sys.path.insert(0, os.getcwd())` in the few examples that require such functionality. I don't mind it because it's explicitly showing that an import needs to be made to some local file.

That being said, I don't mind adding this bit of code. My understanding is that SG running scripts is meant to mimic running them manually as much as possible. This seems like it will improve upon that. So +0.5 from me

Titan-C:
I have experienced the same kind of problem, although I ended up moving
the needed functions to the main package directory instead of a local
module once they where mature enough.

> That being said, I don't mind adding this bit of code. My understanding is that SG running scripts is meant to mimic running them manually as much as possible. This seems like it will improve upon that. So +0.5 from me

I have the same feeling. SG should be equivalent to running a script as
an independent entity and I hope to get one day there. That would make
multiprocessing support possible and will help clearing all this tricks
we put in to emulate that (changing dirs, changing sys.argv, cleaning
seaborn imports).

This is very similar to what we do changing sys.argv to avoid those
problems, refer PR#252. I would thus prefer if we put this changes in
the generate_file_rst next to the sys.argv change. That should make
future maintenance simpler, they are not the nicest fixes but we have
them for now and I would like to keep them together.

lesteve:
This would be nice to have a test for this, ideally as a unit test, but a separate example will do if the former is very inconvenient. Without test, this kind of micro-features tend to be broken very quickly.

NelleV:
Great! I'll write a test for this then (or add an example)

anntzer:
Copying sys.path can be done with just `sys.path[:]` (or `sys.path.copy()` if you want to be super explicit) -- it's always a list of strings, nothing fancier.

tacaswell:
Somewhat related, where does the information for the new style namespace packages get stashed?

GaelVaroquaux:
Is this still up to date? If so, @NelleV : can you fix the merge conflict, so that we merge this.

tacaswell:
Also maybe related to https://github.com/matplotlib/matplotlib/issues/11838 (which breaks building on python3.7).

larsoner:
It would also be good to add a test. @NelleV if you do not have time I'm happy to take over and add one

larsoner:
Rebased and tests added, I'll merge once it's green since folks were happy with it other than needing a test

NelleV:
Thank you :)

larsoner:
Thanks @NelleV

tacaswell:
Thanks!

PR #307, opened 2017-10-20
DOC backreferences_dir is now mandatory

NelleV:
so adding it to the fast set up page.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -128,8 +128,6 @@ is declared with a relative path from the ``conf.py`` file location::
       'examples_dirs' : '../examples',
       # path where to save gallery generated examples
       'gallery_dirs'  : 'auto_examples',
-      # Directory for back-references:
-      'backreferences_dir'  : 'gen_modules/backreferences',
     }
 
 The ``gallery_dirs`` is the folder where Sphinx-Gallery will store the
--- a/README.rst
+++ b/README.rst
@@ -124,10 +124,10 @@ dictionary for Sphinx-Gallery. The examples directory ``../examples``
 is declared with a relative path from the ``conf.py`` file location::
 
     sphinx_gallery_conf = {
-       # path to your examples scripts
-       'examples_dirs' : '../examples',
-       # path where to save gallery generated examples
-       'gallery_dirs'  : 'auto_examples',
+         # path to your examples scripts
+         'examples_dirs': '../examples',
+         # path where to save gallery generated examples
+         'gallery_dirs': 'auto_examples',
     }
 
 The ``gallery_dirs`` is the folder where Sphinx-Gallery will store the
--- a/README.rst
+++ b/README.rst
@@ -124,10 +124,13 @@ dictionary for Sphinx-Gallery. The examples directory ``../examples``
 is declared with a relative path from the ``conf.py`` file location::
 
     sphinx_gallery_conf = {
-	# path to your examples scripts
-	'examples_dirs' : '../examples',
-	# path where to save gallery generated examples
-	'gallery_dirs'  : 'auto_examples'}
+      # path to your examples scripts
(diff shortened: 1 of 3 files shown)


Titan-C:
backreferences_dir is not mandatory. It remains as an option.
We do issue a warning if the variable is not set, but the quick start
guide should not enforce a directory just like that.

I would agree for the sanity of the newcomers to avoid them to see this
warning, and set the use of False in the quickstart guide. Default is
set to None, because we want old users to update their config, and we
need a way for them to see the warning.

The backreferences are a fantastic feature of the gallery, but getting
them to work is a bit more involved, especially in our main use for
generating mini-galleries at the function level. Thus I would put an
explanation why the guide sets this variable to False, and a link to the
documentation of this feature. In such way the more invested user, can
read the docs and enable this feature.

NelleV:
That change doesn't get backreferences to work: it just avoids sphinx-gallery's red warning messages, which I really don't expect when setting up a project. By default, it should work without warnings.

choldgraf:
I agree with @NelleV  that "it should work out of the box without warnings" is a good first-experience with sphinx-gallery.

lesteve:
IIRC the warning was to gently nudge people towards updating their config (old way of doing things was mod_examples_dir). It was done in https://github.com/sphinx-gallery/sphinx-gallery/commit/dd661dd1e29f9f9945c090a01a5f7b41a5c10919 which is in 0.1.9 *6 months ago April 20th 2017). We have released 0.1.13 recently.

Maybe this is enough time to remove the warning. Alternatively we could only warn if we detect the old mod_example_dir key in the sphinx config.

choldgraf:
I'm +1 on warning only if we detect the old key
--

NelleV:
I agree as well.

Titan-C:
Ok. Let's change the warning setup. Put the warning when the old key is found.

Nevertheless I'm still unsure about the default. What is getting backreferences to work? In my most common use case it is to configure `autosummary` to use them to extend the API documentation. In which case just declaring the directory where they are stored is not enough, there are some extra steps you need to be careful about. I rather make the default `False` and direct the user to the docs on how to setup this feature correctly.
- If to work is to just store them in a directory, yes we could suggest a default directory in the quickstart guide in the main readme.

choldgraf:
hey all - looking through the comments, it looks like this is what needs to be done here:

* Raise a warning if `mod_example_dir` is in there
* Set the default value to `None` (or `False`, whichever works better) and include in the comment a link to the docs on how to set this up.

That it? @NelleV is this is something you have time to clean up, or do you want someone else to take over the PR?

NelleV:
Someone can take over the PR

choldgraf:
@NelleV I don't think this should be a tough fix*, I can just push to this PR and we can move from there


*famous last words

choldgraf:
Just pushed changes to the PR, it now reverts the documentation addition, but now makes the warning *only* come up if it detects `mod_example_dir`!

choldgraf:
ok I think this is ready for you to take a look @Titan-C :-)

lesteve:
I pushed a few tweaks:
* do not warn and set a default backreferences_dir when backreferences_dir is None. I think this was supposed to be nice for users that were using the default magical mod_example_dirs. In fact this was actually working for not that many users.
* test that no warning is issued for a simple config
* minor tweaks to warning message when mod_example_dir is used

@choldgraf do you want to have a quick look at this one?

choldgraf:
looks good to me @lesteve ... +1 to merge if you're happy w/ it!

lesteve:
OK let's merge this one then!

choldgraf:
woot! thanks for helping push this one through

choldgraf:
and thanks @NelleV for making the initial push!

PR #308, opened 2017-10-24
Fix __future__ imports across cells

lesteve:
I found this problem in scikit-learn when building the documentation with Python 2. An example showing the problem is:

```py
"""
This example tests that __future__ imports works across cells

"""

####################
# Section
print('test', end='')
```

You get a SyntaxError because the `__future__` import in the first code cell is not taken into account in the second cell.

I am not an expert in the ast/compile flags, so if someone sees a better way to do the same thing, let me know.


larsoner:
Do we know if these `__future__` imports are properly isolated across examples? We could check, at least on Python 2, by doing `assert 1 / 2 == 0` in one file and  `from __future__ import division; assert 1 / 2 == 0.5` in another. Is it related to this / worth adding? (I think I've had something like this come up before with `print_function` not working even though a file had `from __future__ import print_function` before.)

lesteve:
Also any suggestion of testing this without adding a dummy example would be more than welcome!

larsoner:
Add it to the `tinybuild` instead then it won't go in the docs but will be tested

https://github.com/sphinx-gallery/sphinx-gallery/tree/master/sphinx_gallery/tests/tinybuild

lesteve:
> Do we know if these __future__ imports are properly isolated across examples?

I think they are because I used a separate compiler per file but I have not checked.

> Add it to the tinybuild instead then it won't go in the docs but will be tested

I saw that but my understanding was that basically, the test_embed_links function in in sphinx_gallery/tests/test_full.py would test my fix, which is a bit misleading. Ideally it would be better to have a more fine-grained way of testing different things in tiny_build, wouldn't it?

larsoner:
>  Ideally it would be better to have a more fine-grained way of testing different things in tiny_build, wouldn't it?

Sure. Feel free to rename or change as necessary. Either way, it seems nicer to have the regression testing explicitly done by `pytest` rather than needing to run `make doc` (the current PR requirement)

lesteve:
> Sure. Feel free to rename or change as necessary. Either way, it seems nicer to have the regression testing explicitly done by pytest rather than needing to run make doc (the current PR requirement)

I will try this route and see how far I can go.

lesteve:
I am using tinybuild now, I did some work in sphinx_gallery/tests/test_full.py to have a module-level fixture to run the sphinx app only once and split the existing test in two (one that just checks that the sphinx app runs with basic checks, and test_embed_links that just tests the embedding of the links')

tinybuild is a nice enhancement I think we could test more things there ... making gen_rst.generate_file_rst more easily testable would certainly be another improvement going in the same direction.

larsoner:
Since we're checking `__future__` usage, WDYT about:

1. adding a `from __future__ import division; print(1 / 2);` in some example that will be run before a separate example that does `print(1 / 2)`, and
2. on Python 2 check to make sure the former prints 0.5 and the latter 0, and on Python 3 check that they are both 0.5

It seems to get at mostly the same point of this PR, namely that `__future__` imports are handled properly. But it's also reasonable to do it separately later if you want to be done for now.

sphinx_gallery/tests/test_full.py
```diff
@@ -49,6 +60,5 @@ def test_embed_links():
     assert 'scipy.signal.firwin.html' in lines
     assert '#module-numpy' in lines
     assert 'numpy.arange.html' in lines
-    # The matplotlib doc download is a bit unsafe, so skip for now:
-    # assert '#module-matplotlib.pyplot' in lines
-    # assert 'pyplot_api.html' in lines
+    assert '#module-matplotlib.pyplot' in lines
```
larsoner: I had these commented out for a reason :)

https://travis-ci.org/sphinx-gallery/sphinx-gallery/jobs/292615909#L1129

lesteve: Yeah I did not understand why and it was passing locally so I figured I would give it a go ... oh well, maybe I'll comment them out again ... maybe it depends on the sphinx version ?

larsoner: Yeah probably sphinx. I guess you could check whether version 1.5+ is used and only check if that's the case, as I think that Ubuntu build uses 1.4.9

requirements.txt
```diff
@@ -1,5 +1,5 @@
 setuptools
-matplotlib
+matplotlib==2.0
```
larsoner: revert this, too?

lesteve: Yep, done.

sphinx_gallery/tests/test_full.py
lesteve: For the record I could not reproduce this locally with sphinx 1.4.9. I thought maybe the matplotlib version has something to do with it (and the possible mismatch with the matplotlib doc fetched from the internet but I did not manage to reproduce either).

lesteve:
> It seems to get at mostly the same point of this PR, namely that __future__ imports are handled properly. But it's also reasonable to do it separately later if you want to be done for now.

I could do that, the example order makes it not that obvious which example is run first though.  I could make two example folders so that at least the order is easier to guess (examples and second_folder_to_run would work I think). To be honest I would rather make `gen_rst.generate_file_rst` easier to test but the first solution may take less time.

sphinx_gallery/tests/test_full.py
larsoner: Yeah at one point I tried to reproduce, as well. To really find the problem I figured I'd need to get a Trusty system going with system Python packages to mirror those of Travis, which was too much for me to follow through on.

larsoner:
> I could do that, the example order makes it not that obvious which example is run first though.

The default IIRC is "amount of code" so just ensure that the second one to be run has more lines of code. Or explicitly set the order in `conf.py`, as it's a configurable thing now (can be alphabetical if you want to do it that way, or explicit if you just want to list them)

larsoner:
i.e. 

http://sphinx-gallery.readthedocs.io/en/latest/advanced_configuration.html#within-gallery-order

lesteve:
Yeah I remember this. It's just that the test is not as obvious as it could be.

larsoner:
I think it could be made sufficiently explicit with comments in the vicinity of the test. But up to you if you want to try it or not

lesteve:
I think this is ready to go. It was a bit painful but I discovered a small issue with the fix (hence plot_future_imports_broken.py) so it was worth it.

larsoner:
Changes LGTM, glad to see this `__future__` bug fixed!

lesteve:
OK merging this one then, thanks for the review @larsoner!

Issue #269, opened 2017-07-15
Problem compiling from py to rst with unicode when local is not defined on mac


NelleV:
Someone at the matplotlib sprint had a problem compiling the matplotlib documentation because of an example with utf8 character. His local was not set, but the encoding was defined correctly at the top of the file and thus the example ran fine.
When sphinx-gallery attempted to compile this python file into an rst file, it failed to detect the unicode encoding.

This problem was encountered on mac os x. We have not tried to reproduce it.

lesteve:
This kind of problem with encoding has been noted a few times. I think we should add a test to make sure that the tests pass fine with an ASCII locale as I commented in https://github.com/sphinx-gallery/sphinx-gallery/issues/253#issuecomment-313123210.

lesteve:
See how it was done there for example: https://github.com/astropy/sphinx-automodapi/pull/33/files#diff-354f30a63fb0907d4ad57269548329e3R46

Titan-C:
I did stat work on this in my branch https://github.com/Titan-C/sphinx-gallery/tree/encoding
It is not only failing to build because of ASCII encoding, but at the same time recognizing the used functions does not work on python 2(Reason to revert #257 ) and issue #260 currently.

anntzer:
Shouldn't you use https://docs.python.org/3/library/tokenize.html#tokenize.detect_encoding or https://docs.python.org/3/library/tokenize.html#tokenize.open to detect the encoding?  That should work as is on Py3; I see that you need to pass a bytestring to ast on Py2 so I'd suggest just reencoding the unicode string back to UTF-8 for Py2.

lesteve:
Closed by #311.

NelleV:
Thanks!

Issue #260, opened 2017-06-29
Test embeding of links in source blocks


Titan-C:
Link embedding is fragile. It stopped working in python 2 after #257 thus it is reverted in  #259
Need a test with little html to check links are being embedded.

lesteve:
I think we have something like this already:
https://github.com/sphinx-gallery/sphinx-gallery/blob/ebe1ca3cbdd7a410c2a45b22cb374af08d4912d5/sphinx_gallery/tests/test_full.py#L48-L64

Closing, @Titan-C feel free to reopen if I missed something.

PR #311, opened 2017-10-31
[MRG] Encoding problems

Titan-C:
The filesystem's encoding has generated quite some problems  #257, #259, #260, #269
I started this branch long ago, I'll bring it back to live now. This starts the test on Travis-CI for the ASCII encoding.
It does not work yet.


lesteve:
It'd be nice to fix this indeed. Your last commit status is green, does that mean that this is ready for review?

Titan-C:
> It'd be nice to fix this indeed. Your last commit status is green, does that mean that this is ready for review?

Not really. I just arrived to an easy solution. Enforcing reading all
files as byte strings. But I actually want to refactor and put some more
test. This solution is easy and wont bite us back as long as no one
declares function names with characters outside ASCII before terminate
support for Python2.7. But I'm still a bit worried about comments as
normal strings.

I also need a new test that verifies code blocks are correctly parsed
for embedding documentation links, to avoid #259 happening again.

lesteve:
> I also need a new test that verifies code blocks are correctly parsed for embedding documentation links, to avoid #259 happening again.

I think there is a test for that already:
https://github.com/sphinx-gallery/sphinx-gallery/blob/587198a79c6e602e7a0be97f1f06c9ab57a123c2/sphinx_gallery/tests/test_full.py#L48-L64

Although for some reason the matplotlib check was not passing on Travis which is why it is commented out. I could not reproduce, see https://github.com/sphinx-gallery/sphinx-gallery/pull/308#discussion_r146858672 for more details.

larsoner:
> Although for some reason the matplotlib check was not passing on Travis 

I suspect this was due to old `sphinx` being used

Titan-C:
This is finally working. I even found some extra problems that could be originated by different encodings, characters in use and the source file config. I have added comments explaining such events. I also have generated a new example using unicode emoji just to stress test this setup.
https://184-26191147-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/alabaster_html/auto_examples/plot_unicode_everywhere.html

Titan-C:
Are there any comments on this? Can I merge?
@pllim @NelleV does this solve your issues #253 & #269 ?

pllim:
@Titan-C , I don't have time to test this right now but I can give it a try in the future when a new release with this fix is in. Thanks!

sphinx_gallery/py_source_parser.py
```diff
@@ -19,30 +19,57 @@
...
         content = fid.read()
     # change from Windows format to UNIX for uniformity
     content = content.replace(b'\r\n', b'\n')
 
     try:
         node = ast.parse(content)
+        return node, content.decode('utf-8')
     except SyntaxError:
-        return SYNTAX_ERROR_DOCSTRING, content.decode('utf-8'), 1
+        return SyntaxError, content.decode('utf-8')
```
lesteve: This feels a bit kludgy to return an exception class rather than a node. Why don't you SyntaxError through and catch it in `get_docstring_and_rest` and have an except clause at the `parse_file_source` call?

sphinx_gallery/tests/test_backreferences.py
```diff
@@ -89,4 +78,30 @@ def test_identify_names():
...
+from a.b import c
+import d as e
+print(c)
+e.HelloWorld().f.g
+"""
+    expected = {'c': {'name': 'c', 'module': 'a.b', 'module_short': 'a.b'},
+                'e.HelloWorld': {'name': 'HelloWorld', 'module': 'd', 'module_short': 'd'}}
+
+    import locale
+    print(locale.getlocale())
```
lesteve: Looks like debugging print statements that you forgot to remove.

sphinx_gallery/tests/test_backreferences.py
```diff
@@ -89,4 +78,30 @@ def test_identify_names():
...
+
+    with tempfile.NamedTemporaryFile('wb', delete=False) as f:
```
lesteve: Using pytest `tmpdir` fixture is a lot more neat for this kind of things.

sphinx_gallery/py_source_parser.py
```diff
@@ -19,30 +19,57 @@
 """
 
 
-def get_docstring_and_rest(filename):
-    """Separate `filename` content between docstring and the rest
+def parse_file_source(filename):
```
lesteve: Maybe `parse_python_file` or `parse_source_file`?

sphinx_gallery/py_source_parser.py
```diff
@@ -19,30 +19,57 @@
...
     try:
         node = ast.parse(content)
+        return node, content.decode('utf-8')
     except SyntaxError:
-        return SYNTAX_ERROR_DOCSTRING, content.decode('utf-8'), 1
+        return SyntaxError, content.decode('utf-8')
+
+
+def get_docstring_and_rest(filename):
+    """Separate `filename` content between docstring and the rest
```
lesteve: I think you need to use double backticks if you want this to be fixed-width in the HTML reference documentation. Same thing for a few single backticks that you use in a few different places.

sphinx_gallery/tests/test_py_source_parser.py
```diff
@@ -0,0 +1,20 @@
...
+# License: 3-clause BSD
+
+from __future__ import division, absolute_import, print_function
+import sphinx_gallery.py_source_parser as sg
+
+
+def test_get_docstring_and_rest():
+
+    docstring, rest, lineno = sg.get_docstring_and_rest(
+        'sphinx_gallery/tests/unicode.sample')
```
lesteve: Can you write the content into a temporary file instead using `tmpdir`? This way the test is a bit easier to understand.

lesteve:
@Titan-C I am looking at this, please don't pull the trigger yet ;-) !

sphinx_gallery/py_source_parser.py
Titan-C: I agree is not the nicest way. I ended up using this, because I need this function in few more places and I rely those places always having access to the decoded content and to know if the parsing was successful or not.

lesteve: I saw that but I would rather let the SyntaxError go through and catch it where you call `get_docstring_and_rest`. Arguably you could have a SyntaxError somewhere else in your code but I don't think this is that bad. If we care about that we could catch the SyntaxError in `parse_source_file` and raise a custom exception e.g. ParsingError (or a better name) then catch ParsingError where parse_source_file` is called.

Titan-C: This is on the boundaries of my knowledge. But if I raise the exception how do I get access to the returned value of content?

lesteve: Just in case I am missing and the exception approach does not work, can you not return None instead?

sphinx_gallery/tests/test_py_source_parser.py
Titan-C: Not really. If you check on the previous file there is `test_identify_names2` there I indeed write to file. And for everything to work correctly in py2 & py3 I write it as bytes. But in this case and the function `test_identify_names` I really want to read a file from disk, the file is identical. But is the 2 places I need to be sure the file read from disk is correctly parsed.

sphinx_gallery/tests/test_backreferences.py
Titan-C: I just did this of using the fixture, but all tests jobs in travis (except 1) failed

sphinx_gallery/tests/test_py_source_parser.py
lesteve: It seems like this does work:

```py

content = b'\n'.join(
    [b'',
     b'# -*- coding: utf-8 -*-',
     b"'''",
     b'\xc3\x9anicode in header',
     b'=================',
     b'',
     b'U\xc3\xb1icode in description',
     b"'''",
     b'',
     b'# Code source: \xc3\x93scar N\xc3\xa1jera',
     b'# License: BSD 3 clause',
     b'',
     b'import os',
     b"path = os.path.join('a','b')",
     b'',
     b"a = 'hei\xc3\x9f'  # Unicode string",
     b'',
     b'import sphinx_gallery.back_references as br',
     b'br.identify_names',
     b'',
     b'from sphinx_gallery.back_references import identify_names',
     b'identify_names ',
     b''])

open('/tmp/unicode.sample', 'wb').write(content)

# use tmpdir to get a file here instead
open('/tmp/unicode.sample', 'wb').write(content)
```

sphinx_gallery/py_source_parser.py
Titan-C: Of course. I can return None

sphinx_gallery/tests/test_backreferences.py
Titan-C: Any info on this. It's recent commit https://github.com/sphinx-gallery/sphinx-gallery/pull/311/commits/f2f2d04e4fd7401b1ec1ffe6833716505c65c166

sphinx_gallery/tests/test_py_source_parser.py
Titan-C: Are you sure this is simpler than having a unicode heavy python file on disk? Also, I use this file twice, one to test that despite the unicode I `get_docstring_and_rest` and one to test that despite the unicode I `identify_names` of the used functions.

lesteve: > Are you sure this is simpler than having a unicode heavy python file on disk?

Yes, this is a lot more clear because you see exactly what you are writing to the file. Otherwise how do we know how you generated your file. Also test files like this don't get included in the installed package which mean that you can not run the tests on an installed package to make sure that they run. For example this is a problem we already have for sphinx_gallery/tests/reference_parse.txt.

If you use the file content in two tests you can put that in a helper function.

Titan-C: OK then. I implemented a pytest fixture to generate this sample file

Titan-C:
This is again ready for review. 
Travis on ubuntu fails because the system version of six is too old and pytest coverage refuses to load. But this is a particular situation because
`python setup.py test` passes as it installs *for the test* a resent version of six
`pytest sphinx_gallery` fails because it uses the old system version.
Is everyone fine if I pip install a more resent version of six in this test environment? Other ideas?

lesteve:
I saw that a few days ago (basically we have a failure on master at the moment because of this). I just opened #325 to fix this.

I guess (I did not check in details) that this is one of the reason that @choldgraf had troubles in #244. Installing sphinx 1.5 probably pulls in a more recent version of six ...

lesteve:
I added an entry in CHANGES.rst, @Titan-C you probably understand the problem more than me, so feel free to improve/fix it!

Other than this, I think this is ready to be merged, great work on a nasty long-standing issue!

lesteve:
Merging this one, thanks a lot @Titan-C!

ImportanceOfBeingErnest:
When will a version containing this fix be released?    
(Building the matplotlib docs on windows currently blocks on this.)

choldgraf:
Relatively soon, I believe. We are spot-checking the remaining issues before a release. @lesteve just opened a PR for one and I'll plan to hit the others later this week or next. As always, PRs are welcome ;-)

jni:
I just want to note that this PR also fixed an unclosed file warning when building docs:

```pytb
/home/travis/venv/lib/python3.4/site-packages/sphinx_gallery/backreferences.py:144:
ResourceWarning: unclosed file <_io.TextIOWrapper name='/home/travis/build/scikit-image/scikit-image/doc/source/auto_examples/color_exposure/plot_ihc_color_separation.py' mode='r' encoding='UTF-8'>
```

See e.g. https://travis-ci.org/scikit-image/scikit-image/jobs/372237529#L6134

I searched issues and PRs for ResourceWarning and couldn't find anything, so I presumed it was unfixed and *almost* raised a new issue. =)

Thanks for the great work!

jni:
[This](https://github.com/sphinx-gallery/sphinx-gallery/pull/311/files#diff-33215ef937b7c3d86fe0a23425c585e4L144) was the offending line, now properly context-managed elsewhere:

```python
example_code_obj = identify_names(open(example_file).read())
```

PR #312, opened 2017-10-31
Don't install latex recommended in Circle CI

Titan-C:
Installing the texlive distribution in Circle CI, takes some time because Ubuntu by default installs some recommended packages. This PR avoids installing those recommended packages, which reduces the download time and install time. Build time drops by 1:30 mins


larsoner:
Nice, +1 for merge from me (the one who IIRC introduced this timing regression)

PR #324, opened 2017-12-14
[WIP/Experimental] Capturing objects. Bokeh test case

Titan-C:
I gave it a try on how to capture objects. This takes on the work proposed PR#313

I would also like to have PR #311 dealing with encodings, and PR #239 which is only some refactoring of the functions in gen_rst dealing with the rst files generation merged first.

I base this PR on the idea that the object capturer shall return directly the rst representation for direct embedding in the HTML.
I have seen that a big problem with capturing arbitrary objects is that, unlike matplotlib or mayavi there is no state machine to know what is plotted. Thus scrapers of png images where proposed. In this PR I propose to allow the capture function to access the state dictionary of the executed example and the source code of the block. Then I allow myself to specify as a comment in the source code what has to be captured.


CHANGES.rst
```diff
@@ -8,7 +8,7 @@ Bug Fixes
 '''''''''
 
 * Fix __future__ imports across cells. See `#308
-  <https://github.com/sphinx-gallery/sphinx-gallery/pull/308>` for more details.
+  <https://github.com/sphinx-gallery/sphinx-gallery/pull/308>`_ for more details.
```
lesteve: :disappointed: the syntax for rst link is something I will never get right on the first attempt I am afraid. I have fixed that in master.

Titan-C:
This is working on my machine. Getting the dependencies installed correctly in CI is a kind of a pain. But you can test on your machine.

Titan-C:
Dependencies on Circle CI are good. Don't understand why it doesn't work. I'll try other day.

Titan-C:
Figured out the bugs. Here it is 
https://275-26191147-gh.circle-artifacts.com/0/html/auto_examples/plot_bokeh.html#sphx-glr-auto-examples-plot-bokeh-py

larsoner:
I think most of the necessary functionality now exists in `master` so I'll close, feel free to reopen if you want to come back to this. Though the rebase alone will be a nightmare

PR #325, opened 2017-12-15
[MRG] Require six>=1.10 in ubuntu build

lesteve:
to avoid pytest_cov PluginValidationError.


lesteve:
CIs are green, I am going to merge this one.

PR #326, opened 2017-12-16
fix binder doc configuration

choldgraf:
@Titan-C

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -344,7 +344,7 @@ sphinx_gallery_conf = {
     'binder': {'org': 'sphinx-gallery',
                'repo': 'sphinx-gallery.github.io',
                'url': 'https://mybinder.org',
-               'branch': 'gh-pages',
+               'branch': 'master',
                'dependencies': './binder/requirements.txt'
                }
 }


choldgraf:
@titan-c did this fix the problem? The website seems to be the same but not sure if this is deployed on rtd now

Titan-C:
Yes this fixed the problem. The new website is in the github pages and
it works. But, you are right somehow the readthedocs website is not
updated. I'll look at it now.

Titan-C:
No Idea what is going on with read the docs. The build version does correspond to this merged PR. The summary of the build even shows the binder conf is to the master branch. I have no idea why the links still point to the gh-pages. I'll try to see if I can generate a clean build in read the docs.

Titan-C:
The rebuild docs are working. This one more reason to stop using read the docs. I'll prepare a PR to change the read the docs links into github links. This should help migrating our traffic and google page rank.

lesteve:
> The rebuild docs are working. This one more reason to stop using read the docs. I'll prepare a PR to change the read the docs links into github links. This should help migrating our traffic and google page rank.

Something readthedocs is good at is making it clear which version the docs are for. It would be great to keep the docs for a few versions or at least for the last release + master. You could look at how scikit-learn is doing it but there may be another way of doing it. doctr for example seem to be able to do that: https://github.com/drdoctr/doctr/blob/master/docs/recipes.rst#deploy-docs-from-git-tags

choldgraf:
I believe this is how matplotlib does it:

https://github.com/matplotlib/matplotlib.github.com

PR #327, opened 2017-12-17
Migrate documentation links to github pages

Titan-C:
Migration towards hosting the documentation in github pages has some advantages over readthedocs.
- We control the build environment, in circle CI. 
  - Can build the mayavi examples
  - Can test the generated docs before they are published
- Binder can open github repositories not readthedocs pages

Some disadvantages
- As for current configuration. We loose readthedocs feature of displaying latest version from master and latest stable version docs
- A lot of our traffic and page rank is on readthedocs(Changing this links will help evolve this)


lesteve:
Putting my comment from https://github.com/sphinx-gallery/sphinx-gallery/pull/326#issuecomment-352340820

> Something readthedocs is good at is making it clear which version the docs are for. It would be great to keep the docs for a few versions or at least for the last release + master. You could look at how scikit-learn is doing it but there may be another way of doing it. doctr for example seem to be able to do that: https://github.com/drdoctr/doctr/blob/master/docs/recipes.rst#deploy-docs-from-git-tags

choldgraf:
another note: if the plan is to use gh-pages, it's worth looking into `doctr`: https://github.com/drdoctr/doctr

larsoner:
We are now on gh-pages, closing

PR #328, opened 2017-12-18
[MRG] Add codecov support

lesteve:
Some maintenance tasks while I was at it:
- move part of .travis.yml to separate scripts
- use miniconda3 rather than miniconda2 installer
- use INSTALL_MAYAVI env variable
- remove python nightly build (it's been failing for a while and has been ignored because it is in allowed_failure mode). I don't think it makes sense to test on python 3.7 on each PR I would say anyway. Something more useful IMHO would be to have a cron build testing sphinx master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ matrix:
             - python-pandas
     # This environment tests is for mayavi
     - os: linux
-      env: DISTRIB="conda" PYTHON_VERSION="2.7"
+      env: DISTRIB="conda" PYTHON_VERSION="2.7" INSTALL_MAYAVI="true"
     # This environment tests the Python 3 versions
     - os: linux
       env: DISTRIB="conda" PYTHON_VERSION="3.4"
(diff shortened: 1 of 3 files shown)


lesteve:
@Titan-C do you want to have a quick look at this one?

Titan-C:
Packing to leave my apartment. I'll look in a few days

Titan-C:
The changes look fine. Moving from travis.yml to a script is cleaner.
I agree on not testing on the nightly version of python for the moment.
I don't get the codecov report. Pytest-cov also gives a report on coverage, that one is easy to understand.

lesteve:
> I don't get the codecov report. Pytest-cov also gives a report on coverage, that one is easy to understand.

The codecov comment is not always useful I agree. It is more useful once this is merged into master because it then does some kind of diff between master and the PR.

I generally look at things like per-file report https://codecov.io/gh/sphinx-gallery/sphinx-gallery/pull/328/tree?path=sphinx_gallery which is similar to the pytest-cov one.

There is also a very useful codecov browser extension that show you the line in the PR diff that are not covered by any tests. Note: I don't think the codecov extension has been extended for Firefox 57 yet. I am watching https://github.com/codecov/browser-extension/issues/44 and hoping this happens not too far in the future.

One last thing is that there is a codecov status that is red if your absolute coverage or the coverage of your diff is no high enough. The latter is quite useful because it tends to catch our laziness when we forget to add test for some edge cases.

choldgraf:
I'm +1 on this, but won't merge since it seems @Titan-C has some concerns

lesteve:
I am going to merge this one, it's easy to revert or tweak (e.g. turning codecov comments off comes to mind) if needed.

Slightly related: I read about [Optimistic Merging](http://hintjens.com/blog:106) recently. Probably not applicable in practice but thought provoking still.

choldgraf:
+1 on optimistic merging :-)

PR #329, opened 2017-12-19
[MRG] Align binder badge with other download buttons

lesteve:
The binder badge stands a bit on its own and I feel it should be on the same footing as the other download button (download .py + download .ipynb).

About the CSS part, I copied and pasted sphx-glr-download. The thing is that there the rule `div.sphx-glr-download a` was coloring the background around the binder badge and it did not look that great. Any improvements, let me know.

I'll paste links or snapshots once Circle finishes on this PR.

### DIFF
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -108,6 +108,12 @@ div.sphx-glr-footer {
     text-align: center;
 }
 
+div.binder-badge {
+  display: inline-block;
+  margin: 1em auto 1ex 2ex;
+  vertical-align: middle;
+}
+
 div.sphx-glr-download {
   display: inline-block;
   margin: 1em auto 1ex 2ex;
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -109,14 +109,12 @@ div.sphx-glr-footer {
 }
 
 div.binder-badge {
-  display: inline-block;
-  margin: 1em auto 1ex 2ex;
+  margin: 1em auto;
   vertical-align: middle;
 }
 
 div.sphx-glr-download {
-  display: inline-block;
-  margin: 1em auto 1ex 2ex;
+  margin: 1em auto;
   vertical-align: middle;
 }
 
@@ -127,36 +125,10 @@ div.sphx-glr-download a {
   border: 1px solid #c2c22d;
   color: #000;
   display: inline-block;
-  /* Not valid in old browser, hence we keep the line above to override */
-  display: table-caption;
   font-weight: bold;
   padding: 1ex;
   text-align: center;
 }
-
-/* The last child of a download button is the file name */
-div.sphx-glr-download a span:last-child {
-    font-size: smaller;
-}
-
-@media (min-width: 20em) {
-    div.sphx-glr-download a {
-	min-width: 10em;
-    }
-}
-
-@media (min-width: 30em) {
-    div.sphx-glr-download a {
-	min-width: 13em;
-    }
-}
-
-@media (min-width: 40em) {
-    div.sphx-glr-download a {
-	min-width: 16em;
-    }
-}
-
 
 div.sphx-glr-download code.download {
   display: inline-block;
--- a/sphinx_gallery/binder.py
+++ b/sphinx_gallery/binder.py
@@ -82,10 +82,12 @@ def gen_binder_rst(fname, binder_conf):
         The reStructuredText for the Binder badge that links to this file.
     """
     binder_url = gen_binder_url(fname, binder_conf)
-    rst = (".. figure:: https://mybinder.org/badge.svg\n"
-           "      :target: {}\n").format(
-        binder_url)
-    rst += "      :width: 150 px\n      :figclass: binder-badge\n\n"
+    rst = (
+        "\n"
+        "  .. container:: binder-badge\n\n"
+        "    .. image:: https://mybinder.org/badge.svg\n"
+        "      :target: {}\n"
+        "      :width: 150 px\n").format(binder_url)
     return rst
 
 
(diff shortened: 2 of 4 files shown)


lesteve:
[This PR](https://321-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/rtd_html/auto_examples/plot_seaborn.html)

[master](https://sphinx-gallery.github.io/auto_examples/plot_seaborn.html)

lesteve:
Rebased to fix a conflict and screenshots added:

### Master
master

### This PR
pr

lesteve:
@choldgraf any objection about this visual tweak?

choldgraf:
I am definitely +1 on this PR, though in the link I can't get all three to line up since the buttons don't scale w/ the size of the page:

image

I think this problem is less-evident when there are 2 buttons rather than 3.

That said, it sounds like this issue is more one of dynamic scaling rather than one of adding in the binder button, so probably shouldn't be blocking on this PR, no? What do you think?

lesteve:
> That said, it sounds like this issue is more one of dynamic scaling rather than one of adding in the binder button, so probably shouldn't be blocking on this PR, no? What do you think?

I am not an expert on web stuff but I think the behaviour when you adjust the screen size seems fine.

choldgraf:
I guess I'm not making a value judgment on that behavior, just noting that on my 14" laptop screen (probably well within the expected browser sizes of our users) the above image is what they'd see.

lesteve:
> I guess I'm not making a value judgment on that behavior, just noting that on my 14" laptop screen (probably well within the expected browser sizes of our users) the above image is what comes up.

Interesting I can reduce the window to about half the size on my screen (14" laptop screen too) and still see the three buttons on the same row.

In any case I think that even in this case (three buttons not fitting on the screen width) having the three buttons on the same footing makes it look better than before.

choldgraf:
huh, that is weird!  well, either way I like the change, so I'll merge in later today unless other folks chime in w/ dissenting opinions!

GaelVaroquaux:
On Firefox I can reproduce the awkward layout.

GaelVaroquaux:
The right way to do the layout would be either:

* a 3 column layout switching to a 1 column layout when the width isn't
  large enough for three columns

* to put the binder badge on it's own row if the width isn't large enough

choldgraf:
> * a 3 column layout switching to a 1 column layout when the width isn't large enough for three columns

makes sense to me. though I am unsure as to the magical CSS incantation to make this happen

GaelVaroquaux:
You need the "@media" selector. I am in a meeting right now. Can't code.

choldgraf:
hadn't seen that before, very cool!

https://www.w3schools.com/cssref/tryit.asp?filename=trycss3_media_example1

The question is: at what point does the strange stacking happen. I suppose this depends on a combination of resolution + width, but it's weird that @lesteve and I have the same size monitor but 2 different behaviors

larsoner:
If you are on OSX with HiDPI and they are not (standard Linux or Windows,
or non-Retina OSX) this can happen. Or maybe there is some browser
difference.

lesteve:
Hmmm I can actually reproduce the behaviour from https://github.com/sphinx-gallery/sphinx-gallery/pull/329#issuecomment-353549937 on Chromium, so it looks like there is something fishy going on.

Titan-C:
I want to add that we should use the `https` address for the binder badge. I have firefox57 and it does not load the image because the website is https(in github-pages and circle-ci), thus it blocks loading http resources.

choldgraf:
Should this PR be blocked on the CSS three-column to one-column transition problem?

Also I agree we should use HTTPS for the badge...I'll open a PR for this, good spot @Titan-C

lesteve:
I arranged the download buttons and the binder badge vertically. I feel this is a lot simpler and more robust than trying to put them on the same row and making it work with media queries depending on the screen width (The sphinx read the docs theme already has some media query that decides to display the side-bar which further complicates matters. Other themes may have similar quirks as well). 

This is how it looks:
screenshot-2018-1-2 seaborn example sphinx-gallery 0 1 13-git documentation.

choldgraf:
I'm fine w/ this as well

lesteve:
@Titan-C and/or @GaelVaroquaux any objection to this one?

Titan-C:
I'm fine with this. I actually prefer this stacking of the buttons, it really makes the layout more uniform. +1

lesteve:
OK three people agree, so I am going to merge this one. I guess this can always be modified with some custom CSS if needed.

GaelVaroquaux:
In general, +1, though on wide screens it creates a bit of a waste of screen real estate.

The best design would be to have it floating on the side for wide pages, something like (in addition and separate from the other CSS instructions to sphx-glr-footer):

```
div.sphx-glr-footer @media only screen and (min-width: 100ex) {
    position: fixed;
    right: 10px;
    top: 80px;
    background: #f2f2f2;
    padding: 1ex;
    border: solid 1px #ccc;
}
```

I put the 100ex in as a ballpark estimate, maybe something larger is required to avoid overlap with text on too narrow screens. This can be tuned by seeing what happens when the window of the browser is resized.

The benefit is that it would make the binder and download buttons much more visible. Right now, my experience is that users often miss them.

lesteve:
> Right now, my experience is that users often miss them.

I agree with this, a notebook-like example like [this](http://scikit-learn.org/stable/auto_examples/plot_compare_reduction.html) encourages users to copy and paste cell by cell from the HTML rather than go to the bottom of the page. This may lead to confusion as in https://github.com/scikit-learn/scikit-learn/pull/10394, where the `plt.show` is right at the end of the example.

IMO it would be nice to put:
* either a link before the example than goes to the download section at the bottom
* the download buttons before the example. Probably this this takes too much space, so better suggestions welcome!

GaelVaroquaux:
Cooking up a PR right now

PR #331, opened 2017-12-19
refactored py to ipynb

### LINKED ISSUE #322: name string_replace trips on projects with ".py" in path
https://github.com/sphinx-gallery/sphinx-gallery/blob/fb2da9af8c3c9282150b26c356f97fdddfcefb82/sphinx_gallery/notebook.py#L193

This code will trip on projects that have '.py' as a substring.  I found this because I am working on a project called "aibs.pyproject_template".  This will need a more exclusive replacement string.

nicain:
Resolves #322 

All tests passing in py27, py35, and py36

testing updates and bug fix done in two separate commits; if you don't want tox additions, I can just cherry-pick the bug fix commit (bd211e55f9b78ec64d38f25ea2d1ad63ee4d4719) and reissue the PR


nicain:
Also all tests passing locally on my dev environment for py27 py35 and py36 via tox

lesteve:
Thanks a lot for your PR! I remove the controversial changes and pushed a commit with additional minor tweaks.

Can you add a test in `tests/test_utils.py`, testing both the normal condition and the exception + exception message? Can you also add an entry in CHANGES.rst?

I haven't jumped on the tox bandwagon yet although I have heard it is quite good. Minimal googling seems to indicate that conda and tox do not play nice together, oh well this will have to wait then ...

My 2 cents:
* short and focussed PRs tend to be merged more quickly. Avoid doing two unrelated things in a single PR.
* controversial changes even if minor tend to make the review less pleasant for everyone involved (bikeshedding + arguing).

sphinx_gallery/utils.py
```diff
@@ -36,3 +36,15 @@ def __init__(self):  # noqa: D102
 
     def __del__(self):  # noqa: D105
         rmtree(self._path, ignore_errors=True)
+
+
+def replace_py_ipynb(fname):
+    """Replace .py extension in filename by .ipynb"""
+    fname_prefix, extension = fname.rsplit('.', 1)
+    if extension != 'py':
+        raise RuntimeError(
```
lesteve: Thinking about it maybe ValueError is better here

nicain:
@lesteve 

> Can you add a test in tests/test_utils.py, testing both the normal condition and the exception + exception message? Can you also add an entry in CHANGES.rst?
Done.

> I haven't jumped on the tox bandwagon yet although I have heard it is quite good. Minimal googling seems to indicate that conda and tox do not play nice together, oh well this will have to wait then ...

I just learned to use tox myself, and I love it.  I use conda as well, but I can across a handy trick to use tox in a conda development env: create bare conda envs for the python versions you need tox to run against, and add them as symlinks to your path (for example ln -s /foo/envs/toxpy35/bin/python ./python 3.5, executed from ~/local/bin).

Ill stash the tox changes in a branch on my fork, in case you ever want to circle back to them.

sphinx_gallery/tests/test_utils.py
```diff
@@ -0,0 +1,26 @@
...
+        assert utils.replace_py_ipynb(file_name+'.py') == file_name+'.ipynb'
+
+def test_replace_py_ipynb_improper_input():
+    
+    file_name = 'some/file/name'
+
+    with pytest.raises(ValueError) as expected_exception:
+        utils.replace_py_ipynb(file_name+'.txt')
+
+    assert 'Unrecognized file extension' in str(expected_exception.value)
```
choldgraf: could you add a final line at the end?

sphinx_gallery/utils.py
```diff
@@ -36,3 +36,15 @@ def __init__(self):  # noqa: D102
 
     def __del__(self):  # noqa: D105
         rmtree(self._path, ignore_errors=True)
+
+
+def replace_py_ipynb(fname):
+    """Replace .py extension in filename by .ipynb"""
+    fname_prefix, extension = fname.rsplit('.', 1)
```
choldgraf: how about `os.path.splitext` instead: https://docs.python.org/2/library/os.path.html#os.path.splitext

choldgraf: (it seems like the functionality would be the same though...so I don't think it's blocking this PR)

sphinx_gallery/tests/test_utils.py
```diff
@@ -0,0 +1,26 @@
...
+
+from __future__ import division, absolute_import, print_function
+import sphinx_gallery.utils as utils
+import pytest
+
+def test_replace_py_ipynb():
+    for file_name in ['some/file/name', '/corner.pycase']:
+        assert utils.replace_py_ipynb(file_name+'.py') == file_name+'.ipynb'
+
+def test_replace_py_ipynb_improper_input():
```
choldgraf: I'd merge these into a single `test_replace_py_ipynb()` function

choldgraf:
Just a couple of quick thoughts from me - nice catch and thanks for the PR @nicain ! :-)

sphinx_gallery/tests/test_utils.py
nicain: Done.

sphinx_gallery/utils.py
nicain: Done.

sphinx_gallery/tests/test_utils.py
nicain: Done.

nicain:
Updated and ready for re-review.

choldgraf:
Once tests are passing I'm +1 on this. Will hold off on merging until @lesteve can confirm he's happy too!

Thanks @nicain !

sphinx_gallery/utils.py
```diff
@@ -36,3 +37,16 @@ def __init__(self):  # noqa: D102
 
     def __del__(self):  # noqa: D105
         rmtree(self._path, ignore_errors=True)
+
+
+def replace_py_ipynb(fname):
```
Titan-C: Isn't it just simpler to have a regex substitution
```python
re.sub('.py$', '.ipynb', fname)
```
We have already filtered for python files when collecting files for examples. No need for the value error check.

nicain: I like the regex, implemented in commit below.  But a little extra early-checking seems safe.  If you want me to rip it out, I can do that too.

lesteve:
LGTM, thanks a lot @nicain!

sphinx_gallery/tests/test_utils.py
```diff
@@ -0,0 +1,25 @@
...
+# License: 3-clause BSD
+
+from __future__ import division, absolute_import, print_function
+import sphinx_gallery.utils as utils
+import pytest
+
+def test_replace_py_ipynb():
+    
+    # Test behavior of function with expected input:
+    for file_name in ['some/file/name', '/corner.pycase']:
```
lesteve: Very nice that you added a test for the corner case you bumped into by the way!

choldgraf:
woo! way to go @nicain ! 🎉

nicain:
@lesteve Any idea when the next patch revision number bump will occur, and be released to pypi?  I have a project that would like SG to pip-install as a dependency, but need this fix for correct behavior.  I Just need a rough timeline so that I can plan accordingly :)

lesteve:
Not really, you can install sphinx-gallery from master in the mean time if you are willing to take the (small) risk:
```
pip install git+https://github.com/sphinx-gallery/sphinx-gallery
```

PR #332, opened 2017-12-23
https badge

choldgraf:
cc @Titan-C


Titan-C:
This is good. But I just noticed another reason for the image not to show is that I use privacy badger. It tells me mybinder is tracking me and thus blocks the domain.
I'll merge this, https is better anyway

PR #336, opened 2018-01-08
[MRG] CSS: float the download buttons on wide screens

GaelVaroquaux:
To make them more visible to users for download

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,11 @@ New features
   config. Note that this API may change in the future. `#244
   <https://github.com/sphinx-gallery/sphinx-gallery/pull/244>`_
 
+* **CSS** Download and binder buttons are now on the right side for large
+  screens. This might change downstream CSS layouts.
+  "div.sphx-glr-footer-example" can be used to capture these specific
+  case in project-specific CSS.
+
 Bug Fixes
 '''''''''
 
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -129,7 +129,7 @@ div.sphx-glr-footer {
 }
 
 /* Special case the RTD theme, capturing "section.wy-nav-content" */
-@media only screen and (min-width: 150ex) {
+@media only screen and (min-width: 160ex) {
     section.wy-nav-content div.sphx-glr-footer-example  {
 	right: unset;
 	left: 140ex;
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -115,9 +115,9 @@ div.sphx-glr-footer {
 	right: 5px;
 	top: 100px;
 	background: rgba(200, 200, 200, 0.3);
-	padding: 0ex 1ex;
+	padding: 0ex .5ex;
 	border: solid 1px #ccc;
-	max-width: 25ex;
+	max-width: 24ex;
 	width: fit-content;
 	width: -moz-min-content;
     }
@@ -125,23 +125,30 @@ div.sphx-glr-footer {
     /* Avoid overflow of long file names */
     div.sphx-glr-footer-example code {
 	max-width: 100%;
+	overflow: hidden;
     }
 
     div.sphx-glr-footer-example a {
 	max-width: 95%;
+	overflow: hidden;
+	padding: .5ex;
+    }
+
+    div.sphx-glr-footer-example a code span:last-child {
+	font-size: smaller;
     }
 }
 
 /* On very large screen, don't flush it too far */
-@media only screen and (min-width: 160ex) {
+@media only screen and (min-width: 190ex) {
     div.sphx-glr-footer-example  {
 	right: 5%;
-	max-width: 30ex;
+	max-width: 29ex;
     }
 }
 
 /* Special case the RTD theme, capturing "section.wy-nav-content" */
-@media only screen and (min-width: 170ex) {
+@media only screen and (min-width: 190ex) {
     section.wy-nav-content-wrap div.sphx-glr-footer-example  {
 	right: unset;
 	left: 150ex;
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -108,6 +108,19 @@ div.sphx-glr-footer {
     text-align: center;
 }
 
+/* On large screens, float the download and binder buttons on the right */
+@media only screen and (min-width: 130ex) {
+    div.sphx-glr-footer  {
+	position: fixed;
+	right: 5px;
+	top: 100px;
+	background: #f2f2f2;
+	padding: 1ex;
+	border: solid 1px #ccc;
+	max-width: 15%;
+    }
+}
+
 div.binder-badge {
   margin: 1em auto;
   vertical-align: middle;
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -109,7 +109,7 @@ div.sphx-glr-footer {
 }
 
 /* On large screens, float the download and binder buttons on the right */
-@media only screen and (min-width: 130ex) {
+@media only screen and (min-width: 140ex) {
     div.sphx-glr-footer-example  {
 	position: fixed;
 	right: 5px;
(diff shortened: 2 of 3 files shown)


GaelVaroquaux:
Continuation of discussion on #329

GaelVaroquaux:
Links to the rendered pages:
* https://336-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/alabaster_html/auto_examples/plot_colors.html
* https://336-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/rtd_html/auto_examples/plot_colors.html

And screenshot with default sphinx theme:
sphx-glr

I think that this is a usability improvement compared to what we have.

lesteve:
Nice! The floating right is a bit weird on the readthedocs theme (which is the one we used in sphinx-gallery.github.io). From [this example](https://336-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/rtd_html/auto_examples/plot_exp.html#sphx-glr-auto-examples-plot-exp-py):

screenshot-2018-1-8 plotting the exponential function sphinx-gallery 0 1 13-git documentation

GaelVaroquaux:
I don't think that it's that bad. What bother's you? That's it's too far on the right?

lesteve:
> That's it's too far on the right?

Yeah I was wondering if there is an easy way to make it float just a bit right of the example code. If there is no easy way to do that, it's fine with me.

GaelVaroquaux:
> Yeah I was wondering if there is an easy way to make it float just a bit right
> of the example code. If there is no easy way to do that, it's fine with me.

Trying something...

GaelVaroquaux:
OK, I've tried something that should look better (and caught a design bug
in the same time :D).

Tell me how you like it

larsoner:
A bit too late to look at this, but for MNE before #329 we had:

screenshot from 2018-01-08 12-20-29

#329 it broke a bit:

screenshot from 2018-01-08 12-22-12

After this PR (e87efda) it looks just like the above until I make the screen wider, then I get:

screenshot from 2018-01-08 12-28-12

And wider still:

screenshot from 2018-01-08 12-28-26

These problems can be fixed, but it is worth knowing that CSS tweaking can break existing use cases a bit. By customizing our CSS in MNE perhaps we should even expect that upstream changes can do this (maybe we should put it in the docs somewhere if it's not already there?).

This seems to interact nicely when building the SG docs, though -- and thus probably should anywhere CSS is not customized -- so +1 for merge from me.

GaelVaroquaux:
Fixed for RTD theme:
https://341-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/rtd_html/auto_examples/plot_exp.html

GaelVaroquaux:
@larsoner : +1 adding a note for saying that customizing CSS is a bit dangerous (it will always be, CSS is unfortunately brittle). Where do you suggest that we add it?

The fix on your side will be easy. In addition, it the latest commit, I've added a new class that you can capture in CSS.

GaelVaroquaux:
I've added a note in CHANGES

larsoner:
Also here is a possibly different sort of problem with the float -- it can sit on top of the content. It's clearest in MNE:

screenshot from 2018-01-08 12-35-25

But even happens at some widths with the SG builds:

screenshot from 2018-01-08 12-36-35

As for the doc, I'd put it in the "And some things can be tweaked directly in CSS:" here:

http://sphinx-gallery.readthedocs.io/en/latest/advanced_configuration.html

larsoner:
(I'm still +1 for merge even with this overlap issue, I read that "optimistic merging" article and I figure someone can fix it in a follow-up PR if it's truly irritating)

GaelVaroquaux:
The overlap cannot be fixed. If you think that it is an important problem, we can change the media query so that the behavior switches for larger screens. If people agree that the overlap is a problem they should let me know, and I'll change this behavior.

lesteve:
FWIW in Firefox 57 I see an overlap with the text for readthedocs theme in https://341-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/rtd_html/auto_examples/plot_exp.html even in full-screen window (even if there is plenty of space on the right)

screenshot-2018-1-8 plotting the exponential function sphinx-gallery 0 1 13-git documentation 1

Titan-C:
@lesteve  I'm on Firefox 57 and I don't see such overlap on a wide screen window. This PR looks nice enough. 
As a side opinion. What if we put this buttons at the head of the document instead of the footer and save ourselves the pain of having a floating element?

GaelVaroquaux:
I'd rather not have the buttons on the top, as they take a lot of
vertical space, and its a problem if users that click to see the example
have to scroll.

I am working on reducing a bit the overlap. Stay tuned.

larsoner:
> What if we put this buttons at the head of the document instead of the footer and save ourselves the pain of having a floating element?

This would be done via changing the ordering of the elements within the HTML template rather than tweaking the CSS, right? If so, -1 on this as it can't be reverted/fixed via CSS.

GaelVaroquaux:
OK, I worked to reduce the overlap. Here is the link to rendered versions:

* RTD theme: https://344-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/rtd_html/auto_examples/plot_choose_thumbnail.html
* Albany theme: https://344-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/alabaster_html/auto_examples/plot_choose_thumbnail.html

I hope that this is good to go. It can still generate some overlap, but minor, and I don't think that it is easy to do less.

larsoner:
Here is what I get for a few views:

screenshot from 2018-01-08 15-51-21

screenshot from 2018-01-08 15-48-15

This one gets a scroll bar , maybe Chrome is just being weird?

screenshot from 2018-01-08 15-48-19

I'm +0 on this change because it always makes it easier to download files or launch Binder, but depending on the resolution it can either be helpful or harmful for example readability.

GaelVaroquaux:
Darn the scrollbar is strange. Let me see if I can reproduce this and fix it.

GaelVaroquaux:
OK, new iteration, that should fix the bugs of the previous iteration and remove even more overlap:

https://345-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/rtd_html/auto_examples/plot_choose_thumbnail.html
https://345-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/alabaster_html/auto_examples/plot_choose_thumbnail.html

GaelVaroquaux:
And with a tiny extra improvement to please both @larsoner and @lesteve : 

https://346-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/rtd_html/auto_examples/plot_choose_thumbnail.html
https://346-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/alabaster_html/auto_examples/plot_choose_thumbnail.html

choldgraf:
I really like this addition, I think having the buttons float to the side definitely makes them more discoverable. +1 for an optimistic merge (though there have been quite a lot of iterations on this one already!) and spot-checking improvements as people have a chance to get to know the new behavior

larsoner:
LGTM +1 for merge

lesteve:
Merging, thanks a lot @GaelVaroquaux !

GaelVaroquaux:
Great! The iterations (ie y'all being demanding bastards like I am)
really helped, IMHO.

agramfort:
sorry to come late here but I feel this is not an ideal decision. it makes you use a full column for just a button -> it leads to a lot of white spaces.

It also forces you to use a wide screen

larsoner:
@agramfort the button actually floats to the right, so it should (hopefully!) only use already-unused white white space, and when you decrease resolution, it should automatically float to the bottom.

agramfort:
at least here
https://6793-1301584-gh.circle-artifacts.com/0/tmp/circle-artifacts.QIpzD23/html/auto_tutorials/plot_receptive_field.html
the button shows up on the title + is half hidden.

maybe it's an MNE pb though

larsoner:
Yes I get this behavior too at some resolutions:

screenshot from 2018-01-17 15-31-58

> maybe it's an MNE pb though

We do use a width that is larger than the SG build defaults. Not sure how common it will be elsewhere.

lesteve:
In full-width, it looks fine for me (Firefox 57 and Chrome). If I start making the window narrower I get something like @larsoner's snapshot. I don't know if there is a better way of doing it.
* can we float the buttons under the header or at the bottom of the window?
* lower tech solution, maybe we can add link before the title of the example that links to the download section at the bottom of the page?

Other suggestions more than welcome !

GaelVaroquaux:
Typography guidelines for readability say that very wide columns are a bad thing. You should limit the column width in the MNE CSS (eg using max-width).

IMHO: yes it's an MNE  problem that should be fixed there. 

larsoner:
In MNE our width comes from using `sphinx_boostrap_theme`, which sets the column width to 1170px when the width is >= 1200 px, i.e. similar to themes here:

https://bootswatch.com/

So I suspect anyone using `sphinx_boostrap_theme` in conjunction with `sphinx-gallery` will have this problem. 

In any case, at the MNE end we will either modify our column width, or tweak the floating box behavior to jump to the bottom sooner.

PR #337, opened 2018-01-09
[MRG]: MISC: CSS: fix predecence

GaelVaroquaux:
Fix a bug in my previous layout (minor one, but this fix reduces screen
real estate used by download buttons).

### DIFF
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -108,6 +108,45 @@ div.sphx-glr-footer {
     text-align: center;
 }
 
+div.binder-badge {
+  margin: 1em auto;
+  vertical-align: middle;
+}
+
+div.sphx-glr-download {
+  margin: 1em auto;
+  vertical-align: middle;
+}
+
+div.sphx-glr-download a {
+  background-color: #ffc;
+  background-image: linear-gradient(to bottom, #FFC, #d5d57e);
+  border-radius: 4px;
+  border: 1px solid #c2c22d;
+  color: #000;
+  display: inline-block;
+  font-weight: bold;
+  padding: 1ex;
+  text-align: center;
+}
+
+div.sphx-glr-download code.download {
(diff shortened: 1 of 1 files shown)


GaelVaroquaux:
To ease review, here are the rendered versions:
https://350-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/rtd_html/auto_examples/plot_choose_thumbnail.html
https://350-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/alabaster_html/auto_examples/plot_choose_thumbnail.html

I know, the difference is very small: it's a .5ex difference on the side of the buttons. But on both sides, 2 x .5ex = 1ex, and that's one extra character less of overlay (@lesteve got me picky about overlay)

lesteve:
:confused: have you changed something other than moving things around?

GaelVaroquaux:
> 😕 have you changed something other than moving things around?

No. But a rule setting the padding of the floating div to .5ex was
overridden because the "media" block was too high up. This is why I
moved it down. Hence the actual rendered CSS did change by 2 x .5ex when
the div is floating.

lesteve:
Just for completeness what I mean is: if you realign the diff then nothing has changed in your CSS ...

lesteve:
Ah ... some CSS subtleties that maybe one day I will learn. OK fair enough, merging, thanks a lot!

GaelVaroquaux:
> Just for completeness what I mean is: if you realign the diff then
> nothing has changed in your CSS ...

Yes, but order matters in CSS: the last rule wins. And given that rules
have different scoping applications, order is actually semantic, and
not a bug.

Titan-C:
> Yes, but order matters in CSS: the last rule wins. 

I was not aware of this. Maybe that's the reason it's always a pain to get CSS working. I usually end setting up a more specialized class or id. It never occurred to me that order in the file was important.

PR #345, opened 2018-02-09
MRG: Allow title only -> use title as first paragraph

### LINKED ISSUE #222: Make example text optional
I would be nice if the example text (in the example module docstring) was optional.  Of course, an example *title* is necessary e.g. to build the index (well, I guess it could always default to the filename), but requiring more text tends to lead to examples such as https://github.com/matplotlib/matplotlib/pull/8233/files, where the body of the text just paraphrases the title.

### LINKED ISSUE #232: SG incorrectly grabbing the description when a label is defined in docstring
I noticed that SG is incorrectly grabbing the description when you define a label in the docstring before the title. E.g.:

```
"""
.. my_label:

=======
Title
=======

etc etc.
"""
```

SG will correctly find title, but when it tries to pull the description it shows up as `=====Title=====` because it's splitting on the wrong line.

For example see my matplotlib (new) tutorials build here: http://predictablynoisy.com/matplotlib/tutorials/index.html

larsoner:
This makes an example description optional (title only is okay) by using the title as the description if it's not provided. The alternative is to just use `''`, but this seemed less useful given how the intro paragraph is used alter.

#232 was already fixed, this just adds an explicit test of it.

Closes #232.
Closes #222.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -208,11 +208,11 @@ def extract_intro_and_title(filename, docstring):
     # lstrip is just in case docstring has a '\n\n' at the beginning
     paragraphs = docstring.lstrip().split('\n\n')
     # remove comments and other syntax like `.. _link:`
-    paragraphs = [p for p in paragraphs if not p.startswith('.. ')]
-    if len(paragraphs) <= 1:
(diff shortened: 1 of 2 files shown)


choldgraf:
nice! will wait for travis to be happy :-)

PR #346, opened 2018-02-09
MRG: Add ignore_pattern

### LINKED ISSUE #262: Possible to ignore a file with sphinx-gallery?
I'm trying to get SG set up with a project that needs an `__init__.py` file inside of the examples folder. This is giving sphinx-gallery issues because if there's no docstring it will error, and if I add a docstring then I get an empty image in the gallery for the init file. Is there a way to tell SG to totally ignore a file (e.g., with a regex or something?)

larsoner:
Hopefully the doc changes are clear enough, but this allows skipping some files entirely (i.e., not parsing or adding them to the gallery).

Closes #262.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -434,9 +434,11 @@ def generate_dir_rst(src_dir, target_dir, gallery_conf, seen_backrefs):
     # get filenames
     listdir = [fname for fname in os.listdir(src_dir)
                if fname.endswith('.py')]
-    # limit which to look at based on regex
+    # limit which to look at based on regex (similar to filename_pattern)
     listdir = [fname for fname in listdir
-               if re.match(gallery_conf['ignore_pattern'], fname) is None]
+               if re.search(gallery_conf['ignore_pattern'],
+                            os.path.normpath(os.path.join(src_dir, fname)))
+               is None]
     # sort them
     sorted_listdir = sorted(
         listdir, key=gallery_conf['within_subsection_order'](src_dir))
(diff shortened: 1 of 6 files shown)


larsoner:
Readable doc update:

https://366-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/rtd_html/advanced_configuration.html#build-pattern

sphinx_gallery/gen_rst.py
```diff
@@ -439,8 +439,13 @@ def generate_dir_rst(src_dir, target_dir, gallery_conf, seen_backrefs):
 
     if not os.path.exists(target_dir):
         os.makedirs(target_dir)
+    # get filenames
     listdir = [fname for fname in os.listdir(src_dir)
                if fname.endswith('.py')]
+    # limit which to look at based on regex
+    listdir = [fname for fname in listdir
+               if re.match(gallery_conf['ignore_pattern'], fname) is None]
```
lesteve: As far as I remember, filename_pattern uses the full path. It would be great if ignore_pattern was using the full path as well.

For the same reason, I would use `re.search` rather than `re.match`.

sphinx_gallery/tests/tinybuild/examples/__init__.py
```diff
@@ -0,0 +1 @@
+from numpy import pad
```
lesteve: Maybe add a comment here rather than an unused import? Something like this for example?

```py
# This file is ignored through ignore_pattern so it does not need to have a
# docstring following the sphinx-gallery convention (i.e. title + description)
```

lesteve:
I saw the Python 3.4 failure before. I am trying to tackle it in #350.

lesteve:
Merging, thanks a lot @larsoner!

PR #347, opened 2018-02-09
MRG: Add refs from docstring to backrefs

### LINKED ISSUE #205: SG misses classes that aren't imported
UPDATED:

Right now it seems like methods of classes will only be linked with backrefs if the class is imported somehow (e.g., `import matplotlib.pyplot as plt; ax = plt.Axes()` would I think properly link to any methods in `ax` such as `ax.scatter`).

However, sometimes there are objects which have examples in the docs, but are not explicitly imported in those examples. E.g., `fig, ax = plt.subplots`). 

How should we handle these cases?

@Eric89GXL has proposed something like this:

> Maybe a compromise would be to parse the non-code for statements in the that contain Sphinx references, like:
```
  This example makes use of :class:`mne.SourceEstimate`.
```
> Then it's clean RST format, and should/could link nicely?

larsoner:
Adds to backreferences any function/class that gets `:func:` or `:meth:`'ed in the docstring. This should be particularly useful for things like `matplotlib.axes.Axes` and similar classes that get instantiated only indirectly.

Closes #205.

### DIFF
--- a/doc/advanced_configuration.rst
+++ b/doc/advanced_configuration.rst
@@ -252,7 +252,7 @@ typically implicitly returned rather than explicitly instantiated (e.g.,
 within function calls).
 
 For example, we can embed a small gallery of all examples that use or
-refer to :func:`numpy.exp`, which looks like this:
+refer to :obj:`numpy.exp`, which looks like this:
 
 .. include:: gen_modules/backreferences/numpy.exp.examples
 .. raw:: html
@@ -437,7 +437,7 @@ If you host your documentation on a GitHub repository, it is possible to
 auto-generate a Binder link for each notebook. Clicking this link will
 take users to a live version of the Jupyter notebook where they may
 run the code interactively. For more information see the `Binder documentation
-<http://docs.mybinder.org>`_.
+<http://docs.mybinder.org>`__.
 
 .. warning::
 
@@ -481,7 +481,7 @@ Binder links will point to these notebooks.
    Binder link, these files will be used to create the environment.
    For more information on what files you can use, see `preparing your
    repository <https://mybinder.readthedocs.io/en/latest/using.html#preparing-a-repository-for-binder>`_
-   in the `Binder documentation <docs.mybinder.org>`_ for more information on
+   in the `Binder documentation <docs.mybinder.org>`__ for more information on
    what build files are supported.
 
 See the Sphinx-Gallery `Sphinx configuration file <https://github.com/sphinx-gallery/sphinx-gallery/blob/master/doc/conf.py>`_
--- a/doc/advanced_configuration.rst
+++ b/doc/advanced_configuration.rst
@@ -238,20 +238,29 @@ point to the directory containing ``searchindex.js``, such as
 Adding references to examples
 =============================
 
-Sphinx-Gallery enables you, when documenting your modules, to
-reference to the examples that use a particular function. For example
-if we are documenting the :func:`numpy.exp` function its possible to embed
-a small gallery of examples that is specific to this function and
-looks like this:
+When documenting a given function/class, Sphinx-Gallery enables you to link to
+any examples that either:
+
+1. Use the function/instantiate the class in the code.
+2. Refer to that function/class using sphinx markup ``:func:``/``:class:``
+   in a documentation block.
+
+The former is useful for auto-documenting functions that are used and classes
+that are explicitly instantiated. The latter is useful for classes that are
+typically implicitly returned rather than explicitly instantiated (e.g.,
+:class:`matplotlib.axes.Axes` which is most often instantiated only indirectly
+within function calls).
+
+For example, we can embed a small gallery of all examples that use or
+refer to :func:`numpy.exp`, which looks like this:
 
 .. include:: gen_modules/backreferences/numpy.exp.examples
 .. raw:: html
 
         <div style='clear:both'></div>
 
-
-For such behavior to be available, you have to activate it in your
-Sphinx-Gallery configuration ``conf.py`` file with::
+For such behavior to be available, you have to activate it in
+your Sphinx-Gallery configuration ``conf.py`` file with::
 
     sphinx_gallery_conf = {
         ...
--- a/doc/advanced_configuration.rst
+++ b/doc/advanced_configuration.rst
@@ -430,14 +430,14 @@ Generate Binder links for gallery notebooks (experimental)
 ==========================================================
 
 Sphinx-Gallery automatically generates Jupyter notebooks for any
-examples built with the gallery. `Binder <http://mybinder.org>`_ makes it
+examples built with the gallery. `Binder <https://mybinder.org>`_ makes it
 possible to create interactive GitHub repositories that connect to cloud resources.
 
 If you host your documentation on a GitHub repository, it is possible to
 auto-generate a Binder link for each notebook. Clicking this link will
 take users to a live version of the Jupyter notebook where they may
 run the code interactively. For more information see the `Binder documentation
-<http://docs.mybinder.org>`__.
+<https://docs.mybinder.org>`__.
 
 .. warning::
 
@@ -481,7 +481,7 @@ Binder links will point to these notebooks.
    Binder link, these files will be used to create the environment.
    For more information on what files you can use, see `preparing your
    repository <https://mybinder.readthedocs.io/en/latest/using.html#preparing-a-repository-for-binder>`_
-   in the `Binder documentation <docs.mybinder.org>`__ for more information on
+   in the `Binder documentation <https://docs.mybinder.org>`__ for more information on
    what build files are supported.
 
 See the Sphinx-Gallery `Sphinx configuration file <https://github.com/sphinx-gallery/sphinx-gallery/blob/master/doc/conf.py>`_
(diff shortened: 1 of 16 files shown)


larsoner:
Readable updated docs:

https://367-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/rtd_html/advanced_configuration.html#references-to-examples

larsoner:
I'm happy to rebase, IIRC the only remaining thing is reviews (@choldgraf ?)

choldgraf:
holy moly I didn't notice this one in February, but agree it would be *super* useful.  I'll try to give it a review during one of my "I've been talking about code way too much I need to actually code a bit" breaks at SciPy :-)

larsoner:
Rebased and added fixes for our test infrastructure:

1. CircleCI had library problems.
2. Sphinx build needed to be done within the `docutils_namespace` context manager.

Example of `np.sin` manually placed in example docs to be in backrefs:

https://523-25860190-gh.circle-artifacts.com/0/rtd_html/index.html

New docs of feature:

https://523-25860190-gh.circle-artifacts.com/0/rtd_html/advanced_configuration.html#references-to-examples

Ready for review/merge from my end.

lesteve:
> Example of np.sin manually placed in example docs to be in backrefs:

Looks nice! Just curious, would it be easy to make np.sin a link here to the `np.sin` documentation?

doc/advanced_configuration.rst
```diff
@@ -336,10 +345,12 @@ saved in the directory we specified in the *toctree* option of the
 The files we are including are from the ``backreferences_dir``
 configuration option setup for Sphinx-Gallery.
 
+.. _code_tesiii:
```
lesteve: This reference is not used anywhere is it?

sphinx_gallery/tests/test_full.py
```diff
@@ -37,8 +37,8 @@ def ignore(src, names):
     with docutils_namespace():
         app = Sphinx(src_dir, conf_dir, out_dir, toctrees_dir,
                      buildername='html')
-
-    app.build(False, [])
+        # need to build with this mode for automodule and backrefs to work
```
lesteve: Not sure what this comment is about ... what do you mean by "this mode"? Is that related to moving `app.build` inside the `with docutils_namespace()`

sphinx_gallery/tests/test_full.py
```diff
@@ -66,5 +66,15 @@ def test_embed_links(sphinx_app):
     assert 'scipy.signal.firwin.html' in lines
     assert '#module-numpy' in lines
     assert 'numpy.arange.html' in lines
-    # assert '#module-matplotlib.pyplot' in lines
-    # assert 'pyplot.html' in lines
+    assert '#module-matplotlib.pyplot' in lines
```
lesteve: Intersesting, I remember this was failing from time to time, but maybe now this is more stable?

sphinx_gallery/tests/test_full.py
```diff
@@ -66,5 +66,15 @@ def test_embed_links(sphinx_app):
...
+    assert 'pyplot.html' in lines
+
+
+def test_backreferences(sphinx_app):
+    """Test backreferences."""
+    out_dir = sphinx_app.outdir
+    mod_file = op.join(out_dir, 'gen_modules', 'sphinx_gallery.sorting.html')
+    with codecs.open(mod_file, 'r', 'utf-8') as fid:
+        lines = fid.read()
+    assert 'FileNameSortKey' in lines
```
lesteve: Very nice test!

Maybe add a few comments before each assert to explain a bit what they are checking? Proposal (improvements more than welcome!):
* first assert: "check that FileNameSortKey class appears in the reference API of the sphinx_gallery.sorting module"
* second assert: "check that the backreferences work when 
  
Bonus points if you can add a test that the "normal" (i.e. in Python code) backreferences works? (maybe there is one already I haven't checked ...)

lesteve:
> Possibly but that's a separate issue (exists in master too I think) so I'd rather deal with it separately 

Yep completely agree, I was asking more out of curiosity!

doc/advanced_configuration.rst
larsoner: I don't see it added anywhere, @Titan-C what is this for?

sphinx_gallery/tests/test_full.py
larsoner: yes exactly, if the `app.build` is not in the context manager the links are not actually made

sphinx_gallery/tests/test_full.py
larsoner: yeah the CIs are happy so I figure we can try it, if it ends up being inconsistent again we can comment it back out

doc/advanced_configuration.rst
Titan-C: Since I added in `doc/_templates/module.rst` also the instruction to document classes I wanted to seed how this was affected on the line numbers. This reference was for personal use when reloading the HTML should have not been committed. 
As a side issue. It appears that the readthedocs theme on a python3 build(at least mine) cannot align the line numbers with the code lines. Other sphinx themes and our py2 build on CircleCI do display correctly.

larsoner:
@lesteve comments addressed

examples/plot_function_identifier.py
```diff
@@ -3,13 +3,20 @@
 Identifying function names in a script
 ======================================
 
-Calls Sphinx-Gallery identify names function to figure out which
-functions are called in the script and to which module do they belong.
+This demonstrates how Sphinx-Gallery identifies names function to figure out
```
choldgraf: ~names function~ function names ?

examples/plot_function_identifier.py
```diff
@@ -3,13 +3,20 @@
...
+which functions are called in the script and to which module do they belong.
+
+It uses both the code itself, as well as the docstrings (such as this one),
+as adding a reference to :func:`numpy.sin` and :func:`numpy.exp` will create
+proper backreferences even if they are not explicitly used. This is useful
+in particular when functions return classes -- if you add them to the docstring
```
choldgraf: it seems from the regex below that this is true for any text block references yeah? Can we make this a bit clearer in the examples / documentation? They tend to refer to the docstring but this should also work within `#######` blocks right?

choldgraf:
this is a clever way to support functions that aren't directly imported! Just a couple quick thoughts but in general it LGTM. +1 on a merge once my question on whether this works in more than just example docstrings is addressed

larsoner:
@lesteve it turns out NumPy intersphinx [is complicated] but I solved the linking issue by doing `:obj:` rather than `:func:` or `:np:func:` -- bottom line is that the links should work now.

@choldgraf I have addressed your comments and done some rephrasing, can you see if this is better?

https://526-25860190-gh.circle-artifacts.com/0/rtd_html/advanced_configuration.html#references-to-examples

https://526-25860190-gh.circle-artifacts.com/0/rtd_html/auto_examples/plot_function_identifier.html#sphx-glr-auto-examples-plot-function-identifier-py

doc/advanced_configuration.rst
```diff
@@ -481,7 +481,7 @@ Binder links will point to these notebooks.
    Binder link, these files will be used to create the environment.
    For more information on what files you can use, see `preparing your
    repository <https://mybinder.readthedocs.io/en/latest/using.html#preparing-a-repository-for-binder>`_
-   in the `Binder documentation <docs.mybinder.org>`_ for more information on
+   in the `Binder documentation <docs.mybinder.org>`__ for more information on
```
Titan-C: I'm wondering why you need to change the underscore? Trying to figure out(still ignorant), I realized this link is broken. Without `http` sphinx considers it as a local link.

doc/maintainers.rst
```diff
@@ -1,49 +1,52 @@
+:orphan:
```
Titan-C: Did sphinx complained this not being in any toctree? I missed that warning. But this makes me think if we should put this on our docs. Kind of reduce cognitive burden. All docs are online, but the release instructions are only in the repo. 
Also why all the new white space?

larsoner: > Did sphinx complained this not being in any toctree? 

Yes I was getting rid of warnings

> Also why all the new white space?

Getting rid of sphinx warnings about bad spacing (which are in `master`). But really these should all be part of the same bulleting indentation so it's a bugfix of sorts, too.

doc/advanced_configuration.rst
larsoner: The `_` makes it a link that can be used elsewhere, and it's named the same way so it creates a warning. The alternative would be just to not include the link in the second one, but this way makes it so that if one is removed both still work

larsoner: See e.g.

http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#anonymous-hyperlinks

doc/maintainers.rst
larsoner: We can add it to the docs later if someone is motivated, not sure there is much to gain (users don't care and maintainers can just look in the repo)

choldgraf:
I'm planning to merge this once appveyor is happy (and if github isn't broken), speak now or forever hold your peace!

choldgraf:
actually it looks like appveyor *did* pass, but GitHub is just being wonky so the display isn't updating. I'm gonna merge since we've now got a bunch of +1s!

🎉 🎉 🎉 thanks @larsoner !!! 🎉 🎉 🎉

lesteve:
Nice!

PR #352, opened 2018-02-15
Fix compatibility of SphinxDocLinkResolver with Sphinx 1.7

### LINKED ISSUE #351: Error with Sphinx 1.7 due to changes in DOCUMENTATION_OPTIONS
The following code:

https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/docs_resolv.py#L95

Now fails:

```
Exception occurred:
  File "/Users/tom/tmp/sphinx-gallery/sphinx_gallery/docs_resolv.py", line 99, in parse_sphinx_docopts
    raise ValueError('Documentation options could not be found in index.')
ValueError: Documentation options could not be found in index.
```

Due to the following change in Sphinx 1.7:

https://github.com/sphinx-doc/sphinx/blob/6cf0698b0c6bde9fdcf4a606838552e1e7e20663/CHANGES#L237

astrofrog:
This fixes compatibility with Sphinx 1.7

If you are happy with this, would it be possible to do a quick bugfix release? ReadTheDocs has now updated to Sphinx 1.7, so I think any project using sphinx-gallery on RTD is likely to run into this issue.

Fixes #351

### DIFF
--- a/sphinx_gallery/docs_resolv.py
+++ b/sphinx_gallery/docs_resolv.py
@@ -148,9 +148,11 @@ class SphinxDocLinkResolver(object):
                                  'URLs (doc_url cannot be absolute)')
             index_url = doc_url + '/'
             searchindex_url = doc_url + '/searchindex.js'
+            docopts_url = doc_url + '_static/documentation_options.js'
         else:
             index_url = os.path.join(doc_url, 'index.html')
             searchindex_url = os.path.join(doc_url, 'searchindex.js')
+            docopts_url = os.path.join(doc_url, '_static', 'documentation_options.js')
 
         # detect if we are using relative links on a Windows system
         if (os.name.lower() == 'nt' and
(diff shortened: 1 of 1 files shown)


larsoner:
Any idea why this wasn't caught by Travis? It would be good to update so that for versions < 1.7 and >= 1.7 this gets tested moving forward.

astrofrog:
@larsoner - no I'm not sure - it looks like Travis uses conda and sphinx 1.7 isn't available in defaults yet, so it could be due to that? (it's in conda-forge though)

larsoner:
Ahh so this will effectively be automatically implemented once `conda` updates. (We already pin a version at 1.5.5 so the < 1.7 will remain satisfied). +1 for merge from me, then.

astrofrog:
It looks like the Travis build failures are due to Travis issues (not this PR). Would it be possible to restart these? (or just merge as-is if you are happy with it)

lesteve:
I double-checked that this fixes the problem with sphinx 1.7. Merging, thanks a lot @astrofrog!

I am planning to add a Travis Cron job using sphinx master. This would allow us to detect such problems earlier (in my experience they tend to happen relatively often for new sphinx releases, e.g. 1.5 -> 1.6 or 1.6 -> 1.7).

astrofrog:
@larsoner @lesteve - would it be possible to do a bugfix release shortly with this fix? At the moment, RTD builds are failing in quite a few of our packages because RTD doesn't give the option to downgrade sphinx.

lesteve:
Would installing be a good stop gap solution until the next relase?

It seems like a requirements.txt like this works:
```
-e git+http://github.com/sphinx-gallery/sphinx-gallery#egg=sphinx-gallery 
```

together with:
```
pip install -r requirements
```

If you are using conda on RTD, I am hoping there is a similar solution for environment.yml.

astrofrog:
@lesteve - yes, I'll do that in the mean time, thanks!

larsoner:
See conda env pip install (with version pinning) here if you are curious:

https://github.com/mne-tools/mne-python/blob/master/environment.yml#L29

PR #356, opened 2018-02-26
[MRG] Fix title detection logic.

lesteve:
I tried using sphinx-gallery from master in scikit-learn and I found that https://github.com/sphinx-gallery/sphinx-gallery/pull/345 introduces a regression in a edge case, where the title starts with a space, i.e. your example starts with something like this:

```py
"""
======================
        The title
======================
"""
```

I also added more tests for extract_intro_and_title. The detection is based on regex and I feel is as good as the previous one. I tried a docutils based solution using `docutils.core.publish_doctree` (I am guessing we could even check that the first paragraph is a title if we go this way) but I did not get anywhere convincing in a reasonable amount of time.

Note: to make a long story short, the title found by sphinx-gallery was empty before #345 but in practice this does not matter at all (except if you use ordering by title). The title for each example in the gallery is generated automatically by sphinx from a `:ref`: to the automatically added reference into the example rst.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -217,20 +217,22 @@ def extract_intro_and_title(filename, docstring):
     # Title is the first paragraph with any ReSTructuredText title chars
     # removed, i.e. lines that consist of (all the same) 7-bit non-ASCII chars.
     # This conditional is not perfect but should hopefully be good enough.
-    title = paragraphs[0].strip().split('\n')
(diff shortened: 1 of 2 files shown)


lesteve:
OK thanks for the review @larsoner! Let's merge this one then.

lesteve:
> Argh sorry for the regression.

Forgot to say actually, it was more helping a bug to come out of the closet than a regression per se ;-) !

PR #357, opened 2018-03-10
Fix CI

Titan-C:
While working again on PR #239 I noticed travis is not happy anymore due to update in matplotlib's API in matplolib 2.2. Also on our oldest test box the ubuntu one. Python 2.7 is just to old for the requirements of Tornado 5 regarding OpenSSL so I avoid installing that version.


lesteve:
Great, thanks a lot for fixing that!

lesteve:
For the record I added a daily cron job in https://travis-ci.org/sphinx-gallery/sphinx-gallery/settings. This should allow to notice this kind of breakages earlier next time. Also I have some plan to add a build with sphinx master in the daily cron job one day so that we anticipate sphinx changes a bit better.

PR #358, opened 2018-03-10
[MRG] Add Travis sphinx master cron job

lesteve:
This allows to detect possible problems with sphinx before it is released. It will be run as a daily cron job because rather than in PRs because most of the time sphinx master failures are not related to PR changes but to changes in sphinx.

We do something similar in scikit-learn to test against numpy-dev, scipy-dev and cython-dev and this has proved very useful to feed back issues upstream and get them ironed out before the next release.


lesteve:
An example successful cron job on my fork can be seen [here](https://travis-ci.org/lesteve/sphinx-gallery/builds/351603502).

lesteve:
CIs are green, merging this one!

PR #359, opened 2018-03-11
Add dedicated class for timing related block

ThomasG77:
This PR is for addressing the need to style timing block by adding a CSS class `sphx-glr-timing` class to `<p>`.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -159,6 +159,10 @@ CODE_OUTPUT = u""".. rst-class:: sphx-glr-script-out
 
 {0}\n"""
 
+TIMING_CONTENT = """
+.. rst-class:: sphx-glr-timing
+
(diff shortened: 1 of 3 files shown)


Titan-C:
I like that this part of the rst document moves to the variable `TIMING_CONTENT`. Can we have a little test for this? The reason I ask is to have something to remember me to port this in PR #239, a test that fails because the class is not found is a great way to remember this was added and does not get lost in all the reordering I do in PR #239

larsoner:
One way to test would be to look at the `tinybuild` HTML output for the proper class name using a regular expression.

larsoner:
@ThomasG77 can you add the test? If not I can probably take over and do it, would be good to get this into the 0.3 release

larsoner:
I rebased, added tests, and force pushed. Since @Titan-C was +1 for merge I'll merge once CIs are green

larsoner:
Passes other than the failure in `master` fixed by #432

larsoner:
Thanks @ThomasG77

Issue #340, opened 2018-01-24
The "floating buttons box" still needs some work


choldgraf:
Or some documentation on how to disable it / change its behavior.

E.g., the matplotlib theme doesn't quite get the floating correct:

![sphinx_gallery_float](https://user-images.githubusercontent.com/1839645/35309427-556cfcb4-0061-11e8-9ad9-16ffa616729e.gif)

lesteve:
Not 100% great I agree. As someone with a very limited CSS-fu, I would be enclined to follow a low-tech approach, basically put something like this (in a visible yet not too intrusive manner that is left as an exercise to the reader) before the beginning of the example: "To launch binder or download the example code click here" with a link to the bottom of the page where you have all the download links.

lesteve:
As always better suggestions more than welcome!

choldgraf:
Alternatively we could add a section to the configuration page that explains how to modify the CSS for these elements. E.g., it could have a section on removing the "floating right" behavior and defaulting to putting the buttons at the end. Maybe that's a simpler short-term solution than changing the code?

lesteve:
> E.g., it could have a section on removing the "floating right" behavior and defaulting to putting the buttons at the end.

I am OK with this as a stop-gap solution. The not so great part is that it goes against the original goal of the "float right" functionality which was to make the download buttons more prominent because they are easy to miss.

GaelVaroquaux:
Is the matplotlib theme a simple modification of a classic sphinx theme?
If so, we should detect it with CSS rules and change the behavior
accordingly.

If not, we should just document how to override this behavior and let
matplotlib do it in their CSS.

choldgraf:
I believe so, but not sure how to check for it since I do think they've customized it quite a bit

Titan-C:
My CSS is very limited too. But my guess is that trying to get a generic CSS right to any theme our users may want to use might be beyond the time we have for this. I once prepared the build script to test with all available Sphinx themes, needless to say we don't always check that all themes are working in a compatible manner.
My suggestion would be to return to our default and not fancy download buttons at the end by default and document as a user enabled feature how to extend the CSS of the download buttons to get the "floating right" behaviour. We can of course continue to have our dog food and put such CSS override in our read the docs theme.

choldgraf:
yeah I'm +1 on "solve this with documentation" one way or another...I'm .5/.5 on whether we the default floating or make the default at the bottom

[...]

PR #362, opened 2018-03-22
[MRG] Add note at the beginning of the example to the download section at the bottom

lesteve:
This is my attempt at removing the floating right download buttons , which I feel is just too hard to configure right by default for a majority of sphinx themes, sorry @GaelVaroquaux ... 

There is a little bit of (at least for me) advanced CSS so that the link to the example from the gallery show the note (i.e. it behaves as if you go to the example section and then scroll up a bit so that you can see the note).

Comments more than welcome!

See #340 as well.

### DIFF
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -147,54 +147,13 @@ div.sphx-glr-download a:hover {
   background-color: #d5d57e;
 }
 
-/* On large screens, float the download and binder buttons on the right */
-@media only screen and (min-width: 140ex) {
-    div.sphx-glr-footer-example  {
-	position: fixed;
-	right: 0px;
-	top: 100px;
-	background: rgba(200, 200, 200, 0.3);
-	padding: 0ex .5ex;
-	border: solid 1px #ccc;
-	max-width: 24ex;
-	width: fit-content;
-	width: -moz-min-content;
-    }
-
-    /* Avoid overflow of long file names */
-    div.sphx-glr-footer-example code {
-	max-width: 100%;
-	overflow: hidden;
-    }
-
-    div.sphx-glr-footer-example a {
-	max-width: 95%;
-	overflow: hidden;
-	padding: .5ex;
-    }
-
-    div.sphx-glr-footer-example a code span:last-child {
-	font-size: smaller;
-    }
+.sphx-glr-example-title > :target::before {
+  display: block;
+  content: "";
+  margin-top: -50px;
+  height: 50px;
+  visibility: hidden;
 }
-
-/* On very large screen, don't flush it too far */
-@media only screen and (min-width: 190ex) {
-    div.sphx-glr-footer-example  {
-	right: 5%;
-	max-width: 29ex;
-    }
-}
-
-/* Special case the RTD theme, capturing "section.wy-nav-content" */
-@media only screen and (min-width: 190ex) {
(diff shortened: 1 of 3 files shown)


lesteve:
Generated docs:
* [rtd theme](https://405-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/rtd_html/auto_examples/index.html)
* [alabaster theme](https://405-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/alabaster_html/auto_examples/index.html)
* [default theme](https://405-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/doc/_build/html/auto_examples/index.html)

larsoner:
Can we make the `.. note::` at the top effectively optional by giving it a CSS class so it can be set invisible?

GaelVaroquaux:
Actually, I think that it should be a div (directive "container" in sphinx), containing the download buttons and using a CSS similar to the one that I had used for the floating button: flushed on the right, inside a box. Maybe it could even be an exact copy of the CSS that I used, removing the "position: fixed", and using a float to the right.

That would give the best of both worlds.

lesteve:
> Can we make the .. note:: at the top effectively optional by giving it a CSS class so it can be set invisible?

@larsoner this can definitely be done. Does that mean that you would disable the note in mne-python? Maybe you still find it too intrusive?

@GaelVaroquaux I am not sure I follow you. Maybe you can add a quick sketch of what you have in mind?

larsoner:
>  Does that mean that you would disable the note in mne-python? Maybe you still find it too intrusive?

Probably, depending on if @agramfort and others agree. Aethestically I don't like the note being above the title, and don't think it's necessary to make the download links so prominent. But it could still make sense as a default behavior for other packages.

lesteve:
> Aethestically I don't like the note being above the title, and don't think it's necessary to make the download links so prominent.

Would that be fixable with some CSS or content tweaks? Is it that you are happy enough with download links at the bottom of the example?

larsoner:
> Would that be fixable with some CSS or content tweaks?

Maybe putting the note after the title, but that might have its own problems?

> Is it that you are happy enough with download links at the bottom of the example?

Yeah for me it seems good enough, but I see why it being more prominent can be useful, too.

lesteve:
> Maybe putting the note after the title, but that might have its own problems?

The main reason I did not do put the note after the title is mostly technical: at the moment we put the example docstring as is into the rst. Putting the note after the title, would require splitting the docstring in two. We kind of already do that in a hacky way to extract the title but I thought I would try the simplest option first.

lesteve:
OK I had some grand plans of improving the design (e.g. by having the the download links at the top of the example, in a header like fashion, on just one line with less text and more icons to make it fit) . The short story is I just severely lack HTML + CSS skills to do that effectively so I give up for now.

I added a class to the note as suggested by @larsoner so that it can be easily hidden for people that are happy with the status quo.

Comments more than welcome!

lesteve:
For completeness I think I also tried https://github.com/sphinx-gallery/sphinx-gallery/pull/362#issuecomment-375727788 but that did not work very well because the right float buttons could overlap with the code cell when the example description is not big enough. I did that on my laptop so I don't have a screenshot to make this more clear but I'll try to find something when I get home later today.

lesteve:
Good to see this merged, although I agree there is definitely room for improvements here ...

PR #364, opened 2018-04-17
Add 'python -msphinx_gallery.clean' command.

anntzer:
This reads out the correct directories to clean from the conf, rather
than having to hardcode them in three different places (conf.py,
Makefile and make.bat).


lesteve:
Is that different from what the `clean` target in the `Makefile` is currently doing?

Would it be possible to add a test for `sphinx_gallery.clean.main()` ?

lesteve:
> The modern way of doing the Makefile with newer sphinx would be to call sphinx rather than write shell scripts (look at the Makefile of the dirty_cat project, for instance).

I am fine with any changes/improvements as long as `make clean` works as well before. In other words in 99% of the cases I interact with sphinx stuff via `make` and I'd rather keep it this way.

anntzer:
Added test.

sphinx_gallery/tests/test_full.py
```diff
@@ -62,3 +62,9 @@ def test_embed_links(sphinx_app):
     assert 'numpy.arange.html' in lines
     # assert '#module-matplotlib.pyplot' in lines
     # assert 'pyplot.html' in lines
+
+
+def test_clean(sphinx_app):
```
lesteve: Do we want to make sure this function is run as the last test, otherwise you may break other tests by cleaning some of the generated artifacts?

I don't know off the top of my head if there is a way to ensure that with pytest but it's worth looking at it. Maybe a module teardown could work also that seems like a hack.

anntzer: Can't find it the docs right now but I am almost certain pytest runs the tests in the source order (the fact that the tests pass on travis despite e.g. hash randomization on Py3 clearly indicates (well, strongly suggests :-)) that that's the case).

lesteve: OK the only thing I found is [this SO answer](https://stackoverflow.com/a/44867455). I tested this a bit and the only tricky case where ordering does not seem to be respected is if you use a class test that derives from unittest.TestCase.

If that ends up being too brittle, we can always try to use [pytest-ordering](https://github.com/ftobia/pytest-ordering).

sphinx_gallery/tests/test_full.py
```diff
@@ -62,3 +64,11 @@ def test_embed_links(sphinx_app):
...
+    subprocess.check_call(
+        [sys.executable, '-msphinx_gallery.clean'], cwd=sphinx_app.srcdir)
```
lesteve: Could you use `sphinx_gallery.clean.main()`. This seems like a good candidate to make codecov happy.

sphinx_gallery/tests/test_full.py
```diff
@@ -62,3 +64,11 @@ def test_embed_links(sphinx_app):
     assert 'numpy.arange.html' in lines
     # assert '#module-matplotlib.pyplot' in lines
     # assert 'pyplot.html' in lines
+
+
+def test_clean(sphinx_app):
```
lesteve: In case there is no better suggestions for ensuring that this tests runs last, can you add a comment mentioning that this test should be at the end of the file in order not to interfere with other tests.

anntzer:
Added a comment.  Did not switch to calling main() directly, as that would need 1) to manipulate the cwd and 2) to manipulate sys.argv, which are both doable but quite overkill IMO.

ImportanceOfBeingErnest:
Is there any way this could determine the actual location of the `conf.py` file? I think you may often have the `conf.py` in a subfolder compared to `make`. So this would lead to 

```
make clean
cd source
python -msphinx_gallery.clean
cd..
```

anntzer:
Indeed, that's a better option (`conf.py` is in srcdir, which is not necessarily the same as the directory containing the makefile).
Given that the correct solution requires passing srcdir to `python -msphinx_gallery.clean`, I decided to just make it a full replacement of `python -msphinx -Mclean` instead.

sphinx_gallery/clean.py
```diff
@@ -0,0 +1,29 @@
...
+    parser = ArgumentParser(
+        description=
+        "Clean intermediate files generated by sphinx-gallery, and call "
+        "'python -msphinx -M clean' with the same arguments.")
+    parser.add_argument("sourcedir")
+    parser.add_argument("builddir")
+    args, _ = parser.parse_known_args()
+    d = {}
+    with open(os.path.join(args.sourcedir, "conf.py")) as file:
+        exec(file.read(), d)
```
Titan-C: `file` is a builtin in python 2 please use: `sg_conf_file`, `sg_conf_py` or other reasonable variable name.

sphinx_gallery/clean.py
```diff
@@ -0,0 +1,29 @@
...
+import subprocess
+import sys
+
+
+def main():
+    parser = ArgumentParser(
+        description=
+        "Clean intermediate files generated by sphinx-gallery, and call "
+        "'python -msphinx -M clean' with the same arguments.")
+    parser.add_argument("sourcedir")
```
Titan-C: Very ignorant question but which are the defaults?

sphinx_gallery/clean.py
```diff
@@ -0,0 +1,29 @@
...
+    d = {}
+    with open(os.path.join(args.sourcedir, "conf.py")) as file:
+        exec(file.read(), d)
+    sg_conf = d["sphinx_gallery_conf"]
+    for path in sg_conf["gallery_dirs"]:
+        shutil.rmtree(os.path.join(args.sourcedir, path), ignore_errors=True)
+    shutil.rmtree(os.path.join(args.sourcedir, sg_conf["backreferences_dir"]),
+                  ignore_errors=True)
+    subprocess.check_call(
+        [sys.executable, "-msphinx", "-M", "clean"] + sys.argv[1:])
```
Titan-C: why `argv` and not the directories that where just parsed?

sphinx_gallery/clean.py
anntzer: fixed, also needed to backport exec for Py2...

sphinx_gallery/clean.py
anntzer: It's a required argument, so there's no default.

sphinx_gallery/clean.py
anntzer: Because there can be a bunch of other args that are meaningful to sphinx:
```
optional arguments:
  -h, --help        show this help message and exit
  --version         show program's version number and exit

general options:
  -b BUILDER        builder to use (default: html)
  -a                write all files (default: only write new and changed
                    files)
  -E                don't use a saved environment, always read all files
  -d PATH           path for the cached environment and doctree files
                    (default: OUTPUTDIR/.doctrees)
  -j N              build in parallel with N processes where possible (special
                    value "auto" will set N to cpu-count)

build configuration options:
  -c PATH           path where configuration file (conf.py) is located
                    (default: same as SOURCEDIR)
  -C                use no config file at all, only -D options
  -D setting=value  override a setting in configuration file
  -A name=value     pass a value into HTML templates
  -t TAG            define tag: include "only" blocks with TAG
  -n                nit-picky mode, warn about all missing references

console output options:
  -v                increase verbosity (can be repeated)
  -q                no output on stdout, just warnings on stderr
  -Q                no output at all, not even warnings
  --color           do emit colored output (default: auto-detect)
  -N, --no-color    do not emit colored output (default: auto-detect)
  -w FILE           write warnings (and errors) to given file
  -W                turn warnings into errors
  -T                show full traceback on exception
  -P                run Pdb on exception
```
most of them probably don't matter for cleaning, but we may as well make everything work even if sphinx decides to start pulling stuff from these args when cleaning too.

anntzer:
The instruction for the makefile is as well documented as it was before:
```
Extending your Makefile
-----------------------
Once your gallery is working you might need remove completely all generated files by
sphinx-gallery to have a clean build, or you might want to build the gallery without
running the examples files. For this you need to extend your ``Makefile`` with:

.. code-block:: bash

    clean:
            python -msphinx_gallery.clean
```
(previously this had `rm -rf $(BUILDDIR)/*` and `rm -rf auto_examples/`, which is literally what this PR is replacing).

anntzer:
Issues should all be fixed now.

Titan-C:
I don't see any pushed changes

Titan-C:
> The instruction for the makefile is as well documented as it was before:

I mean changing the makefiles themselves. As @lesteve says, we want the make files to continue working and use this new feature.

anntzer:
Ah, I see.  Done.

sphinx_gallery/clean.py
Titan-C: where is this fixed?

larsoner:
Not sure why this got stuck / unmerged, @anntzer do you still want to work on this?

anntzer:
I can't say I really care anymore, but feel free to pick up the PR if you want.

larsoner:
I'll close for now but if someone wants to pick this up again, feel free

PR #367, opened 2018-04-23
Don't create spurious examples/ directory.

anntzer:
When using the "separate source and build" layout, sphinx-gallery
creates a spurious (and ultimately empty) "examples/" directory next to
the Makefile, whereas the correct place for that directory is
"source/examples".  It is in fact not necessary to pre-create that
directory, because generate_dir_rst also contains the snippet

        os.makedirs(target_dir)

which will create the *correct* examples/ directory.


lesteve:
LGTM, I am going to merge this one.

I am not sure how well we test "separate source and build" layout but it would be good to double-check. Historically we have incorporated fixes (e.g. from scikit-image IIRC) without thorough tests.

Issue #368, opened 2018-04-25
Create `output` folder configurable option


choldgraf:
In both #353 and #338 there are requests to allow for an "output" folder in the sphinx gallery configuration. This would be a folder where all generated SG content would live *in addition to* the traditional sphinx "downloads" folder.

An important difference is that this output folder should mirror the structure that the examples folders take.

lesteve:
In my mind this should be only for generated notebooks. Do you have other kind of content in mind?

choldgraf:
closed by #371 !

PR #371, opened 2018-04-27
adding binder custom tree option

### LINKED ISSUE #353: JupyterLab support on binder
I just heard about urlpath=lab. Maybe it would be nice to support at one point. I could not find a way to point to a given file with JupyterLab, i.e. using both urlpath=lab&filepath=myipnb. Is this correct @choldgraf?

Thinking about it out loud, in a multi-notebook context (at the moment we are just opening a single notebook in binder but you could imagine pointing to a folder e.g. in the main gallery or it seems like urlpath=lab works like this by default) it feels like the notebooks should be in auto_examples as well as in _downloads. This way the notebook folder structure mirrors the example folder (maybe that does get rid of the duplicated filename problem).

I am not sure why .ipynb can not be seen in https://github.com/sphinx-gallery/sphinx-gallery.github.io/tree/master/auto_examples but this can probably be fixed (at the cost of duplicating .ipynb in the repo, once in _downloads, once in auto_examples which is probably fine ...).

choldgraf:
This PR adds the ability to collect jupyter notebooks in a way that preserves their relative paths/folders. It also generates Binder URLs that point to these paths.

## The challenge

To my knowledge, sphinx gallery doesn't have access to the `_downloads` folder until _after_ the site is built. To that extent, I don't think we can modify / duplicate those files after they've been created. I also don't think it's possible to tell the `:download:` directive to put a file in some relative path in `_downloads`. It only dumps it into the root of the directory.

So I think our only option (used in this PR) is kind of hacky: putting the folders/files in `_static` just before the site is built.

So, this PR does:

1. If the binder configuration `static_folder` is defined (relative to `conf.py`) it will trigger the following process:
2. Each ipynb file will be copied to a sub-tree within `<staticfolder>/binder/path/to/ntbk.ipynb` during the build process
3. Binder links will point to this sub-tree rather than the `_downloads` folder

It also adds in the ability to specify `use_lab=True` in the binder config, in which case it will change the end of the link from:

`?filepath=_downloads/myfile.ipynb` to `?urlpath=lab/tree/_downloads/myfile.ipynb`

LMK what folks think! It's a bit hacky but I think our only option. In the docs, we should recommend that people put the output folder under their `gitignore`.

Here's a binder link that shows what the final folder/file structure looks like:

https://mybinder.org/v2/gh/choldgraf/sphinx-gallery/gh-pages?urlpath=lab/tree/_static/binder/auto_examples/plot_colors.ipynb

LMKWYT!

cc @lesteve 

closes #353 #368


doc/conf.py
```diff
@@ -341,10 +341,15 @@ def setup(app):
     'find_mayavi_figures': find_mayavi_figures,
     'expected_failing_examples': ['../examples/no_output/plot_raise.py',
                                   '../examples/no_output/plot_syntaxerror.py'],
-    'binder': {'org': 'sphinx-gallery',
-               'repo': 'sphinx-gallery.github.io',
+    'binder': {#'org': 'sphinx-gallery',
```
choldgraf: I will update these back to their respective values when we're +1 on a merge...this is just for testing

lesteve:
Thanks a lot for the PR! I have not looked at the details yet, but the main question I have is: is there a good reason that we *have* to use `_static`? I would have thought that the variable should be called `notebooks_output_folder` or something similar and that and that the user could chose any folder he wants but I may be missing a sphinx quirk.

lesteve:
Also some CI failures seem genuine, but I don't think that is the main problem right now.

choldgraf:
@lesteve as far as I could figure out, sphinx won't copy over anything that isn't RST or markdown (e.g., it doesn't copy over anything other than this in the "auto_examples" folder, but correct me if I'm wrong) so the only option is to put it in `_static/myfolder`

If somebody knows a hack around this, lemme know!

lesteve:
Can we not copy the notebooks ourselves? This was my naive strategy originally:
* we may be able to know the html output folder (`_build/html` in the sphinx-gallery case): if `app` is a `Sphinx` object, it looks like `app.outdir` is what we want.
* once we know the html output folder we add `notebooks_output_folder` to it so that the notebook are inside the generated html. That means that the notebooks are part of the repo and we can use binder with the appropriate URLs.

lesteve:
Something like this, copying the notebooks rather than moving seems safer. Just to be clear, in my mind `notebook_output_dir` is a subfolder or `app.outdir`. Just to take an example in the sphix-gallery case, if I have `notebook_output_dir='my/notebooks'` I should have the notebooks in `_build/html/my/notebooks`.

I agree that this is a bit working around sphinx but I feel this is a bit what sphinx-gallery is already doing to some respect (e.g. the backreferences aka mini-galleries in the reference API doc comes to mind). Of course, if there are ways to do it the sphinx way for this PR or others, this would be more than welcome!

choldgraf:
lemme give this a shot and we'll see how it looks :-)

choldgraf:
@lesteve what do you think about the latest push? You can now specify an arbitrary folder, and the notebooks will be copied to that folder within `_build`.

e.g.:

http://predictablynoisy.com/sphinx-gallery/auto_examples/plot_seaborn.html#sphx-glr-auto-examples-plot-seaborn-py

I think part of the errors are because I'm using a python 3 function somewhere, I'll try and run that down, but in the meantime WDYT about the current pattern?

Also - do you think that we should actually do this automatically, rather than have people provide a config to enable it? E.g., we could basically make `notebooks_dir` to default to `notebooks`.

sphinx_gallery/binder.py
```diff
@@ -45,20 +45,39 @@ def gen_binder_url(fname, binder_conf):
...
+    link_base = binder_conf.get('notebooks_folder', '_downloads')
+    if link_base != '_downloads':
+        # In this case, we want to keep the relative path to sub-folders
+        # Split so we have the relative path starting after the gallery dir
+        relative_link = _get_gallery_dir_path(fname, gallery_conf)
+        path_link = os.path.join(
+            link_base, replace_py_ipynb(relative_link))
+    else:
+        # Assume that we're in the root of _downloads
+        fname = os.path.basename(fname)
```
Titan-C: why do you need this extra `basename`?

sphinx_gallery/gen_rst.py
```diff
@@ -709,7 +711,19 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
 
     time_m, time_s = divmod(time_elapsed, 60)
     example_nb = jupyter_notebook(script_blocks)
-    save_notebook(example_nb, replace_py_ipynb(example_file))
+    path_ipynb = replace_py_ipynb(example_file)
+    save_notebook(example_nb, path_ipynb)
+
+    # Copy the ipynb file to the static folder, preserving folder structure
+    if binder_conf.get('notebooks_folder'):
```
Titan-C: I don't like doing the copy part here nor passing the app object this deep, this breaks my desire of making sphinx-gallery more modular. This copy task should happen at the end of the build in the same way we implement `embed_code_links`.
That is have in setup something like this
```python
app.connect('build-finished',copy_jupyter_nb)
```
And then in binder.py the funtion that does the copy.

Titan-C: maybe for the the copy just use an `os.walk` and copy everything that is a notebook.

sphinx_gallery/binder.py
choldgraf: I believe that fname here is a full path, so this just grabs the name of the file itself

sphinx_gallery/gen_rst.py
choldgraf: yeah I think os.walk is the path forward here...lemme give a shot at modularizing this a little bit more.

choldgraf:
ok I *think* the latest push should address much of @lesteve 's concerns. It factors out the notebook copying step to the very end as you suggested. Also uses `os.walk` now, so perhaps that means the python 2 builds will be happy again

choldgraf:
oops, that was @Titan-C 's concerns not @lesteve 's concerns :-)

Latest version sets `notebooks_folder` to default to `notebooks/`, and lets people change this if they wish.

LMK if there's anything else I should tackle other than making tests pass.

choldgraf:
OK I think we're looking good w/ tests and coverage. This is ready for another review, or ready to 🚢

choldgraf:
huh - yes I see the same thing too! I didn't check to see whether *extra* files and folders were created haha...lemme check on this. And I agree with `_dir` being consistent with the broader project

choldgraf:
OK I think I fixed it up. Here's the my build tree now:

```
doc/_build/html/notebooks/
├── auto_examples
│   ├── no_output
│   │   ├── just_code.ipynb
│   │   ├── plot_raise.ipynb
│   │   ├── plot_strings.ipynb
│   │   └── plot_syntaxerror.ipynb
│   ├── plot_choose_thumbnail.ipynb
│   ├── plot_colors.ipynb
│   ├── plot_exp.ipynb
│   ├── plot_function_identifier.ipynb
│   ├── plot_gallery_version.ipynb
│   ├── plot_quantum.ipynb
│   ├── plot_seaborn.ipynb
│   ├── plot_sys_argv.ipynb
│   ├── plot_unicode_everywhere.ipynb
│   └── sin_func
│       ├── plot_sin_black_background.ipynb
│       └── plot_sin.ipynb
└── tutorials
    ├── plot_notebook.ipynb
    ├── plot_parse.ipynb
    └── seaborn
        └── plot_seaborn_notebook.ipynb
```

also replaced `_folder` with `_dir`

choldgraf:
friendly ping :-)

doc/advanced_configuration.rst
```diff
@@ -442,19 +442,27 @@ dictionary following the pattern below::
...
          'repo': '<github_repo>',
          'url': '<binder_url>',  # Any URL of a binder server. Must be full URL (e.g. https://mybinder.org).
          'branch': '<branch-for-documentation>',  # Can be any branch, tag, or commit hash. Use a branch that hosts your docs.
          'dependencies': '<list_of_paths_to_dependency_files>',
-         'filepath_prefix': '<prefix>' # Optional, a prefix to append to any filepaths in Binder links.
-                            use this if you move your built documentation to a sub-folder of your repository (e.g., "v2.1")
+         # Optional keys
+         'filepath_prefix': '<prefix>' # A prefix to append to any filepaths in Binder links.
+                            use this if you move your built documentation to a sub-folder of your repository (e.g., "v2.1").
+         'notebooks_dir': '<notebooks-directory-name>' # A directory name where jupyter notebooks that Binder links point to will be moved.
```
Titan-C: I find this hard to read. 
Also, I know this is not  "runnable" code, but I would appreciate that multi-line comments do start with `#`.

doc/conf.py
```diff
@@ -341,10 +341,15 @@ def setup(app):
...
+               'org': 'choldgraf',
+               'repo': 'sphinx-gallery',
                'url': 'https://mybinder.org',
-               'branch': 'master',
-               'dependencies': './binder/requirements.txt'
+               'branch': 'gh-pages',
+               # 'branch': 'master',
+               'dependencies': './binder/requirements.txt',
+               'notebooks_dir': 'notebooks',
+               'use_lab': True
```
Titan-C: Nitpick. Terminate this with a comma.

sphinx_gallery/binder.py
```diff
@@ -95,17 +118,61 @@ def gen_binder_rst(fname, binder_conf):
...
+def _copy_binder_notebooks(app, gallery_conf, binder_conf):
+    """Copy Jupyter notebooks to the output directory.
+
+    Walk through each output gallery directory
+    For each one, check if the file ends in ipynb
+    If it does, copy it to `notebooks_dir` with the same folder structure.
+    """
+    gallery_dirs = gallery_conf.get('gallery_dirs')
+    if not isinstance(gallery_dirs, (list, tuple)):
+        gallery_dirs = [gallery_dirs]
```
Titan-C: This control is already done in the generation of the gallery. No need to redo it here

sphinx_gallery/binder.py
```diff
@@ -95,17 +118,61 @@ def gen_binder_rst(fname, binder_conf):
...
+    for i_folder in gallery_dirs:
+        for i_path, _, i_files in os.walk(os.path.join(app.srcdir, i_folder)):
+            for i_file in i_files:
+                if len(i_file) > 0 and i_file.endswith('.ipynb'):
+                    path_rel = _get_gallery_dir_path(i_path, gallery_conf)
+                    path_out_dir = os.path.join(
+                        app.outdir, notebooks_dir, path_rel)
+                    if not os.path.exists(path_out_dir):
+                        os.makedirs(path_out_dir)
+                    sh.copyfile(os.path.join(i_path, i_file),
```
Titan-C: can we use shutil instead of just sh.

sphinx_gallery/binder.py
```diff
@@ -95,17 +118,61 @@ def gen_binder_rst(fname, binder_conf):
...
+    if not len(binder_conf) > 0:
+        return
+
+    logger.info('copying binder requirements...', color='white')
+    _copy_binder_reqs(app, gallery_conf, binder_conf)
+    logger.info('copying binder notebooks...', color='white')
+    _copy_binder_notebooks(app, gallery_conf, binder_conf)
+
+
+def _copy_binder_reqs(app, gallery_conf, binder_conf):
```
Titan-C: Why is there a need to change this function? `gallery_conf` is not used, remove it. Also use `shutil.copy` not `sh.copy`

sphinx_gallery/gen_rst.py
```diff
@@ -709,7 +710,9 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
 
     time_m, time_s = divmod(time_elapsed, 60)
     example_nb = jupyter_notebook(script_blocks)
-    save_notebook(example_nb, replace_py_ipynb(example_file))
+    path_ipynb = replace_py_ipynb(example_file)
+    save_notebook(example_nb, path_ipynb)
```
Titan-C: Why this change? Please keep in mind I also try to get pr #239 and I don't want too many things changing unnecessarily as I need to keep track of the changes.

sphinx_gallery/tests/test_gen_gallery.py
```diff
@@ -273,3 +273,33 @@ def test_collect_gallery_files(config_app, tmpdir):
...
+
+@pytest.mark.conf_file(content="""
+import os
+import sphinx_gallery
+extensions = ['sphinx_gallery.gen_gallery']
+# General information about the project.
+project = u'Sphinx-Gallery <Tests>'
+
+sphinx_gallery_conf = {
+    'mod_example_dir' : os.path.join('modules', 'gen'),
```
Titan-C: `mod_example_dir` is deprecated use `backreferences_dir`

sphinx_gallery/binder.py
```diff
@@ -158,3 +227,15 @@ def check_binder_conf(binder_conf):
             'for sphinx-gallery. A path to at least one of these files '
             'must exist in your Binder dependencies.')
     return binder_conf
+
+
+def _get_gallery_dir_path(fname, gallery_conf):
```
Titan-C: I don't really understand why this function is needed. Also, I don't really understand what it is trying to do. Can you do a unit test of this and parametrize to many inputs? So that I can get an idea of what this function is trying to do and be sure we need it. I apologize not yet running the code and thus having little idea what goes on in the building of the binder directory.

sphinx_gallery/gen_rst.py
choldgraf: just cruft from an earlier iteration of this PR, reverting

sphinx_gallery/binder.py
choldgraf: removed the gallery conf (was from an earlier version of the PR I think). I modified the function so that it has the same general structure as the "copy notebooks" function

sphinx_gallery/binder.py
choldgraf: actually I think that I can accomplish the same thing with `relpath` now that I think about it, lemme give that a shot

Titan-C:
This looks good to me. Thank you for the quick refactoring. Happy to see you use `relpath` now and not your own function. It makes the code more readable now. I took the liberty to update the function that copies the jupyter notebooks. Please review it.
This is ready for a merge if everything looks good on @choldgraf live website, then we can restore the old config.

choldgraf:
Thanks for the patch! I cleaned up the filtering function a little bit since the multi-part boolean logic was confusing to me. Other than that, I think this is ready to go from my end. Here's a link to the live build:

https://predictablynoisy.com/sphinx-gallery/auto_examples/sin_func/plot_sin.html#sphx-glr-auto-examples-sin-func-plot-sin-py

choldgraf:
Sounds good - I am all for beautifully complex list comprehensions, but I must admit it took me like 10 minutes to figure out what that one was doing :-) 

I think the main thing that confused me was these lines:

`entry.endswith('.ipynb') != os.path.isdir(os.path.join(path, entry))`

anyway feel free to suggest alternate namings etc...

Titan-C:
> `entry.endswith('.ipynb') != os.path.isdir(os.path.join(path, entry))`

You have to see != as an exclusive OR. Then it works fine and you can
read: either a Jupyter notebook or a directory. Then the NOT in front
transforms it to neither ... nor.

choldgraf:
I tried updating the functions name/description/structure to make it a bit more clear...LMK what you think (or feel free to make edits)

lesteve:
Sorry I haven't had time to look at this which I think is really important. I'll try to find some time beginning of next week.

choldgraf:
friendly ping on this :-)

Titan-C:
I'm fine with this one. Just update the config values to point to
sphinx-gallery repo and not yours.

lesteve:
I am going to look at this one today.

sphinx_gallery/binder.py
```diff
@@ -95,17 +117,76 @@ def gen_binder_rst(fname, binder_conf):
...
+    binder_conf = gallery_conf.get('binder')
+    notebooks_dir = os.path.join(app.outdir, binder_conf.get('notebooks_dir'))
+    shutil.rmtree(notebooks_dir, True)
+    os.makedirs(notebooks_dir)
+
+    iterator = sphinx_compatibility.status_iterator(
+        gallery_dirs, 'copying binder notebooks...', length=len(gallery_dirs))
+
+    for i_folder in iterator:
+        shutil.copytree(os.path.join(app.srcdir, i_folder),
```
lesteve: In terms of readability I feel it would be clearer to walk the folder and copy all the .ipynb files explicitly. You would have to change the the beginning of the path from `app.srcdir` to `i_folder` by hand.

In other words if we only want to select `*.ipynb` files, I find the `ignore` API is not that well suited.

Reflecting about this though, I am wondering whether notebooks need other files to be fully functional. One example that comes to mind is if you do an import of a local python module (I seem to remember someone, maybe matplotlib, was doing it).

sphinx_gallery/tests/test_gen_gallery.py
```diff
@@ -273,3 +273,33 @@ def test_collect_gallery_files(config_app, tmpdir):
...
+# General information about the project.
+project = u'Sphinx-Gallery <Tests>'
+
+sphinx_gallery_conf = {
+    'backreferences_dir' : os.path.join('modules', 'gen'),
+    'examples_dirs': 'src',
+    'gallery_dirs': ['ex'],
+    'binder': {'url': 'http://test1.com', 'org': 'org',
+               'repo': 'repo', 'branch': 'branch',
+               'dependencies': 'requirements.txt'}
```
lesteve: It would be nice to have a non-default notebook folder here, just to double-check that the notebook output directory is fine.

sphinx_gallery/binder.py
```diff
@@ -95,17 +117,76 @@ def gen_binder_rst(fname, binder_conf):
...
+    """Given a list of files in `contents`, remove all files named `ipynb` or
+    directories named `images` and return the result.
+
+    Used with the `shutil` "ignore" keyword to filter out non-ipynb files."""
+    contents_return = []
+    for entry in contents:
+        if entry.endswith('.ipynb'):
+            # Don't include ipynb files
+            pass
+        elif (entry != "images") and os.path.isdir(os.path.join(path, entry)):
```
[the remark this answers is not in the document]
choldgraf: not sure, this one was @Titan-C 's , I was just adapting it w/o changing functionality :-)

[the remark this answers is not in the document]
Titan-C: We keep images because is an ignore list and I don't care about them. We need the notebooks and directories containing them. The image folder just adds to the noise of folders copied.

sphinx_gallery/binder.py
Titan-C: I first proposed to use os.walk but after looking at @choldgraf's use of it I though it was a very bad idea to push forward, I needed few nested loops and you are joining paths all the time manually. I find the copytree implementation way cleaner. It copies the entire tree structure of the examples but only keeping the notebook files. It may feel counter intuitive to think about the ignore list but it's simpler to extend later on.
The ignore list is easier to extend and test as it is a separate function.

Regarding your use case of needing to copy other files that are not an example, the ignore list makes it easier. Unfortunately we will meet other roadblocks on that area regarding the way Sphinx-Gallery works. Examples are run in the `examples_dirs`, and thus look for file local to that place, in that place we put local modules, data files(like nilearn), or what ever else needed to run the script. But the `gallery_dirs` was only planned for web display, thus has only source code of executed file, rst representation, jupyter notebook representation and output images.

That means if we need to copy other files to give a full reproducible experience where we need local files, we would need to add an extra copy procedure to copy from the `examples_dirs`

Titan-C: No offence is meant to @choldgraf for his implementation. That was probably the way to use os.walk, it just didn't look good to have it.

choldgraf: No worries - I just want a clear and short path forward to get this PR merged. What shall we agree on for this particular point? Keep it the same?

choldgraf:
I believe that I addressed all of @lesteve 's comments (there are two remaining ones where it wasn't clear if/what I should change). Let me know what you guys think!

sphinx_gallery/binder.py
lesteve: I think it is fine to keep it like this for now.

lesteve:
I ran this on scikit-learn and I am seeing this warning:
```
WARNING: Duplicate file name(s) found. Having duplicate file names will break some links. List of files: ['../examples/tree/plot_iris.py']
```

Unless I am missing something, we are now mirroring the folder structure, so duplicate file names are not a problem anymore and we should remove this logic.

larsoner:
IIUC it still creates problems with image file names from one example
affecting the file names for the other (sphinx controls this to avoid
collisions). If so we should keep it.

>

Titan-C:
Here we are copying the folder structure to generate the binder links. Be it for single notebook or binder lab. In this way we should not see any problems with links, which was the unpredictable part as name collisions would change the filenames when they were moved to the `_downloads` folder. Regardless of that, keeping this as a warning is very useful and we should keep it.

lesteve:
OK fine for leaving it as is. I checked and found https://github.com/sphinx-gallery/sphinx-gallery/pull/244#issuecomment-335531679 for example. Although the warnings was introduced in a binder context but you could argue that it makes sense in general.

larsoner:
Yes the problem is more general than the binder context.

>

choldgraf:
so.......is there anything for me to do here? :-)

lesteve:
I pushed some tweaks, merging, thanks a lot @choldgraf, great stuff !

choldgraf:
woooo!!!! 🎉 🎉 🎉 🎉

PR #372, opened 2018-04-29
[MRG] Fix: Gallery name for paths ending with '/'

GaelVaroquaux:
If the user enters in the conf.py paths that end by '/', such as
'examples/', the filename of the downloads was '_python.zip'. This
commit fixes it.

### DIFF
--- a/sphinx_gallery/downloads.py
+++ b/sphinx_gallery/downloads.py
@@ -64,14 +64,7 @@ def python_zip(file_list, gallery_path, extension='.py'):
         zip file name, written as `target_dir_{python,jupyter}.zip`
         depending on the extension
     """
-    path_components = os.path.split(gallery_path)
-    if path_components[-1]:
-        zipname = path_components[-1]
-    elif len(path_components) <= 1:
-        zipname = 'gallery'
-    else:
-        # The path entered by the user ends with '/'
-        zipname = path_components[-2]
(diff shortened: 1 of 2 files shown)


larsoner:
Maybe it's worth changing the `tinybuild` test to use the name with `/` at the end, and leaving the sphinx-gallery repo's `conf.py` as it is. That way both forms will be tested. I don't know an easy way to hit the third branch, unless you want to make a little test specifically for `python_zip`.

Titan-C:
I would argue for simplicity and just use the `normpath` function to solve this issue. Does this work for your case @GaelVaroquaux ?
```diff
-zipname = os.path.basename(gallery_path)
+zipname = os.path.basename(os.path.normpath(gallery_path))
```
maybe we could write a test on this one.

GaelVaroquaux:
> -zipname = os.path.basename(gallery_path)
> +zipname = os.path.basename(os.path.normpath(gallery_path))

Yes, that's a much better way of writing this. I didn't know about
normpath!

> maybe we could write a test on this one.

It's not clear to me how to do it.

larsoner:
You could a path here to have `/` at the end:

https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/tests/tinybuild/conf.py

GaelVaroquaux:
Thanks Oscar! And sorry for not finishing this PR.

PR #374, opened 2018-05-10
Show peak memory usage if memory_profiler is installed

tamasgal:
This is just a small modification which will add an extra line below the "Total running time of the script..." showing the peak memory usage.

The line will only be displayed if `memory_profiler` is installed.

Example output:

```
Total running time of the script: ( 0 minutes 0.363 seconds)

Peak memory usage: 167.664MB
```

### DIFF
--- a/continuous_integration/travis/install.sh
+++ b/continuous_integration/travis/install.sh
@@ -39,6 +39,7 @@ elif [ "$DISTRIB" == "ubuntu" ]; then
     pip install -U requests[security]  # ensure SSL certificate works
     pip install "tornado<5"
     pip install -r requirements.txt
+    pip uninstall memory_profiler  # test show_memory=True without memory_profiler
     pip install seaborn sphinx==1.5.5 pytest "six>=1.10.0" pytest-cov
 else
     echo "invalid value for DISTRIB environment variable: $DISTRIB"
--- a/continuous_integration/travis/install.sh
+++ b/continuous_integration/travis/install.sh
@@ -39,7 +39,7 @@ elif [ "$DISTRIB" == "ubuntu" ]; then
     pip install -U requests[security]  # ensure SSL certificate works
     pip install "tornado<5"
     pip install -r requirements.txt
-    pip uninstall memory_profiler  # test show_memory=True without memory_profiler
+    pip uninstall -y memory_profiler  # test show_memory=True without memory_profiler
     pip install seaborn sphinx==1.5.5 pytest "six>=1.10.0" pytest-cov
 else
     echo "invalid value for DISTRIB environment variable: $DISTRIB"
--- a/continuous_integration/travis/install.sh
+++ b/continuous_integration/travis/install.sh
@@ -18,7 +18,8 @@ if [ "$DISTRIB" == "conda" ]; then
     # force no mkl because mayavi requires old version of numpy
     # which then crashes with pandas and seaborn
     conda create --yes -n testenv python=$PYTHON_VERSION pip nomkl numpy\
-        setuptools matplotlib pillow pytest pytest-cov coverage seaborn
+        setuptools matplotlib pillow pytest pytest-cov coverage seaborn\
+        memory_profiler
     source activate testenv
     if [ "$INSTALL_MAYAVI" == "true" ]; then
         conda install --yes mayavi
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -730,8 +730,11 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
         if time_elapsed >= gallery_conf["min_reported_time"]:
             example_rst += ("**Total running time of the script:**"
                             " ({0: .0f} minutes {1: .3f} seconds)\n\n"
-                            "**Peak memory usage:** {2: .3f}MB\n\n".format(
-                                time_m, time_s, peak_mem))
+                            .format(time_m, time_s))
+        if PROFILE_MEMORY:
+            example_rst += ("**Peak memory usage:** {2: .3f}MB\n\n"
+                            .format(time_m, time_s, peak_mem))
+
         # Generate a binder URL if specified
         binder_badge_rst = ''
         if len(binder_conf) > 0:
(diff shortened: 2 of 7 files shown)


larsoner:
Perhaps there should be a `show_memory=False` default config value. If `True` and no `memory_profiler` is installed, a warning is emitted (so we don't prevent users from building docs for repos that want this option on for CI-built docs, for example). WDYT?

larsoner:
(And this will be great to have, I will enable it once it's in!)

sphinx_gallery/gen_rst.py
```diff
@@ -707,15 +721,20 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
...
                      mode='w', encoding='utf-8') as f:
         if time_elapsed >= gallery_conf["min_reported_time"]:
             example_rst += ("**Total running time of the script:**"
-                            " ({0: .0f} minutes {1: .3f} seconds)\n\n".format(
-                                time_m, time_s))
+                            " ({0: .0f} minutes {1: .3f} seconds)\n\n"
+                            .format(time_m, time_s))
+        if PROFILE_MEMORY:
+            example_rst += ("**Peak memory usage:** {2: .3f}MB\n\n"
+                            .format(time_m, time_s, peak_mem))
```
larsoner: Why have `time_m` and `time_s` here at all rather than just using `{0:` and one `arg`?

Also is it worth adapting `sizeof_fmt` from [here](https://github.com/mne-tools/mne-python/blob/master/mne/utils.py#L2027) so that large numbers of MB get rendered as GB? You could just look for `// 1000 > 0` and format as `GB` directly, but I'm thinking such a function could also be useful elsewhere eventually.

tamasgal: Ah that was a mistake, sorry.

I'll also figure out how to add that option, thanks for the comments!

larsoner:
... it would be good to add this `memory_profiler` to the CIs (Travis, CircleCI, AppVeyor) so we can see it in action and have it tested. If you add the `show_memory` config option, you can change the `tinybuild` test `conf.py` to have `show_memory=True` to explicitly test it with Travis, especially if you add some `assert "Peak memory" in ...` statement in an appropriate place.

tamasgal:
I just realised what you meant with `tinybuild`. Let me fix that quickly.

Btw. I have not added a test case for the `show_memory` yet; I need to figure out how the test suite works...

larsoner:
If you add it to `tinybuild` your test can consist of inspecting the HTML for any example output to see if it contains `Peak memory`

larsoner:
i.e. add another `assert` statement here

https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/tests/test_full.py#L62

larsoner:
Can you post a link to the CircleCI build once it's green?

(If you don't know how, click the "Details" link to the right of the CI, log in to CircleCI with your GitHub account, then navigate to the "Artifacts" tab at the top of the build page for your PR in the `sphinx-gallery` repo, and paste a link to one example with the output.)

larsoner:
For some reason `memory_profiler` does not show up in the Travis log for installs

https://travis-ci.org/sphinx-gallery/sphinx-gallery/jobs/377451729

larsoner:
In addition to fixing Travis so `memory_profiler` is actually installed, in your test it would be better to do:
```
try:
    import memory_profiler
except ImportError:
    assert "Peak memory" not in lines
else:
    assert "Peak memory" in lines
```
Then for one of the travis builds, add a call to `pip uninstall memory_profiler` so the `except` branch of the test gets hit

continuous_integration/travis/install.sh
```diff
@@ -18,7 +18,8 @@ if [ "$DISTRIB" == "conda" ]; then
     # force no mkl because mayavi requires old version of numpy
     # which then crashes with pandas and seaborn
     conda create --yes -n testenv python=$PYTHON_VERSION pip nomkl numpy\
-        setuptools matplotlib pillow pytest pytest-cov coverage seaborn
+        setuptools matplotlib pillow pytest pytest-cov coverage seaborn\
+        memory_profiler
```
larsoner: for example you could add to the `DISTRIB == ubnutu` case below:
```
pip uninstall memory_profiler
```
to hit the `except` lines you have now added

tamasgal: Done, let's see what the CI says...

tamasgal:
While waiting on Travis and AppVeyor, here is the requested link to a gallery example with shown peak memory usage:
https://444-25860190-gh.circle-artifacts.com/0/home/ubuntu/sphinx-gallery/tiny_html/auto_examples/plot_numpy_scipy.html#sphx-glr-auto-examples-plot-numpy-scipy-py

sphinx_gallery/gen_rst.py
```diff
@@ -714,8 +731,13 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
                      mode='w', encoding='utf-8') as f:
         if time_elapsed >= gallery_conf["min_reported_time"]:
             example_rst += ("**Total running time of the script:**"
-                            " ({0: .0f} minutes {1: .3f} seconds)\n\n".format(
-                                time_m, time_s))
+                            " ({0: .0f} minutes {1: .3f} seconds)\n\n"
+                            .format(time_m, time_s))
+        if show_memory:
+            peak_mem = max(memory_measurements)
+            example_rst += ("**Peak memory usage:** {0: .3f}MB\n\n"
```
larsoner: I think `.1f` (or even `d`) is sufficient here

larsoner: ... and there should be a space between the number and `MB`

tamasgal: I'll take care when I am back at my desk ;)

GaelVaroquaux:
I was thinking of merging this PR, but unfortunately, there are many conflicts. Any chance that you can fix them?

larsoner:
I would also appreciate this feature. @tamasgal let me know if you need help fixing the conflicts

larsoner:
The rebase was a pain :) I hope it's okay I took care of it @tamasgal. New PR up at #415

tamasgal:
Sorry for coming late to the RB party!

> The rebase was a pain :) I hope it's okay I took care of it @tamasgal. New PR up at #415

Absolutely, thank you @larsoner ;)

PR #375, opened 2018-05-11
Support for custom default thumbnails in 'RGBA' space

padix-key:
This PR adds support for custom default thumbnails with transparency values. Up to now, the alpha value in RGBA PNG images is simply discarded, resulting in side effects like a black box surrounding the image.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -385,7 +385,7 @@ def scale_image(in_fname, out_fname, max_width, max_height):
     # width_sc, height_sc = img.size  # necessary if using thumbnail
 
     # insert centered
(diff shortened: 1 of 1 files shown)


Titan-C:
This is probably good thing to have. I notice nevertheless that our `broken` stamp for the thumbnails becomes barely readable I tried putting a black background but it remains barely readable. Maybe we need to redo that image file.

padix-key:
I like the transparent version of ``broken`` more than the one with the black background, but alternatively one could use something like a [forbidden symbol](https://commons.wikimedia.org/wiki/File:Forbidden_Symbol_Transparent.svg).

lesteve:
I don't fully understand the change but I have to admit I prefer the "Broken" image like this. Let's merge this one, thanks @padix-key!

PR #384, opened 2018-05-29
FIX: Use unstyled pygments for output

larsoner:
We should probably use `none` as the markup language for outputs. This is more reflective of how they are rendered in the terminal (a key goal of SG). It also takes care of weird, unexpected formatting of log messages that the default `python` interpretation causes.

The only downside I can see is that `repr`s that were previously formated to look like Python objects no longer will be. But I'd argue this is reasonable behavior since this is more like what a terminal would display when you actually execute the commands.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -173,7 +173,9 @@ SINGLE_IMAGE = """
 # This one could contain unicode
 CODE_OUTPUT = u""".. rst-class:: sphx-glr-script-out
 
- Out::
+ Out:
+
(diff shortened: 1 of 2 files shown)


lesteve:
This change seems fine to me. There are some failures that seem genuine though (not sure why).

larsoner:
Better code coverage than I thought meant there was a test to update. CIs are happy now.

lesteve:
Merging, thanks!

lesteve:
> Better code coverage than I thought meant there was a test to update.

Always better this way than the other I guess ;-)

PR #385, opened 2018-05-30
FIX: Workaround for temp Sphinx 1.8.0 bug

larsoner:
The nightly cron job failed, this is a quick workaround so we can stay green.


lesteve:
Thanks a lot for this! If you believe this is a sphinx bug it would be very nice to open a sphinx issue (I quickly looked and did not find anything).

larsoner:
https://github.com/sphinx-doc/sphinx/issues/5038

lesteve:
Great thanks a lot!

PR #387, opened 2018-06-05
FIX: No more workaround

larsoner:
I suspect that this will die on older versions of Sphinx, in which case I'll probably add a context manager to the `sphinx_compatibility` submodule.

### DIFF
--- a/sphinx_gallery/tests/test_gen_gallery.py
+++ b/sphinx_gallery/tests/test_gen_gallery.py
@@ -64,8 +64,12 @@ class SphinxAppWrapper(object):
         self.kwargs = kwargs
 
     def create_sphinx_app(self):
-        app = Sphinx(self.srcdir, self.confdir, self.outdir,
-                     self.doctreedir, self.buildername, **self.kwargs)
+        # Avoid warnings about re-registration, see:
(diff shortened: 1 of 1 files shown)


Titan-C:
This look good. I think this can be merged before the release today.

lesteve:
Nice, thanks a lot for this! I saw the response on the sphinx issue but I was not sure how easy it was to use it.

larsoner:
+10/-8 easy :)

lesteve:
CIs are green merging, thanks a lot!

lesteve:
I should have tested this on sphinx master, it looks like the problem is still here: https://travis-ci.org/sphinx-gallery/sphinx-gallery/jobs/388594944 (sphinx master is only tested during the daily Travis cron job).

lesteve:
> Any objection to adding sphinx master as a standard Travis job? We don't seem to be anywhere near having problems with testing time.

The reason I put it in a daily Travis cron job is that in 99% of the cases, it's not the fault of the PR if the build with sphinx master fails at all. Also it's not really urgent generally so we can leave it fail for a few days, whereas I find having all PRs red for a few days without a good reason is annoying.

To be honest: I just followed what we do in scikit-learn for numpy, scipy, cython master and that has worked quite well for us so far.

PR #391, opened 2018-06-12
MRG: Fix for 1.8.0 dev

larsoner:
Better fix than #387. Passes on my machine™

### DIFF
--- a/sphinx_gallery/tests/test_gen_gallery.py
+++ b/sphinx_gallery/tests/test_gen_gallery.py
@@ -8,6 +8,7 @@ Test Sphinx-Gallery
 from __future__ import (division, absolute_import, print_function,
                         unicode_literals)
 import codecs
+from functools import partial
 import os
 import sys
 import re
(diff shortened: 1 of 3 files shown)


sphinx_gallery/tests/test_gen_gallery.py
```diff
@@ -68,9 +69,8 @@ def __init__(self, srcdir, confdir, outdir, doctreedir, buildername,
     def create_sphinx_app(self):
         # Avoid warnings about re-registration, see:
```
lesteve: You probably want to remove the comment here and move it where it belongs.

larsoner:
> Do you understand why it works?

Yes basically you need to use `docutils_namespace` context manager during each build. Before it was only being used when instantiating the `Sphinx` instance, which is not enough. And we weren't doing it for all `Sphinx` instances (one was missing for `tinybuild`).

Other comment addressed in latest commit.

larsoner:
@lesteve okay to merge this? That way we stop the flow of daily Travis-is-sad emails :)

larsoner:
> Before it was only being used when instantiating the Sphinx instance, which is not enough.

I thought this was the case but your commit appears to work on my machine with 1.8.0.dev. So this should be good to go with your reversion to the simpler version + tweaks

larsoner:
That's great, I didn't like having to manually call the `__enter__` and `__exit__` methods

PR #392, opened 2018-06-12
MRG: Move to CircleCI V2

larsoner:
Let's see if this works


larsoner:
Looks like it works:

https://circleci.com/gh/sphinx-gallery/sphinx-gallery/495#artifacts/containers/0

Can't check the deploy step until it's in `master` but in theory that should work since it was just a script.

lesteve:
Had a quick look at it and it looks fine, let's merge and see what happens for the deploy step! Thanks a lot!

lesteve:
So crossing fingers was not enough ... I pushed some fixes in master but it still fails for some reason. I think the best thing to do in this case is to setup CircleCI on one's fork and debug the problem.

larsoner:
I'm going to "rerun job with SSH". One of the wonderful features of CircleCI is being able to debug this way :)

larsoner:
When running with SSH I found that it was a key issue. From looking at docs (and what SciPy does) I added this:

https://github.com/sphinx-gallery/sphinx-gallery/commit/ed99ab5bd7ce55383d09d9d9bb393529f8358fd5

And it's green now

lesteve:
Great thanks a lot!

PR #400, opened 2018-07-25
ENH: update only changed examples

### LINKED ISSUE #394: Unnecessary regeneration of API docs

jklymak:
Closes #394, #395 (or I think it does)...

This speeds things up a bit on the sphinx-gallery docs.  Not a huge improvement, though...

### Master:

First run: 10.69 s
second run 3.55 s

### This PR:

first run: 10.65 s
second run: 3.11 s

### DIFF
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -155,8 +155,11 @@ def scan_used_functions(example_file, gallery_conf):
     example_code_obj = identify_names(example_file)
     if example_code_obj:
         codeobj_fname = example_file[:-3] + '_codeobj.pickle'
-        with open(codeobj_fname, 'wb') as fid:
-            pickle.dump(example_code_obj, fid, pickle.HIGHEST_PROTOCOL)
+        if (not os.path.isfile(codeobj_fname) or
+                os.stat(example_file).st_mtime >
+                os.stat(codeobj_fname).st_mtime):
+            with open(codeobj_fname, 'wb') as fid:
+                pickle.dump(example_code_obj, fid, pickle.HIGHEST_PROTOCOL)
 
     backrefs = set('{module_short}.{name}'.format(**entry)
                    for entry in example_code_obj.values()
@@ -217,13 +220,16 @@ def write_backreferences(seen_backrefs, gallery_conf,
         include_path = os.path.join(gallery_conf['src_dir'],
                                     gallery_conf['backreferences_dir'],
                                     '%s.examples' % backref)
-        seen = backref in seen_backrefs
-        with codecs.open(include_path, 'a' if seen else 'w',
-                         encoding='utf-8') as ex_file:
-            if not seen:
-                heading = '\n\nExamples using ``%s``' % backref
-                ex_file.write(heading + '\n')
-                ex_file.write('^' * len(heading) + '\n')
-            ex_file.write(_thumbnail_div(build_target_dir, fname, snippet,
-                                         is_backref=True))
-            seen_backrefs.add(backref)
+        # only update .example file if example_file is newer:
+        if (not os.path.isfile(include_path) or
+                os.stat(example_file).st_mtime > os.stat(include_path).st_mtime):
+            seen = backref in seen_backrefs
+            with codecs.open(include_path, 'a' if seen else 'w',
+                             encoding='utf-8') as ex_file:
+                if not seen:
+                    heading = '\n\nExamples using ``%s``' % backref
+                    ex_file.write(heading + '\n')
+                    ex_file.write('^' * len(heading) + '\n')
+                ex_file.write(_thumbnail_div(build_target_dir, fname, snippet,
+                                             is_backref=True))
+                seen_backrefs.add(backref)
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -17,6 +17,8 @@ from __future__ import division, print_function, absolute_import
 from time import time
 import ast
 import codecs
+import distutils
+from distutils import file_util
 import hashlib
 import os
 import re
@@ -611,7 +613,7 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf):
     binder_conf = check_binder_conf(gallery_conf.get('binder'))
     src_file = os.path.normpath(os.path.join(src_dir, fname))
     example_file = os.path.join(target_dir, fname)
-    shutil.copyfile(src_file, example_file)
+    distutils.file_util.copy_file(src_file, example_file, update=1)
     file_conf, script_blocks = split_code_and_text_blocks(src_file)
     intro, title = extract_intro_and_title(fname, script_blocks[0][1])
 


sphinx_gallery/gen_rst.py
```diff
@@ -17,6 +17,8 @@
 from time import time
 import ast
 import codecs
+import distutils
```
larsoner: You shouldn't need this import. You can do just `import distutils.file_util`

jklymak:
> Have you tried it with your actual use case (matplotlib)? What kind of build speedup do you see there when modifying one example?

Yes, I tried with matplotlib.  I still don't fully understand the sphinx build, but the real slow down is the API pages that get generated.  In matplotlib the build makes `api/_as_gen/api/_as_gen/matplotlib.pyplot.get_fignums.rst` and many thousand others, as well as `api/_as_gen/matplotlib.pyplot.get_fignums.examples`.  With this change, these are both "old" when I do an incremental build, yet the html still gets rebuilt:

```
-rw-r--r-- 1 jklymak staff   0 Jul 24 15:37 api/_as_gen/matplotlib.pyplot.get_fignums.examples
-rw-r--r-- 1 jklymak staff 241 Jul 24 15:31 api/_as_gen/matplotlib.pyplot.get_fignums.rst
```
and
```
-rw-r--r-- 1 jklymak staff 11K Jul 25 06:41 build/html/api/_as_gen/matplotlib.pyplot.get_fignums.html
```

Maybe its just my ignorance, but I thought that the html would not be built if nothing changed in the repo.

larsoner:
So you don't get any meaningful gains?

What if you disable SG entirely so that the .examples does not need to be incorporated, do you get fast subsequent builds in that case? If so maybe it has to do with including one file in another always causing a rebuild (Sphinx bug or limitation).

Titan-C:
Are you sure the API generation is the bottle neck. To me this is a fast process. I feel slow process are more on the side of embedding links in the code, or sphinx itself reading all available rst to be processed.

sphinx_gallery/backreferences.py
```diff
@@ -155,8 +155,11 @@ def scan_used_functions(example_file, gallery_conf):
     example_code_obj = identify_names(example_file)
     if example_code_obj:
         codeobj_fname = example_file[:-3] + '_codeobj.pickle'
-        with open(codeobj_fname, 'wb') as fid:
-            pickle.dump(example_code_obj, fid, pickle.HIGHEST_PROTOCOL)
+        if (not os.path.isfile(codeobj_fname) or
+                os.stat(example_file).st_mtime >
+                os.stat(codeobj_fname).st_mtime):
```
Titan-C: We abandoned checking against timestamps long ago. They lead to lots of problems. I Try using the change against the md5sum of the sourcefile.

jklymak:
I’ll test properly, there may have indeed been gains.  But the matpltolib build is still pretty slow despite nothing changing in the source tree.   If the example files were all changing every build that made sense to me, but now I guess something else triggers the API rebuild.

jklymak:
> Are you sure the API generation is the bottle neck. 

Ha, no, not sure at all.  With this change, it spends a lot of time on the `writing output...` step in Matplotlib.  

As for how to best detect if a file is modified or not, more than happy to do what works well across different platforms.  

Timing on matplotlib incremental build:

### Before

Lots of time on `reading sources...` and `writing output...`

101 s

### After

Almost no time on `reading sources...`  still lots of time on `writing output...`
71.56s

sphinx_gallery/backreferences.py
jklymak: ... actually walk me through this.  The source and destination files are not the same in the `.example` case.  Are you sugesting caching the `md5sum` of the source files between runs somewhere?

jklymak:
OK, so my tests on this are that I run the `sphinx-gallery` doc build, not changing the sources between builds.   

If I disable *all* the sphinx.extensions including sphinx-gallery, the incremental build behaves as expected and doesn't output new html files.  If I enable `sphinx-gallery` it rebuilds all the html files, even with the changes in this PR.  So I'm still not clear what causes this rebuild step just within `sphinx-gallery`.

sphinx_gallery/backreferences.py
larsoner: @Titan-C I'm not sure how this could be done unless we embed the MD5s from the previous run somewhere. I guess we could start doing this.

It seems like we could at least allow a default `rebuild = True` default config arg (current behavior), or `rebuild='mtime'` (proposed changes here), and later we could add `rebuild='md5'` which would somehow keep track of the MD5 sums. That would provide a way forward without adding too many new complications for people who are satisfied with mtime.

larsoner:
> If I disable all the sphinx.extensions including sphinx-gallery, the incremental build behaves as expected and doesn't output new html files. If I enable sphinx-gallery it rebuilds all the html files, even with the changes in this PR. So I'm still not clear what causes this rebuild step just within sphinx-gallery.

Could it be `sphinx.ext.autosummary`? It might always regenerate API pages no matter what.

I would try disabling all to get a fast build, then reenabling them one by one to see what causes the slowdown.

Titan-C:
> @Titan-C I'm not sure how this could be done unless we embed the MD5s from the previous run somewhere. I guess we could start doing this.

Indeed, we can store that in the codeobj.pickle. Although, If my memory
is not failing that file is saved in the same place as the md5 hashfile,
so we could just read from disk without the need of changing to many things.

NicolasHug:
Hi everyone, I'm super interested in this feature because it makes the scikit-learn doc build much slower than it could be.

I may have missed something important but it seems that the following changes do pretty much what I need: rebuilding the docs go from about 2min to a few seconds.

Tests are passing.

It is using mtime, but that's also what sphinx is doing.

```diff
diff --git a/sphinx_gallery/gen_rst.py b/sphinx_gallery/gen_rst.py
index 9e5829d..c5388c0 100644
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -325,6 +325,10 @@ def generate_dir_rst(src_dir, target_dir, gallery_conf, seen_backrefs):
         length=len(sorted_listdir))
     clean_modules(gallery_conf, src_dir)  # fix gh-316
     for fname in iterator:
+        src_file = os.path.normpath(os.path.join(src_dir, fname))
+        target_file = os.path.join(target_dir, fname)
+        if os.path.exists(target_file) and os.path.getmtime(src_file) <= os.path.getmtime(target_file):
+            continue
         intro, time_elapsed = generate_file_rst(
             fname, target_dir, src_dir, gallery_conf)
         clean_modules(gallery_conf, fname)
```

NicolasHug:
BTW I'm happy to submit a PR if this is a valid solution (but of course @jklymak feel free to use it here)

larsoner:
I'm fine with using `mtime` even if it's not perfect. Especially if it's what `sphinx` itself does, we're even being consistent.

larsoner:
(see API proposal [above](https://github.com/sphinx-gallery/sphinx-gallery/pull/400#discussion_r233594218) about allowing different values of a new `rebuild` config var)

jklymak:
Feel free to close this PR and open another.  Or to add the two together. My change didn’t make things that much faster for matplotlib  but if the above makes things faster I think it would be nice if it were an option.

jklymak:
Closing in lieu of #448!

PR #401, opened 2018-08-13
adding ability to customize first cell of notebooks

### LINKED ISSUE #388: matplotlib inline vs notebook
Currently sphinx-gallery puts ["% matplotlib inline"](https://github.com/sphinx-gallery/sphinx-gallery/blob/830a11c6ba5a80a4cbbaff7e2932dd37e81321a5/sphinx_gallery/notebook.py#L113) at the top of all notebooks.  Would it be possible to add an option to use "% matplotlib notebook" instead?  Interactive elements in figures seem to work better with notebook.

choldgraf:
This adds a configuration parameter `first_notebook_cell` that lets users choose what text goes in the first cell of created notebooks. By default this remains the same (`%matplotlib inline`).

This could be useful for either other packages that need their own special configuration in a notebook context, or for people that want to run the notebook with the "interactive" matplotlib setting (`%matplotlib notebook`).

closes #388 and relevant to https://github.com/matplotlib/matplotlib/issues/11415

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -348,5 +348,8 @@ sphinx_gallery_conf = {
                'dependencies': './binder/requirements.txt',
                'notebooks_dir': 'notebooks',
                'use_jupyter_lab': True,
-               }
+               },
+    'first_notebook_cell': ("# This cell is added by sphinx-gallery\n"
+                            "# It can be customized to whatever you like\n"
+                            "%matplotlib inline")
 }
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -349,7 +349,4 @@ sphinx_gallery_conf = {
                'notebooks_dir': 'notebooks',
                'use_jupyter_lab': True,
                },
-    'first_notebook_cell': ("# This cell is added by sphinx-gallery\n"
-                            "# It can be customized to whatever you like\n"
-                            "%matplotlib inline")
 }
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -62,6 +62,7 @@ DEFAULT_GALLERY_CONF = {
     'binder': {},
     'image_scrapers': ('matplotlib',),
     'reset_modules': ('matplotlib', 'seaborn'),
+    'first_notebook_cell': '%matplotlib inline'
 }
 
 logger = sphinx_compatibility.getLogger('sphinx-gallery')
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -104,6 +104,7 @@ def parse_config(app):
     gallery_conf = _complete_gallery_conf(
         app.config.sphinx_gallery_conf, src_dir, plot_gallery,
         abort_on_example_error)
+
     # this assures I can call the config in other places
     app.config.sphinx_gallery_conf = gallery_conf
     app.config.html_static_path.append(glr_path_static())
@@ -172,16 +173,11 @@ def _complete_gallery_conf(sphinx_gallery_conf, src_dir, plot_gallery,
     del resetters
 
     # Ensure the first cell text is a string if we have it
-    first_cell = gallery_conf.get("first_notebook_cell", "%matplotlib inline")
+    first_cell = gallery_conf.get("first_notebook_cell")
     if not isinstance(first_cell, str):
         raise ValueError("The 'first_notebook_cell' parameter must be type str"
                          "found type %s" % type(first_cell))
     gallery_conf['first_notebook_cell'] = first_cell
-
-    # this assures I can call the config in other places
-    app.config.sphinx_gallery_conf = gallery_conf
-    app.config.html_static_path.append(glr_path_static())
-
     return gallery_conf
 
 
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -171,6 +171,17 @@ def _complete_gallery_conf(sphinx_gallery_conf, src_dir, plot_gallery,
     gallery_conf['reset_modules'] = tuple(resetters)
     del resetters
 
+    # Ensure the first cell text is a string if we have it
+    first_cell = gallery_conf.get("first_notebook_cell", "%matplotlib inline")
+    if not isinstance(first_cell, str):
+        raise ValueError("The 'first_notebook_cell' parameter must be type str"
+                         "found type %s" % type(first_cell))
+    gallery_conf['first_notebook_cell'] = first_cell
+
+    # this assures I can call the config in other places
+    app.config.sphinx_gallery_conf = gallery_conf
+    app.config.html_static_path.append(glr_path_static())
+
     return gallery_conf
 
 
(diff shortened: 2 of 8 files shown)


doc/conf.py
```diff
@@ -348,5 +348,8 @@ def setup(app):
                'dependencies': './binder/requirements.txt',
                'notebooks_dir': 'notebooks',
                'use_jupyter_lab': True,
-               }
+               },
+    'first_notebook_cell': ("# This cell is added by sphinx-gallery\n"
+                            "# It can be customized to whatever you like\n"
+                            "%matplotlib inline")
```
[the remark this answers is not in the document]
choldgraf: Isn't the documentation of sphinx-gallery only for developers? Who do we imagine would be browsing the sphinx-gallery documentation that is not a developer?

sphinx_gallery/tests/test_notebook.py
```diff
@@ -81,7 +81,8 @@ def test_convert():
 def test_jupyter_notebook():
     """Test that written ipython notebook file corresponds to python object"""
     file_conf, blocks = sg.split_code_and_text_blocks('tutorials/plot_parse.py')
-    example_nb = jupyter_notebook(blocks)
+    gallery_conf = {}
+    example_nb = jupyter_notebook(blocks, gallery_conf)
```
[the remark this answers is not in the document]
choldgraf: do you mean the `def conf_file(request)` functions? or the `sphinx_app_wrapper` function in `test_gen_gallery`?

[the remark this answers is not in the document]
Titan-C: If memory does not fail me you should be able to do:
```python
import pytest
def test_jupyter_notebook(gallery_conf):
    example_nb = jupyter_notebook(blocks, gallery_conf)
```

doc/conf.py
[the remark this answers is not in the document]
Titan-C: Yes, Sphinx-Gallery docs are for developers. But putting this in the `conf.py` makes the comment appear in every notebook we generate. If we play the role of having sensible defaults, I imagine is not a good idea to put this reminder as a default. Developers might think they need to configure it always. Just show the plain defaults.

sphinx_gallery/tests/test_utils.py
```diff
@@ -10,10 +10,11 @@
 
 from __future__ import division, absolute_import, print_function
 import sphinx_gallery.utils as utils
+from sphinx_gallery.tests.test_gen_gallery import sphinx_app_wrapper
```
Titan-C: Ignorance speaking. Why do you need this here?

sphinx_gallery/gen_gallery.py
```diff
@@ -170,6 +171,17 @@ def _complete_gallery_conf(sphinx_gallery_conf, src_dir, plot_gallery,
...
+    # Ensure the first cell text is a string if we have it
+    first_cell = gallery_conf.get("first_notebook_cell", "%matplotlib inline")
+    if not isinstance(first_cell, str):
+        raise ValueError("The 'first_notebook_cell' parameter must be type str"
+                         "found type %s" % type(first_cell))
+    gallery_conf['first_notebook_cell'] = first_cell
+
+    # this assures I can call the config in other places
+    app.config.sphinx_gallery_conf = gallery_conf
+    app.config.html_static_path.append(glr_path_static())
```
Titan-C: Doesn't this crash? `app` is not in scope of this function. Or is this a rebase issue. This lines belong in the `parse_config` function

choldgraf: I have no idea how these lines got here lol, you are correct that it crashes...but for some reason it wasn't crashing on my machine. I'll remove those last two lines and rewrite this so it's more similar to the other code in the section

sphinx_gallery/tests/test_utils.py
choldgraf: was a remnant from an earlier part of this PR - removing in next push

choldgraf:
@Titan-C I think I got those comments addressed, LMK what you think!

sphinx_gallery/notebook.py
```diff
@@ -180,6 +183,7 @@ def python_to_jupyter_cli(args=None, namespace=None):
 
     Takes the same arguments as ArgumentParser.parse_args
     """
+    from . import gen_gallery  # To avoid circular import
```
Titan-C: Not sure where the circular import is, but I'll trust you on this one.

choldgraf:
ping :-)

sphinx_gallery/notebook.py
```diff
@@ -191,5 +195,6 @@ def python_to_jupyter_cli(args=None, namespace=None):
     for src_file in args.python_src_file:
         file_conf, blocks = split_code_and_text_blocks(src_file)
         print('Converting {0}'.format(src_file))
-        example_nb = jupyter_notebook(blocks)
+        gallery_conf = copy.deepcopy(gen_gallery.DEFAULT_GALLERY_CONF)
```
larsoner: Sorry if I missed it, why doesn't this actually use the currently active gallery conf and just use the default one instead? It's worth adding a comment to justify because it's strange

choldgraf: Maybe I'm not understanding the CLI, but it seemed to me that there is no "gallery configuration" when using the CLI, so I was just copying the default one. Maybe put another way, what do you mean by "use the currently active gallery conf" in the context of the CLI?

larsoner: What do you mean by CLI? You mean calling `sphinx_build` or something else?

sphinx_gallery/gen_gallery.py
```diff
@@ -170,6 +172,12 @@ def _complete_gallery_conf(sphinx_gallery_conf, src_dir, plot_gallery,
     gallery_conf['reset_modules'] = tuple(resetters)
     del resetters
 
+    # Ensure the first cell text is a string if we have it
+    first_cell = gallery_conf.get("first_notebook_cell")
+    if not isinstance(first_cell, str):
```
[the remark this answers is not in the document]
Titan-C: Yes, this has to be fixed. I missed this. I also want to drop py2 support.

sphinx_gallery/notebook.py
Titan-C: This is just an utility function. It is not really related to the Sphinx-Gallery build, but a way to circumvent it just for one task. What it does is to directly parse the python script and output a jupyter notebook. I think @choldgraf is right in doing the copy of the default config here. Also this function has no access to the configuration at build time, and there is no guarantee this would even be executed within a project with Sphinx-Gallery for us to search for that config file. Finally I could be wrong, but I guess not many people use this feature, although the reverse (from notebook to script) has been requested. Thus we can leave this default for now.

larsoner: Ahh okay, makes sense.

sphinx_gallery/gen_gallery.py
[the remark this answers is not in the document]
larsoner: I think we should drop support the same time NumPy and SciPy do. New releases in 2019+ will not support it.

choldgraf:
ok, latest push fixes the basestring suggestion from @larsoner and I think that was the last comment...RTM from my end.

sphinx_gallery/notebook.py
larsoner: I know the TLA @choldgraf :) I just had no idea SG had a CLI aside from using it with `sphinx-build` calls!

larsoner:
Thanks @choldgraf

PR #410, opened 2018-08-29
FIX: Travis

larsoner:
Sometimes `conda` does not resolve dependencies fully. Let's see if this makes CIs green. Currently we have one failing from the cron job:

https://travis-ci.org/sphinx-gallery/sphinx-gallery/jobs/421905303

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 sudo: false
+dist: trusty
 addons:
   apt:
     packages:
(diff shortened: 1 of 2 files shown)


larsoner:
@choldgraf this one looks like it might finally come back clean. Okay with you if I merge once it is?

larsoner:
In it goes to make things green, but if there are problems I'm happy to follow up with another PR (though things will change quite soon after we release, so probably won't matter)

choldgraf:
Just following up on this that Travis looks happy!

Issue #407, opened 2018-08-29
Only support the latest 2 or 3 Sphinx versions


choldgraf:
In https://github.com/sphinx-gallery/sphinx-gallery/issues/406 we decided that we can safely have Sphinx-Gallery only support the latest 2 or 3 Sphinx versions. We can be a bit more aggressive with dropping support for old versions of Sphinx because Sphinx-Gallery is more "developer-facing" than many other analysis packages, and so we assume our users are more familiar with installing many versions of Python if need be.

We weren't quite clear yet on whether we'll support the latest 2 or the latest 3 Sphinx versions. I'm +1 on supporting the last 2. We should figure that out before this issue is closed!

Issue #405, opened 2018-08-29
Drop Python 2.X support


choldgraf:
in https://github.com/sphinx-gallery/sphinx-gallery/issues/404 we decided that in the release-after-next we'll drop Python 2.x support. This issue is to discuss and/or keep track of this decision!

PR #416, opened 2018-09-10
MRG: Remove outdated cron job

larsoner:
Our CRON job [failed](https://travis-ci.org/sphinx-gallery/sphinx-gallery/jobs/426532591#L761) because:
```
ERROR: Sphinx requires at least Python 3.5 to run.
```
This PR thus removes the 2.7 + Sphinx `master` cron job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,9 +37,6 @@ matrix:
     - os: linux
       env: DISTRIB="conda" PYTHON_VERSION="3.6" LOCALE=C
     - os: linux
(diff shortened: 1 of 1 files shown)


Titan-C:
True. Should not enforce legacy python on the dev versions. Good to see we are dropping support on Py2.7. 
In accordance to #405 and #407, this is OK. +1

larsoner:
Just to be clear, we are not fully dropping support yet. This just removes the 2.7 + sphinx dev run. There are other 2.7 runs. My understanding is that we will drop support for 2.7 after the next release, which is more or less waiting on the Bokeh PR so we can make any necessary tweaks to the scraping API.

PR #417, opened 2018-09-17
Scraper documentation and an image file path scraper

choldgraf:
This PR does two things:

1. Following discussion in #150, it factors out and expands the instructions for creating a custom image scraper into the `advanced.rst` section.
2. It also implements an `ImageFileScraper` as a default scraper. This will check the folder of an example for any image files that have been modified since the script first started running. If so, it will embed that image file after the latest-executed cell.

I mostly did 2 because I had to figure out how the image scraping stuff worked in order to do number 1, and thought this might be useful both for GIFs as well as for other packages that can save to images.

@larsoner what do you think since you are the image scraper guru :-)  If people are +1 on the ImageFileScraper I'll write up some tests soon as well.

## Links

[New advanced usage section](https://602-25860190-gh.circle-artifacts.com/0/alabaster_html/advanced.html#writing-a-custom-image-scraper)
[Example of new ImageFileScraper functionality](https://602-25860190-gh.circle-artifacts.com/0/alabaster_html/auto_examples/plot_gif.html#sphx-glr-auto-examples-plot-gif-py)


t-makaro:
I haven't tried this out yet but:
```
Sphinx-gallery will only display an image file if it has been modified since the execution
of the Python file. It will not display a pre-existing image file (though you can always
refer to these by directly embedding them in your rST). 
```
So if I run my build, it will properly collect the images and build my docs, but if I run my build a second time without cleaning out the old images, then it will miss them? 

Since sphinx-gallery already has restrictions on the file names that are run, maybe a restriction on files that are collected by default. (File name: ```plot_<name>.py``` gets run, so maybe files ```plot_<name><number>.<format>``` get collected.)

choldgraf:
> So if I run my build, it will properly collect the images and build my docs, but if I run my build a second time without cleaning out the old images, then it will miss them? If I refer to the image file by manually specifying in reST, then will the file be displayed twice if I did clear out the old image files?

As long as the example is run a second time, then it'll update the image. It just checks whether the "last modified" timestamp has changed from when the script first executed. If you manually refer to the image with rST as well as creating the file in the example, then yeah it'd be double-included.

larsoner:
Let me know when I should look, thanks @choldgraf

choldgraf:
@larsoner you can take a look now, the main thing I'd add if we're +1 on this approach is tests

t-makaro:
So, I've been [trying this out](https://github.com/t-makaro/animatplot/pull/14).
Here is a [link](https://119-87026896-gh.circle-artifacts.com/0/home/circleci/project/docs/build/html/auto_examples/plot_sqwell.html#sphx-glr-auto-examples-plot-sqwell-py) to what I got working.

Few things I've noticed:
1. The gif does get picked up and copied to the gallery directory. Huzzah!

2. The gif does not automatically get added to the .rst file. I have to manually add:
```
.. image:: plot_sqwell.gif
```
to the string at the top of the ```.py``` file. Which I could live with. (I like the gif at the top more anyway)

3. I am using the ImageFileScape to replace the matplotlib scraper, so I don't get thumbnails. Is there anyway to either:
 a. Use the gifs as the thumbnails? or
 b. Use the matplotlib scraper just for the thumbnail and not automatically add the image to the generated .rst file?

4. The first time I run my build I get:
```
docs\source\auto_examples\plot_sqwell.rst:20: WARNING: image file not readable: auto_examples\plot_sqwell.gif
```
but it builds fine.

choldgraf:
latest push adds a tests and should make the behavior a bit more reliable.

@t-makaro good question re: thumbnails. @larsoner , do you have an intuition for the best way to embed something like a GIF as a thumbnail? Maybe check if a file is a GIF, and if so save the middle frame as a PNG?

larsoner:
> Maybe check if a file is a GIF, and if so save the middle frame as a PNG?

Yeah sure. Though there is no guarantee that any frame will be a good thumbnail, one other option would be to use the frame with the highest standard deviation (of the `.ravel()` data). This at least should help prevent the thumbnail from being a blank or hard-to-see image from within the GIF. But I have no idea how people usually handle this.

choldgraf:
ugh I just realized that the "ImagePathIterator" will also break with this. It assumes that the images are always PNGs and so the copied images always end in `.png` (e.g. `image1.png`, `image2.png` etc) even if they originally did *not* have this extension (such as `.gif`).

This doesn't actually break the embedding in the sphinx docs, but it is a bit confusing. I am starting to wonder if this would actually take quite a lot of refactoring to support anything other than PNGs :-(

t-makaro:
@choldgraf Interesting.
image

image

I seem to be getting slightly different results building locally (on windows) than on circleci, but in either case the ```.gif``` file is in the gallery directory. It appears that the image gets copied twice. Once as a gif and sometimes again as a png to the sub directory. If you don't append the scrapped file (the png) to the rst, then I can add it manually just fine. It also appears that the png does animate....

t-makaro:
Ok, now I'm getting that same results localling as I was on circleci. (I'm not sure what changed). But the ```.png``` files does appear to play as a gif (and gets selected as the thumbnail properly).

t-makaro:
Some more testing. It almost works exactly as I expect.

Here is my readthedocs: https://animatplot.readthedocs.io/en/latest/auto_examples/index.html
Thumbnail is from the image_scraper, and when you click on the example the file does appear, but it does not play the gif (probably because it is actually a png). It just displays the first frame.

The strange thing is that if I build locally or on circleci instead of RTD, then the gif (which is actually a png) plays properly.

choldgraf:
that is really weird...so RTD keeps the same file names etc, but only the first frame is there?

t-makaro:
This is hard to diagnose since RTD doesn't let me download the built source files, only the a static version of the html.

When I build locally I get the follow auto_example directory:
image

Notice that I have ```plot_sqwell.gif```, and ```sphx_glr_sqwell_001.png```. Locally, the ```.png``` can be rendered as an animated gif. 

From the RTD logs, I can tell that both of these files exist, and are copied to the build directory. When I download a copy of the static html from RTD, only the ```sphx_glr_sqwell_001.png``` is there and it is only 1 frame. I think RTD isn't properly producing the original ```plot_sqwell.gif```.

t-makaro:
I opened an issue on RTD
https://github.com/rtfd/readthedocs.org/issues/4661

t-makaro:
I can now get the original ```gif``` off RTD, and it is indeed one frame. So I believe this is in issue with saving the image on RTD. Other than the RTD issue, and that it changes files to end in ```png``` this seems to work as expected.

jni:
Any chance of getting this merged (and released) before the RTD issue is fixed? We have immediate use for it in the skimage documentation... =)

larsoner:
@jni does it work for you?

jni:
I was wondering whether it would be possible to merge it without RTD support, or whether RTD is absolutely mandatory before merging.

@larsoner I haven't tried it yet, sorry, have a couple of students working on the example. I'll try to give some feedback on whether I can get it to work soon.

GaelVaroquaux:
> I was wondering whether it would be possible to merge it without RTD support,
> or whether RTD is absolutely mandatory before merging.

Many people use RTD (including sphinx-gallery itself), so it seems a bit
awkward to merge this without RTD support.

choldgraf:
@GaelVaroquaux what if we cut a release and merged this into master afterwards? Then @jni could be our first betatester as opposed to blocking the next version on this.

GaelVaroquaux:
Fine with that. 

sphinx_gallery/scrapers.py
```diff
@@ -127,9 +129,55 @@ def mayavi_scraper(block, block_vars, gallery_conf):
...
+            return ''
+        else:
+            return figure_rst(image_names, gallery_conf['src_dir'])
+
+
+def _find_images(path, image_extensions=['jpg', 'jpeg', 'png', 'gif']):
+    """Find all unique image paths for a set of extensions."""
+    image_files = set()
+    for ext in image_extensions:
+        this_ext_files = set(glob(os.path.join(path, '*.'+ext)))
```
Titan-C: I'm against of using solutions with globing in out supported ways of capturing objects. Users can have their plugins the way they want, but I disagree on having this pattern as part of our supported codebase.
I once proposed to parse the code block for additional capture comments and based on those capture the object specifically.

choldgraf: can you propose another way of accomplishing the same thing? direct pushes to this PR are welcome

larsoner:
@choldgraf can you remove the "official image scraper" from this PR? If so then this seems good to put in the 0.3.0 release. Then in the 0.4 cycle we can add the official scraper after some testing.

choldgraf:
@larsoner the latest commit removes all of the image scraper code from this PR, and I think I removed all references to it in the docs etc. It's fine to just not implement this, I get the feeling that folks are not super enthusiastic about the feature (with the exception of @jni...maybe you can read the docs on writing your own custom scraper, and then copy/paste the image scraper code from the previous commit on this PR for scikit-image? I'm happy to help with that process)

larsoner:
@Titan-C okay for you now? If so feel free to merge

jni:
I'm certainly happy to package what we need into scikit-image, thanks @choldgraf for pointing the way. But I think it's pretty crazy that people aren't more enthusiastic about this. What library couldn't use animations in their galleries? I just noticed, for example, that the Matplotlib animation gallery is just *sad*:
https://matplotlib.org/gallery/index.html#animation

As to RTD, my feeling is that as long as we're not *breaking* any existing features, people can go on happily using RTD the way they were. (Though it probably would need a slightly awkward note in the release notes, this is true.)

Titan-C:
Hi sorry for the late reply, I feel like the bottleneck here, but I have
less time to dedicate to the project. I love the idea of having this in
the docs and encouraging people to try it out under a note(I would prefer
if its a warning) that this scraper API is experimental.

In the example two, I miss a parallel to the first example. The closing
of the images to avoid future use. For example if you use a glob to get
the images. You will notice on the first run of the first example things
will go fine, the seen variable will avoid repetition. But for the
second example because seen is local to the scraper, and that is
initialized per example(I could be wrong, but even if is initialized per
build that just delays the problem), there will be images around from
the first example and all will be copied again. Say that doesn't happen,
then for the second build, the examples that are rerun again will find a
new bunch of files on disk which are not in the seen variable, and will
be copied. I would really prefer on that example at least to change from
copyfile to a move. So that images vanish from disk after
execution.

The contributions. Yes, I love contributions. But please, encourage the
custom scrapers to go to a separate repository, call it
sphinx-gallery-extra or sphinx-gallery-contrib, issue #427. Everything
that goes into the main repo is something we aim to support, let us not
pull to many things in there.

As a final thought. I'm sorry I don't have time at the moment, but if
the main goal of this PR is to support matplotlib animations. I would
encourage to capture animations the same way we capture the matplotlib
images. Through matplotlibs state variables. I'm sorry I can't propose a
draft on it for the moment.

larsoner:
... it sounds like you just want this changed to `.. warning::` instead of `.. note::` then this is okay?

choldgraf:
How about we do the following:

1. Create a new repository called "sphinx-gallery-extra" that has no "official" support from the SG team, and can be used as a sandbox to try new things.
2. Decide whether @Titan-C is happy with these changes (purely as doc changes now...more specific to @larsoner 's point, @Titan-C you are currently 'blocking changes' via the github UI)
3. I can move the actual code for the image scraper that @jni is interested in to the `sphinx-gallery-extra` repo. I'm thinking that this repo could have one folder per sub-project, so I'd likely create a folder called `image_scraper` and put the code for this class in there. Then @jni and I can iterate on this.

larsoner:
+1

choldgraf:
ok, I'm gonna create the new repository process since I have +1s from @larsoner and @GaelVaroquaux and (I think) @Titan-C

GaelVaroquaux:
Explicit +1 from me. 

doc/advanced.rst
```diff
@@ -59,3 +59,173 @@ its reference
 ``:ref:`sphx_glr_auto_examples_plot_gallery_version.py```. The image
 it generated has the name ``sphx_glr_plot_gallery_version_thumb.png``
 and its thumbnail is ``sphx_glr_plot_gallery_version_thumb.png``
+
+.. _custom_scraper:
+
+Writing a custom image scraper
+==============================
+
+.. note:: The API for custom scrapers is currently experimental.
```
choldgraf:
```suggestion
.. warning:: The API for custom scrapers is currently experimental.
```

choldgraf:
ok, just changed note to warning

t-makaro:
I like the idea of a ```sphinx-gallery-extra```. I want the image scrapper too, and it'll be nice not to have to implement it myself.

choldgraf:
for interested folks, come on over to https://github.com/sphinx-gallery/sphinx-gallery-extra/pull/1 and let's iterate there!

Titan-C:
Sorry about the block, I had no access to github. My own paranoia blocks me from knowing my password. 
Yes, I agree this change being a doc PR. Thank you for the changing the note to a warning. If you can change `shutil.copyfile` to `shutil.move` in the second example, I would be even happier.
Thank you for creating `sphinx-gallery-extra`, that would alleviate my nightmares of providing support and encourage testing things.

doc/advanced.rst
```diff
@@ -59,3 +59,173 @@ its reference
...
+
+           # Iterate through PNGs, copy them to the sphinx-gallery output directory
+           image_names = list()
+           image_path_iterator = block_vars['image_path_iterator']
+           for png in pngs:
+               if png not in seen:
+                   seen |= set(png)
+                   this_image_path = image_path_iterator.next()
+                   image_names.append(this_image_path)
+                   shutil.copyfile(png, this_image_path)
```
choldgraf:
```suggestion
                   shutil.move(png, this_image_path)
```

choldgraf:
ok I changed copyfile to "move"

choldgraf:
hey all, can somebody please merge this? It's been in PR limbo for 3 months and I think @jni would benefit from the instructions.

GaelVaroquaux:
Merging. I had given my +1 a while ago. I reviewed it again today, still +1.

PR #420, opened 2018-10-12
POC: ipyvolume support

maartenbreddels:
This is just a proof of concept, which I managed to get working just for https://github.com/maartenbreddels/ipyvolume/, and shows where we want to go with ipywidget support (and more).
I will follow up with an issue that will discuss more the idea we (@SylvainCorlay and I) have for how to include more rich output.
sphinx-gallery-poc

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -413,6 +413,21 @@ def _get_memory_base(gallery_conf):
     return memory_base
 
 
+_captured_display_calls = []
+def _capture_display(*args, **kwargs):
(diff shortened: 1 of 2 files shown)


agramfort:
I am really excited about this. May I ask how this is going to work? when you do a `make html` with sphinx you can have the rendering work without any browser? no need for headless browser?

agramfort:
just saw #421 .... sorry for the noise

maartenbreddels:
I think you could make the starting/stopping of chrome headless part of the machinery, but it does require an install of chrome.

Issue #421, opened 2018-10-12
Executing and capturing rich output using notebook / nbconvert


maartenbreddels:
@lesteve : @SylvainCorlay suggested I ping you to include you in the conversation

Dear all,

Let me start with saying this is a really awesome project, and I've already managed to get it working with ipyolume (#420) quite easily.
Together with @SylvainCorlay we have been thinking of how we can have general ipywidgets support, and even xwidget (using C++ kernels).
In parallel to this, there is this PR https://github.com/jupyter-widgets/jupyter-sphinx/pull/22 that uses the notebook/nbconvert system to execute code using an arbitrary kernel. If we also take into account the discussion about including HTML output from for instance pandas dataframes, https://github.com/sphinx-gallery/sphinx-gallery/issues/396 you can probably see an obvious direction to take.

# Using the notebook/nbconvert to execute code blocks
Leveraging nbconvert to execute code seems like a very good way forward, since you get many things for free:
 * Rich output, html, widgets, images (matplotlib included)
 * Auto display of the last expression
 * Different languages/kernels support

Note that we do not require the examples to be in a notebook, just that the code blocks as extracted by sphinx-gallary from the `.py` files to be injected into a (temporary in memory?) notebook, which can be executed by nbconvert as done in https://github.com/jupyter-widgets/jupyter-sphinx/pull/22. 

# Screenshots/thumbnails using headless chrome
Since one of the really nice things about the sphinx-gallary is the thumbnail preview, we probably also want to have screenshots generated for rich output (such as widgets). This should be possible to do with headless chrome, as shown in #420. The method should be made more general, and also requires a headless browser (chrome?) to run at readthedocs. This 

# Summary
The ideal future we see would be:
 * less code in sphinx-gallary about executing code which should help maintainability (this should be something nbconvert is good at).
 * maybe improvements to nbconvert to make this happen.
 * More language support (Python, C++, ...)
 * Rich output: html, jupyter-widgets, ...
 * A way to generate thumbnails for rich output using headless chrome (not sure where this part wil go)
 * Full backward compatibily, meaning matplotlib and mayavi will continue to work.

I/we would like to know if people are on board with this idea, and maybe get input from dev main developers who may have already thought of these ideas. Looking forward to your ideas/comments/blessings/rants.

Regards,

Maarten Breddels

GaelVaroquaux:
Thanks for the positive feedback.

One challenge is that I really don't like adding dependencies. More
[...]

PR #422, opened 2018-10-12
WIP: Using a notebook + nbconvert to execute code blocks

maartenbreddels:
This is the initial attempt to use a notebook + nbconvert to produce rich output, as explained in #421. It does not yet do any attempt to make a snapshot of the html or widget output. 
I understand the issues with the dependency in nbconvert/nbformat, but this is more of an exploration of what is possible. Note that the dependency could be totally optional.

It renders (most) of the gallery already, except for mayavi:
screen shot 2018-10-12 at 16 49 58

I added an pandas dataframe example (rendering html) and ipyvolume (rendering a live widget). It would be interesting to connect the widget to a live mybinder kernel!.

# Pandas dataframe / html
screen shot 2018-10-12 at 16 54 57
screen shot 2018-10-12 at 16 55 28

# ipyvolume / ipywidgets
screen shot 2018-10-12 at 16 56 49

To use notebook+nbconvert you need to set the configuration option:
`'executor': 'notebook'`, and optionally set the `jupyter_kernel` to something else then the default value of 'python3'.

For other languages, such as C++, we need some language specific template that are injected into the notebook for the timing and memory usage.

### DIFF
--- a/continuous_integration/travis/install.sh
+++ b/continuous_integration/travis/install.sh
@@ -16,7 +16,7 @@ if [ "$DISTRIB" == "conda" ]; then
     conda update -y conda
 
     # Force conda to think about other dependencies that can break
-    export CONDA_PKGS="python=$PYTHON_VERSION pip numpy scipy setuptools matplotlib pillow pytest pytest-cov coverage seaborn sphinx_rtd_theme memory_profiler"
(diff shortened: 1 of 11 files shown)


choldgraf:
hey @maartenbreddels - looks interesting! Do you think you could give an idea for the main "pros" of doing things this way? E.g., since Sphinx-Gallery already has its own "run the notebook and collect results" script, what would the main benefit be here? Simplification of the build process by offloading onto a dependency?

PR #220, opened 2017-04-07
Add optional pypandoc rst parsing

chsasank:
Please see #219 for discussion on this. Fix #219.

Can somebody point to me where I should add optional requirements.

PR #423, opened 2018-10-14
adding a notebook to python converter

### LINKED ISSUE #245: Support jupyter notebooks
I've had a few people ask me whether they can simply write their examples in jupyter notebooks, rather than writing them as `.py` files first. Is there any plan to support things like `plot_example.ipynb`?

choldgraf:
Over the months we've gotten several requests to support people storing examples as Jupyter Notebooks. Recently I also found there's even a fork of sphinx-gallery whose *primary* feature is just to support docs w/ notebooks (https://github.com/Chilipp/sphinx-nbexamples). 

In #245 it was mentioned that a relatively easy solution would be to support a notebook -> .py converter. @chsasank helpfully provided a script that they use to convert notebooks (using the `json` module so it's relatively lightweight). I've modified/updated that script a bit, and included in this PR as a helper function, along with tests and an example that shows the `.ipynb -> .py -> .ipynb` roundtrip.

closes #245 

note that this was attempted once last year by @chsasank (https://github.com/sphinx-gallery/sphinx-gallery/pull/220) and it ultimately didn't make it through. My hope is that making this a "helper function" instead of part of the core build process, we have a bit more wiggle-room for inconsistencies. 

Here's a link to the built page from a notebook: https://613-25860190-gh.circle-artifacts.com/0/alabaster_html/auto_examples/sin_func/plot_notebook.html#sphx-glr-auto-examples-sin-func-plot-notebook-py


.circleci/config.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - restore_cache:
           keys:
             - cache-pip
-      - run: pip install --user numpy scipy matplotlib seaborn sphinx_rtd_theme pillow sphinx pytest vtk traits traitsui pyface pyqt5 mayavi memory_profiler
+      - run: pip install --user -r doc/doc-requirements.txt
```
choldgraf: note - this PR also splits out the docs requirements into their own file to make things clearer for would-be developers or users.

Titan-C:
Thank you for taking time to do this @choldgraf 
As I also said in https://github.com/sphinx-gallery/sphinx-gallery/issues/245#issuecomment-303669468
can we make this converter part of a separate repository please, one for more open contributions we don't guaranty support. I know it is a lot of extra work, and I haven't had time for this, but I don't want to mix. 
Unfortunately for now Sphinx only supports rst, and Jupyter notebooks are amazing but they work on markdown. I don't want to throw us into problems we barely have time to support. We'll have to deal with a lot of conversions, which pandoc does not even support. Reason why #220 was turned down in the end.

larsoner:
@choldgraf what should we do with this? Move to SG-extra?

I have a feeling it will need some reworking to fit with the new scraping changes.

Does this supersede #422 and #420?

choldgraf:
yeah probably, and anyway at this point I'd probably recommend that people use Jupytext anyway...if we want to support jupyter notebooks more effectively I'd rather it happen by improving the story of two-way sync between ipynb and `.py` in that package. I'll close this one

PR #428, opened 2018-10-30
DOC: Missing perenthisis in PNGScraper

ksunden:


choldgraf:
Nice! Thanks for noticing, and thanks for the patch 🎉

PR #430, opened 2018-11-13
MRG: Fix for newer sphinx

larsoner:
As I was eradicating warnings during sphinx builds for MNE, I came across a DeprecationWarning and a warning about `\`. This fixes those.

### DIFF
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -39,7 +39,7 @@ except NameError:
 
 DEFAULT_GALLERY_CONF = {
     'filename_pattern': re.escape(os.sep) + 'plot',
-    'ignore_pattern': '__init__\.py',
+    'ignore_pattern': r'__init__\.py',
     'examples_dirs': os.path.join('..', 'examples'),
     'subsection_order': None,
     'within_subsection_order': NumberOfCodeLinesSortKey,
@@ -435,7 +435,10 @@ def setup(app):
     for key in ['plot_gallery', 'abort_on_example_error']:
         app.add_config_value(key, get_default_config_value(key), 'html')
 
-    app.add_stylesheet('gallery.css')
+    try:
+        app.add_css_file('gallery.css')
+    except AttributeError:  # Sphinx < 1.8
+        app.add_stylesheet('gallery.css')
 
     # Sphinx < 1.6 calls it `_extensions`, >= 1.6 is `extensions`.
     extensions_attr = '_extensions' if hasattr(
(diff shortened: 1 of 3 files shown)


larsoner:
I was also seeing with every build:
```
generating gallery...
Exception ignored in: <bound method Popen.__del__ of <subprocess.Popen object at 0x7f38ae492fd0>>
  File "/usr/lib/python3.6/subprocess.py", line 766, in __del__
    ResourceWarning, source=self)
```
The second commit fixes that problem.

Third commit gets rid of a bunch of traits-related warnings when doing `pytest sphinx_gallery`.

sphinx_gallery/gen_gallery.py
```diff
@@ -39,7 +39,7 @@
 
 DEFAULT_GALLERY_CONF = {
     'filename_pattern': re.escape(os.sep) + 'plot',
-    'ignore_pattern': '__init__\.py',
+    'ignore_pattern': r'__init__\.py',
```
choldgraf: hmmm - does this mean we should add something to the docs like "if you're getting warnings you should put `r"` in front of your ignore pattern?

sphinx_gallery/gen_gallery.py
```diff
@@ -435,7 +435,10 @@ def setup(app):
     for key in ['plot_gallery', 'abort_on_example_error']:
         app.add_config_value(key, get_default_config_value(key), 'html')
 
-    app.add_stylesheet('gallery.css')
+    try:
+        app.add_css_file('gallery.css')
+    except AttributeError:  # Sphinx < 1.8
+        app.add_stylesheet('gallery.css')
```
choldgraf: ugh I swear sphinx is the new pandas lol

sphinx_gallery/gen_gallery.py
larsoner: This is just a general Python thing that `\` when you want it to be `\` and not a special char should either be `\\` in strings or you should use `r''` strings

choldgraf:
couple nits but looks good to me

larsoner:
Made a small tweak to hopefully make AppVeyor happy. Assuming it comes back green I'll merge. Thanks for the quick reviews

agramfort:
+1

larsoner:
Also at some point we should bump the version in `__init__.py` from `0.2.0` to `0.3.0.dev0` or so

larsoner:
(I can do it here if someone agrees before I otherwise merge)

choldgraf:
@larsoner are we into v0.3 territory yet? I thought there was some stuff @Titan-C wanted to take care of before we bumped?

also @agramfort what do you think about renaming sphinx-gallery to "pandas-gallery"?

larsoner:
We have 0.2 in pip already as of [June 5](https://pypi.org/project/sphinx-gallery/#history), we presumably should have bumped to 0.3 as soon as that happened

PR #437, opened 2018-12-06
FIX: Fix for latest sphinx-dev

larsoner:
Fixes things locally after reproducing the error with sphinx-dev that we have gotten [recently on Travis](https://travis-ci.org/sphinx-gallery/sphinx-gallery/jobs/463698115#L1174). See for explanation:

https://github.com/sphinx-doc/sphinx/pull/5718#issuecomment-444739024


choldgraf:
nice - this doesn't change any user-facing API, seems to solve the problem, and tests are all happy, so I'm merging! Thanks for the patch @larsoner :-)

PR #439, opened 2018-12-13
[MRG] updating binder instructions

### LINKED ISSUE #438: Improve binder button instructions
I just gave a shot at using the binder instructions and had a few thoughts for improvements:

1. It's not clear what some of the configuration options are used for (e.g. username/org/branch are used *only* to create the binder links)
2. Specifically, the 'dependencies' option was a bit confusing especially if you're *already* using a `binder/` folder.
3. We should update to use the new binder badge ()
4. List of paths to dependencies isn't clear that it's relative to doc (maybe it should be relative to root). Maybe this should just be "optional" in case people already set up their repo w/ the necessary dependencies.
5. There's a bug if `gallery_dirs` is a string and not a list (it iterates through the string, not the list)
6. Unclear what the "prefix" part of the config does, need to make sure there isn't a bug + clarify in docs
7. notebooks_dir is unclear if "site" refers to the built HTML site (which it does) vs. the root of the repository
8. A lot of the assumptions above are that you're not building your site with something like readthedocs, but instead are posting the whole HTML on a branch (using something like gh-pages). Need to see if there's a way to make this compatible with readthedocs

keeping this list here for future ref!

choldgraf:
closes #438 

This PR improves a lot of the instructions / some code around Binder deployment...I'll turn my list of items from #438 into a todo list here:

- [x] It's not clear what some of the configuration options are used for (e.g. username/org/branch are used *only* to create the binder links)
- [x] Specifically, the 'dependencies' option was a bit confusing especially if you're *already* using a `binder/` folder.
- [x] We should update to use the new binder badge ()
- [x] List of paths to dependencies isn't clear that it's relative to doc (maybe it should be relative to root). Maybe this should just be "optional" in case people already set up their repo w/ the necessary dependencies.
- [x] There's a bug if `gallery_dirs` is a string and not a list (it iterates through the string, not the list)
- [x] Unclear what the "prefix" part of the config does, need to make sure there isn't a bug + clarify in docs
- [x] notebooks_dir is unclear if "site" refers to the built HTML site (which it does) vs. the root of the repository
- ~~[ ] A lot of the assumptions above are that you're not building your site with something like readthedocs, but instead are posting the whole HTML on a branch (using something like gh-pages). Need to see if there's a way to make this compatible with readthedocs~~

This is largely a docs PR, with some notable code/API changes:

1. the binder configuration parameter `url` is now `binderhub_url` to make it more clear
2. the `notebook_first_cell` now also accepts `None` as an input, in which case it just won't add an extra first cell.

I think this is ready for a review


choldgraf:
here's the updated instructions: https://665-25860190-gh.circle-artifacts.com/0/rtd_html/configuration.html#binder-links

sphinx_gallery/binder.py
```diff
@@ -64,8 +64,8 @@ def gen_binder_url(fpath, binder_conf, gallery_conf):
     path_link = path_link.replace(os.path.sep, '/')
 
     # Create the URL
-    binder_url = binder_conf['url']
-    binder_url = '/'.join([binder_conf['url'],
+    binder_url = binder_conf['binderhub_url']
```
larsoner: For backward compat perhaps you should detect the presence of the `url` key, emit a warning, and put it in `binderhub_url`?

sphinx_gallery/gen_gallery.py
```diff
@@ -186,7 +186,7 @@ def _complete_gallery_conf(sphinx_gallery_conf, src_dir, plot_gallery,
 
     # Ensure the first cell text is a string if we have it
     first_cell = gallery_conf.get("first_notebook_cell")
-    if not isinstance(first_cell, basestring):
+    if not isinstance(first_cell, (basestring, type(None))):
         raise ValueError("The 'first_notebook_cell' parameter must be type str"
```
larsoner: should say "str or None" now?

choldgraf:
see latest push - how does that look?

sphinx_gallery/gen_gallery.py
choldgraf: do these need to be in two different parts? e.g. `isinstance(` followed by `is None`?

sphinx_gallery/tests/test_binder.py
```diff
@@ -75,6 +76,18 @@ def test_binder():
         url = check_binder_conf(conf6)
     excinfo.match(r"`dependencies` value should be a list of strings")
 
+    # Missing requirements file should raise an error
+    with pytest.raises(ValueError) as excinfo:
```
larsoner: why not simpler `ValueError, match="Couldn't find the Binder requirements")`?

Also I'd just put the line that is expected to fail in the context handler, even though in principle it should be the last one anyway

sphinx_gallery/gen_gallery.py
larsoner: I usually do `isinsance(x, y) and x is not None` but what you have here is fine. My comment is about the text in the error to the user, `None` should also be offered as an option

doc/configuration.rst
```diff
@@ -493,36 +495,60 @@ dictionary following the pattern below::
...
+2. The built Jupyter Notebooks from the documentation will be copied to a folder called ``<notebooks_dir/>`` at the root of
+   your built documentation (they will follow the same folder hierarchy within the notebooks directory folder.
+2. The rST output of each Sphinx-Gallery example will now have a ``launch binder`` button in it.
+4. That button will point to a binder link with the following structure::
+
+       <binderhub_url>/v2/gh/<org>/<repo>/<ref>?filepath=<filepath_prefix>/<notebooks_dir>/path/to/notebook.ipynb
+
+Below is a more complete explanation of each field.
+
+**org** (type: string)
```
larsoner: no need for `**` to bold here, def lists become bold already

https://667-25860190-gh.circle-artifacts.com/0/html/configuration.html

sphinx_gallery/tests/test_binder.py
choldgraf: >Also I'd just put the line that is expected to fail in the context handler, even though in principle it should be the last one anyway

what does this mean exactly? :-)

larsoner: Answered in commit form

https://github.com/sphinx-gallery/sphinx-gallery/pull/439/commits/a69592a7c62d549aa63b9a2163c44651243e7826

sphinx_gallery/tests/test_binder.py
```diff
@@ -95,11 +91,10 @@ def apptmp():
         assert iconf == {}
 
     # Assert extra unkonwn params
-    with pytest.raises(ValueError) as excinfo:
-        conf7 = deepcopy(conf1)
-        conf7['foo'] = 'blah'
+    conf7 = deepcopy(conf1)
+    conf7['foo'] = 'blah'
+    with pytest.raises(ValueError, match='Unknown Binder config key'):
```
choldgraf: I had no idea about this

choldgraf:
thx for the push, I'll try to get to the other comments soon when I have a moment

choldgraf:
ok I think your comments are addressed @larsoner

sphinx_gallery/gen_gallery.py
larsoner: Still need to say this can be `str or None` in the error message :)

Pushed a commit for this

doc/conf.py
```diff
@@ -350,8 +350,8 @@ def setup(app):
     'min_reported_time': min_reported_time,
     'binder': {'org': 'sphinx-gallery',
                'repo': 'sphinx-gallery.github.io',
-               'url': 'https://mybinder.org',
                'branch': 'master',
+               'binderhub_url': 'https://mybinder.org',
```
GaelVaroquaux: Are we breaking backward compatibility?

GaelVaroquaux:
I'm overall +1 for merge, aside from the fact that I am uneasy that it is not clear to my that backward compatibility is maintained.

If it's not, we need at least to state it clearly in the changelog.

doc/conf.py
choldgraf: currently we are not: https://github.com/sphinx-gallery/sphinx-gallery/pull/439/files#diff-2983a963fa400530c600da28315ae7b7R208

should we put in the changelog that the deprecation is scheduled for version XX?

choldgraf:
(trying to clean up my stale PRs)

I just added a note to the changelog to let people know about the configuration key deprecation (https://github.com/sphinx-gallery/sphinx-gallery/pull/439/commits/bbcf8aa36341b861ce93161897145419dae7fdee). Other than that, I think this is ready to go.

larsoner:
Thanks @choldgraf

PR #443, opened 2019-02-05
Activating travis on pre-release of python

NelleV:
Activating travis on 3.7 and nightly build of latest dev (with allowed failure, currently python 3.8 which is in pre-release, and apparently breaks sphinx-gallery (unchecked)).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,12 @@ matrix:
     - python: "nightly"
       env: PYTHON_VERSION="nightly"
       os: linux
+      addons:
+        apt:
+          packages:
+            libblas-dev
+            liblapack-dev
+            gfortran
   allow_failures:
     - python: "nightly"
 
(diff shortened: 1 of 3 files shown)


NelleV:
It fails, but not for the right reasons. I'll try to fix this later.

NelleV:
The remaining errors on the sphinx-gallery are real.

NelleV:
I'm not sure why the coverage went down: I've added matrix entries, and haven't removed any.

larsoner:
> I'm not sure why the coverage went down: I've added matrix entries, and haven't removed any.

Agreed. Codecov has some questionable statistics, I think we can safely ignore it here.

If you're done, +1 for merge from me, and we can fix the nightly errors in a subsequent PR. I can try to look in the coming days if it would help.

larsoner:
Okay in it goes, let's fix the issues in another PR

larsoner:
Thanks @NelleV

choldgraf:
wooo, thanks @NelleV !

(on a side-note, I accidentally started with `:` when I was looking for nelle's username above, and I discovered that apparently there is now a "neckboard" emoji: :neckbeard: )

PR #445, opened 2019-02-11
MAINT: Modernize requirements

### LINKED ISSUE #192: Lack of `install_requires`

larsoner:
- [x] Adds to the `install_requires` in `setup`
- [x] removes `pytest` from `requirements.txt` (since it's needed for testing, not using SG)
- [x] adds a "minimal" test build with appropriate skips
- [x] removes unnecessary SciPy from `tinybuild`
- [x] removes SciPy from the nightly build (building from source takes > 20 min, all other runs are < 5 min so this is too painful)

Closes #192.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,8 @@ jobs:
       - restore_cache:
           keys:
             - cache-pip
-      - run: pip install --user numpy scipy matplotlib seaborn sphinx_rtd_theme pillow sphinx pytest vtk traits traitsui pyface pyqt5 mayavi memory_profiler
+      # PyQt5 5.12 causes problems with libxcb
+      - run: pip install --user numpy scipy matplotlib "pyqt5<5.12" seaborn sphinx_rtd_theme pillow sphinx pytest vtk traits traitsui pyface mayavi memory_profiler
       - save_cache:
           key: cache-pip
           paths:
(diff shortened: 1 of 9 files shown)


larsoner:
... and I did not add `mayavi` as an optional thing, since it seems silly to add the capability of `pip install sphinx_gallery[mayavi]` when you can just as easily do `pip install sphinx_gallery mayavi` (seems mostly redundant). But if there are upstream packaging constants then it's simple enough to add.

larsoner:
@Titan-C can you delete `sphinx-gallery` from your AppVeyor account so I can properly set up a `sphinx-gallery` organizational account? That way I (and other admins) can administer the account, kill jobs, etc.

larsoner:
@Titan-C never mind, I was manually able to delete the webhook, create a `sphinx-gallery` account, and reenable the webhook. I'll add the updated badge as part of this PR.

GaelVaroquaux:
LGTM. This a clear improvement.

Merging!

PR #446, opened 2019-02-20
[MRG] ENH: update only changed examples -- continued

### LINKED ISSUE #394: Unnecessary regeneration of API docs

### LINKED ISSUE #395: Unnecessary regeneration of example pages

NicolasHug:
Closes #394 
Closes #395
Closes #400

This PR introduces a `rebuild` parameter which can take 2 values:
- `'changed_only'` (new default)
- `'always'` (previous behaviour)

This makes the scikit-learn rebuild go from a few minutes to a few seconds (558 changed files to 71).

I don't see any reason not to make this the default behaviour but feel free to disagree.

Also, we check the files mtime because this is very easy, and sphinx is also doing this (ping @Titan-C who seemed to disaprove this.)

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -122,11 +122,6 @@ you would do::
 Here, one should escape the dot ``r'\.'`` as otherwise python `regular expressions`_ matches any character. Nevertheless, as
 one is targeting a specific file, it would match the dot in the filename even without this escape character.
 
-.. note::
-    If the ``rebuild`` option is set to 'mtime' and the example you want to
-    run haven't changed, they won't be executed again. See
-    :ref:`rebuild_changed_only`.
-
 Similarly, to build only examples in a specific directory, you can do::
 
     sphinx_gallery_conf = {
@@ -600,10 +595,9 @@ Only rebuild newly changed examples
 ===================================
 
 By default, Sphinx-Gallery will always generate rst files and backreferences
-for all examples, even if they aren't executed. This can make the
-re-building process slow for projects with a lot of examples. If you want to
-only generate rst files for newly updated examples, you can set the
-``rebuild`` parameter to ``'mtime'``::
+for all examples. This can make the re-building process slow for projects with
+a lot of examples. If you want to only generate rst files for newly updated
+examples, you can set the ``rebuild`` parameter to ``'mtime'``::
 
     sphinx_gallery_conf = {
         ...
@@ -756,9 +750,6 @@ build. Change your `conf.py` accordingly::
 Here you list the examples you allow to fail during the build process,
 keep in mind to specify the full relative path from your `conf.py` to
 the example script.
-
-Note that if an example is expected to fail, sphinx-gallery will error if
-the example runs without error.
 
 
 .. _setting_thumbnail_size:
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -28,7 +28,7 @@ file:
 - ``line_numbers`` (:ref:`adding_line_numbers`)
 - ``download_all_examples`` (:ref:`disable_all_scripts_download`)
 - ``plot_gallery`` (:ref:`without_execution`)
-- ``rebuild`` (:ref:`always_rebuild`)
+- ``rebuild`` (:ref:`rebuild_changed_only`)
 - ``image_scrapers`` (and the deprecated ``find_mayavi_figures``)
   (:ref:`image_scrapers`)
 - ``reset_modules`` (:ref:`reset_modules`)
@@ -589,20 +589,19 @@ a default::
 The highest precedence is always given to the `-D` flag of the
 ``sphinx-build`` command.
 
-.. _always_rebuild:
+.. _rebuild_changed_only:
 
-Always rebuild examples
-=======================
+Only rebuild newly changed examples
+===================================
 
-By default, Sphinx-Gallery will only generate rst files and backreferences
-for newly updated examples. If you want to always generate new rst files, you
-can set the ``rebuild`` parameter to ``'always'``. Note that this will
-likely make your sphinx build significantly slower if you have a lot
-examples and with backreferences::
+By default, Sphinx-Gallery will always generate rst files and backreferences
+for all examples. This can make the re-building process slow for projects with
+a lot of examples. If you want to only generate rst files for newly updated
+examples, you can set the ``rebuild`` parameter to ``'mtime'``::
 
     sphinx_gallery_conf = {
         ...
-        'rebuild': 'always',
+        'rebuild': 'mtime',
     }
 
 .. _image_scrapers:
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -122,6 +122,11 @@ you would do::
 Here, one should escape the dot ``r'\.'`` as otherwise python `regular expressions`_ matches any character. Nevertheless, as
 one is targeting a specific file, it would match the dot in the filename even without this escape character.
 
+.. note::
+    If the ``rebuild`` option is set to 'mtime' and the example you want to
+    run haven't changed, they won't be executed again. See
+    :ref:`rebuild_changed_only`.
+
 Similarly, to build only examples in a specific directory, you can do::
 
     sphinx_gallery_conf = {
@@ -595,9 +600,10 @@ Only rebuild newly changed examples
 ===================================
 
 By default, Sphinx-Gallery will always generate rst files and backreferences
-for all examples. This can make the re-building process slow for projects with
-a lot of examples. If you want to only generate rst files for newly updated
-examples, you can set the ``rebuild`` parameter to ``'mtime'``::
+for all examples, even if they aren't executed. This can make the
+re-building process slow for projects with a lot of examples. If you want to
+only generate rst files for newly updated examples, you can set the
+``rebuild`` parameter to ``'mtime'``::
 
     sphinx_gallery_conf = {
         ...
@@ -750,6 +756,9 @@ build. Change your `conf.py` accordingly::
 Here you list the examples you allow to fail during the build process,
 keep in mind to specify the full relative path from your `conf.py` to
 the example script.
+
+Note that if an example is expected to fail, sphinx-gallery will error if
+the example runs without error.
 
 
 .. _setting_thumbnail_size:
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -28,6 +28,7 @@ file:
 - ``line_numbers`` (:ref:`adding_line_numbers`)
 - ``download_all_examples`` (:ref:`disable_all_scripts_download`)
 - ``plot_gallery`` (:ref:`without_execution`)
+- ``rebuild`` (:ref:`always_rebuild`)
 - ``image_scrapers`` (and the deprecated ``find_mayavi_figures``)
   (:ref:`image_scrapers`)
 - ``reset_modules`` (:ref:`reset_modules`)
@@ -588,6 +589,21 @@ a default::
 The highest precedence is always given to the `-D` flag of the
 ``sphinx-build`` command.
 
+.. _always_rebuild:
+
+Always rebuild examples
+=======================
+
+By default, Sphinx-Gallery will only generate rst files and backreferences
+for newly updated examples. If you want to always generate new rst files, you
+can set the ``rebuild`` parameter to ``'always'``. Note that this will
+likely make your sphinx build significantly slower if you have a lot
+examples and with backreferences::
+
+    sphinx_gallery_conf = {
+        ...
+        'rebuild': 'always',
+    }
 
 .. _image_scrapers:
 
--- a/sphinx_gallery/tests/test_gen_gallery.py
+++ b/sphinx_gallery/tests/test_gen_gallery.py
@@ -211,6 +211,16 @@ def test_duplicate_files_warn(sphinx_app_wrapper):
     assert msg.format(m) in build_warn
 
 
+@pytest.mark.conf_file(content="""
+sphinx_gallery_conf = {
+    'rebuild': 'wrong_value',
+}""")
+def test_config_rebuild(sphinx_app_wrapper):
+    """Test ValueError is raised for invalid rebuild values"""
+    with pytest.raises(ValueError, match="The 'rebuild' parameter must be"):
+        sphinx_app_wrapper.create_sphinx_app()
+
+
 def _check_order(sphinx_app, key):
     index_fname = os.path.join(sphinx_app.outdir, '..', 'ex', 'index.rst')
     order = list()
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,13 @@ Change Log
 git master
 ----------
 
+New features
+''''''''''''
+
+* Added the ``rebuild`` option, which by default will only rebuild newly
+  changed files, leading to faster incremental builds. Use ``rebuild =
+  'always'`` to always update files (previous behaviour). TODO link PR.
+
 Developer changes
 '''''''''''''''''
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,9 +7,10 @@ git master
 New features
 ''''''''''''
 
-* Added the ``rebuild`` option, which by default will only rebuild newly
-  changed files, leading to faster incremental builds. Use ``rebuild =
-  'always'`` to always update files (previous behaviour). TODO link PR.
+* Added the ``rebuild`` option, which allows to only rebuild newly
+  changed examples, leading to faster incremental builds. Use ``rebuild =
+  'mtime'`` to enable this. `#446
+  <https://github.com/sphinx-gallery/sphinx-gallery/pull/446>`_.
 
 Developer changes
 '''''''''''''''''
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -357,4 +357,5 @@ sphinx_gallery_conf = {
                'use_jupyter_lab': True,
                },
     'show_memory': True,
+    'rebuild': 'mtime',
 }
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -357,5 +357,4 @@ sphinx_gallery_conf = {
                'use_jupyter_lab': True,
                },
     'show_memory': True,
-    'rebuild': 'mtime',
 }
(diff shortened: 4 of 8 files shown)


larsoner:
Instead of `changed_only` I'd rather call it `mtime`, since it makes explicit what it uses. Later we might also want a `changed_only`-like option that uses a cached `md5sum`, for example, and that could be called `'md5'`

larsoner:
... and for backward compat, we probably should leave the default as `always`. People can have lots of workflows (e.g., rebuild doc while modifying a library) that could break by such a change. And it's easy for repos that want the `mtime` flavor to set it.

NicolasHug:
Thanks for the feedback, I'll update.

Personally I find that `updapte: 'mtime'` is a bit obscure but I agree it allows for more flexibility in the future

larsoner:
Looks like the CIs are not happy. Locally you can test with `pytest sphinx_gallery`

sphinx_gallery/tests/test_gen_gallery.py
```diff
@@ -211,6 +211,16 @@ def test_duplicate_files_warn(sphinx_app_wrapper):
...
 
 
+@pytest.mark.conf_file(content="""
+sphinx_gallery_conf = {
+    'rebuild': 'wrong_value',
+}""")
+def test_config_rebuild(sphinx_app_wrapper):
+    """Test ValueError is raised for invalid rebuild values"""
+    with pytest.raises(ValueError, match="The 'rebuild' parameter must be"):
+        sphinx_app_wrapper.create_sphinx_app()
```
larsoner: You can use the `tinybuild` infrastructure in `test_full.py` to do two builds in a row, ensuring that no examples are run the second time around.

NicolasHug:
The tests are passing, what's failing is this and I'm a bit confused:

```
Exception occurred:
  File "/home/travis/miniconda3/envs/testenv/lib/python3.6/site-packages/sphinx_gallery-0.3.dev0-py3.6.egg/sphinx_gallery/gen_gallery.py", line 466, in summarize_failing_examples
    "\n" + "-" * 79)
ValueError: Here is a summary of the problems encountered when running the examples
Examples expected to fail, but not failing:
Please remove these examples from
sphinx_gallery_conf['expected_failing_examples']
/home/travis/build/sphinx-gallery/sphinx-gallery/examples/no_output/plot_syntaxerror.pyin your conf.py file
/home/travis/build/sphinx-gallery/sphinx-gallery/examples/no_output/plot_raise.py
-------------------------------------------------------------------------------
The full traceback has been saved in /tmp/sphinx-err-0d22hikh.log, if you want to report the issue to the developers.
Makefile:76: recipe for target 'html' failed
make: *** [html] Error 2
```

larsoner:
SG has the ability to mark some examples as expected to fail. If you build the doc and they did not fail, then it raises an error. This change makes it so those (failing) examples are not run at all, so it errantly says that they "did not fail".

The thing to do is to update the check for "did all examples we wanted to have fail, actually fail" to check for "did all the example we wanted to have fail, out of the examples we actually tried to run, actually fail".

NicolasHug:
Thanks, looks like reverting the default made the tests pass

larsoner:
Great, now we need a test that the functionality actually work (see comment above about using `tinybuild` to check)

NicolasHug:
Sorry to ask @larsmans but I've been unsuccessfully trying for over an hour to use tinybuild:

- How can I set the rebuild parameter to 'mtime'?
`        app.config.sphinx_gallery_conf['rebuild'] = 'mtime'` doesn't seem to be working

- How can I re-run the app? The app seems to be run in the fixture with  
`app = Sphinx(src_dir, conf_dir, out_dir, toctrees_dir,
                     buildername='html', status=MixedEncodingStringIO())`. I tried calling this twice but results aren't as expected, and that requires creating a second fixture.

- How can I check that the examples aren't run twive? Would checking the status work?
```
    status = sphinx_app._status.getvalue()
    assert 'executed 0 out of 4' in status
```

larsoner:
> How can I set the rebuild parameter to 'mtime'?

Change it in this file, it should be fine to do it

https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/tests/tinybuild/conf.py#L33

> I tried calling this twice but results aren't as expected, and that requires creating a second fixture.

It might be worth creating a second fixture for this specific use case. Or using no fixture at all and just doing the app creation and build steps as necessary in a new test function, at least to start / get it working.

> How can I check that the examples aren't run twive? Would checking the status work?

Yes I think it should. You could also check that the `mtime` of some other file is not changed (the HTML? Some RST file somewhere?)

NicolasHug:
Thank you for your help @larsoner 

I came up with a test in 8859ab7 following your guidelines.

Also, I had to add a new set `not_updated_examples` in the config dict that gets passed around: we need to remember which examples aren't run (because they haven't been updated), so that those that were marked as `should_fail` don't raise an error (because they don't actually fail since they aren't run).

doc/configuration.rst
```diff
@@ -588,6 +589,20 @@ a default::
...
+
+By default, Sphinx-Gallery will always generate rst files and backreferences
+for all examples. This can make the re-building process slow for projects with
+a lot of examples. If you want to only generate rst files for newly updated
+examples, you can set the ``rebuild`` parameter to ``'mtime'``::
+
+    sphinx_gallery_conf = {
+        ...
+        'rebuild': 'mtime',
+    }
```
larsoner: Can you also add a `.. note::` at the bottom of the `Parsing and executing examples via matching patterns` section where it talks about building a specific example that if people's `conf.py` has `'rebuild': 'mtime'`, they might also need to set `-D sphinx_gallery_conf.rebuild=always`?

Also maybe a complementary note in this section would help. I know there are some repos that use this `-D` pattern to build specific files in a `Makefile`, and it would be good to give them a hint that if what they want is to build e.g. one file no matter what (say they are modifying the library, not the example) then they'll want to add another `-D` to override `rebuild`, too.

larsoner:
@jklymak can you see if this solves your problem?

larsoner:
Also if you want to update the `Don't fail the build on exit` section to say that examples expected to fail that actually pass will cause the build to fail, feel free. Otherwise we can add it separately. But it occurred to me you were at least caught by that lack of docs working on this.

NicolasHug:
Ok I'll update the docs and add notes.

Would you have any clue why the mtimes are still updated in some travis build (3 out of 10)?
The nightly build seems to have other issues as well.

I find the test suite a bit tricky to work with. Tests aren't independent one from another. For example after removing the `with docutils_namespace()` context in `test_full.py`, tests still pass in `test_full.py` but some tests break in `est_gen_gallery.py`.

larsoner:
> Tests aren't independent one from another. For example after removing the with docutils_namespace() context in test_full.py, tests still pass in test_full.py but some tests break in est_gen_gallery.py.

Yes this is a Sphinx quirk, `docutils_namespace` was their suggested workaround/fix. The vast majority of the time people use sphinx from the command line, so some of their global-value-changing stuff has to be worked around. Removing the `docutils_namespace` is thus expected to cause problems. Not much we can do about it that I know of, outside of a big refactoring of `sphinx` itself.

> Would you have any clue why the mtimes are still updated in some travis build (3 out of 10)?

Probably `sphinx` version differences. You could set up `venv`s (or just temporarily `pip uninstall` and `pip install` different versions) to test this. If it's a sphinx verison problem we can `warn` that it's only properly supported on some versions, once we know which those are.

> The nightly build seems to have other issues as well.

Yes that's an allowed fail, and it is failing in `master`. So probably not your fault.

NicolasHug:
Looking at the first 2 Python 2.7 builds:

- one is passing, one is failing
- both use sphinx 1.8.4

So it does not look like a sphinx version issue. :/

I'll try something but I'm not very confident.

------

To rule out the possibility that the others travis builds pass for the wrong reason, I checked with adding `sleep(1)` between the 2 successive sphinx builds and the tests still pass. So the files aren't updated which is the correct behavior.

larsoner:
This error is really strange. I don't know what would cause it.

I'm tempted just to say that this feature is only (officially) supported on Python 3.5+. Major packages are moving to 3.5+ for new features anyway, so it's perhaps not a huge loss.

larsoner:
(we should probably drop 3.4 from our CIs regardless since it's EOL is in the next month)

jklymak:
This PR gives me: `document isn't included in any toctree` for every file for matplotlib's build.

NicolasHug:
Even after a `make clean`  @jklymak  ?

I can't reproduce because building the mpl docs will error with `Warning, treated as error:
/home/nico/dev/matplotlib/doc/api/toolkits/axes_grid1.rst::image file not readable: api/toolkits/../../gallery/axes_grid1/images/sphx_glr_demo_axes_grid_001.png
`
for me. (which is not related to the PR, I get the same on the master branch)

jklymak:
Yes, sorry for the noise - it seems to be something wrong with how I have matplotlib set up...

NicolasHug:
So do you confirm the PR works as expected?

jklymak:
Not yet :wink:...

I'm doing `git clean xdf` on the doc directory, so that should get me to a pristine state.  Trying the build ignoring warnings...

NicolasHug:
Cool thanks for taking the time

NicolasHug:
Tests are green @larsoner \o/ !

The fix was to check the src dir (with rst files) instead of the output dir (_build/, with html files).

It makes more sense this way.

It still feels strange to me why sphinx would still generate new html files in some of the builds, but I won't argue...

I'll clean this up and will update the docs ASAP

jklymak:
## On master:

### First run:
 - `generating gallery for gallery/style_sheets...` long time
 - total: 610 s

### Second run (no edits)
  - `generating gallery for gallery/style_sheets...` quick
  - total: 113s

## This PR:

`git clean -xdf` 

### First build: `time make SPHINXOPTS= html`
  - long time on generating gallery
  - total: 620s

### Second build:
  - `WARNING: document isn't included in any toctree` pops up for every file.  
  - Gallery gen is fast
  - still long tme on writing output
  - 66.61 s

So, not 100% convinced this works properly.

NicolasHug:
Well that almost halved the second build right?

The other files might come from other sphinx extensions? I know in scikit learn there are still other "chnged" files as well.

the warnings are concerning though

larsoner:
> 113s ... 66.61 s

I agree this is a substantial speedup. The rest of the time is hopefully/probably some other `sphinx` operation's fault.

> * `WARNING: document isn't included in any toctree` pops up for every file.

> the warnings are concerning though

@NicolasHug do you see this locally e.g. when building `sklearn`?

Is it possible it depends on the `sphinx` version?

larsoner:
Congrats on conquering the CIs, BTW :)

jklymak:
Yeah, the whole doc/gallery tree doesn't end up in the built docs this way for matplotlib

larsoner:
@jklymak can you try the one-two build process again, but this time change exactly one example before running the second build? If that reduces or eliminates the warnings, maybe there is something SG only writes out if it runs >= 1 example, that we should always write out when using `mtime`.

NicolasHug:
I can confirm the `not included in toctree` warnings when building `sphinx-gallery` docs.

larsoner:
Okay here are some possible hints.

1. Even after changing `'rebuild'='mtime'` in `conf.py` I see in the output:
    ```
    after excluding 0 files that had previously been run (based on MD5).
    ```
    This should probably say "based on mtime" now... I think we're doing some double-exclusion.
2. On second run, same message, but it should probably say "after excluding N files that had previously been run (based on mtime)".
3. After doing `touch ../examples/plot_seaborn.py` (which just modifies `mtime`, not `md5`), I see in the output:
    ```
    Sphinx-gallery successfully executed 0 out of 0 files subselected by:
    ...
    after excluding 1 file that had previously been run (based on MD5).
    ```
    However, this line disappears:
    ```
    /home/larsoner/python/sphinx-gallery/doc/auto_examples/plot_seaborn.rst: WARNING: document isn't included in any toctree
    ```

So it looks like:

1. There is MD5-based exclusion, *and* `mtime`-based exclusion performed when `rebuild='mtime'`. And
2. The exclusion that occurs due to MD5-based exclusion does not generate the `toctree` warnings.

This makes me think that the `md5` mode should be fixed and added separately, keeping the default as `'always'` for backward compat. That can be a separate PR, but this one should probably not *also* use the MD5-based checks and messaging. There is some hint here that the MD5 code is doing something different that does not elicit these warnings.

If I actually then modify `plot_seaborn.py`, the `mtime` mode actually kicks in (and of course the `md5` is satisfied so it does not exclude):
```
Sphinx-gallery successfully executed 1 out of 1 file subselected by:
...
after excluding 0 files that had previously been run (based on MD5).
```
and the warning about `toctree` is not there.

So there is still some unification to do here.

larsoner:
... in other words, you can `assert` in the `tinybuild` test that the exclude statement in the tests starts out saying "executed 4 out of 4" and "after excluding 0 files .*based on mtime", and on the second run says "executed 0 out of 0" and "after excluding 4 files .*based on mtime" on the second run (assuming there are 4 total examples).

jklymak:
(Glad it shows the bug on a shorter doc build.  My machine was gettting hit running the insane matplotlib build)

jklymak:
... for matplotlib, if I edit `examples/animated_histogram.py` it still spits out the `* WARNING: document isn't included in any toctree` for every file, *except* `/animated_histogram`, which also properly renders in the TOC at `gallery/index.html` (matplotlib maps `examples`->`gallery` for whatever reason).

NicolasHug:
Sorry @larsoner  I'm not sure I entirely follow the md5 thing. What are md5 sums used for in sphinx-gallery?

----------

In any case, I think that I understand where the warning come from: the index file toctree isn't populated with the not-updated examples since these are filtered out of the iterator.

I pushed a super hacky fix (that doesn't entirely work anyway) in my last commit if you wanna have a look...

There might still be a problem with backreferences as well :(

larsoner:
In `master` see for example `stale_examples`:

https://github.com/sphinx-gallery/sphinx-gallery/blob/40aec52de43e4c27b000e054ae50eae6fb0f30e2/sphinx_gallery/gen_gallery.py#L61

These computations:

https://github.com/sphinx-gallery/sphinx-gallery/blob/40aec52de43e4c27b000e054ae50eae6fb0f30e2/sphinx_gallery/gen_rst.py#L227-L249
https://github.com/sphinx-gallery/sphinx-gallery/blob/40aec52de43e4c27b000e054ae50eae6fb0f30e2/sphinx_gallery/gen_rst.py#L623-L626

And this message:

https://github.com/sphinx-gallery/sphinx-gallery/blob/40aec52de43e4c27b000e054ae50eae6fb0f30e2/sphinx_gallery/gen_gallery.py#L443-L453

It seems like you should be hooking into this system and modifying how it decides to run these using `mtime` instead of the `md5` hash. Your current PR seems to implement a separate but seemingly-equivalent-functionality system, except it does not output the correct messages.

Granted, the `md5` method clearly does not (always) work, since everything gets regenerated every time. But it seems like both `mtime` and `md5` should be using this system, and it should be fixed.

NicolasHug:
So the md5 system is supposed to do exactly what we want, but doesn't?

jklymak:
No, I think the md5 system works - generation drops from ~600s to ~113s on master on the second run; the `generating gallery` step goes much faster.

Until it successfully builds the docs, I'm not sure if this PR speeds things up at all - the 60s quoted above may be a red herring.    

I found this all quite hard to time on the sphinx-gallery docs because they are so small.

NicolasHug:
If it works, what's its intended purpose? If its goal is to not generate examples that haven't been updated it's clearly failing...

NicolasHug:
Also: the generation drops because all files non related to sphinx-gallery aren't checked again by sphinx. Sphinx is (hopefully) doing its job properly. Our problem here is that sphinx-gallery is unnecessarily generating files and sphinx will believe they have been updated.

larsoner:
> So the md5 system is supposed to do exactly what we want, but doesn't?

Yes I think so. I only realized it after looking at the code a bit. So somewhere between what it does, and what you do, there is probably a solution to be found. Sorry for the run-around :(

NicolasHug:
Maybe @Titan-C can join as you're the one that wrote the md5 thing? What is it supposed to be doing exactly?

---------------

On my last commit (that will fail on master but that's OK I just didn't adjust the checks), I think I fixed the 'not in toctree' issues, and the galleries are rendered properly. The rebuild of sklearn goes from 556 "changed" files to only 322. Before the "fix", I only had about 50 changed files left... So the speed gain will not be great. I'm wondering if that thing is actually even possible.

jklymak:
partly works?  The rst files are not regenerated.  

Working out of `sphinx-gallery/doc`:

## On master: 

### First build:
13.4 s

### second build (no changes)
5.03 s. Note that most file in `doc/tutorials` or `doc/gen_modules` do not get changed when this run is run.  So the `generating gallery` step is much faster.

larsoner:
My understanding is that the MD5 is designed to prevent things from being re-run that do not need to be re-run. @Titan-C might not respond soon (IIRC he did not respond recently to another PR ping), so it might be good to proceed under that assumption, and try to integrate the two approaches.

At the very least, it seems like:

1. the MD5-based checks should be skipped if `mtime` is done, and vice-versa.
2. there is something to learn about which `sphinx` steps are included/excluded based on these two approaches, and
3. the messages given by `sphinx-gallery` are incorrect currently -- it should say how many examples were not run because of `mtime` (and not `md5`)

NicolasHug:
Ok I see.

I'll try to get back to this, but I won't be able to for about at least 2 weeks. there's a scikit-learn sprint and then I'll be off for a few days. Thanks for your help so far!

(Meanwhile I'll use the quick hack from https://github.com/sphinx-gallery/sphinx-gallery/pull/400#issuecomment-465195893, it works OK as a local workaround as long as you don't need the gallery plots)

larsoner:
Thanks for working through all the issues!

larsoner:
@NicolasHug I might take a stab at integrating the two if you don't mind

NicolasHug:
Of course!!

larsoner:
Closing in favor of #448

PR #447, opened 2019-02-21
MAINT: Drop 3.4, add mayavi to one

larsoner:
Python 3.4 reaches EOL in a few weeks, and it would be good to use Mayavi on a Py3k run.


choldgraf:
I'm all for it

larsoner:
Right, no dropping 2.7 yet. We should probably cut a release, then do it.

larsoner:
I'll merge this for now, and once 0.2.1 is out, we can do #405 and remove 2.7 from CIs

GaelVaroquaux:
> Right, no dropping 2.7 yet. We should probably cut a release, then do it.

+1

PR #448, opened 2019-02-25
MRG: Only rebuild necessary parts

### LINKED ISSUE #394: Unnecessary regeneration of API docs

### LINKED ISSUE #395: Unnecessary regeneration of example pages

### LINKED ISSUE #449: sphinx-gallery doesn't play nice with sphinx's ability to detect new files...
As pointed out by @larsoner (https://github.com/sphinx-gallery/sphinx-gallery/pull/448#issuecomment-467608061_):

I think I have narrowed it down to these to `conf.py` entries:
```
    'subsection_order': gallery_order.sectionorder,
    'within_subsection_order': gallery_order.subsectionorder,
```
I'm guessing that it's because Sphinx can't tell if these config values have changed or not, so decides to re-write everything. With those two lines as they are in `mpl` master:

... what we could do is allow these entries to be strings and do some `importlib` magic with them, rather than allowing them to be classes directly. That *should* allow `sphinx` to treat them as static, rather than unknowns. In this case making changes to your classes will not be reflected in the build, but this should be (much) rarer than the case of changing a single example, and also can be worked around with a simple `make clean` (or removing the quotes from the name) when you want to do this.

_Originally posted by @larsoner in https://github.com/sphinx-gallery/sphinx-gallery/pull/448#issuecomment-467608061_

larsoner:
This writes a few files with a `.new` extension and replaces the old one only if necessary (e.g., `.py`, `.rst` files) to avoid Sphinx rebuilding them. Most of the added lines are new tests adapted from #466.

@NicolasHug @jklymak can you see if it works for you?

Closes #449
Closes #446
Closes #400
Closes #395
Closes #394


jklymak:
Does this need conf.py to have a change or does it just work?

larsoner:
It should just work

jklymak:
On sphinx-gallery: 
  - first call: 11.44 s
  - second call: 4.4s
  - change one tutorial: 4.8 s

Checking matplotlib; update in about 30 minutes :wink:

jklymak:
In matplotlib, after running `git clean -xdf` in the `doc` directory, I get an exception when I run 
`time make SPHINXOPTS= html`:

```
Exception occurred:
  File "/Users/jklymak/sphinx-gallery/sphinx_gallery/utils.py", line 98, in get_md5sum
    with open(src_file, 'rb') as src_data:
FileNotFoundError: [Errno 2] No such file or directory: '/Users/jklymak/matplotlib/doc/api/_as_gen/matplotlib.pyplot.Subplot.examples.new'
```

Note that there *are* lots of *.new files in that directory, so not entirely clear what didn't work...

NicolasHug:
On sklearn sphinx tells me that it's seeing about 60 changed files (instead of hundreds) and the second build runs in a few seconds.
In other words: looks like it  works \o/

jklymak:
I wonder if the above issue is something funky about the cases:

```
ls api/_as_gen/matplotlib.pyplot.Subplot*
matplotlib.pyplot.Subplot.examples              matplotlib.pyplot.subplot_tool.rst
matplotlib.pyplot.subplot.rst                   matplotlib.pyplot.subplots.examples
matplotlib.pyplot.subplot2grid.examples         matplotlib.pyplot.subplots.rst
matplotlib.pyplot.subplot2grid.rst              matplotlib.pyplot.subplots_adjust.examples.new
matplotlib.pyplot.subplot_tool.examples         matplotlib.pyplot.subplots_adjust.rst
```

Note that `Subplot.examples` corresponds with `subplots.rst`.

larsoner:
@jklymak I can't replicate that error in building the `matplotlib` doc. First build I get `real	5m7.390s` and second `real	1m1.422s`, no errors. And this is what I see on my file system:
```
$ ls api/_as_gen/matplotlib.pyplot.Subplot*
api/_as_gen/matplotlib.pyplot.Subplot.examples
```
I don't know why you see any other files since the case does not match. Are you on Windows? Maybe there is some file-case-insensitivity problem...?

Is it possible you are on some broken `matplotlib` branch and not clean latest `master`?

Looking at the code, I'm not sure how it's possible you see this error. Can you paste the full traceback? Is it in `finalize_backreferences`?

> Note that Subplot.examples corresponds with subplots.rst.

Regardless of where `Subplot.examples` is included in the RST files, the `Subplot.examples` filee should be first written as an empty file, then with meaningful content as `Subplot.examples.new`, then renamed (if md5sum differs).

Nonetheless can you try the latest commit?

larsoner:
Commit pushed to avoid the error and emit a warning.

larsoner:
Ready for review/merge from my end. @choldgraf feel free to have a look if you have time.

jklymak:
I'm on a mac, so there are definite case issues (`subplot` = `Subplot`)....

jklymak:
Works now, so long as I don't fail on warning for matplotlib.  OTOH:

### This PR:

 - 563s for first build,
 - 85s for second build no changes
 - 250s  changed one file in `tutorials`; Most of time is spent in `reading sources... api/_as_gen/*`

### Master:

 - 122s second build no changes; lots of time spent in `reading sources... `
 - 258s for one file changed in `tutorials`; Most of time is spent in `reading sources... api/_as_gen/*`

So, this doesn't particularly speed up incremental building (8s has to be in the noise compared to other stuff I'm doing on my machine)...

larsoner:
Maybe your `sphinx` is old / bad at detecting changes? For me I get:

* 311s for first build
* 61s for second build no changes (almost all time in writing output)
* 62s for third build after adding 1 char to the intro paragraph of `tutorials/introductory/images.py` (which SG reports as being rebuilt, and I can verify to be the case viewing the output)

I can confirm this behavior makes sense by:
```
$ ls -alt api/_as_gen/ | more
total 9356
-rw-r--r-- 1 larsoner larsoner  69798 Feb 25 16:43 matplotlib.pyplot.figure.examples
-rw-r--r-- 1 larsoner larsoner   8924 Feb 25 16:43 matplotlib.pyplot.colorbar.examples
-rw-r--r-- 1 larsoner larsoner   9462 Feb 25 16:43 matplotlib.pyplot.imshow.examples
-rw-r--r-- 1 larsoner larsoner    890 Feb 25 16:43 matplotlib.image.imread.examples
-rw-r--r-- 1 larsoner larsoner   2076 Feb 25 16:43 matplotlib.pyplot.hist.examples
-rw-r--r-- 1 larsoner larsoner      0 Feb 25 16:32 mpl_toolkits.mplot3d.Axes3D.text.examples
-rw-r--r-- 1 larsoner larsoner      0 Feb 25 16:32 mpl_toolkits.mplot3d.Axes3D.quiver.examples
-rw-r--r-- 1 larsoner larsoner      0 Feb 25 16:32 mpl_toolkits.mplot3d.Axes3D.bar.examples
...
```
You can clearly see only five `.examples` were changed (the ones used by that example) when I ran it the for that third build.

Moreover, if instead of changing the *introductory paragraph* or title (which SG turns into a tooltip in the `.examples`, forcing it to change on the run) I change something in the *main body text* of the script (e.g., `specific argument` to `specificc argument`) there are no `.examples` files changed at all, so it has even less of an effect. But even changing a few of these `.examples` files should not be catastrophic.

Did you change some other example? If so, maybe it has worse consequences and we can track down why. Can you see if you can replicate what I see by just changing `tutorials/introductory/images.py`? If it builds fast for you in that case, I'm curious which example you changed that made it take so much time in the reading sources stage. If it does not build quickly for you, perhaps try updating Sphinx to see if it helps.

larsoner:
... and as a further diagnostic, you can look at e.g. `ls -alt tutorials/introductory/` to see that only a few files have changed (the `*codeobj.pickle` files will have changed but should not matter since Sphinx does not use these)

jklymak:
I just updated to sphinx 1.8.4.  This is on a Mac.  

### This PR
- 67s second run no changes...  reading sources very fast!
- 202s: one source changed (added a character to `tutorials/intermediate/artists.py`):  [reading sources...] goes very slowly. See the output below...

```
computation time summary:
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 2264 source files that are out of date
updating environment: [config changed] 2264 added, 4 changed, 0 removed
reading sources... [ 52%] api/_as_gen/mpl_toolkits.axes_grid1.axes_divider.make_axes_area_auto_adjustablreading sources... [ 57%] api/_as_gen/mpl_toolkits.axisartist.grid_helper_curvelinear.FixedAxisArtistHelreading sources... [ 57%] api/_as_gen/mpl_toolkits.axisartist.grid_helper_curvelinear.FloatingAxisArtistreading sources... [ 57%] api/_as_gen/mpl_toolkits.axisartist.grid_helper_curvelinear.GridHelperCurveLinreading sources... [100%] users/whats_new
```

- third run no changes; note the `0 added` (and reading sources goes extremely quickly)

```
computation time summary:
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 2264 source files that are out of date
updating environment: [] 0 added, 35 changed, 0 removed
reading sources... [100%] tutorials/toolkits/sg_execution_times
```

jklymak:
The only weird thing about `tutorials/intermediate` is that it has two png files in it.  Could those be triggering the big rebuild somehow?  

```
$ ls -halt ../tutorials/intermediate/
total 140K
-rw-r--r--  1 jklymak staff  11K Feb 25 14:56 gridspec.py
-rw-r--r--  1 jklymak staff  28K Feb 25 14:52 artists.py
-rw-r--r--  1 jklymak staff  11K Feb 25 14:29 CL02.png
-rw-r--r--  1 jklymak staff  11K Feb 25 14:29 CL01.png
-rw-r--r--  1 jklymak staff  30K Feb 25 14:10 constrainedlayout_guide.py
-rw-r--r--  1 jklymak staff  11K Feb 25 11:56 tight_layout_guide.py
drwxr-xr-x 12 jklymak staff  384 Feb 25 11:56 .
-rw-r--r--  1 jklymak staff  11K Feb 25 11:56 legend_guide.py
-rw-r--r--  1 jklymak staff 9.7K Feb 25 11:56 imshow_extent.py
-rw-r--r--  1 jklymak staff 4.0K Feb 25 11:56 color_cycle.py
drwxr-xr-x  9 jklymak staff  288 Sep 25 08:52 ..
-rw-r--r--  1 jklymak staff  213 Sep 25 08:52 README.txt
```

jklymak:
Confirmed, removing the `*.png` files

```
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 2264 source files that are out of date
updating environment: [] 0 added, 43 changed, 0 removed
```

So, why do those png files trigged rebuilding 2264 source files if anything else in that directory is modified?

larsoner:
No idea, but it sounds like this PR is at least working then

jklymak:
@larsoner Its certainly much faster 👍   Thanks a lot for coming up with the correct solution!

OTOH, I am still not sure why it takes over 30 s to re-write all the `html` files.  Some of the rest of the 60 seconds or so is taken up w/ cruft Matplotlib has added, but it seems that if the *.rst file doesn't change, then sphinx shouldn't re-generate the html.

larsoner:
Yes agreed, but I think that's a Sphinx issue. In principle it's only supposed to write for files that are added or changed.

jklymak:
OK, if I run `make build` on sphinx's docs the second time I get:

```
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 0 source files that are out of date
updating environment: 0 added, 0 changed, 0 removed
looking for now-outdated files... none found
no targets are out of date.
build succeeded.
```

If I run it on sphinx-gallery the second time...

```
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 46 source files that are out of date
updating environment: 46 added, 0 changed, 0 removed
reading sources... [100%] utils
/Users/jklymak/sphinx-gallery/doc/index.rst:113: WARNING: toctree contains reference to nonexisting document 'auto_mayavi_examples/index'
looking for now-outdated files... none found
```

and then it re-writes all the html files.  Since there is only 46 of them, its pretty fast, but for matplotlib, there are 4000 of them, so it takes a while.  

A jaunt through `sphinx-doc` code, says that this coms about because somehow the `.buildinfo` file isn't correct: `if self.build_info != buildinfo` gets triggered in `html.py` and fallback to all the html files getting rergenerated.  I have no idea if this is because of something `sphinx-gallery` is doing, or one of the modules shared by sphinx-gallery and Matplotlib.  

Feel free to open a new issue if this is the incorrect place to discuss this.  This PR removes a big performance drop, but there are others that seem fixable...

jklymak:
Removing that check in sphinx drops the matplotlib compile time from 71 s to 24 s.  So the question is why does the `buildinfo` file get out of sync....

larsoner:
At least in that case, it's probably because you changed the `conf.py` file. The contents of `buildinfo` for me are just:
```
# Sphinx build info version 1
# This file hashes the configuration used when building these files. When it is not found, a full rebuild will be done.
config: b87786b3dc86aa52b49b20323bb1e240
tags: 645f666f9bcd5a90fca523b33c5a78b7
```

larsoner:
... and if I comment out `sphinx_gallery.gen_gallery` from the `extensions`, the `config` line changes, causing a full rebuild. That build takes 150 sec. A second build takes only 9 sec, though. So indeed something about including `sphinx_gallery.gen_gallery` makes it rebuild all outputs no matter what.

jklymak:
What do you mean I changed the conf file?  I don’t touch that between builds.  

When you run make do you get 46 files added?  If so then the buildings is still not registering properly somehow.

larsoner:
Ahh sorry I misunderstood what you did. I'll take a look to see what's causing the rebuild and see if I can address it here. I have a sneaking suspicion it's our "Sphinx hack" `clean_gallery_out` function. We are deleting all SG images. We should probably only delete ones for examples that we run (i.e., make it something we clean up before running a given example only, not something we always do). But in any case I think you're right that there is probably still something wrong at the SG end.

larsoner:
@jklymak I think I have narrowed it down to these to `conf.py` entries:
```
    'subsection_order': gallery_order.sectionorder,
    'within_subsection_order': gallery_order.subsectionorder,
```
I'm guessing that it's because Sphinx can't tell if these config values have changed or not, so decides to re-write everything. With those two lines as they are in `mpl` master:

- first run: 328s
  ```
  building [html]: targets for 876 source files that are out of date
  updating environment: 2824 added, 0 changed, 0 removed
  Sphinx-gallery successfully executed 521 out of 521 files subselected by:
  ```
- second run: 70s
  ```
  building [html]: targets for 2824 source files that are out of date
  updating environment: [] 0 added, 35 changed, 0 removed
  Sphinx-gallery successfully executed 0 out of 0 files subselected by:
  ```
- change `tutorials/intermediate/artists.py`: 73s
  ```
  building [html]: targets for 2824 source files that are out of date
  updating environment: [] 0 added, 42 changed, 0 removed
  Sphinx-gallery successfully executed 1 out of 1 file subselected by:
  ```

If I comment out those two lines and do `make clean`, then:

- first run: 328s
  ```
  building [html]: targets for 876 source files that are out of date
  updating environment: 2824 added, 0 changed, 0 removed
  Sphinx-gallery successfully executed 521 out of 521 files subselected by:
  ```
- second run: 19s
  ```
  building [html]: targets for 34 source files that are out of date
  updating environment: [] 0 added, 35 changed, 0 removed
  Sphinx-gallery successfully executed 0 out of 0 files subselected by:
  ```
- change `tutorials/intermediate/artists.py`: 24s
  ```
  building [html]: targets for 35 source files that are out of date
  updating environment: [] 0 added, 42 changed, 0 removed
  Sphinx-gallery successfully executed 1 out of 1 file subselected by:
  ```
@jklymak can you confirm?

If so, what we could do is allow these entries to be strings and do some `importlib` magic with them, rather then allowing them to be classes directly. That *should* allow `sphinx` to treat them as static, rather than unknowns. In this case making changes to your classes will not be reflected in the build, but this should be (much) rarer than the case of changing a single example, and also can be worked around with a simple `make clean` (or removing the quotes from the name) when you want to do this.

jklymak:
> @jklymak can you confirm?

I can confirm the above, except for the fact that my computer is slower than yours :wink:

larsoner:
Okay I think I fixed it. Restore `conf.py` to its previous glory, then in `gallery_order.py` for `mpl` add the following to the two classes there:
```
    def __repr__(self):
        return '<%s: %s>' % (self.__class__.__name__, self.ordered_list)
```
And see if everything works properly. If it does, we just need to update our doc to mention the need for stable `str(obj)` for custom sorters.

jklymak:
... ok is this a mpl problem?  The same issue seems to be in sphinx-gallery builds as well, but maybe I'm mis-diagnosing...

jklymak:
I made a new issue for this, because I think its orthogonal to this PR.

For this PR, I'm concerned about the `Warning` for path case.   Matplotlib typically runs with warnings stop the compile, and asking users to turn that off for their local builds is asking for trouble.  Is there a better way?  Even just logging at a lower level would be fine?

jklymak:
@larsoner: yes including the `__repr__` in `gallery_order.py` allows the old `conf.py` to work and give fast incremental updates.

jklymak:
The changes here also now allow changes in `tutorials/intermediate` directory to recompile as fast as other changes!

doc/configuration.rst
```diff
@@ -188,6 +188,22 @@ the converse does not hold.
...
+             ``sphinx/builders/html.py``. Default class instances
+             in Python have their memory address in their ``__repr__`` which
+             will in general change for each build. For ``ExplicitOrder``
+             for example, this is fixed via::
+
+                 def __repr__(self):
+                     return '<%s: %s>' % (self.__class__.__name__, self.ordered_list)
+
+             Thus the files are only all rebuilt if the specified ordered list
+             is changed.
```
jklymak: Should downstream packages just subclass `ExplicitOrder`?

larsoner:
> I made a new issue for this, because I think its orthogonal to this PR.

I've fixed it here. See [updated documentation](https://730-25860190-gh.circle-artifacts.com/0/html/configuration.html#sub-gallery-order) and [fixed `ExplicitOrder.__repr__`](https://github.com/sphinx-gallery/sphinx-gallery/pull/448/files#diff-2c6a2142e40b92389192f8e472327b64R53).

> ok is this a mpl problem? 

Since `mpl` subclasses `ExplicitOrder`, it will actually be fixed by this PR's addition of `__repr__` to the `ExplicitOrder` class. And `MplExplicitSubOrder` is a class, so its `__repr__` should already be stable, So actually no new changes in `mpl` should be necessary once it's using the changes from this PR to build docs.

> For this PR, I'm concerned about the Warning for path case. Matplotlib typically runs with warnings stop the compile, and asking users to turn that off for their local builds is asking for trouble. Is there a better way? Even just logging at a lower level would be fine?

Which warning are you referring to?

doc/configuration.rst
larsoner: `mpl` does already

sphinx_gallery/backreferences.py
```diff
@@ -227,3 +232,21 @@ def write_backreferences(seen_backrefs, gallery_conf,
...
+
+    for backref in seen_backrefs:
+        path = os.path.join(gallery_conf['src_dir'],
+                            gallery_conf['backreferences_dir'],
+                            '%s.examples.new' % backref)
+        if os.path.isfile(path):
+            _replace_md5(path)
+        else:
+            logger.warning('Could not find backreferences file: %s' % (path,))
+            logger.warning('The backreferences are likely to be erroneous '
```
jklymak: This warning stops compilation if compiled with `-W`.

larsoner: The `warning` level seems appropriate for most cases. We could add a new `config` option to allow you to turn down the `backreferences` warning level for repos that really want to do so for people who have code that requires case-sensitivity building on file systems that are case insensitive.

doc/configuration.rst
jklymak: Right, I just meant it could be explicitly mentioned here; i.e. either subclass `ExplicitOrder` or make sure you have this `__repr__`....

jklymak:
So, this PR leads to a factor of five speed up in compiling the matplotlib docs under a single change for me!  I think it actually speeds up the SG parts much more than that, and the rest of the overhead is some other extensions that are not optimized to key off changes only...

sphinx_gallery/backreferences.py
jklymak: Well, lets see how it works in the wild...

doc/configuration.rst
larsoner: I don't want to suggest people subclass `ExplicitOrder` to get this functionality; and by the same token, people who subclass it for the right reasons (using its intended purpose) should immediately understand that they'll get `__repr__` for free by subclassing.

larsoner:
@jklymak can you see if the new config var is okay for you?

jklymak:
> @jklymak can you see if the new config var is okay for you?

This works fine.  Set to 'info', and MPL build works out of the box.  

OTOH, this seems a bit hacky.  I'm still a little confused about what is happening; I'm on APFS, which is case-aware, but probably case insensitive because I migrated from HFS+ on this machine.   It seems that the code to make the "*.new" files might be able to figure out that "Subplot" should be "subplot"?  It would be a strange library that had case-conflicts in their module names like that.

larsoner:
> OTOH, this seems a bit hacky.

It seems like the right behavior to `warn` by default, and the mechanism added here (dict of override values) is pretty standard (see `np.errstate` and similar functionality).

A sphinx build on a case insensitive file system with a lib that requires case sensitivity in its submodules seems like it *must* break `autosummary` already. Do the docs for `plt.subplot` and `plt.Subplot` actually look correct on your system? If so, we can look into what `autosummary` does to deal with this problem, but not in this PR. It should be a pretty rare corner case (libs that do this are not common AFAIK, and devs building on Windows, the most frequent of case-insensitive file systems, are fairly rare, too -- so the intersection shouldn't be huge).

jklymak:
Just to be clear, I'm on a mac... :wink:

jklymak:
OK, well maybe this is a different bug:

- there is no `pyplot.Subplot` in matplotlib, its `pyplot.subplot`.  
- For some reason on my machine, sg makes `/Users/jklymak/matplotlib/doc/api/_as_gen/matplotlib.pyplot.Subplot.examples`.  

I tracked this down to a few cases where `matplotlib` gallery entries called `plt.Subplot` instead of `plt.subplot`.  Removing those un-broke the doc build for me.  So...  Maybe this warning should stay in, because it indicated a bug?

larsoner:
Yes I think `warn` is the correct default, but I don't mind starting a `log_level` config var to control the logging levels. It's only a few lines and people might find it useful.

sphinx_gallery/utils.py
```diff
@@ -90,3 +91,27 @@ def replace_py_ipynb(fname):
...
+    with open(src_file, 'rb') as src_data:
+        src_content = src_data.read()
+        return hashlib.md5(src_content).hexdigest()
+
+
+def _replace_md5(fname_new, fname_old=None, method='move'):
+    assert method in ('move', 'copy')
+    if fname_old is None:
+        assert fname_new.endswith('.new')
+        fname_old = fname_new[:-4]
```
choldgraf: small nitpick, in my experience this is a bit more dangerous than just doing `os.path.splitext(fname_new)[0]`...it seems trivial but I've been bitten by it before when I change code in the future

choldgraf:
this looks really nice to me in general - and to be clear, it sounds like this doesn't require any new behavior on the part of the user, yeah? If that's the case, once we're sure that @jklymak is happy with the current behavior and since tests are passing, then I'm +1 on merge.

(I added a small nitpick comment but it shouldn't block on merging)

jklymak:
Seems great to me!

choldgraf:
will merge tomorrow unless anybody speaks up otherwise! :-)

larsoner:
> this doesn't require any new behavior on the part of the user, yeah?

Correct.

> in my experience this is a bit more dangerous than just doing ...

Done!

choldgraf:
had another appveyor config conflict...I think I resolved it properly, let's see if tests are happy then I'll merge if so!

choldgraf:
ok tests are happy so I'm 🚢 ing it

jklymak:
Thanks a lot @larsoner - this will really speed things up!

PR #451, opened 2019-02-27
BUG: Fix execution time writing

larsoner:
Tiny bugfix for names that can end in underscores like `my_example_.py`, which can cause RST parsing problems. Fix it by rendering it as code (which looks better anyway IMO).

@drammock can you see if this fixes your problem with names that end in underscores in MNE?


drammock:
confirmed: after pulling `sphinx-gallery/larsoner:fix-exec`, I ran a full doc build (`make html_dev`) while on the mne branch in mne-tools/mne-python#5944, and it completes without errors.

choldgraf:
woot! thanks for the patch

Issue #427, opened 2018-10-27
Create sphinx-gallery-extra


choldgraf:
There have been a few moments now where it seems like a particular feature would be useful (e.g. allowing for notebook inputs utils as in #423 ) but it's stalled because there's disagreement about adding new dependencies etc.

The idea has been thrown around before for a repository that can be a bit more loose in its restrictiveness. Are people still interested in this?

GaelVaroquaux:
+1 for creating a sphinx-gallery-extra

Note that the lack of reply on some of the interesting proposals that
fell in recently is nothing more than a lack of time, rather than a lack
of interest. Some of the constraints on requirements could be lifted to
some point.

Issue #91, opened 2016-01-19
Write a contributing.md file


Titan-C:
Based on https://github.com/sphinx-gallery/sphinx-gallery/pull/84#issuecomment-170370186
Some installation, configuration and testing, guidance related to development process.

melissawm:
Are you folks still open to this? I'd be happy to do it 😄

PR #455, opened 2019-03-06
adding contributing guide for releases

choldgraf:
Since we'd like to release 2.1, and since there were no contributing guides for this, I'm adding a contributing.md file to the repository, along with a to-do list for cutting a new release.

@larsoner and @Titan-C , it would be great if you can review this list and confirm that it is correct. Once we all agree on the steps, I can follow them to cut a new release and see how it goes.

## ToDo

- [ ] Rename `CONTRIBUTING.md` to `RELEASES.md`


larsoner:
I usually see Contributing having to do with how people can contribute. This is more like a how-to for maintainers, which I'd just put in the wiki

CONTRIBUTING.md
```diff
@@ -0,0 +1,27 @@
...
+
+Sphinx Gallery is [hosted on the pypi repository](https://pypi.org/project/sphinx-gallery/).
+To create a new release of Sphinx Gallery, you need to do these things:
+
+1. Ensure that you have push access to the [Sphinx Gallery pypi repository](https://pypi.org/project/sphinx-gallery/)
+2. Install [the twine package](https://twine.readthedocs.io/en/latest/). This is a package that helps you
+   bundle and push new Python package distributions to pip.
+3. Ensure that new additions, bugs, etc for this release are updated in
+   [the CHANGELOG](https://github.com/sphinx-gallery/sphinx-gallery/blob/master/CHANGES.rst)
+ release of Sphinx Gallery has
```
larsoner: cruft?

choldgraf:
there is no wiki though, is there? In general I like to keep information in a single repository, and cutting a new release is an important part of contributing...but if you think it would confuse things too much I can put it elsewhere...

larsoner:
> there is no wiki though, is there?

https://github.com/sphinx-gallery/sphinx-gallery/wiki

You would just create the first page as maintainer instructions.

>  In general I like to keep information in a single repository, and cutting a new release is an important part of contributing...

I would say it's an important part of *maintaining*, which is a small subset of (and not of most people think of when they hear) *contributing*. Most contributions come in the form of PRs. If we're going to have a `CONTRIBUTING.md`, its primary focus should be on how to make code changes to the repo, which makes it relevant to anyone who wants to make code changes. A `CONTRIBUTING.md` that focuses on how to make a release is only relevant to a handful of people a couple of times a year, and will waste the time of the much larger set of people who, with greater frequency, will want to make a PR and reasonably think "I should check the CONTRIBUTING doc to see if I'm doing things correctly".

So for now, to me the name `CONTRIBUTING.md` seems misleading for this document. I think we need to do one of:

1. Fill in the how-to-make-code-changes bits,
2. Change the name to `RELEASING.md` until we add those sections, or
3. Move the content to the wiki

larsoner:
FWIW regarding content, I have never used `twine`, so I can't verify if the instrutions are correct. Maybe amend the docs to include first an iteration over `test.pypi.org` before pushing to the real one? But if we end up having to release 3.0.1 or 3.0.2 because `twine` or the instructions are wrong, I wouldn't worry about it.

choldgraf:
that's fair - to me the problem is that we don't have good content in the `CONTRIBUTING.md` file, not that there is some content related to releases in it :-)  but fine, I will rename to RELEASES.md

larsoner:
Thanks, a `git mv RELEASES.md CONTRIBUTING.md` + edits is easy later and in the meantime we won't annoy/confuse/waste the time of people who want to add code. Probably best done as part of #427

Titan-C:
Hi,
A long time ago I wrote with @lesteve a release guide in https://github.com/sphinx-gallery/sphinx-gallery/blob/master/doc/maintainers.rst
Certainly that location has not been an evident place to look for. I also discussed about putting that on a wiki but at that time it like an extra place to look for. I would combine the maintainers.rst with your current proposition @choldgraf and then keep it as maintainers.md or releases.md in the main folder, a bit out of the docs although you can search for release on the read the docs and get the link to those docs.

choldgraf:
d'oh, I didn't realize that this was there, my bad! How do you get to that contributing page in the website? I don't see links anywhere to it

In the name of not changing too much, once I get a link to the contributing docs we can always just update the content there, and then add a `CONTRIBUTING.md` that simply links to where the actual contributing docs are.

Titan-C:
Yes, there is no link in the docs, you need to use the search bar for
"release". Then you'll jump into that page.

https://sphinx-gallery.readthedocs.io/en/latest/maintainers.html?highlight=release

The contributing.md is more a github thing, that issues and pull
requests read to help users fill those. It's a good thing if we can get
that one done too. It has been an old todo #91

choldgraf:
This is ready to go!

larsoner:
Thanks @choldgraf

PR #457, opened 2019-03-22
bump to dev version

choldgraf:


choldgraf:
since we're on a dev version now, I'll merge this once tests are happy

larsoner:
There is a conflict and I think we want a different version format, see #462

PR #461, opened 2019-03-22
Remove links to read the docs

GaelVaroquaux:
We need to stop linking to readthedocs, partly to increase the Google rank of the new site.


larsoner:
Thanks @GaelVaroquaux

PR #468, opened 2019-03-30
FIX: Minor fixes for memory profiling

larsoner:
When building the MNE doc on OSX I sometimes got:
```
  File "/Users/larsoner/python/sphinx-gallery/sphinx_gallery/gen_rst.py", line 410, in _get_memory_base
    memory_base = max(memory_usage(proc, interval=1e-3, timeout=timeout))
TypeError: '>' not supported between instances of 'NoneType' and 'float'

Exception occurred:
  File "/Users/larsoner/python/sphinx-gallery/sphinx_gallery/gen_rst.py", line 410, in _get_memory_base
    memory_base = max(memory_usage(proc, interval=1e-3, timeout=timeout))
TypeError: '>' not supported between instances of 'NoneType' and 'float'
```
It turns out sometimes the last entry in the list returned by `memory_usage` was `None`, so now we cull that from the list. It's difficult to write a test for this, so no new test has been added.

Once it worked, sometimes I would get a warning about unclosed file descriptors, so I added `close_fds=True` to `Popen`. Once we are Py3k-only, we can use `Popen` as a context manager, too, which will be nice.

Also fixes a typo where we recommend `memory_profile` instead of `memory_profiler`.

### DIFF
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -131,7 +131,7 @@ def _complete_gallery_conf(sphinx_gallery_conf, src_dir, plot_gallery,
         try:
             from memory_profiler import memory_usage  # noqa, analysis:ignore
         except ImportError:
-            logger.warning("Please install 'memory_profile' to enable peak "
+            logger.warning("Please install 'memory_profiler' to enable peak "
                            "memory measurements.")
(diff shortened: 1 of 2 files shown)


choldgraf:
tests, codecov are happy, and the code looks 👍 , thanks @larsoner !

PR #471, opened 2019-04-03
MRG: Add SVG support

### LINKED ISSUE #463: Change output format of images
I haven't found a way to change the output format for images yet. My goal would be SVG. How to do this?

larsoner:
Closes #463.

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -75,7 +75,8 @@ scraper for other python packages. This section describes how to do so.
 Image scrapers are functions (or callable class instances) that do two things:
 
 1. Collect a list of images created in the latest execution of code.
-2. Write these images to disk
+2. Write these images to disk in PNG or SVG format (with .png or .svg
+   extensions, respectively)
 3. Return rST that embeds these figures in the built documentation.
 
 The function should take the following inputs (in this order): ``block``,
@@ -86,8 +87,12 @@ a description of the inputs/outputs.
 
 This function will be called once for each code block of your examples.
 
-Example one - a Matplotlib and Mayavi-style scraper
----------------------------------------------------
+.. note:: Sphinx-gallery will take care of scaling images for the gallery
+          index page thumbnails. PNG images are scaled using Pillow, and
+          SVG images are copied.
+
+Example 1: a Matplotlib and Mayavi-style scraper
+------------------------------------------------
 
 For example, we will show sample code for a scraper for a hypothetical package.
 It uses an approach similar to what :func:`sphinx_gallery.scrapers.matplotlib_scraper`
@@ -128,8 +133,8 @@ scraper would look like::
         'image_scrapers': ('matplotlib', my_module_scraper),
     }
 
-Example two - detecting image files on disk
--------------------------------------------
+Example 2: detecting image files on disk
+----------------------------------------
 
 Here's another example that assumes that images have *already been written to
 disk*. In this case we won't *generate* any image files, we'll only generate
@@ -146,6 +151,9 @@ package in a module called ``scraper``. Here is the scraper code::
    class PNGScraper(object):
        def __init__(self):
            self.seen = set()
+
+       def __repr__(self):
+           return 'PNGScraper'
 
        def __call__(self, block, block_vars, gallery_conf):
            # Find all PNG files in the directory of this example.
@@ -168,12 +176,37 @@ package in a module called ``scraper``. Here is the scraper code::
 Then, in our ``conf.py`` file, we include the following code::
 
    from mymodule import PNGScraper
-   my_scraper_instance = PNGScraper()
 
    sphinx_gallery_conf = {
        ...
-       'image_scrapers': ('matplotlib', my_scraper_instance),
+       'image_scrapers': ('matplotlib', PNGScraper()),
    }
+
+Example 3: matplotlib with SVG format
+-------------------------------------
+The :func:`sphinx_gallery.scrapers.matplotlib_scraper` supports ``**kwargs``
+to pass to :meth:`matplotlib.figure.Figure.savefig`, one of which is the
+``format`` argument. Currently sphinx-gallery supports PNG (default) and SVG
+output formats. To use SVG, you can do::
+
+    from sphinx_gallery.scrapers import matplotlib_scraper
+
+    class matplotlib_svg_scraper(object):
+
+        def __repr__(self):
+            return self.__class__.__name__
+
+        def __call__(self, *args, **kwargs):
+            return matplotlib_scraper(*args, format='svg', **kwargs)
+
+    sphinx_gallery_conf = {
+        ...
+        image_scrapers = (matplotlib_svg_scraper(),)
+        ...
+    }
+
+You can also use different formats on a per-image basis, but this requires
+writing a customized scraper class or function.
 
 Contributing scrapers back to Sphinx-gallery
 --------------------------------------------
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -68,8 +68,8 @@ Writing a custom image scraper
 .. warning:: The API for custom scrapers is currently experimental.
 
 By default, Sphinx-gallery supports image scrapers for Matplotlib
-(:func:`sphinx_gallery.scrapers.matplotlib_scraper`) and Mayavi
-(:func:`sphinx_gallery.scrapers.mayavi_scraper`). You can also write a custom
+(:func:`~sphinx_gallery.scrapers.matplotlib_scraper`) and Mayavi
+(:func:`~sphinx_gallery.scrapers.mayavi_scraper`). You can also write a custom
 scraper for other python packages. This section describes how to do so.
 
 Image scrapers are functions (or callable class instances) that do two things:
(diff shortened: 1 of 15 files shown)


larsoner:
@steffengraber @mdeff can you see if this addresses your SVG use case? To test, use my `larsoner:svg` branch and see if the docs make it clear what to do with your doc build:

https://821-25860190-gh.circle-artifacts.com/0/rtd_html/advanced.html#example-3-matplotlib-with-svg-format

@choldgraf no tests were added because I actually changed the mode of the main SG docs to be SVG. (It works locally but there is some bug with the thumbnails on CircleCI I need to debug.) So it's at least implicitly tested. I took it this far, I trust you can add some unit tests when you add GIF support? :)

Want to add GIF support here, or in a separate PR?

doc/advanced.rst
```diff
@@ -168,13 +176,38 @@ package in a module called ``scraper``. Here is the scraper code::
...
+
+        def __repr__(self):
+            return self.__class__.__name__
+
+        def __call__(self, *args, **kwargs):
+            return matplotlib_scraper(*args, format='svg', **kwargs)
+
+    sphinx_gallery_conf = {
+        ...
+        image_scrapers = (matplotlib_svg_scraper(),)
```
mdeff:
```suggestion
        'image_scrapers': (matplotlib_svg_scraper(),)
```

mdeff:
Mostly works for me! Great improvement in terms of quality and space. :)

The images are generated as SVG. Thumbnails are however not found as sphinx-gallery looks for the PNG version (while an SVG was generated).


Two additional remarks:
1. With SVG, I don't think it's necessary to generate thumbnails anymore. They're the same files.
2. Shouldn't the choice of image format be easier for the user? Like a configuration option rather than a custom scraper. Or make SVG the default. ;) To me, SVG is strictly superior to PNG. Except maybe for very complex graphics?

larsoner:
> With SVG, I don't think it's necessary to generate thumbnails anymore. They're the same files.

Yes they're the same. However, doing this triaging will make the code (even) more complex to make this decision in the code that deals with thumbnails, which hurts maintainability. We can consider this for a future optimization but as it's not trivial and the benefits will likely be limited (just saving some hopefully small server space) let's leave it for later.

>  Thumbnails are however not found

This showed up on CircleCI. I have a fix. Just adding proper tests at the moment.

> Shouldn't the choice of image format be easier for the user? Like a configuration option rather than a custom scraper. Or make SVG the default. ;) To me, SVG is strictly superior to PNG. Except maybe for very complex graphics?

Personally I think a ~6 line class is pretty simple already :) But I guess we could add some string alias for this (`'matplotlib_svg'`?) if this is asking too much.

> To me, SVG is strictly superior to PNG. Except maybe for very complex graphics?

We try hard to maintain backward compatibility in SG. People don't want to fight with deps like ours making breaking changes to keep their (typically automated) doc builds working.

By way of example, I have a repo where we routinely plot hundreds of lines with (tens of?) thousands of points apiece in examples that get rendered by SG. SVG versions would be 10s- of MB in size and take seconds to render in the browser on a good desktop machine (forget about mobile browsing), so changing this would be a brutal backward compat break for us.

mdeff:
Makes sense to leave thumbnails as a future optimization.

> SVG versions would be 10s- of MB in size and take seconds to render in the browser on a good desktop machine

Yeah that's what I thought could happen. I guess matplotlib doesn't do drawing simplification for SVGs.

Regarding configuration, you're better positioned than me to decide whether forcing the user to define their own scraper, defining a string alias, or any other solution is best aligned with the overall design of SG. ;) I'm fine either way.

larsoner:
> Yeah that's what I thought could happen. I guess matplotlib doesn't do drawing simplification for SVGs.

Also, another caveat to be aware of (that CircleCI showed me) is that this breaks `latex` sphinx builds for PDF documentation. So I've reverted SG to use the standard `matplotlib` scraper for now.

> you're better positioned than me to decide whether forcing the user to define their own scraper, defining a string alias, or any other solution is best aligned with the overall design of SG

I think we should think about simplifying later and just support classes for now. @choldgraf is probably going to add support for GIF, we might want some "auto" option that tries to decide based on image complexity, and so forth. These use cases should eventually factor into how we design the API to choose the image type. So I'd rather wait to see how these use cases develop before settling on how to provide the simplified interface.

@choldgraf feel free to review and merge -- tests added  and CIs are green. Then you can add GIF support :)

doc/advanced.rst
```diff
@@ -147,6 +154,9 @@ package in a module called ``scraper``. Here is the scraper code::
        def __init__(self):
            self.seen = set()
 
+       def __repr__(self):
+           return 'PNGScraper'
+
        def __call__(self, block, block_vars, gallery_conf):
```
choldgraf: Just a quick note that I ran into when figuring out gif stuff with @stefanv - we should add somewhere an explanation of the inputs ("block, block_vars, gallery_conf"). We couldn't easily find this information and it was causing confusion for the scikit-image folks

doc/conf.py
```diff
@@ -312,22 +314,24 @@ def setup(app):
...
 gallery_dirs = ['auto_examples', 'tutorials']
 
+
+image_scrapers = ('matplotlib',)
 try:
     # Run the mayavi examples and find the mayavi figures if mayavi is
     # installed
     from mayavi import mlab
-    image_scrapers = ('matplotlib', 'mayavi')
+except Exception:  # can raise all sorts of errors
```
choldgraf: lol ain't that the truth

doc/advanced.rst
```diff
@@ -168,13 +178,38 @@ package in a module called ``scraper``. Here is the scraper code::
...
+
+    from sphinx_gallery.scrapers import matplotlib_scraper
+
+    class matplotlib_svg_scraper(object):
+
+        def __repr__(self):
+            return self.__class__.__name__
+
+        def __call__(self, *args, **kwargs):
+            return matplotlib_scraper(*args, format='svg', **kwargs)
```
choldgraf: so the GIF scraper would be a little bit more complex, right? Because savefig probably works differently for GIFs than for static images?

sphinx_gallery/gen_rst.py
```diff
@@ -275,7 +278,10 @@ def save_thumbnail(image_path_template, src_file, file_conf, gallery_conf):
         img = gallery_conf.get("default_thumb_file", img)
     else:
         return
-    scale_image(img, thumb_file, *gallery_conf["thumbnail_size"])
+    if ext == 'svg':
+        copyfile(img, thumb_file)
```
choldgraf: for GIF, I wonder if there's an easy way to just pull out the first frame

sphinx_gallery/scrapers.py
```diff
@@ -214,7 +239,8 @@ def save_figures(block, block_vars, gallery_conf):
                             % (scraper, type(rst), rst))
         n_new = len(image_path_iterator) - prev_count
         for ii in range(n_new):
-            current_path = image_path_iterator.paths[prev_count + ii]
+            current_path, _ = _find_image_ext(
```
choldgraf: so image path iterator outputs `.png`, but find_image_paths just doesn't use the extension when searching for images?

sphinx_gallery/tests/test_full.py
```diff
@@ -169,8 +203,8 @@ def test_embed_links_and_styles(sphinx_app):
     # highlight language
     fname = op.join(src_dir, 'auto_examples', 'plot_numpy_matplotlib.rst')
     assert op.isfile(fname)
-    with open(fname, 'rb') as fid:
-        rst = fid.read().decode('utf-8')
+    with codecs.open(fname, 'r', 'utf-8') as fid:
```
choldgraf: is `codecs` better than just `open`? I've never seen that used before

sphinx_gallery/scrapers.py
larsoner: the iterator always gives `.png` filenames yes but SG will search for outputs ending in `.png` or `.svg` currently

sphinx_gallery/tests/test_full.py
larsoner: It's the preferred method in SG (look around the codebase and you'll see it a lot). Everywhere else in the Python ecosystem I tend to see `with open(x, 'rb') as fid: fid.read().decode('utf-8')` but this was frowned upon when I did it this way a while ago.

sphinx_gallery/gen_rst.py
larsoner: I would try to actually scale it properly, it's probably not much more effort (loop rather than take first) and would have a nice payoff. But this is for the next PR :)

doc/advanced.rst
larsoner: You need to figure out how to actually save a GIF, yes. This depends entirely upon what animation you're trying to extract. We could start with an easy case of MNE animated topomap if you want to. This would involve somehow finding the matplotlib `Animation` objects.

doc/advanced.rst
larsoner: Agreed, let's do it separately though. Feel free to add an issue.

doc/advanced.rst
choldgraf: I had actually found that animated GIFs were way easier with moviepy...though perhaps that'd add more complexity in the form of a new dependency we don't want...

larsoner: Optional dependency for optional functionality seems fine

choldgraf:
ok this LGTM and we've had quite a bit of time for others to comment etc, so let's ship it! :shipit::shipit::shipit: thanks so much @larsoner :-)

larsoner:
Now it's your turn to deal with GIF / movie @choldgraf :)

PR #475, opened 2019-04-16
FIX: Output from code execution is not stripped

padix-key:
This PR removes the whitespace stripping from code output. Therefore, indent at the beginning of the output will be retained.

Example:
With this PR:
```
  A B
A 1 2
B 3 4
```
Without this PR:
```
A B
A 1 2
B 3 4
```

Is there any case, where stripping is neccessary?


larsoner:
Can you add a small test to `test_gen_rst.py` that would fail on `master` but pass on this PR?

> Is there any case, where stripping is neccessary?

I would worry that an indentation of the first line could be interpreted as the intended indentation "level" for the code block. But maybe as long as they are all indented by at least 4 spaces (as they will be) it won't matter if the first one is indented by, say, 8 spaces. Have you tried it / does it work?

If it doesn't, there is probably some RST trick to make it work.

One way it probably won't is if there is only a single line, and there is starting whitespace. In this case, we could probably do something fancy with unicode spacing to force it to actually space it out, but this might be overkill.

padix-key:
I added the requested test.

> I would worry that an indentation of the first line could be interpreted as the intended indentation "level" for the code block. But maybe as long as they are all indented by at least 4 spaces (as they will be) it won't matter if the first one is indented by, say, 8 spaces. Have you tried it / does it work?

I tested this and I found that this is not a problem.

> One way it probably won't is if there is only a single line, and there is starting whitespace. In this case, we could probably do something fancy with unicode spacing to force it to actually space it out, but this might be overkill.

I think indentation for a single output line is not necessary,  since indentation is usually used to align text in multiple lines to each other.

padix-key:
Could you have a look, what went wrong here? The failure is in `test_full.py`, a file that I have not touched and seems to happen on Windows at Python 2.7 only. I do not quite understand how my added test caused this error.

larsoner:
Thanks @padix-key !

PR #480, opened 2019-04-27
Fix ReST block after docstring

### LINKED ISSUE #473: Adding a ReST block immediately after the module docstring breaks the generated .rst file
## Bug

It appears that both module docstring and ReST blocks are stripped of blank lines at the beginning and end.

Thus the example from https://sphinx-gallery.github.io/syntax.html

~~~
"""
This is my example script
=========================

This example doesn't do much, it just makes a simple plot
"""

# This is a section header
# ------------------------
~~~

would result in an .rst file with
~~~
This example doesn't do much, it just makes a simple plot
This is a section header
------------------------
~~~

resulting in incorrect rendering of the section due to a missing line before it. A similar effect happens if you have just plain text paragraphs: They will be joined.

This appears to be a common pitfall:
https://github.com/matplotlib/matplotlib/pull/13911
https://github.com/matplotlib/matplotlib/pull/13950

## Proposed fix

Probably the correct fix would be to always add a blank line after the module docstring text in the generated .rst file.

timhoffm:
Fixes #473.

I'm not sure if there is some assumption on the termination of `bcontent`. Text blocks from rST seem to already end with a newline, docstrings don't. This patch does not rely on any assumption and generates correct rST for both cases.

The test seems a bit longish, But I would like to test the full source to rST code conversion. Just creating blocks manually and passing them to `rst_blocks()` seems a bit fragile, because the code is sensitive to newlines at the end. OTOH calling `generate_file_rst()` seems too broad since it generates a lot more stuff like the thumbnail and writes everything to files. Please advise if I should modify the test.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -714,7 +714,8 @@ def rst_blocks(script_blocks, output_blocks, file_conf, gallery_conf):
                     example_rst += "\n\n|\n\n"
                 example_rst += code_rst
         else:
-            example_rst += bcontent + '\n'
+            block_separator = '\n\n' if not bcontent.endswith('\n') else '\n'
+            example_rst += bcontent + block_separator
     return example_rst
 
 
--- a/sphinx_gallery/tests/test_gen_rst.py
+++ b/sphinx_gallery/tests/test_gen_rst.py
@@ -97,6 +97,41 @@ def test_direct_comment_after_docstring():
     assert result == expected_result
 
 
+def test_rst_block_after_docstring(gallery_conf, tmpdir):
+    """Assert there is a blank line between the docstring and rst blocks."""
+    filename = str(tmpdir.join('temp.py'))
+    with open(filename, 'w') as f:
+        f.write('\n'.join(['"Docstring"',
+                           '####################',
+                           '# Paragraph 1',
+                           '',
+                           '####################',
+                           '# Paragraph 2',
+                           '']))
+    file_conf, blocks = sg.split_code_and_text_blocks(filename)
+
+    assert file_conf == {}
+    assert blocks[0][0] == 'text'
+    assert blocks[1][0] == 'text'
+    assert blocks[2][0] == 'text'
+
+    script_vars = {'execute_script': ''}
+
+    output_blocks, time_elapsed = sg.execute_script(blocks,
+                                                    script_vars,
+                                                    gallery_conf)
+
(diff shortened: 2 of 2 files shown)


timhoffm:
Appveyor seems to have failed during fetching some external data (probably unrelated to this PR).

sphinx_gallery/gen_rst.py
```diff
@@ -714,7 +714,8 @@ def rst_blocks(script_blocks, output_blocks, file_conf, gallery_conf):
                     example_rst += "\n\n|\n\n"
                 example_rst += code_rst
         else:
-            example_rst += bcontent + '\n'
+            block_separator = '\n\n' if not bcontent.endswith('\n') else '\n'
+            example_rst += bcontent + block_separator
```
larsoner: Can't we always just insert `\n\n`? Too many newlines probably isn't a problem

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -97,6 +97,42 @@ def test_direct_comment_after_docstring():
     assert result == expected_result
 
 
+def test_rst_block_after_docstring(gallery_conf):
+    """Assert there is a blank line between the docstring and rst blocks."""
+    with tempfile.NamedTemporaryFile('w', delete=False) as f:
```
larsoner: A way to do this more consistent with what we typically do (and that does not require try/except) is to use the `pytest` pattern:
```
def test_rst_block_after_docstring(gallery_conf, tmpdir):
    fname = os.path.join(tmpdir, 'temp.py')
    with open(fname, 'w') as fid:
         f.write(...)
    ...
```
And `pytest` will clean things up for you eventually.

sphinx_gallery/gen_rst.py
timhoffm: That is true. However, it seems that all blocks except the docstring do actually have a trailing newline. While having more lines is not a problem. I didn't want the PR to introduce more newlines in the majority of cases. If you want to, I can still simplify to always add `\n\n`.

Note: This is the result of the loose definition of the block content. If one wants to do this really clean, trailing newlines on blocks should be consistent (probably be absent).

sphinx_gallery/tests/test_gen_rst.py
timhoffm: Took the other examples as a template. But you are right, tmpdir is simpler. Changed.

larsoner:
Thanks @timhoffm !

PR #482, opened 2019-04-27
Mention pytest-coverage dependency

timhoffm:
Added to README.rst.


choldgraf:
many thanks, good catch! 🎉

PR #484, opened 2019-05-01
Fix 'Out:' label position in html output block

timhoffm:
Previously the 'Out:' label was aligned at the top of the block and not with the first line.

This results in an awkward appearance in particular for single line output:

grafik

grafik

This PR fixes this:

grafik

grafik

### Explanation

The layout appears to be a bit of a hack anyway because the rST code is

~~~
 Out:

 .. code-block:: none

{0}\n
~~~

So 'Out:' is its own paragraph that would naturally appear above the code. The margin shifting in `.sphx-glr-script-out .highlight` puts it back up. I've shifted the output text down by 0.7em and conversely shifted the code back up by the same amount because it's relative to the Out-paragraph. Effectively, the code stays in the same position, only the Out text is rendered a bit lower.

### DIFF
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -83,10 +83,13 @@ thumbnail with its default link Background color */
   color: #888;
   margin: 0;
 }
+p.sphx-glr-script-out {
+    padding-top: 0.7em;
+}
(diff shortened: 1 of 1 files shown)


choldgraf:
nice 👌 ! thanks for the patch!

PR #490, opened 2019-05-17
Add list of external image scrapers

banesullivan:
I think it would be useful to start a list of image scrapers that are maintained by external projects. This helps new users realize that they can use this power tool in their own projects when leveraging plotting libraries other the Matplotlib and Mayavi.

I also think that as this tools becomes more widely adopted that plenty of other plotting libraries will maintain their own image scrapers. Sphinx-Gallery should let the world know what plotting tools are supporting this project so more projects can leverage this new way of showing examples

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -62,6 +62,11 @@ Sphinx-Gallery has also support for packages like:
 * Seaborn
 * Mayavi
 
+Other packages that maintain their own Sphinx-Gallery image scrapers include:
+
+* `PyVista <https://github.com/pyvista/pyvista>`_
+* `PyGMT <https://github.com/GenericMappingTools/pygmt>`_
+
 Install as a developer
 ----------------------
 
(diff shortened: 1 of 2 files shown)


banesullivan:
@larsoner, that's a good point. I'll close this PR and add them there

banesullivan:
In this section: https://sphinx-gallery.github.io/configuration.html#custom-scrapers

There is a note:

> If you’ve developed a custom scraper for Sphinx-gallery that would be useful to the broader community, we encourage you to contribute it to the list of natively-supported scrapers located in [the scrapers module](https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/scrapers.py). We welcome PRs!

I don't see a list in [the scrapers module](https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/scrapers.py)... should I add anything there?

PR #495, opened 2019-05-20
MRG, ENH: Catch and write warnings

### LINKED ISSUE #491: BUG: warnings/stderr not captured
Messages sent to `sys.stderr` are not captured, but they probably should be. I think that the fix is to do to `stderr` everything that we do to `stdout` here:

https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/gen_rst.py#L427

Feel free to chime in if anyone has thought of this already, otherwise I'll take care of this next week.

larsoner:
Closes #491.

It wasn't as simple as just overriding `sys.stderr` (which I did do), I also had to monkey-patch `warnings.showwarning` to use the default implementation, presumably because something else (Sphinx probably?) was overriding it already. This monkey patches it only during code execution, which seems like the right thing to do.

@drammock can you see if this fixes your problem? No config change should be necessary.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -15,11 +15,13 @@ Files that generate images should start with 'plot'.
 from __future__ import division, print_function, absolute_import
 from time import time
 import copy
+import contextlib
 import ast
 import codecs
 import gc
 import os
 import re
+import warnings
 from shutil import copyfile
 import subprocess
 import sys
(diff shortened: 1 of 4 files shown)


larsoner:
(Someday we might want to change the warning/`stderr` text color in the output, but it does not seem trivial to do this with RST formatting. This can be a follow-up PR if someone is motivated to get it working.)

drammock:
Testing now. when I was messing around with this last friday, I ended up with `stdout` and `stderr` as two separate output boxes (rather than merged and interleaved as they would be in REPL).  If that were acceptable it would be easier to style them differently.

larsoner:
Whatever we do, I think we should preserve the order of the messages. So separate boxes would only work if we had (potentially) multiple ones per code block.

drammock:
yeah, that's what I was thinking too

larsoner:
By-eye evidence that it at least captures now:

https://881-25860190-gh.circle-artifacts.com/0/tiny_html/auto_examples/plot_numpy_matplotlib.html

drammock:
It might be a nightmare to get multiple out boxes to work with the save-as-ipynb functionality. I'll be content if the warnings just show up, as it looks like they are in your test.

drammock:
I confirm that this branch works as expected on the new MNE-Python config tutorial: warnings are shown in the output, and warnings that are filtered in `conf.py` do not bomb the build process. :+1:

larsoner:
Okay, I'll wait a day or two to see if anyone else has comments, then merge as it seems like a straightforward oversight that we did not capture these before, examples and tests say it's working, and CIs are (about to come back) green.

PR #500, opened 2019-05-31
fix wrong keyword in docs for "binder"

sappelhoff:
see: https://github.com/sphinx-gallery/sphinx-gallery/blob/a69ad570fd175f593b428e8838e54b72211fe1af/sphinx_gallery/binder.py#L216

apparently the 'ref' keyword was used previously ... but it's now called "branch"


larsoner:
@choldgraf merge if you're happy

choldgraf:
ah thanks for catching this :-)

PR #501, opened 2019-05-31
DOC: Update docs

larsoner:
This PR:

1. Update docs for PyVista, cleans up the `scrapers` section. While doing this I noticed we needed to...
2. Fix warnings related to bad links. This made me think we needed to...
3. Change the default sphinx build options to `-nWT --keep-going`, that is nitpicky with warnings as errors, but not stopping, and showing a full traceback on error. This made me realize that we needed to...
4. Fix some term references in docstrings, and...
5. Triage the `logger.warning` or `logger.info` call when an example fails to build based on whether or not we expected it to fail (it should just `info` if so, not `warn`). While doing this I realized we also could...
6. Set the output color also based on expected result, so unexpected error is red just like in master but expected failure is blue (calmer, does not scream "this is wrong" but still distinct from other outputs).

Sorry for the compound changes, but at least points 2-6 should probably be done at once since they "hang together". I can separate point (1) out if necessary but this PR is still only +73/-50 so review shouldn't be too bad.


larsoner:
@banesullivan okay with https://892-25860190-gh.circle-artifacts.com/0/rtd_html/configuration.html#image-scrapers ?

banesullivan:
looks good to me - thanks for handling this!!

choldgraf:
the changes LGTM too - though circle doesn't recognize the `--keep-going` flag

choldgraf:
Here's what's breaking in travis, just one build: https://travis-ci.org/sphinx-gallery/sphinx-gallery/jobs/539977028#L334

larsoner:
Failure was spurious, restarting fixed it

choldgraf:
Ah my bad, I looked at the error and it didn't seem like a travis hiccup. You ready to merge this?

larsoner:
Yeah should be ready

choldgraf:
boom!

PR #502, opened 2019-06-07
ENH: Expose example globals

larsoner:
It is easier to create custom scrapers if we have access to the variables in the script.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -600,6 +600,7 @@ def execute_script(script_blocks, script_vars, gallery_conf):
         '__name__': '__main__',
         # Don't ever support __file__: Issues #166 #212
     }
+    script_vars['example_globals'] = example_globals
 
     argv_orig = sys.argv[:]
     if script_vars['execute_script']:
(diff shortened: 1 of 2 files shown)


choldgraf:
Could you explain a bit more what this would give us? It looks from

https://github.com/sphinx-gallery/sphinx-gallery/blob/e91e08beda412d7111401673625c60426702b32f/sphinx_gallery/gen_rst.py#L590

like this is a pre-set dictionary, but something tells me I am missing something...

larsoner:
`example_globals` is mutable and gets changed as the code runs. Basically it gives you access to the variables that were created during execution.

Concretely I'm using it to embed HTML reports from MNE using a custom scraper that looks for `mne.Report` instances, and checks to see if they have written HTML to disk yet:

https://github.com/mne-tools/mne-python/pull/6433
https://13624-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/plot_make_report.html

The tricky part so far isn't actually `sphinx-gallery` (this PR is enough to allow it) but figuring out the nightmare that seems to be proper `iframe` embedding.

choldgraf:
sounds good - 🚢 'ed away!

PR #511, opened 2019-06-21
ENH: Remove some Sphinx markup from text

### LINKED ISSUE #497: ENH: remove API crossrefs from hover text
Currently if the first sentence of an example/tutorial/whatever contains an API cross-ref, the hover text can end up looking like this:

Screenshot_2019-05-28_11-33-48

Enhancement would be to add a regex to the hover text generator that would strip out the cross-reference-generating text and leave just the class/function/method/attribute name. Ideally would be smart enough to behave the same way the cross-refs do w/r/t the presence of the leading `~`.

larsoner:
Closes #497.

@drammock can you review and see if this fixes the problem for you? Locally for me it fixes the issue you posted about.

### DIFF
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -196,7 +196,7 @@ BACKREF_THUMBNAIL_TEMPLATE = THUMBNAIL_TEMPLATE + """
 
 def _thumbnail_div(target_dir, src_dir, fname, snippet, is_backref=False,
                    check=True):
-    """Generates RST to place a thumbnail in a gallery"""
+    """Generate RST to place a thumbnail in a gallery."""
     thumb, _ = _find_image_ext(
         os.path.join(target_dir, 'images', 'thumb',
                      'sphx_glr_%s_thumb.png' % fname[:-3]))
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -196,16 +196,26 @@ def codestr2rst(codestr, lang='python', lineno=None):
     return code_directive + indented_block
 
 
+def _regroup(x):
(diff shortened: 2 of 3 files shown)


drammock:
Thanks for doing this (I meant to do it myself, but hadn't gotten to it yet).

I didn't do a thorough test of all the possibilities, but it does work for all the current MNE-Python tutorials.  In MNE-Python examples there is one tooltip ("plot a cortical parcellation") that still renders as this:

```
In this example, we download the HCP-MMP1.0 parcellation [1]_ and show it on ``fsaverage``. We ...
```

and another one ("show noise levels from empty room data") that has a tooltip

```
This shows how to use :meth:`mne.io.Raw.plot_psd` to examine noise levels of systems. See [1]_ ...
```

Would be nice if it stripped links, backticks, and `:meth:` roles too (not sure how it handles `:attr:`, `:func:`, `:doc:`, or `:math:` roles, to name the ones we most often use), but if you don't feel like doing that now I'm fine to merge as-is and tackle that improvement myself in a separate PR.

drammock:
actually I'm not sure stripping links is such a good idea. Just saw a tooltip that says:

```
This example runs the analysis described in [1]_. It illustrates how one can fit a linear class...
```

I'd say that's a poor job of writing the opening sentence, but stripping the `[1]_` will just make it worse.

larsoner:
Okay the two cases for `plot_psd` and `fsaverage` should now be fixed, can you try again @drammock ?

drammock:
@larsoner I confirm that this fixes those two cases.

larsoner:
@choldgraf feel free to merge and I'll cut a release

choldgraf:
🚢

PR #515, opened 2019-07-01
Add FURY to the sphinx-gallery users list

skoudoro:
Thank you for this nice package!


choldgraf:
travis is unhappy because python nightly is broken, but that's broken on master as well so I'm gonna merge away 🚢

PR #518, opened 2019-07-11
[MRG] EHN support #%% cell separators

### LINKED ISSUE #370: Feature request: Add an option for different cell separations

lucyleeow:
closes #370 

Support `# %%` as cell separators. 
Amend 'test_bug_cases_of_notebook_syntax.py' to test 2 input docs - the original one that uses only #'s as  a cell separator and one that uses both #'s and `# %%` as cell separators.
Amend docs to advise that `# %%` can also be used.


tutorials/plot_parse_both.py
```diff
@@ -0,0 +1,100 @@
+# -*- coding: utf-8 -*-
+"""
+The Header Docstring
+====================
+
+When writting latex in a Python string keep in mind to escape the backslashes
+or use a raw docstring
```
agramfort: this file looks too much a copy paste from https://sphinx-gallery.github.io/tutorials/plot_parse.html
I would suggest to see how to integrate both so users find the information more easily.

my 2c

lucyleeow: It IS a copy paste from that. 
The original test I modified (`test_bug_cases_of_notebook_syntax()` from https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/tests/test_gen_rst.py) tests that `split_code_and_text_blocks()` is performed correctly on the file 'tutorials/plot_parse.py'. I just amended that test to also test a version of 'plot_parse.py' ('plot_parse_both.py') which uses `# %%` as separators as well as `#`'s. 

Sorry, what do you mean integrate both? I can change the test to test only the 'new' plot_parse_both.py file which uses both `#`'s and `# %%` as a separator?

agramfort: now 2 almost identical files are going to show up on the documentation. I agree that it does the job for the testing but from a narrative doc perspective I would suggest this new example is as easy to discover as the old one OR to integrate them into one (at least for the documentation perspective).

my 2c

lucyleeow: Ah I see the problem now. I will integrate them into one.

choldgraf:
shit why didn't we think of this before haha, I think it's a good patch, seems reasonable to me. So if I wanted to add rST underneath it would look something like:

```python
# %%
# my rst
# ======

my = "code"
```

?

tutorials/plot_parse.py
```diff
@@ -11,7 +11,7 @@
 Closing this string quotes on same line"""
 
 
-##############################################################################
+# %%
```
agramfort: I would add a note that at the top of the file that 2 types of separators are possible.  Say you'll use
the ### at the beginning of the file, put a note saying that you'll use in what follows the alternative
approach with # %%

does it make sense?

lucyleeow: Yes I realised above when you talked about the double up  that the plot_parse.py file is actually a page in the sphinx-gallery documentation (whoops). Am amending this page to add the relevant information.

doc/syntax.rst
```diff
@@ -39,8 +39,9 @@ Jupyter Notebooks are structured (in fact, Sphinx-Gallery also **creates** a
 Jupyter Notebook for each example that is built).
 
 You can embed rST in your Python examples by including a line of ``#`` symbols
-that spans >= 20 columns. We recommend using 79 columns, like
-this::
+that spans >= 20 columns or ``# %%`` (consistent with Jupyter VS code plugin, 
+Pycharm, Hydrogen plugin and Spyder). If using ``#``'s, we recommend using 79
```
GaelVaroquaux: I would add jupytext on that list. Also, I would try to make these links.

lucyleeow: There are two 'v0.4.0' in the changelog (https://sphinx-gallery.github.io/changes.html#id1). Is this correct?

larsoner: No, the topmost one should be 0.5. Feel free to change it here

lucyleeow:
After speaking with @GaelVaroquaux we think both `# %%` and `#%%` (with and without space) should be supported:

- [Spyder](https://docs.spyder-ide.org/editor.html) supports both because Eclipse's automatic code formatting changes `#%%` into `# %%` (see [this](https://github.com/spyder-ide/spyder/issues/852#issuecomment-74599857) issue)
- [Hydrogen](https://nteract.gitbooks.io/hydrogen/docs/Usage/NotebookFiles.html#cell-markers) states that `# %%` is supported in their documentation but on testing both syntaxes work and exports correct cell formatting to jupyter notebook.
- [VSCode](https://code.visualstudio.com/docs/python/jupyter-support) states that `#%%` is supported but on testing both syntaxes work. 
- Could not test in PyCharm as this functionality is only available in their professional version.

If there are no objections, I will change the code and documentation to support both syntaxes. I think I will state that a line of `#`'s or `#%%` can be used and just note that `# %%` is also acceptable for compatibility.

larsoner:
Works for me

sphinx_gallery/tests/reference_parse.txt
```diff
@@ -1,33 +1,39 @@
 [('text',
   '\nThe Header Docstring\n====================\n\nWhen writting latex in a Python string keep in mind to escape the backslashes\nor use a raw docstring\n\n.. math:: \\sin (x)\n\nClosing this string quotes on same line',
   1),
- ('text', 'Direct first comment\nwith second line\n', 15),
- ('code', '\nimport numpy as np\n\n', 17),
- ('code', 'A = 1\n\nimport matplotlib.pyplot as plt\n\n', 21),
- ('text',
+  ('text',  'Direct first comment\nwith second line\n',  15),
```
larsoner: These look PEP8 incompatible (style regressions)

larsoner: (would be better to change only what you must to get tests to pass, helps with `blame`)

larsoner:
> I think that an entry in the changelog is needed.

I just noticed this bit. FYI we don't need changelog entries because we autogenerate them based on PR labels. But it's fine to leave the one you added in there

sphinx_gallery/tests/reference_parse.txt
agramfort:
```suggestion
  ('text', 'Direct first comment\nwith second line\n',  15),
```

lucyleeow: I think I fixed it. Apologies - there was one or two typos I couldn't resist fixing.

sphinx_gallery/py_source_parser.py
```diff
@@ -161,7 +161,7 @@ def split_code_and_text_blocks(source_file):
     file_conf = extract_file_config(rest_of_content)
 
     pattern = re.compile(
-        r'(?P<header_line>^#{20,}.*)\s(?P<text_content>(?:^#.*\s)*)',
+        r'(?P<header_line>^#{20,}.*|^#%%.*|^# %%.*)\s(?P<text_content>(?:^#.*\s)*)',
```
larsoner:
```suggestion
        r'(?P<header_line>^#{20,}.*^# ?%%.*)\s(?P<text_content>(?:^#.*\s)*)',
```
I think this should work and is more compact

lucyleeow: Thanks, don't know why I didn't think of that!

lucyleeow:
Amended the doc to add a note on using sphinx gallery with code blocks in various IDE's. Hopefully this makes sense.

choldgraf:
This seems good to me! @GaelVaroquaux you're still blocking changes just FYI :-)

larsoner:
> @GaelVaroquaux you're still blocking changes just FYI :-)

From the above I'm pretty sure it's just for the changelog update and link updating which are done. So I'll merge but if there was something still wrong @GaelVaroquaux feel free to chime in and we'll fix it.

Thanks @lucyleeow !

GaelVaroquaux:
>     @GaelVaroquaux you're still blocking changes just FYI :-)

Aaaaaaarg!

> From the above I'm pretty sure it's just for the changelog update and link
> updating which are done. So I'll merge but if there was something still wrong
> @GaelVaroquaux feel free to chime in and we'll fix it.

No, this is great. Thank you!

rth:
Thanks for this @lucyleeow ! Is going to make scikit-learn examples much more readable in plain text mode https://github.com/scikit-learn/scikit-learn/pull/17068

Issue #488, opened 2019-05-17
Matplotlib raises warning when 'pyplot.show()' is called


padix-key:
Whenever `matplotlib.pyplot.show()` is called in an example script, the following warning is displayed:

`UserWarning: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.`

I am using `matplotlib 3.0.3` and `sphinx-gallery 0.3.1` at Python 3.6.

padix-key:
Possible solutions could be
- filtering the warning out
- temporarily replacing the `show()` function with an empty function

larsoner:
I think we should add a note to the docs about how to use simplefilter or something similar to do this (for projects that want to).

PR #521, opened 2019-07-30
[MRG] Remove matplotlib agg backend + plt.show warnings from doc

lesteve:
The warnings happen when using agg backend and using plt.show(). This was taken from scikit-learn. This is related to https://github.com/sphinx-gallery/sphinx-gallery/issues/488. This PR only removes the warnings from our documentation and does not add any documentation from a sphinx-gallery user perspective.

See for example https://sphinx-gallery.github.io/auto_examples/plot_exp.html#sphx-glr-auto-examples-plot-exp-py.

image

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -15,8 +15,9 @@
 import sys
 import os
 from datetime import date
+import warnings
+
 import sphinx_gallery
-from sphinx_gallery.scrapers import matplotlib_scraper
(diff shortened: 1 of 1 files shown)


larsoner:
We do this in MNE as well, seems like the right thing to do.

AppVeyor is unrelated, merging, thanks @lesteve

PR #524, opened 2019-08-01
MAINT: Fix AppVeyor

larsoner:


larsoner:
Looks like it might not be crashing anymore. And the CI-helpers folks [say they don't use the latest anaconda because it tends to cause problems](https://github.com/astropy/ci-helpers/issues/353#issuecomment-472524091), so removing our specification to use the latest conda version seems reasonable.

PR #525, opened 2019-08-02
ENH: Add live object refs and methods

### LINKED ISSUE #402: Inspiration from new gallery package for sphinx: sphinx-exhibit
hey folks - I just found out about an interesting project that @anntzer from the Matplotlib community is working on. It's called "sphinx-exhibit" and replicates many of the things that sphinx-gallery does with a few improvements:

https://github.com/anntzer/sphinx-exhibit

The main differences to me seem to be:

1. sphinx-exhibit does more introspection of the code to generate intersphinx links for methods etc, similar to what @larsoner was working on in a recent PR (don't remember which one).
2. sphinx-exhibit exposes some sphinx directives so you can embed a gallery of specific examples files.

Anyway, worth keeping an eye on the project! Maybe the build process of sphinx-exhibit is different enough that it isn't possible to mimic in sphinx-gallery, but worth checking nonetheless!

larsoner:
This adds resolution of (most/many) runtime objects using intersphinx. In `master` we only inspected the AST and imports to look for function calls from imported modules. This PR looks at the runtime objects in `globals` to see if they (or their subclasses) are documented somewhere in intersphinx. It greatly increases the number of things in the code snippets that become links. I'll paste a link to an updated example once CircleCI renders so that people can see what I mean.

It's not perfect but it gets us at least some of the way. For example:

1. It does not handle subclasses properly in all cases. ~~For `mne.io.Raw` for example, it properly documents `raw.plot_psd` calls, but does not get `raw.filter` (despite there being documentation for `mne.io.Raw.filter`).~~
2. If objects are overwritten with the same names, e.g. `raw = ...` at the beginning then later `raw = ... <some other class>` then whichever is first ~~last in the global namespace at the end of the script~~ will be used.

But we can probably handle this sort of thing in a subsequent PR -- the links here are already a good improvement IMO. (1) can be addressed by improving the logic that is in `backreferences.rst`. (2) can be addressed by a much larger PR (that is probably not worth it) refactoring when the objects are inspected and replacements made/set up. To do it properly, the code would need to be executed one line at a time, and name substitutions queued on a line-by-line-within-example basis instead of a per-example basis.

Closes #402.

### DIFF
--- a/continuous_integration/travis/test_script.sh
+++ b/continuous_integration/travis/test_script.sh
@@ -6,7 +6,7 @@
 
 set -e
 
-pytest sphinx_gallery
+pytest sphinx_gallery -vv
 cd doc
 if [ "$DISTRIB" != "minimal" ] && [ "$PYTHON_VERSION" != "nightly" ]; then
     make SPHINXOPTS= html-noplot
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -22,7 +22,7 @@ file:
 - ``subsection_order`` (:ref:`sub_gallery_order`)
 - ``within_subsection_order`` (:ref:`within_gallery_order`)
 - ``reference_url`` (:ref:`link_to_documentation`)
-- ``backreferences_dir`` and ``doc_module`` (:ref:`references_to_examples`)
+- ``backreferences_dir``, ``doc_module``, and ``inspect_global_variables`` (:ref:`references_to_examples`)
 - ``default_thumb_file`` (:ref:`custom_default_thumb`)
 - ``thumbnail_size`` (:ref:`setting_thumbnail_size`)
 - ``line_numbers`` (:ref:`adding_line_numbers`)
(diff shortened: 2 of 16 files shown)


larsoner:
Example:

https://975-25860190-gh.circle-artifacts.com/0/rtd_html/auto_examples/plot_function_identifier.html

choldgraf:
Thanks for this PR! I agree this functionality is super useful. It's a little be beefy for me to be able to get to quickly...what is your "eric confidence this won't break everything" level on this one? I am trying to figure out how much of a close eye to put on this one :-)

larsoner:
So far I built the MNE docs, SG docs, and SG tinybuild with it and it seems to work in those cases (with the exceptions noted above).

@banesullivan can you try this branch and see if PyVista would benefit from this, too?

banesullivan:
Hey @larsoner - Its been a busy week, I'll give this a try sometime today!

larsoner:
> what is your "eric confidence this won't break everything" level on this one? I am trying to figure out how much of a close eye to put on this one :-)

Just to follow up -- it's low enough that I have now [added a config variable](https://998-25860190-gh.circle-artifacts.com/0/rtd_html/configuration.html#references-to-examples) to disable the global variable inspection :)

I built the `matplotlib` documentation on this PR, and I get 8 pages worth of examples instead of just 1 page for `matplotlib.axes.Axes.plot`. This is a lot, but is probably still usable (?).

In the meantime, I fixed some bugs, including the one where `mne.io.Raw.filter` was not properly found. This is what it now does on the MNE docs:

https://14662-1301584-gh.circle-artifacts.com/0/dev/generated/mne.io.Raw.html#mne.io.Raw.filter

choldgraf:
nice - I agree that this would be really helpful, and disabling it seems like a reasonable option. Maybe we can set it as "disabled" by default and then enable by default in a subsequent release?

larsoner:
> Maybe we can set it as "disabled" by default and then enable by default in a subsequent release?

I think in some sense it's meant as a bug fix -- we always tried to do this but didn't do it so well. You want to try it with `mpl`, see how well it works, and decide based on that? `mpl` probably covers most, if not all, of the complicated cases we'd want to support.

choldgraf:
Was the main concern that it'd just generate a **ton** of pages some methods in MPL? If that's not the main concern and we're more concerned about it missing some things, I'd err on the side of just merging this because, as you say, it's an improvement over the behavior that most people probably *assume* will be the case...

larsoner:
> Was the main concern that it'd just generate a ton of pages some methods in MPL? 

Originally I was worried about that a little bit. But even the page for `Axes.plot` is not so terrible, you can tweak the thumbnails to be smaller with CSS if you want, and each method has its own page (and all examples are listed at the bottom) so it seems okay even there.

choldgraf:
OK, I'm merging this one - the travis errors are on 3.7 and seem to be related to VTK and not this PR! Onward and upward 🚀

PR #526, opened 2019-08-03
ENH: Speed up builds

larsoner:
When working on #525 I noticed that `make html_dev-noplot` builds for MNE-Python were spending a lot of time in the SG section. I profiled it and there was ~43 sec spent in `generate_gallery_rst` and things that it calls, even though nothing is executed.

Turns out most of this is due to unnecessary `clean_modules` and `_memory_usage` calls -- we only need to call those before actually executing files. I've changed the code to avoid unnecessary calls (i.e., only call them before attempting to actually execute a script). Now `generate_gallery_rst` takes ~3 sec.

### DIFF
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -165,9 +165,9 @@ THUMBNAIL_TEMPLATE = """
 
 .. only:: html
 
-    .. figure:: /{thumbnail}
+ .. figure:: /{thumbnail}
 
-        :ref:`sphx_glr_{ref_name}`
+     :ref:`sphx_glr_{ref_name}`
 
 .. raw:: html
 
@@ -177,7 +177,7 @@ THUMBNAIL_TEMPLATE = """
 BACKREF_THUMBNAIL_TEMPLATE = THUMBNAIL_TEMPLATE + """
 .. only:: not html
 
-    * :ref:`sphx_glr_{ref_name}`
+ * :ref:`sphx_glr_{ref_name}`
 """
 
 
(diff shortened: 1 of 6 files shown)


larsoner:
@choldgraf this one is also ready

sphinx_gallery/backreferences.py
```diff
@@ -165,9 +165,9 @@ def scan_used_functions(example_file, gallery_conf):
 
 .. only:: html
 
-    .. figure:: /{thumbnail}
+ .. figure:: /{thumbnail}
```
larsoner: FYI these indentation changes are useful because they prevent whitespace from being added in the HTML. Instead of having:
```
   .. figure::
```
this comes out with no leading whitespace as:
```
.. figure::
```
This is better RST, and also necessary if we want to replace the header with `.. rubric::` (which is what I'm using over in MNE to get methods and attributes linked using backrefs)

choldgraf:
This all looks pretty reasonable to me - it's mostly removing (unnecessary) steps and tests are still happy, so I am gonna merge away :-) thanks @larsoner !!!

sphinx_gallery/gen_rst.py
```diff
@@ -592,10 +588,12 @@ def execute_script(script_blocks, script_vars, gallery_conf):
         # for more details.
         sys.argv[0] = script_vars['src_file']
         sys.argv[1:] = []
+        gc.collect()
```
lucyleeow, line 591: @larsoner sorry this PR is very old but will this clean up `global_variables` ?

larsoner: I think that was the goal, yeah

lucyleeow: Thanks! So running gc.collect in a resetter would be redundant then?

larsoner: I'd have to think about execution order, but in principle I think so yes. But also running it again shouldn't hurt. Personally I do `gc.collect` in a resetter in MNE-Python then check that there are no objects of specific types lingering around, e.g.:

https://github.com/mne-tools/mne-python/blob/66c40caea1fd343fbe49a34815542065ceecdf3d/doc/conf.py#L478-L491

lucyleeow: Thanks! I ask because I noticed in MNE you run `gc.collect` in the resetter and again in `_assert_no_instances`, but I guess that is because `_assert_no_instances` is run outside of the resetter. Anyway thank you!

PR #527, opened 2019-08-06
BUG: Fix kwargs modification in loop

larsoner:
@drammock can you see if this fixes https://github.com/mne-tools/mne-python/issues/6634#issuecomment-518774399 for you?

Basically we modify `kwargs` within the loop over figures, which is bad since it means that `kwargs` for figures later down the list can be modified by however we modified them for the preceding ones.


drammock:
confirmed, fixes for me locally.

larsoner:
This is a pretty unambiguous bug and @drammock confirmed it works so I'll go ahead and merge

ImportanceOfBeingErnest:
This fixes https://github.com/matplotlib/matplotlib/issues/15004 for us. Thanks.

PR #530, opened 2019-08-12
MAINT: Work around Anaconda bug

larsoner:
Our recent failures on `master` are probably due to https://github.com/conda/conda/issues/9084, let's set the Python version to work around it


larsoner:
They say that this will be fixed imminently:

https://github.com/ContinuumIO/anaconda-issues/issues/11195

So I'll close this for now

PR #535, opened 2019-09-04
look for both 'README' and 'readme'

### LINKED ISSUE #534: consider looking also for 'readme.*' instead of only 'README.*'
Is there any reason to consider only 'README.*' in uppercase and not include any other variation in lower case?

It would be of great help to all our projects which are constructed around 'readme.rst' files.

I assume it's only a matter of changing the function '_get_readme' in 'gen_rst.py' but maybe there is a non technical reason for not doing it.

Thanks

revesansparole:
Closes #534


larsoner:
Thanks @revesansparole

Issue #143, opened 2016-08-31
DOC: each example should start by explaining why it's there


GaelVaroquaux:
The examples are actually doc / illustrations for sphinx-gallery. In this logic, we should start each example with a short paragraph that explains what the example demonstrates.

In this paragraph, we should probably push the user to look at the Python source code. Indeed, as sphinx-gallery becomes of interest to users that are less and less experts, they will start landing on the gallery of example without reading the big picture, and hence they won't realize that everything is generated from the Python file.

GaelVaroquaux:
By the way, some example already explain why they are there, others don't.

PR #541, opened 2019-09-24
[MRG] ENH: Capture 'repr's of last expression

### LINKED ISSUE #540: ENH capture 'repr's of last expression
Relates to #396 (pretty html rendering) which can be implemented in 2 steps:

1. (this issue) `eval()` last line if it is an expression. This removes the need to use `print()` to output the value of expressions.
2. Capture html output if there is a `__repr_html__`. This allows nice rendering of e.g. pandas dataframes, plots and reports (currently implemented via a 'image' scraper in [nilearn](https://github.com/nilearn/nilearn/pull/2019), [nistats](https://github.com/nistats/nistats/pull/387), [mne](https://github.com/mne-tools/mne-python/pull/6433))

There has been discussion at #421 (and a WIP PR #422) about using nbconvert to execute code blocks, which would address both issues above, however there were concerns about adding dependencies and no work around was agreed upon.

The plan for this issue would be to implement the same ['exec/eval' logic ](https://github.com/jupyter/jupyter-sphinx/blob/b1d5a4568fc697e8ffb07d669432592e5f0f3aa0/jupyter_sphinx/embed_widgets.py#L65) from jupyter-sphinx (as suggested by @maartenbreddels) in [`execute_code_block()`](https://github.com/sphinx-gallery/sphinx-gallery/blob/master//sphinx_gallery/gen_rst.py#L456:5). Thoughts welcome.

(cc @GaelVaroquaux)

lucyleeow:
Closes #540 

First attempt. Works but is not pretty.

First, I could not use logic in [Jupyter-Sphinx](https://github.com/jupyter/jupyter-sphinx/blob/b1d5a4568fc697e8ffb07d669432592e5f0f3aa0/jupyter_sphinx/embed_widgets.py#L65) because the value of the last node in body is made into an expression (`last = ast.Expression(block.body.pop().value)`) and evaluated regardless of whether it is an expression or not. This means that if the last node is a statement (e.g. `b=2`) the value (`2`) will be made into an expression and `2` will be returned (which it shouldn't be) and `b=2` is never run (and the variable not available later). The `pop()` removes this node from the body, which is later executed (let me know if I missed something).

Notes:
* If the last line is an expression, it needs to be removed from 'body', so that it is not run twice. e.g. a `print()` expression should not be run twice, as both outputs would be captured.
* The 'body' needs to be executed before the last expression, so variables assigned in the 'body' are available.

I could not work out how to memory profile the last expression. I think ideally memory profiler should be run on both the 'body' and the last expression and the max mem of the 2 used. I tried using `_memory_usage()` with [`_exec_once()`](https://github.com/sphinx-gallery/sphinx-gallery/blob/33220c8148613107f0fc771b8b8b4461408385a0/sphinx_gallery/gen_rst.py#L403) (with `exec()` changed to `eval()` within the function) but variables assigned in the body were not available. Is it because of `globals_` being set within `_exec_once()` (?):

```
def __init__(self, code, globals_):
        self.code = code
        self.globals = globals_
        self.run = False
```
Suggestions? @GaelVaroquaux ?


lucyleeow:
Worked out the memory profiling - you need to `return` the `eval()` at the end of `_eval_once()`, otherwise the output was always `None`.

Everything works - but the code is not pretty. Review?

larsoner:
I should be able to look tomorrow. Glad you were able to figure out the eval/exec business in the meantime!

lucyleeow:
(I think this is a AppVeyor issue as all I changed were some variable names)

larsoner:
Looks good!

This should probably also be used at least once in the official docs. We use the gallery of `sphinx-gallery` itself to show functionality, so could you modify some example to show it, then once CircleCI renders it, paste a link here?

larsoner:
I would also remove the `Closes` tag at the top since this only takes care of the first of the two points in that issue

lucyleeow:
Thanks!

> This should probably also be used at least once in the official docs. We use the gallery of `sphinx-gallery` itself to show functionality, so could you modify some example to show it, then once CircleCI renders it, paste a link here?

I can simply alter [Local module](https://sphinx-gallery.github.io/auto_examples/local_module.html) so the code is:

```Python
N=100
N
```

It feels trivial but it shows the main point of the ENH. I would also like to add a note about the `eval()` in the getting started section.

While looking into this I have realised that unwanted text is now output while making matplotlib plots - 
`eval("plt.ylabel('$\sin(x)$')")` outputs `Text(0, 0.5, '$\\sin(x)$')` (as well as the graph), whereas `exec("plt.ylabel('$\sin(x)$')")` does not. This is also a problem in IPython/Jupyter (see [SO](https://stackoverflow.com/questions/14506583/suppress-output-of-object-when-plotting-in-ipython)). The workarounds are:

* to assign the output to a temp variable (`_=plt.ylabel('$\sin(x)$')`)
* add `;` to the end of the expression
* use `plt.show()` as the last line

I have tested these and they all work here. I think the best/easiest solution is just to document this, either in the first gallery with a plot: [Colourmaps](https://sphinx-gallery.github.io/auto_examples/plot_colors.html) or the [Sine function](https://sphinx-gallery.github.io/auto_examples/sin_func/plot_sin.html#sphx-glr-auto-examples-sin-func-plot-sin-py)

(On this note I would like to tidy up the galleries (e.g. Sine not Sin!) and address #143 and I might do this before I tackle the `__repr_html__`)




> I would also remove the `Closes` tag at the top since this only takes care of the first of the two points in that issue

I wanted to make this into 2 issues and PRs but I can understand that it makes sense to discuss in 1 issue - in which case I could have just added to the end of the old issue..! Alas, I will remove 'closes'.

larsoner:
> It feels trivial but it shows the main point of the ENH.

Sure, and once we support `__html_repr__` you can add another or change it to an object (`pandas` dataframe?) with a more interesting one to show off that functionality.

> I would also like to add a note about the eval() in the getting started section. WDYT?

Sure

> While looking into this I have realised that unwanted text is now output while making matplotlib plots

This is going to be a potentially nasty backward compat break, as most libraries I work with do many such statements, and having the `__repr__` printed every time will probably be annoing. We'll want a backward-compat config var for this, maybe:
```
print_eval_repr = ()
```
Does what `master` does (prints nothing ever based on the last line), while:
```
print_eval_repr = ('__repr_html__', '__repr__')
```
will print the `__repr_html__` (to be added in another PR) if available, and if not available, will print the `__repr__` of the line (to be done in this PR) of any line that does not contain an assignment/semicolon. Someday we can add finer-grained control by allowing things like `# sg-(no-)print-eval-html-repr` or something to enable disabling/enabling it on a line-by-line basis. But a global option to start is probably pretty good.

In MNE for example we will probably add `__html_repr__` to our `Report` class, then set `sphinx_gallery_conf['print_eval_repr'] = ('__html_repr__',)`, and have things work a reasonable way.

lucyleeow:
> Sure, and once we support __repr_html__ you can add another or change it to an object (pandas dataframe?) with a more interesting one to show off that functionality.

Yes that is a good idea. I think it would be useful to make a new ``.py`` file for this, as it doesn't seem to really fit with any of the other example galleries.

With regards to the Matplotlib problem - this bothered me all weekend but I could not think of a better solution. I have added the configuration.

A few points:
* In my mind, once we support ``__repr_html__``, there would be 2 config options:
    * exec everything, as with master
    * exec everything except the last statement. If it is an expression eval it and capture ``__repr_html__`` if it exists and if not just ``__repr__``

Thus it does not make sense to me why we would make the config equal to ``()`` for the first option and ``('__repr_html__', '__repr__')`` for the second. Unless you wanted 3 options, with the 3rd option being eval the last expression and capture ``__repr__`` only, and never capture ``__repr_html__``. Maybe I am missing something?

* With regard to finer grain control I think per code block control is adequate, since the config only changes what happens to the last statement in a code block anyway. 

> In MNE for example we will probably add __repr_html__ to our Report class, then set sphinx_gallery_conf['print_eval_repr'] = ('__repr_html__',), and have things work a reasonable way.

I am not sure how this would work. I understand fine grain control to work like 
[``# sphinx_gallery_line_numbers``](https://sphinx-gallery.github.io/configuration.html#adding-line-numbers), which can be set on a file-by-file basis (except we would make it work by a per code block basis). Would the Report class add something like ``# sg-(no-)print-eval-html-repr`` to the source ``.py`` file?

For an existing Sphinx-Gallery project, I feel like it would be annoying to add this configuration (in new example ``.py`` files) to every code block, even if you just wanted to nicely print a pandas df nicely or not have to add ``print()``, for example. File-by-file would be nice in this case, though I understand something like the Report class solution would allow easy migration to the new system for old ``.py`` files... Not sure of the best solution.

larsoner:
> In my mind, once we support __repr_html__, there would be 2 config options:
> 1. exec everything, as with master
> 2. exec everything except the last statement. If it is an expression eval it and capture __repr_html__ if it exists and if not just __repr__

I would not branch the code this way. I would *always* execute the second way, then the config option just determines which of the results we actually display in the output. Seems simpler to me.

>  Unless you wanted 3 options, with the 3rd option being eval the last expression and capture __repr__ only, and never capture __repr_html__. Maybe I am missing something?

We need `()` for backward compat. In MNE we would probably use `('__repr_html__`,)` so we only display objects with a nice HTML repr. Other packages will want `('__repr_html__', '__repr__')` (display repr for every last line). And it's future compatible with displaying other things, e.g. people could do `('__repr_html__', '__str__', '__repr__')` if they want. Your code just needs to try to run the options they give in order and check if the result is `str`.

> I am not sure how this would work.

Hopefully it makes sense based on what I said above

> Would the Report class add something like # sg-(no-)print-eval-html-repr to the source .py file?

No we will add a `__repr_html__` method to the `Report` class, set `print_eval_repr = ('__repr_html__',)` to our config, and add:
```
report = Report(...)
report  # <-- this line
```
To our examples.

larsoner:
> I am not sure how this would work. I understand fine grain control to work like
# sphinx_gallery_line_numbers, ... File-by-file would be nice in this case ...

Let's worry about this when we actually need it. For now let's work on sorting out how the main config option should work, then when someone hits a use case let's figure out the file/block/line-based fine-grained control.

larsoner:
> I would not branch the code this way. I would always execute the second way, then the config option just determines which of the results we actually display in the output. Seems simpler to me.

To lay out more clearly, this is what I would do:
```
exec(all_but_last_line)
obj = eval(last_line)
last_str = None
for meth in sg_config['print_eval_repr']:
    try:
        last_str = getattr(obj, meth)()
    except Exception:
        pass
    else:
        if isinstance(last_str, str):
            break
if isinstance(last_str, str):
    append_as_rst(str)
```

lucyleeow:
> I would not branch the code this way. I would always execute the second way, then the config option just determines which of the results we actually display in the output. Seems simpler to me.

Yes that makes much more sense. And trying the ``__repr__`` in order of preference given by the config sounds elegant solution. I will amend this.

With regard to the fine grain control, I was mostly asking how the config would be changed by the ``.py`` file but as you said we can worry about this later.

doc/configuration.rst
```diff
@@ -957,41 +957,42 @@ you can do::
 Controlling what output is captured
 ===================================
 
-Sphinx-Gallery can run your code blocks in two ways:
+The ``print_eval_repr`` configuration allows the user to control what output of
+the last expression in a code block, is captured and rendered in the built
+documentation. By default ``print_eval_repr`` is an empty tuple. With this
+setting, the last statement in a code block is executed with ``exec()`` and
```
larsoner: How this is accomplished (`eval` vs `exec`) is an implementation detail SG users don't need to know/care about. I would just describe that:
```
x = pandas.DataFrame(...)
```
is never rendered, but:
```
x
```
will print depending on the value of `print_eval_repr`.

doc/configuration.rst
```diff
@@ -957,41 +957,42 @@ you can do::
...
-  standard output is captured.
-* **Exec eval** - ``exec()`` everything in the code block *except* the last
-  statement. If the last statement is an expression, ``eval()`` it and if it is
-  not an expression, ``exec()`` it. Both data directed to standard output
-  and the value of the last statement, if it is an expression, is captured.
+If you added ``__repr__`` to the tuple like so::
+
+        sphinx_gallery_conf = {
+            ...
+            'print_eval_repr': ('__repr__',),
```
larsoner: I would make the example `('__html_repr__', '__repr__')` and describe that it will go through the tuple in order and try to execute one that returns a `str` (which `__repr__` should always do) so in this case something will always be renderered for a non-assigment last line.

doc/getting_started.rst
```diff
@@ -171,7 +174,7 @@ Sphinx source ``.rst`` file with an ``.. include::`` statement.
 Build the documentation
 -----------------------
 
-In your Sphinx source directory, (e.g., ``myproject/doc``) execute:
+In your Sphinx source directory, (e.g. ``myproject/doc``) execute:
```
larsoner: See e.g., https://www.aje.com/arc/editing-tip-using-eg-and-ie/

> Most style guides suggest the use of a comma after both e.g. and i.e.

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -469,6 +469,35 @@ def test_empty_output_box(gallery_conf):
     assert output.isspace()
 
 
+@pytest.mark.parametrize('code,out', [
+    ('a=2\nb=3', ''),
+    ('a=2\na', '2'),
+    ('a=2\nprint(a)', '2'),
+    ('print("hello")\na=2\na', 'hello\n\n2')])
```
larsoner: You could also parametrize over whether `print_eval_repr` is `('__repr__',)` or `()` and triage the `output_test_string` check below based on (in the empty case) whether the last line contains `print()` or not

doc/configuration.rst
lucyleeow: > How this is accomplished (eval vs exec) is an implementation detail SG users don't need to know/care about. 

You're right and I was thinking this when I wrote it.

> will print depending on the value of print_eval_repr.

This is a better explanation. The problem is that I have not implemented capturing `_repr_html_` yet and I was trying to make this PR and thus the documentation in it, applying only to the 'exec/eval'.

larsoner: If you want, you can make the docs for now just about `()` and `('__repr__',)`, and add `__html_repr__` support separately

lucyleeow:
I was initially thinking to separate the 2 steps (as per #540) into 2 PRs (to keep it modular..?) but I can see also that it makes sense to do it all in 1 PR. WDYT?

The problem is that I am having trouble working out how to capture anything else apart from ``__repr__()``. At the moment I am evaluating (``eval()``) ``ast.Expression(body=code_ast.body.pop().value)`` (the ``value`` of last node of the parsed ``code_ast`` as a ``ast.Expression()``). It returns the ``__repr__()`` by default. I have not worked out how to capture the other methods using ``ast``.

In the #420 POC PR, it is done using ``IPython.display.display`` (which I think is [here](https://github.com/ipython/ipython/blob/4a2240fb440c497e9e195d9ff43d96973b7ae33f/IPython/core/display.py#L131)). I am not sure how this works and if I should be trying to trying to implement in a similar way...

Suggestions welcome, otherwise I will be banging my head against ``ast`` documentation for the rest of the week...

larsoner:
`code_ast.body.pop().value` is the last line of code, right? If so then maybe the thing to do is see:

1. if there is an assignment (`ast` can probably tell you this?), then do nothing
2. if there is no assignment, make one to a temporary variable, then interrogate the methods of that temporary variable

No idea if this makes sense without digging into `ast` and `eval`, so feel free to ignore if it's nonsensical

larsoner:
Errant close, sorry

lucyleeow:
I can definitely check if it an assignment. An expression is anything on the RHS of a variable assignment (i.e., `var1 = (anything that you can put here is an expression)`). This means variable assignment, for loop, try/except are all NOT expressions (see [SO](https://stackoverflow.com/questions/2220699/whats-the-difference-between-eval-exec-and-compile/29456463#29456463)).

The problem is that I am evaluating the compiled the ``code_ast`` (code block parsed to ast) so I don't know how to do the assignment. My 'hacky' ideas have involved altering the source code string but I think this is too dangerous. I will can ask around and explore more tomorrow...

larsoner:
> My 'hacky' ideas have involved altering the source code string but I think this is too dangerous. I will can ask around and explore more tomorrow...

This is basically what I was thinking. Use `'_'` or `'__'` (or however many `_` you need to get a var not in the namespace), assign to it, use it, then delete it. Agreed it's a hack but might be worth pursuing just to see that something can work.

sphinx_gallery/tests/test_gen_rst.py
lucyleeow: Sorry, I wasn't quite sure if you meant to parametrize twice, e.g.:

```Python
@pytest.mark.parametrize('config', [(),('__repr__',)])
@pytest.mark.parametrize('code', ['code1','code2'])
```
I tried to do this but couldn't figure out a good way because the output for all options (2x2 in the case above) was different - the only solution I thought of was to write out all combinations. Also I wasn't sure what you meant by triage...

(I will explain the tests I made below)

larsoner: You can do just pairs rather than 2x2 with:
```
pytest.mark.parametrize('config, code', [((), 'code1'), (('__repr__',), 'code2')])
```

lucyleeow:
I think (hope) this solution is an acceptable level of 'hacky'.

I worked out how to assign the last statement, if it is an expression to a variable called `___`, using `ast`. Essentially what is happening is:

```Python
a = 2
a
```
turns into:
```Python
a = 2
___ = a
```
Technically, since only statements that can be on the RHS of an assign are expressions, I think I am safe in assigning any expression to a variable. Doing it with `ast` is also safer than trying to alter the code block string.

`___` is now able to be accessed with `example_globals['___']`. (I realised that I could also use `ast` to get the name of the variable called in the last expression and then access it with `example_globals`, however not all last expressions will be a single variable (e.g., `a+b` or `1+2`)). 

I can then use `___` and the logic you described above to look for the 'reprs' as per the config tuple. A few points:

* the user cannot use the variable name `___` (I think this is reasonable)
* the memory profiler is giving the max memory of the assignment of the expression instead of just the expression (e.g., `___ = a` instead of `a`). I don't know if how much difference there may be with this...?
### html header
The html header I used was just the very simple:

```
.. only:: builder_html

    .. raw:: html

```

I think it is best for users to include in their raw html extra things (e.g., if they want their html in an iframe). Open to suggestions. 

### Tests
I appreciate that I may have gone overboard with the tests. I tried to make it clean (thank you @massich for your help) but please suggest better solutions. Here are the cases I wanted to capture:

* Difference between `()` and `('__repr__',)` with code including a print statement (in body and as last expression), with and without expression as last statement.
    * Side note, is it common knowledge that if you assign a print statement to a variable (e.g., `a = print('hello')`) and then try to get the `__repr__` attribute (`getattr(a, '__repr__')()`) you get `'None'` as a string? What is the reason for this?
* I could not think of any Python objects with a `_repr_html_` (without having to add a dependency) so my code to test this includes defining a class with both `__repr__` and `_repr_html_` and just `__repr__`. I wanted to include cases that test that the config tuple order determines preference and that nothing is returned if the object did not have the methods included in the tuple.
* In cases where there is standard output (e.g. a `print()`) from the body and html output, there would need to be a code block header and html header.

The test is only failing for Python_version 'nightly' - and it is failing for many tests...I have no idea why...

doc/configuration.rst
larsoner: @lucyleeow don't forget to address this

doc/getting_started.rst
larsoner: still unaddressed

examples/local_module.py
```diff
@@ -6,3 +6,4 @@
 """
 
 N = 100
+N
```
larsoner: this example is actually not executed by SG (it does not match the build pattern) so this shouldn't do anything. Can you modify an example that runs, e.g., https://1042-25860190-gh.circle-artifacts.com/0/rtd_html/auto_examples/plot_function_identifier.html you could write `font_size` at the end and we should see it (if you also update our `doc/conf.py` to match, which would be worthwhile)

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -469,6 +469,189 @@ def test_empty_output_box(gallery_conf):
...
+        output_test_string = "\n".join(output.strip().split("\n")[-5:])
+        return output_test_string.strip()
+    
+
+@pytest.mark.parametrize('print_eval_repr, code, expected_out', [
+    pytest.param(
+        tuple(),
+        'a=2\nb=3',
+        '',
+        id='assign,()'
```
larsoner: I would not bother with `id`, then these can all easily fit on a single line like:
```
    [tuple(), 'a=2\nb=3', '']
```
Having the more compact representation will be more readable

lucyleeow: Yes they do all fit on one line, the only problem is that some of the input code, which use variables assigned elsewhere, are really long - and this makes interpreting the pytest output difficult.

examples/local_module.py
lucyleeow: Yes, I can do that. Just to confirm I can change our `doc/conf.py` `print_eval_repr` to `('__repr__',)`?

Also, as you suggested it would be nice to give an example of this in action like a pandas dataframe - but would I have to import this if I were to use it? I could instead define my own class like in the test and use this to produce some html - maybe in a separate gallery as I am not sure it fits into any of the current galleries.  I would also have to change the config for our documentation to `('_repr_html_', '__repr__')` (which still might work fine for the other galleries, I will check).

I will also update the documentation to talk about capturing the other repr's 

Finally, I just realised that we don't `eval()` anything now and maybe we should change the name of this config. Maybe `'capture_repr'` or something similar...?

larsoner: Sure `capture_repr` seems good

Yes you can change `doc/conf.py`. I think it's also fine to add a pandas dependency for the doc build for the purposes of showing the `__html_repr__` (eventually) and/or `__repr__` (in this PR). Basically we want to show a use case that will be common, and maybe that will be the most common one.

sphinx_gallery/tests/test_gen_rst.py
larsoner: Ahh. Well in that case I'd still try to make it as compact as possible by not putting each var in a new line.

examples/local_module.py
lucyleeow: But this PR will now be able to show `_repr_html_` now?

larsoner: In that case I'd change our `doc/conf.py` to have `__repr_html__` and `__repr__` in it, demo both in some example, and also update the docs in this PR to mention `__repr_html__` since it's not in there currently

lucyleeow: Excellent.

Sorry, can you add pandas as dependency for just the doc build or do you have to add it for the whole package?

larsoner: Just for the doc build is fine. Add it here:

https://github.com/sphinx-gallery/sphinx-gallery/blob/master/.circleci/config.yml#L29

And in `intersphinx` here:

https://github.com/sphinx-gallery/sphinx-gallery/blob/master/doc/conf.py#L312

as `'pandas': ('https://pandas.pydata.org/pandas-docs/stable', None),`

larsoner: ... and looking at our existing examples, making a dedicated example for this probably makes sense. You can show that `__html_repr__` and `__repr__` both work, depending on what the object actually exposes. And that assigments do not print.

doc/getting_started.rst
```diff
@@ -90,10 +90,18 @@ this folder must have the following things:
...
+    * The output that is captured while executing the ``.py`` files and
+      subsequently incorporated into the built documentation, can be finely
+      tuned. See :ref:`print_eval_repr`.
     * You can have sub-directories in your ``examples`` directory. These will be
       included as sub-sections of your gallery. They **must** contain their own
       ``README.txt`` or ``README.rst`` file as well.
 
+.. warning::
+
+   The variable name ``___`` (3 underscores) should never be used in your
```
larsoner: In principle eventually we can relax this by actually looking in the `ast` to see if `___` exists and if it does, use `____`, if it exists use `_____`, ..., etc. Eventually there will be enough underscores that one will be unused :)

But this can be another PR if we ever hit this case, unless you're really interested in digging through `ast` to figure it out

doc/getting_started.rst
```diff
@@ -90,10 +90,18 @@ this folder must have the following things:
       presented in a rich literate programming fashion, without any output. To
       change the default pattern for execution and capture see
       :ref:`build_pattern`.
+    * The output that is captured while executing the ``.py`` files and
+      subsequently incorporated into the built documentation, can be finely
+      tuned. See :ref:`print_eval_repr`.
```
larsoner: Wrong link now

GaelVaroquaux:
The example does demo very well the functionality, and shows that you
have solved the problem. Congratulations!

One suggestion: frame the example as a demo of capturing the output, and
discuss the configuration options at the end. This will be more useful
for the end user.

lucyleeow:
(You were too quick for me! But here is my summary anyway, for clarity)
I have:

* changed the config name to `capture_repr`
* indented the html properly
* amended the html header to have line breaks at end, otherwise subsequent text was too close
* updated the configuration and getting started docs
* changed the config to `'capture_repr': ('_repr_html_', '__repr__')
* updated the examples that include matplotlib to avoid text output/explain why `plt.show()` was added
* added exampled to for the `'capture_repr'` config

Unfortunately, the pandas dataframe used in the example does not look nice because the `_repr_html_` refers to external CSS. I am not sure where this CSS can be found and even if I could find it, I am not sure if this would be uniform for other packages. When inspecting a pandas dataframe rendered with jupyter notebook, it seemed that jupyter defines their own CSS (but I am not good with this so not certain). I think for MNE reports, the CSS is included with the html, so the report document is standalone, which would be great. Not sure of the best solution for this...

The test failing is because I added html line breaks. But I've also noticed an indenting problem. Might ponder overnight and do it tomorrow. 

@GaelVaroquaux nice to hear from you! :p

larsoner:
The styling for pandas appears code-wise somewhere here:

https://github.com/pandas-dev/pandas/blob/master/pandas/io/formats/html.py
https://github.com/pandas-dev/pandas/blob/master/pandas/io/formats/style.py

But really it looks like their output already includes it. Looking at the HTML I see:

<details><summary>HTML</summary>

```HTML
<style scoped>
    .dataframe tbody tr th:only-of-type {
        vertical-align: middle;
    }

    .dataframe tbody tr th {
        vertical-align: top;
    }

    .dataframe thead th {
        text-align: right;
    }
</style>
<table border="1" class="dataframe">
  <thead>
    <tr style="text-align: right;">
      <th></th>
      <th>col1</th>
      <th>col2</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>0</td>
      <td>1</td>
      <td>3</td>
    </tr>
    <tr>
      <td>1</td>
      <td>2</td>
      <td>4</td>
    </tr>
  </tbody>
</table>
```

</details>

How did you think it should look? If it looks better in Jupyter, it might be worth playing around a little bit with the code and/or pinging the pandas folks (listserv?) to get a hint on how to get `_repr_html_` to include the necessary CSS styling required to look good. It might be a matter of calling some private function to make it do Jupyter-style outputs. We might end up needing to include some external CSS with the `_repr_html_` calls, which we could do with another config var `capture_repr_css` that points to some file to use to inject more `<style scoped>` when rendering `_repr_html_`. It would be nice if we could get Pandas to do it for us, but if Jupyter is actually what makes it look nice, then we might need to inject their CSS.

(As we do more "it almost works but if only we could ..." fixes, we presumably get closer to replicating the functionality of `nbconvert`, though -- so at some point we might want to reconsider whether or not to use it.)

lucyleeow:
The dataframe looks slightly better in jupyter notebook:

image

I thought that `class="dataframe"` might refer to a CSS class defined elsewhere? Also `<style scoped>` seems to be a style, defined somewhere else?

If we do decide to just use nbconvert I might be slightly upset - (┛ಠ_ಠ)┛彡┻━┻  :smile: But a good proof of concept is right here: #422 

I think CSS addition should be another PR...? But I am happy to ask about pandas.

larsoner:
If, within that Jupyter Notebook, you use Chrome to inspect the page element, you can see where e.g. the bold of the leftmost column comes from (which CSS class and which CSS style sheet)

doc/getting_started.rst
lucyleeow: At the moment...no. But I can look into it if someone complains :)

lucyleeow:
@GaelVaroquaux I have amended the example:https://1052-25860190-gh.circle-artifacts.com/0/rtd_html/auto_examples/plot_capture_repr.html
Is this what you meant?

I think the only fail is due to #542 - should I just leave this or fetch to this branch...?

@larsoner I am not sure what listserv is...? But I create an issue about CSS.

larsoner:
> @larsoner I am not sure what listserv is...? But I create an issue about CSS.

https://mail.python.org/mailman/listinfo/pandas-dev

lucyleeow:
Would be acceptable/annoying to message all the pandas devs.....(?) :/

larsoner:
> I think the only fail is due to #542 - should I just leave this or fetch to this branch...?

It's probably worth rebasing to make sure that the new changes you've made don't make new failures.

larsoner:
> Would be acceptable/annoying to message all the pandas devs.....(?) :/

It seems like a relevant enough thing to post, but I'm not on the listserv so don't really know. Maybe check the archives to see what the posts are like, then use your best judgment about that vs a GitHub issue

larsoner:
(and check to see what the Pandas docs say about which communication channels to use)

lucyleeow:
I hope I rebased properly. Not sure why it's still failing for Python nightly build...

larsoner:
Needed the same fix that was in #542 (tested 3.8-RC1 locally to find it), pushed, hopefully CIs will be happy now

lucyleeow:
@larsoner thanks but I am a little confused - I understand c36ed7f and #542 was due to https://bugs.python.org/issue35894 but are they both doing the same thing but in a different way? And why was this required, since #542 seemed to fix it. Is this because nightly changed again...?

And are we just waiting for @GaelVaroquaux to approve the example amendment?

larsoner:
> Is this because nightly changed again...?

No, it's because the bug before was doing (effectively) a call like `Module(body)` is no longer allowed, you have to do `Module(body, ignore_types=[])`. #542 fixed this in the existing code, but then this PR added another `Module(body)` call.

> are they both doing the same thing but in a different way?

Yes. When we had one effective call to `Module` it made sense to just do the call in the conditional (i.e., what is in `master`). Once we had two calls to `Module` to follow DRY it made sense to make a callable that we could use multiple times (now in this PR).

> And are we just waiting for @GaelVaroquaux to approve the example amendment?

The title says WIP so I wasn't sure if it's ready. In general once it's ready from your end it's good to both change the title and ping folks for final reviews, otherwise it's unclear if you're still working on something.

lucyleeow:
@larsoner thank you - that makes much more sense (especially after I realised the new code was part of `execute_code_block`)

ping @GaelVaroquaux

GaelVaroquaux:
I'm having a look at this right now.

For reference: link to the rendered docs: https://1054-25860190-gh.circle-artifacts.com/0/rtd_html/index.html

doc/configuration.rst
```diff
@@ -950,4 +951,86 @@ you can do::
...
+.. _capture_repr:
+
+Controlling what output is captured
+===================================
+
+The ``capture_repr`` configuration allows the user to control what output
+is captured, while executing the example ``.py`` files, and subsequently
+incorporated into the built documentation. By default ``capture_repr`` is an
+empty tuple. With this setting, only data directed to standard output is
+captured.
```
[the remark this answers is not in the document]
lucyleeow: Yes - unfortunately this is for backwards compatibility due to Matplotlib. Like in IPython, if we add `__repr__` to the config tuple, matplotlib calls would output annoying text (e.g. plt.plot() will return something like `[<matplotlib.lines.Line2D at 0x7fcd58a89a90>]` as well as the figure)

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -474,6 +474,121 @@ def test_empty_output_box(gallery_conf):
...
+    'class repr_only_class():\n'
+    '    def __init__(self):\n'
+    '        pass\n\n'
+    '    def __repr__(self):\n'
+    '        return "This is the __repr__"\n\n'
+    'class_inst = repr_only_class()\n\n'
+    'class_inst'
+)
+
+code_repr_and_html = (
```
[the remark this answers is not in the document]
lucyleeow: I tried to do this but it wouldn't work! I think it may have been an indentation issue (from some SO questions) but I am open to suggestions, esp if you have experience with this kind of thing before.

doc/conf.py
```diff
@@ -364,6 +365,7 @@ def setup(app):
                },
     'show_memory': True,
     'junit': os.path.join('sphinx-gallery', 'junit-results.xml'),
+    'capture_repr': ('_repr_html_', '__repr__')
```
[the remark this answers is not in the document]
lucyleeow: I can do this

GaelVaroquaux:
> Yes - unfortunately this is for backwards compatibility due to Matplotlib. Like
> in IPython, if we add __repr__ to the config tuple, matplotlib calls would
> output annoying text (e.g. plt.plot() will return something like
> [<matplotlib.lines.Line2D at 0x7fcd58a89a90>] as well as the figure)

Right. I still think that we should change the default, and maybe add a
clear note in the changelog about the changed default, and how to: 1)
change it back, 2) work around the problem (eg with plot.show(), or
assigning to a variable.

GaelVaroquaux:
> I tried to do this but it wouldn't work! I think it may have been an indentation issue

Did you try:

```
code_repr_only = '''
class repr_only_class():
    def __init__(self):
        pass

    def __repr__(self):
        return "This is the __repr__"

class_inst = repr_only_class()

class_inst
'''
```

ie without leading indentation?

larsoner:
I can live with changing the default

lucyleeow:
> Right. I still think that we should change the default, and maybe add a clear note in the changelog about the changed default, and how to: 1) change it back, 2) work around the problem (eg with plot.show(), or assigning to a variable.

I understand - this bothered me all weekend when I realised the Matplotlib problem. I see both sides of the argument but I don't think it's fair on existing projects that already use SG to change the behaviour. I appreciate that for new projects, the goal would be for them to use `'capture_repr': ('__repr__')` - maybe I can note this more strongly in 'getting started'?

Slightly related, @larsoner and I discussed the possibility of the `'capture_repr'` configuration being able to be altered on a file-by-file basis/code-block basis (like the [line number config](https://sphinx-gallery.github.io/configuration.html#adding-line-numbers
)) to make it easier for existing projects to change to the new configuration - this would mean that they don't have to go through all their examples and change the matplotlib calls.

> Did you try:

I'm pretty sure I did try that but I will check it again.

lucyleeow:
> I'm pretty sure I did try that but I will check it again.

Nevermind it does work - I think I tested it in a rush before going somewhere - I shall blame this.

larsoner:
> I misread what you said @larsoner - for MNE for example, would you change the config back (to ()) and wait for for finer control to be implemented, like above, or do you would go through all the examples and amend matplotlib calls?

I would just change ours to `capture_repr=('_repr_html_',)` and use this instead of our custom `Report` scraper. I don't think we want `('__repr__',)` because we have a lot of (tens? hundreds?) of `matplotlib` and other calls at the end of blocks that return things that we do not escape. It wouldn't be the end of the world to have this in the output, but I don't see much advantage to adding it for us yet.

lucyleeow:
@GaelVaroquaux would your suggestion be to change the default to `('_repr_html_', '__repr__')`? Happy to do this and change the documentation.

GaelVaroquaux:
> @GaelVaroquaux would your suggestion be to change the default to ('_repr_html_', '__repr__')? Happy to do this and change the documentation.

Yes it would be. And make a clear note / warning in the release notes.

Maybe add an FAQ item addressing "I don't like the matplotlib outputs showing" (or something better worded)

The reasoning behind this, is that it think that this is the best default for new projects.

lucyleeow:
@GaelVaroquaux I have changed the default and amended the documentation as suggested.

I don't know how release notes work though...? Should I do something?

larsoner:
> I don't know how release notes work though...? Should I do something?

No we create these at release time

doc/getting_started.rst
lucyleeow: I've thought about this and I don't think we could do it with `ast`. This converts the code in a code block into AST. It doesn't tell you what global variables exist - this would be found in `example_globals`. `example_globals` would cumulatively include all variables assigned in a `.py` file. This means that for the first code block we can safely check to see if `___` exists in `example_globals` but for subsequent code blocks in the same `.py` file, a variable `___` may exist simply because *we* assigned it previously.

For now, I can't think of a solution.

doc/conf.py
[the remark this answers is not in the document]
larsoner: @lucyleeow looks like this still needs a comment

doc/getting_started.rst
larsoner: That's fine, we can deal with it later if need be but YAGNI for now

doc/conf.py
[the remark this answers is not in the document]
lucyleeow: Hmm I did this but not sure why it's not showing up here. It is showing up in 
e72e260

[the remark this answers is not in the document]
lucyleeow: whoops, this is an old commit.

larsoner:
@lucyleeow while we wait for @GaelVaroquaux to have one last look and merge, can you rebase so that AppVeyor will build properly?

GaelVaroquaux:
Don't wait for another review from me. This is great!!

larsoner:
Thanks @lucyleeow !

GaelVaroquaux:
Woot! Congratulations, Lucy, this is a major feature!!

I think that we are going to have to think about a release soon...

banesullivan:
A release coming anytime soon so we can use this?

lucyleeow:
@banesullivan yes it's a good point, documentation is ahead of the current release - I've been meaning to create an issue about this.

If you do wish to use this feature you can [install as developer](https://github.com/sphinx-gallery/sphinx-gallery#install-as-a-developer). Sorry for the confusion.

larsoner:
I think it's fine to release now. @lucyleeow do you want to try following our instructions for how to make a release?

https://sphinx-gallery.github.io/maintainers.html#how-to-make-a-release

lucyleeow:
I should be able to get onto it this week but from reading the instructions, I may have some questions. What's the best way to ask? Make a new issue?

Should I do the items that don't require push access to SG or?

larsoner:
@lucyleeow send me your PyPi address and I'll add you there. Working on giving you write permissions here but I don't have admin access yet (I think?). Going to bother @agramfort about it

lucyleeow:
:open_mouth: that is a lot of trust.

I just made a PyPi account: https://pypi.org/user/lucyleeow/

larsoner:
Done. But let me get #559 taken care of before release

lucyleeow:
If you are taking care of #559 I would like to do #143 before the release (I've already started).

PR #548, opened 2019-10-14
[MRG] Fix terminal rst block last word 

### LINKED ISSUE #546: Last word in rst used as code
If an example `.py` file ends with an rst block, Sphinx-Gallery thinks the last word of the rst block is code and tries to execute it, unless there is an empty line after the rst block.

For example:

```Python
"""
Test
####

test
"""

print('hello')

#%%
# This is a sentence.
```

Gives: 
Screenshot_2019-10-09 Test — mwe 0 documentation

If there is an empty line at the end, this does not happen. Is it expected to just add a new line to the end of the document?

MWE [here](https://github.com/lucyleeow/Sphinx-Gallery-MWE).

lucyleeow:
closes #546 

Includes test.

### DIFF
--- a/sphinx_gallery/tests/test_gen_rst.py
+++ b/sphinx_gallery/tests/test_gen_rst.py
@@ -103,17 +103,16 @@ def test_direct_comment_after_docstring():
     assert result == expected_result
 
 
-def test_final_rst_last_word():
+def test_final_rst_last_word(tmpdir):
     """Tests last word in final rst block included as text"""
-    with tempfile.NamedTemporaryFile('w', delete=False) as f:
+    filename = str(tmpdir.join('temp.py'))
+    with open(filename, 'w') as f:
         f.write('\n'.join(['"Docstring"',
                            '# comment only code block',
                            '#%%',
                            '# Include this whole sentence.']))
-    try:
-        file_conf, result = sg.split_code_and_text_blocks(f.name)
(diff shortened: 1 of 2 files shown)


sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -103,6 +103,26 @@ def test_direct_comment_after_docstring():
...
+    """Tests last word in final rst block included as text"""
+    with tempfile.NamedTemporaryFile('w', delete=False) as f:
+        f.write('\n'.join(['"Docstring"',
+                           '# comment only code block',
+                           '#%%',
+                           '# Include this whole sentence.']))
+    try:
+        file_conf, result = sg.split_code_and_text_blocks(f.name)
+    finally:
+        os.remove(f.name)
```
larsoner: The modern way would be:
```
def test_final_rst_last_word(tmpdir):
    fname = str(tmpdir.join('whatever.py'))
    with open(fname, 'w') as fid:
        ...
    ...
```
`tmpdir` is a pytest fixture that will auto-delete when the run is over, makes all this boilerplate `tempfile` stuff unnecessary

lucyleeow: Thanks! Done.

larsoner:
Thanks @lucyleeow

PR #551, opened 2019-10-18
BUG: Fix index check

larsoner:
Should fix https://circleci.com/gh/mne-tools/mne-python/16316

@lucyleeow can you look and see if you agree with the change? Basically there is no guarantee that `code_ast.body` is non-empty (it can be empty if a code block contains only comments for example), so we need to protect against it


lucyleeow:
LGTM!

larsoner:
Okay, in it goes

PR #556, opened 2019-10-24
Fix JPEG thumbnail generation

rgov:
Since Pillow 4.2, an exception is thrown when trying to save an image with an alpha channel to JPEG (because it cannot store this channel).

This simply checks for an exception and tries again without the alpha channel. Unfortunately Pillow just uses IOError instead of a specific exception type, and I think that checking the exception message is fragile. If the IOError has another cause, such as the disk being full, then the second save will probably fail the same way.


larsoner:
See also #557

rgov:
It is hit when building pytorch/tutorials with Pillow 4.2 or newer.

larsoner:
Ahh they use a bunch of our functions in their own way:

https://github.com/pytorch/tutorials/blob/master/custom_directives.py

Not exactly something we aim to support but with this and #557 it's reasonable enough

PR #557, opened 2019-10-24
ENH: Add official JPG support and better tests

larsoner:
The goal was just to add a test for #556, but I realized that it (probably) needed `jpg` as a supported format. It was easy enough to add, so I did it, and it should fail with the error reported in #556.

I propose we merge #556, I rebase this, then we merge this.


larsoner:
Rebased and ready for review/merge. Maybe @lucyleeow you could look?

lucyleeow:
Sure! I will do it today or tomorrow.

larsoner:
Thanks for the review @lucyleeow

PR #563, opened 2019-11-13
[MRG] DOC: Explain each example

### LINKED ISSUE #143: DOC: each example should start by explaining why it's there

lucyleeow:
closes #143 

- An explanation is added for each example
- Update the note at the beginning of each example, saying it is useful to compare to the source code with the output below (let me know if you think it is too long now)

Looking at general examples, it seems that several examples demonstrate the same thing. Specifically, the following examples simply show a plot output and some math in the text:

- [Colormaps alter your perception](https://sphinx-gallery.github.io/auto_examples/plot_colors.html)
- [Some Quantum Mechanics, filling an atomic orbital](https://sphinx-gallery.github.io/auto_examples/plot_quantum.html)
- [Plotting the exponential function](https://sphinx-gallery.github.io/auto_examples/plot_exp.html)
- [Sphinx-Gallery introduction](https://sphinx-gallery.github.io/auto_examples/plot_gallery_version.html)
The Sin function sub-gallery:
- [Plotting simple sin function](https://sphinx-gallery.github.io/auto_examples/sin_func/plot_sin.html)
- [Plotting simple sin function on a black background](https://sphinx-gallery.github.io/auto_examples/sin_func/plot_sin_black_background.html)
(this example includes a reference to another example, but this addition can be easily combined as well)

I think it would be a good idea to consolidate these into just two examples (it is interesting to demonstrate that if there is only one code block in an example, the output appears before the code block). It helps users find useful information if there is less redundant examples that all show the same thing. One thing it would affect is the number of examples in the [mini-galleries](https://sphinx-gallery.github.io/configuration.html#examples-using-numpy-exp) part configuration documentation.

Further I propose we order the examples - general examples first and put local module and Plotting the exponential function next to each other.


larsoner:
@drammock this would be a good one to review

larsoner:
> Looking at general examples, it seems that several examples demonstrate the same thing. Specifically, the following examples simply show a plot output and some math in the text:

Feel free to restructure as you see fit @lucyleeow. I don't think anyone has made a unifying/cleaning up pass in a long time so it would be great if you could do it

lucyleeow:
I accidentally committed all the changes in one commit.

- Update examples to explain what they show
- Consolidate examples so redundant examples are removed and any useful information moved to another example. Removed:
   - `plot_colors`, `plot_gallery_version`, `plot_quantum` and `plot_notebook` from the secondary gallery
   - the [sin function ](https://sphinx-gallery.github.io/stable/auto_examples/index.html#the-sin-function )sub-gallery 
   - the [seaborn notebooks](https://sphinx-gallery.github.io/stable/tutorials/index.html#seaborn-notebooks) sub-gallery
- update references in the documentation to the `plot_gallery_version` example, they now point to the introductory example
- order examples by filename
- updated `plot_parse.py` - thus needed to update 2 tests

drammock:
@lucyleeow is this ready to be reviewed?

lucyleeow:
@drammock if travis passes! Otherwise I will fix the problem first! Thanks!

examples/plot_0_sin.py
```diff
@@ -0,0 +1,81 @@
...
+To include displayed math notation, use the directive ``.. math::`` and to
+include inline math notation use ``:math:``. For example, we wish to plot the
+following function:
+
+.. math::
+
+    x \\rightarrow \\sin(x)
+
+Here the function :math:`\\sin` is evaluated at each point the variable
+:math:`x` is defined. When writting latex in a Python string, ensure that you
```
drammock:
```suggestion
:math:`x` is defined. When including LaTeX in a Python string, ensure that you
```

examples/plot_0_sin.py
```diff
@@ -0,0 +1,81 @@
...
+escape the backslashes or use a raw docstring. You do not need to do
```
drammock:
```suggestion
escape the backslashes or use a :py:std:label:`raw docstring <strings>`.
You do not need to do
```

examples/plot_0_sin.py
```diff
@@ -0,0 +1,81 @@
...
+import matplotlib.pyplot as plt
+
+x = np.linspace(0, 2 * np.pi, 100)
+y = np.sin(x)
+
+plt.plot(x, y)
+plt.xlabel('$x$')
+plt.ylabel('$\sin(x)$')
+# To avoid matplotlib text output
+plt.show()
```
drammock: This shouldn't be necessary; you could delete these two lines if the prior line was an assignment (`_ = plt.ylabel('$\sin(x)$')`).

examples/plot_0_sin.py
```diff
@@ -0,0 +1,81 @@
...
+#    \sin
+#
+# Cross referencing
+# ^^^^^^^^^^^^^^^^^
+#
+# You can refer to an example from any part of the documentation,
+# including from other examples. Sphinx-Gallery automatically creates reference
+# labels for each example. The label consists of the ``.py`` file name,
+# prefixed with ``sphx_glr_`` and the name of the
+# folder(s) the example is in. In this case, the example is in
```
drammock:
```suggestion
# folder(s) the example is in. In this case, the example we want to
# cross-reference is in
```

examples/plot_0_sin.py
```diff
@@ -0,0 +1,81 @@
...
+# ``auto_examples``, the ``gallery_dirs`` (see
```
drammock:
```suggestion
# ``auto_examples`` (the ``gallery_dirs`` variable; see
```

examples/plot_0_sin.py
```diff
@@ -0,0 +1,81 @@
...
+# :ref:`configure_and_use_sphinx_gallery`), then the subdirectory ``no_output``
+# (since the example is within a sub-gallery). The file name of the example is
+# ``plot_syntaxerror.py``. We can thus cross-link to the example 'SyntaxError'
+# using:
+# ``:ref:`sphx_glr_auto_examples_no_output_plot_syntaxerror.py\```.
```
drammock:
```suggestion
# ``:ref:`sphx_glr_auto_examples_no_output_plot_syntaxerror.py```.
```

examples/plot_1_exp.py
```diff
@@ -3,9 +3,11 @@
 Plotting the exponential function
 =================================
 
-A simple example for ploting two figures of a exponential
-function in order to test the autonomy of the gallery
-stacking multiple images.
+This example demonstrates how to import a local module and how images are
+stacked when two plots are created in one code block. ``N`` from the example
```
drammock:
```suggestion
stacked when two plots are created in one code block. The variable ``N`` from
the example
```

examples/plot_5_unicode_everywhere.py
```diff
@@ -3,8 +3,9 @@
 Using Unicode everywhere 🤗
 ===========================
 
-This examples includes unicode characters, mostly emoji 🎉 to stress test the
-build and test environments that parse the example files. """
+This examples demonstrates how to include unicode characters, mostly emoji 🎉
```
drammock:
```suggestion
This example demonstrates how to include non-ASCII characters, mostly emoji 🎉
```

doc/conf.py
```diff
@@ -345,7 +345,8 @@ def setup(app):
     'subsection_order': ExplicitOrder(['../examples/sin_func',
                                        '../examples/no_output',
                                        '../tutorials/seaborn']),
```
drammock: two of these folders don't exist anymore right?

examples/plot_3_capture_repr.py
```diff
@@ -5,10 +5,10 @@
...
-documentation. The output that is captured with this setting can be observed in
-this example. Differences in outputs that would be captured with other
-``capture_repr`` settings is also explained.
+``capture_repr: ('_repr_html_', '__repr__')`` and was used when building the
+Sphinx-Gallery documentation. The output that is captured with this setting
+is demonstrated in this example. Differences in outputs that would be captured
+with other ``capture_repr`` settings is also explained.
 """
 #%%
 # Nothing is captured for the code block below because no data is directed to
```
drammock: I would end this sentence by saying "the last statement is an assignment, not an expression."

examples/plot_6_function_identifier.py
```diff
@@ -49,7 +49,7 @@
              size=fontsize,
              transform=fig.transFigure,
              bbox=dict(boxstyle='larrow', fc="w", ec="k"))
-#
+
 plt.draw()
 # To avoid matplotlib text output
 plt.show()
```
drammock: again, this shouldn't be needed?  Also the rendered output shows all the text boxes overlapping, might as well reduce the `fontsize` variable while we're overhauling things.

examples/plot_0_sin.py
```diff
@@ -0,0 +1,81 @@
...
+
+Example files should start with a header docstring. Anything before the
+docstring is ignored by Sphinx-Gallery and will not appear in the rendered
+output, nor will it be executed. This docstring requires an rST header, used
+at the title of the example and to correctly build reference links.
+
+Once you close the docstring you will be writing Python code. This code gets
+executed by Sphinx-Gallery and any output, including plots will be captured.
+You can also alternate between code blocks and text blocks that provide
+explanations about the code.
```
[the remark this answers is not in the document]
lucyleeow: I think I will use the term text and code blocks because we've used these terms in other parts of the documentation. And I think comment is a bit confusing because you can still have comments in code blocks. But I have amended the text - thanks.

examples/plot_0_sin.py
lucyleeow:
```
:py:std:label:`raw docstring <strings>`
```
this doesn't render as anything. What is it meant to be..?

I can simply add an example i.e., `r'<string>'`

examples/plot_0_sin.py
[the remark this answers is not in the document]
lucyleeow: I also specified that default behaviour is that anything following the docstring is taken as Python code, unless you add a splitter.

examples/plot_0_sin.py
lucyleeow: True, but it would be good to still make a comment on why I am assigning it (so it would only be deleting 1 line) - and then I should probably be consistent and change the other examples. Any reason you are opposed to `plt.show()`?

doc/conf.py
lucyleeow: Good point

examples/plot_6_function_identifier.py
lucyleeow: Good point - reduced font size and padding of the arrows.

examples/plot_0_sin.py
lucyleeow: Now I know why I can't add ``r"""<docstring>"""`` - SG no longer picks up the docstring if I put that in the docstring. I tried a hacky substitution (e.g. `.. |docstring| replace:: """`) but it didn't work.
Suggestions?

larsoner: Probably needs to be explicit about where the intersphinx ref is (`python`) and/or remove the `:py:` label (it's just under `:std:label:` in intersphinx), so:
```
:std:label:`raw docstring <python:strings>`
```
Verified that it's likely to work by doing:
```
python -m sphinx.ext.intersphinx https://docs.python.org/3/objects.inv > ~/Desktop/python.txt
```
Then looking to see that it's under the `std:label` category.

I would make this change locally then `make html` to see if it works.

lucyleeow:
```
:std:label:`raw docstring <python:strings>`
```
didn't work. Am I doing something wrong?

lucyleeow: I don't need to update `intersphinx_mapping` in the `conf.py` file right?

lucyleeow: `:python:std:label:`raw docstring <python:strings>` also didn't work.

larsoner: Shouldn't need to as there is already an entry for `python` there. I'll try a few things locally in a little while

lucyleeow: @larsoner 
```
:ref:`raw docstring <python:strings>`
```
works and links to [this page](https://docs.python.org/3/reference/lexical_analysis.html#strings), which I think is correct (unfamiliar with intersphinx and versions)

larsoner: That looks correct

lucyleeow:
Note I pretty printed `reference_parse.txt` for legibility but I suspect this is not PEP8 compliant. I can change back to the old format.

larsoner:
@lucyleeow please set title to MRG if this is ready for review/merge from your end

larsoner:
(and ping us for another round of review as GH does not notify for title changes)

lucyleeow:
Thanks - just 2 unresolved things. One is should I change formatting of `reference_parse`?

examples/plot_0_sin.py
lucyleeow: ping @drammock 
I will amend `plot_capture_repr` to show the text output and the two ways of avoiding it. I do think I should otherwise be consistent in all the examples though.

larsoner:
> One is should I change formatting of reference_parse?

If all you did is make it more readable (which I think is the case?) then it should be fine

lucyleeow:
> If all you did is make it more readable (which I think is the case?) then it should be fine

I just used ``pprint.pprint``. Great, I will leave it as is.

lucyleeow:
ping @drammock @larsoner good to go

@drammock, you suggested using assignment to avoid matplotlib output, instead of `plt.show()` but since the additions to `plot_capture_repr` I think we can keep `plt.show()` since this is used in all the examples.

> This shouldn't be necessary; you could delete these two lines if the prior line was an assignment (`_ = plt.ylabel('$\sin(x)$')`).

examples/plot_0_sin.py
drammock: Good point about needing to explain it either way. And there is value in being consistent.  I guess my feeling is that if we consistently do `plt.show()` then people will assume it's required (which it's not). But since it's currently consistent, and you added a good explanation in the `capture_repr` file, let's call it good enough.  If it still bothers me in a week I'll do a new PR to address it.

lucyleeow:
@drammock good point. I can see how using `plt.show` can mislead people into thinking that it is necessary for the plot to show. Feel free to amend it before I make the new release, which I will work on now (since we've separated dev and stable docs now).

lucyleeow:
Minor amendments to the `plot_capture_repr` example. All green so merging.

larsoner:
Time to push out a release them I think. Want to try @lucyleeow ?

lucyleeow:
@larsoner working on it now!

PR #566, opened 2019-11-18
[MRG] Release v0.5.0

lucyleeow:
- Updated changelog
- Updated maintainers documentation (there is no `between-tags` parameter for `github_changelog_generator` anymore)
- Added tl;dr note to `configuration.rst` advising how to revert capturing behaviour

Wasn't sure about:

> Update the version in `sphinx_gallery/__init__.py`. It should end in dev0. You should remove this text, and the remaining number will become the version for this release.

I removed the 'dev0'. If this is correct, I will amend this note to:

> Update the version in `sphinx_gallery/__init__.py`, which should end in dev0. You should remove 'dev0', and the remaining number will become the version for this release.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 Change Log
 ==========
 
-v0.5.0.stable.0
+v0.5.0
 ---------------
 
 Developer changes
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,23 +1,91 @@
 Change Log
 ==========
 
-v0.5.0
-------
+v0.5.0.stable.0
+---------------
 
 Developer changes
 '''''''''''''''''
 
-- N/A
+- Separated 'dev' documentation, which tracks master and 'stable' documentation,
+  which follows releases.
+- Added official jpeg support.
 
 Incompatible changes
 ''''''''''''''''''''
 
 - Dropped support for Sphinx < 1.8.3.
 - Dropped support for Python < 3.5.
+- Added ``capture_repr`` configuration with the default setting 
+  ``('_repr_html_', __repr__')``. This may result the capturing of unwanted output
+  in existing projects. Set ``capture_repr: ()`` to return to behaviour prior
+  to this release.
 
 **Implemented enhancements:**
 
--  ENH: Support ``#%%`` cell separator `#370 <https://github.com/sphinx-gallery/sphinx-gallery/pull/518>`__ (`lucyleeow <https://github.com/lucyleeow>`_)
+-  Explain the inputs of the image scrapers `#472 <https://github.com/sphinx-gallery/sphinx-gallery/issues/472>`__
+-  Capture HTML output as in Jupyter `#396 <https://github.com/sphinx-gallery/sphinx-gallery/issues/396>`__
+-  Feature request: Add an option for different cell separations `#370 <https://github.com/sphinx-gallery/sphinx-gallery/issues/370>`__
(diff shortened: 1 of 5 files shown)


CHANGES.rst
```diff
@@ -1,23 +1,91 @@
 Change Log
 ==========
 
-v0.5.0
-------
+v0.5.0.stable.0
```
larsoner: We don't do this `.stable` business and it's not in the instructions, why change?

sphinx_gallery/__init__.py
```diff
@@ -6,7 +6,7 @@
 import os
 # dev versions should have "dev" in them, stable should not.
 # doc/conf.py makes use of this to set the version drop-down.
-__version__ = '0.5.0.dev0'
+__version__ = '0.5.0.'
```
larsoner: no trailing `.`

CHANGES.rst
lucyleeow: I don't quite understand how releases are tagged and how the distinction between stable and dev is made...
Changed.

larsoner: Stable and releases are synonymous, and are just `X.Y.Z`. Dev is `X.Y.Z.dev0`

lucyleeow: Is tagging done via `__version__ ` in the `__init__.py` file? And I assume stable is differentiated from dev in the CircleCI config file via the line:

```
            tags:
              only: /^Release.*$/
```

But I don't see what this regex is looking for...?

larsoner: > Is tagging done via __version__ in the __init__.py file?

No, it's done via `git tag`, but this is done automatically when you do tell GitHub to make a release. 

> I don't see what this regex is looking for...?

Good call -- it was meant to look for the title standard we've been following when doing a GitHub release:

https://github.com/sphinx-gallery/sphinx-gallery/releases

But it looks like the tags actually land as `vX.X.X` so the regex is incorrect. Can you fix it here? I think `^v\d+\.\d+\.\d+$` might do it?

lucyleeow: > But it looks like the tags actually land as `vX.X.X` so the regex is incorrect. Can you fix it here? I think `^v\d+\.\d+\.\d+$` might do it?

That looks correct to me. Thanks, it all makes sense now.

larsoner: Excellent, I'll merge once you push a commit to fix this (or I can do it) and CIs are happy

lucyleeow: Last question, what do the backslashed (e.g. `/^Release.*$/`) around the regex do?

larsoner: Not sure, but they are always in the CircleCI config

https://circleci.com/docs/2.0/workflows/#executing-workflows-for-a-git-tag

lucyleeow: Thanks!

larsoner:
Thanks @lucyleeow

larsoner:
FYI had to redo the release after deleting the tag, pushing a couple of commits to master to fix the CircleCI conditional, then re-doing the release on GH. Now all is well with `stable`:

https://sphinx-gallery.github.io/stable/

I'll go ahead and bump `master` to 0.6.0.dev0

larsoner:
And dev docs are up to date showing 0.6.0.dev0

https://sphinx-gallery.github.io/dev/

GaelVaroquaux:
Very nice! Thank you all!

Will someone do a nice tweet, that I will retweet?

agramfort:
beautiful !

thx @lucyleeow and @larsoner !

>

lucyleeow:
@GaelVaroquaux done! I note @larsoner doesn't have Twitter...

PR #569, opened 2019-11-20
URLError

### LINKED ISSUE #568: AttributeError: 'URLError' object has no attribute 'url'
Hi,

I updated to the latest 0.5.0 and I've an error message when building the documentation : 
```
File "/home/etienne/anaconda3/lib/python3.7/site-packages/sphinx_gallery/docs_resolv.py", line 260, in _handle_http_url_error
    msg, e.url, extra, e.msg)
AttributeError: 'URLError' object has no attribute 'url'
```

Error is occuring inside the function [_handle_http_url_error](https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/docs_resolv.py#L260). I also tried to comment the logger.warning line and the doc is correctly build again

EtienneCmb:
The function [_handle_http_url_error](https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/docs_resolv.py#L257) seems to be used for displaying an error message when there's url related issues. However, the function can either take URLError or HTTPError and the URLError doesn't have attributes `url` and `msg`.

closes #568

### DIFF
--- a/sphinx_gallery/docs_resolv.py
+++ b/sphinx_gallery/docs_resolv.py
@@ -255,9 +255,11 @@ class SphinxDocLinkResolver(object):
 
 
 def _handle_http_url_error(e, msg='fetching'):
-    extra = e.code if isinstance(e, HTTPError) else e.reason
-    logger.warning('The following HTTP Error has occurred %s %s: %s (%s)',
-                   msg, e.url, extra, e.msg)
+    if isinstance(e, URLError):
+        error_msg = '%s: %s' % (msg, e.reason)
+    elif isinstance(e, HTTPError):
+        error_msg = '%s %s: %s (%s)' % (msg, e.url, e.code, e.msg)
+    logger.warning('The following HTTP Error has occurred %s' % error_msg)
 
 
 def _embed_code_links(app, gallery_conf, gallery_dir):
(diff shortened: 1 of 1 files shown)


larsoner:
You need to update the `test_run_sphinx` test check, see Travis

https://travis-ci.org/sphinx-gallery/sphinx-gallery/jobs/614767797#L1301

EtienneCmb:
@larsoner yes I saw it yesterday. I think the error is coming from somewhere else. As far as I understand, the regular expression that is tested should target the bad url defined for the [scipy package](https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/tests/tinybuild/conf.py#L41). However, the warning doesn't contains this url. The `url` attribute is None, that's why it's failing.

So I think the true error is why the url is not present in the HTTPError. Do you've an idea why?

EtienneCmb:
Ok, I found the issue. `isinstance(HTTPError, URLError)` returns `True`

lucyleeow:
I had a look at this locally. It is the last assertion in the test that is failing `assert re.match(want, warning, re.DOTALL) is not None, warning`. The variable `warning` did not contain '*fetching' or '*wrong_url' as per the regex. 
I checked `warning` using the master branch code and it did contain those words. I guess it is due to the new `error_msg`.

I have no idea why there are warning about not finding `local_module` though.

lucyleeow:
I think `HTTPError` inherits from `URLError` - good pick up.

larsoner:
Thanks @EtienneCmb

PR #573, opened 2019-12-02
Remove the note link to downloads from pdf output

### LINKED ISSUE #572: Remove the note linking to the download section at the beginning of the example from latex/pdf output
For latex/pdf output, this note at the beginning of the example linking to the download section does not link to anything (since there are no download buttons). Would it be possible to disable it in that case? The straightforward way would be wrap it into a `.. only:: html` directive.

benvial:
- wrap the note with a `.. only:: html` directive
- add a test

closes #572

### DIFF
--- a/sphinx_gallery/tests/test_gen_rst.py
+++ b/sphinx_gallery/tests/test_gen_rst.py
@@ -417,6 +417,16 @@ def test_remove_config_comments(gallery_conf):
     assert '# sphinx_gallery_thumbnail_number = 1' not in rst
 
 
+def test_download_link_note_only_html(gallery_conf):
+    """Test if the download_link."""
+    rst = _generate_rst(gallery_conf, 'test.py', CONTENT)
+    download_link_note = (".. only:: html\n\n"
+                          "    .. note::\n"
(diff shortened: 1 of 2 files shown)


benvial:
Great. I couldn't find the PDF in the build  artifacts. Just to be sure the note is actually removed

Also the test docstring could be a more descriptive.

larsoner:
Looks okay:

https://1157-25860190-gh.circle-artifacts.com/0/latex/python.pdf

I wouldn't worry about making the description better, seems fine to me

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -417,6 +417,16 @@ def test_remove_config_comments(gallery_conf):
     assert '# sphinx_gallery_thumbnail_number = 1' not in rst
 
 
+def test_download_link_note_only_html(gallery_conf):
+    """Test if the download_link."""
```
lucyleeow:
```suggestion
    """Test html only directive for download_link."""
```

lucyleeow:
Thanks @benvial! Merging.

benvial:
Brilliant. Thanks @lucyleeow and @larsoner

PR #574, opened 2019-12-03
[MRG] Add ignore pattern to check dups

### LINKED ISSUE #474: check_duplicated does not respect the ignore_pattern
It seems that check_duplicate_filenames does not respect the ignore_pattern.

The proper fix would probably be to have collect_gallery_files respect the ignore_pattern.

lucyleeow:
closes #474


larsoner:
AppVeyor failure is unrelated and this looks straightforward so in it goes, thanks @lucyleeow

PR #577, opened 2019-12-09
Add ability to ignore repr of specific types

banesullivan:
I'm loving the new repr stuff to output the repr of objects at the end of each code block, but we are getting the repr output in cases where we don't actually want it in the PyVista gallery. These changes allow users to specify classes/types that should not be repr'd.

For example, in the PyVista gallery, we get outputs containing the camera position because that is what `Plotter.show()` returns:

Screen Shot 2019-12-08 at 9 18 32 PM

This happens with matplotlib also which is outlined in SG's docs: https://sphinx-gallery.github.io/stable/configuration.html#controlling-what-output-is-captured

These changes make it so that users can specify what classes not to repr and this goes along with a new PR in PyVista to make our camera positions an actual class so that we can ignore them (cc: @akaszynski) per https://github.com/pyvista/pyvista/pull/491

So now if we add `'ignore_repr_classes' : (pyvista.CameraPosition,)` in PyVista's `conf.py`, then the camera positions are ignored! :tada:

Screen Shot 2019-12-08 at 9 16 29 PM

-----
## Help Please

It would be great if someone could help me set up the default arguments for this handle all the matplotlib types and PyVista's new `CameraPosition` class by default perhaps as a string list of class types. Anyone up for helping? pinging @lucyleeow since it looks like you originally implement all the repr stuff


banesullivan:
Or would it be better to use regex to have certain expressions ignored like `*.show(*` or `plt.plot*`, etc. so that way objects of these types could still be output when desired?

lucyleeow:
This seems reasonable to me, thanks @banesullivan. I think the regex would more likely result in unexpected consequences (the user can import modules in many different ways and code can include user-defined functions etc). It would also be more difficult to implement because we would have to convert the ast of the last expression back to the original code. Note that it would be difficult to obtain just the last expression of a code block using the original text code, which is why we do it using ast.

The `type` implementation looks good though. With regard to which types should be ignored by default, I'm not heavily involved in a project that uses SG so @larsoner would be better to offer suggestions here. I would just agree that Matplotlib objects makes sense - but it may be difficult to get a list of all Matplotlib objects to ignore?

banesullivan:
> but it may be difficult to get a list of all Matplotlib objects to ignore?

perhaps we could find some of the parent classes by MPL's inheritance diagrams like [this one](https://matplotlib.org/1.5.1/api/artist_api.html) which shows that `matplotlib.artist.Artist` would cover quite a bit. But still, there would be quite a lot of these

lucyleeow:
Nice diagram! That is a good idea. Maybe `isinstance` is better than `type` as it caters for inheritance?

I was looking at the return value of `plt.plot` and it is just a list - so that would be difficult to exclude...

banesullivan:
> I was looking at the return value of plt.plot and it is just a list - so that would be difficult to exclude...

Ah, good point...

lucyleeow:
Maybe just implement with `None` as the default? It would be a really useful configuration to have, especially for existing projects.

lucyleeow:
@larsoner @choldgraf feel free to chip in!

larsoner:
Seems reasonable to have this sort of option. Agreed it should use `isinstance` instead of `type`.

The other thing we could support is a `callable` (i.e., if the entry is not a `tuple` but is `callable`) that takes the object as an arg and returns True or False for whether or not to decorate. But maybe this is overkill. It would for example allow for custom ignores like `plt.plot()` because in the `callable` you could check if the object is a list and all entries in the list are artists.

lucyleeow:
@banesullivan I'm happy to take over this if you don't have time!

lucyleeow:
ping @banesullivan

banesullivan:
Hey, sorry for dropping the ball on this one. I don’t think I’ll be able to come back to this for a while still. Feel free to take over or leave it!

lucyleeow:
This is more complicated than I thought because:
```

type(matplotlib.text.Text)
```
gives me this error:
```
NameError                                 Traceback (most recent call last)
<ipython-input-9-35c6ddb0be07> in <module>
----> 1 type(matplotlib.text.Text)

NameError: name 'matplotlib' is not defined
```

So for classes from imported packages, the way you imported the package affects the types you have access to via `type()`. Am I missing something? @larsoner

larsoner:
If we have people specify types to skip based on a regex, at the end of the block execution we have the variable itself and we can just do `str(type(var))` and check if there is a match with the supplied regex

lucyleeow:
Ah of course, thanks @larsoner! I don't know why I got stuck on it needing to be a 'type'.

lucyleeow:
@larsoner hopefully the test is adequate?

sphinx_gallery/gen_gallery.py
```diff
@@ -68,6 +68,7 @@
     'junit': '',
     'log_level': {'backreference_missing': 'warning'},
     'inspect_global_variables': True,
+    'ignore_repr_classes': r'',
```
larsoner: Would `ignore_repr_types` be clearer since we check `type(...)` ?

lucyleeow: I am confused between [class and type](https://www.python.org/download/releases/2.2/descrintro/) now. Should I then change the documentation to use the term 'type' as well then?

larsoner: Yeah I'm not sure either. But I think in this case "type" is safer/clearer because we don't traverse all subclasses, we just check `type(x)`. One might assume if we named it `ignore_repr_classes` and they had `ignore_repr_classes='object'`, then *no* `__repr__`s would ever be printed. However, based on our implementation only in the (weird) case that someone returned `object()` nothing would get printed.

larsoner: ... In other words, asking "is this thing class X?", which is implied by `classes` in the current naming scheme, is really like doing `isinstance(thing, X)` rather than `type(thing) == X`. We currently check the latter, which I think is desired behavior, but makes `_classes` a misnomer.

lucyleeow: Ah I see, thanks. Changing now

lucyleeow:
I hope the doc makes sense...

larsoner:
Thanks @lucyleeow @banesullivan !

PR #582, opened 2019-12-31
mention literalinclude in the doc

emmanuelle:
Limiting code duplication makes documentation more robust, so I added an example of `literalinclude` in the doc, I hope this is the right place.

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -64,6 +64,20 @@ its reference
 it generated has the name ``sphx_glr_plot_0_sin_001.png``
 and its thumbnail is ``sphx_glr_plot_0_sin_thumb.png``
 
+You can also include part of a gallery script elsewhere in your documentation
+using the `literalinclude directive <https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude>`_, in order to limit code duplication::
+
+   .. literalinclude:: ../examples/plot_0_sin.py
+      :language: python
+      :lines: 43-53
+
+The above directive inserts the following block:
+
+.. literalinclude:: ../examples/plot_0_sin.py
+    :language: python
+    :lines: 43-51
+
+
 .. _custom_scraper:
 
 Write a custom image scraper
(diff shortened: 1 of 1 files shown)


emmanuelle:
Following discussions with Gaël, I added a warning and replaced line number with `start-after` and `end-before`.

doc/advanced.rst
```diff
@@ -64,6 +64,27 @@ its reference
 it generated has the name ``sphx_glr_plot_0_sin_001.png``
 and its thumbnail is ``sphx_glr_plot_0_sin_thumb.png``
 
+You can also include part of a gallery script elsewhere in your documentation
+using the `literalinclude directive <https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude>`_, in order to limit code duplication::
```
larsoner: Is it possible to make this an intersphinx doc or ref link? It's more robust than a pure URL

doc/advanced.rst
```diff
@@ -64,6 +64,27 @@ its reference
...
+   .. literalinclude:: ../examples/plot_0_sin.py
+      :language: python
+      :start-after: # License: BSD 3 clause
+      :end-before: # To avoid matplotlib
+
+The above directive inserts the following block:
+
+.. literalinclude:: ../examples/plot_0_sin.py
+    :language: python
+    :lines: 43-51
```
larsoner: I expected this to be a verbatim copy of the above but it's not, why use line numbers here but before and after above?

emmanuelle: oops, good catch, I modified one but not the other

emmanuelle:
Thank you for the review @larsoner , I made the requested changes.

emmanuelle:
I don't think Travis CI fail is related to this PR?

choldgraf:
agreed re: the test, and thanks for addressing the comments - this LGTM so I'm gonna give it a merge assuming that @larsoner is +1 after his comments are addressed! Thanks very much @emmanuelle :-)

emmanuelle:
Thank you all :-).

PR #584, opened 2020-01-03
[MRG] Fix backreferences for functions not directly imported

### LINKED ISSUE #583: BUG backreferences not working for functions if not imported directly
Creation of [mini-galleries](https://sphinx-gallery.github.io/stable/configuration.html#add-mini-galleries-for-api-documentation) with `backreferences_dir` and `doc_module` generates an empty rst file if you use a function in an example like this:

```
import package.module

package.module.my_fun()
```

It works if you directly import the function in the example:

```
from package.module import my_fun

my_fun()
```

or if you import the function in the `__init__.py` file of the package (i.e.:

```
from .module import (
    my_fun,
    )
```

Note that this is only for functions. Classes work fine in all cases.

Refer to https://github.com/sphinx-gallery/sample-project/pull/1#issuecomment-566664628 for suggestion on how to fix.

lucyleeow:
closes #583 

If an example uses a function like this:
```
import h.i

h.i.j()
```
[`get_mapping`](https://github.com/sphinx-gallery/sphinx-gallery/blob/44ed8000930a8b8cea70dcfda40a799750edee7d/sphinx_gallery/backreferences.py#L62) will check if `local_name`, in this case `h`, is in the `self.imported_names` dict. The key for this import would be `h.i` so it doesn't match.

I've added a for loop which will run for 'n' times, where 'n' is the max number of splits (`split('.')`) of all keys in `self.imported_names`. In each loop another split of the locally used name is added onto `local_name` which is checked to see if it exists in `self.imported_names`. e.g., for the above example the loop would run twice, first time `local_name` would be `h` and second time `local_name` would be `h.i`.

Not very elegant - open to changes @larsoner 

To remind myself, classes worked (as noted in #583) because when creating a class instance `local_name` existed in `self.global_variables` ([this line](https://github.com/sphinx-gallery/sphinx-gallery/blob/44ed8000930a8b8cea70dcfda40a799750edee7d/sphinx_gallery/backreferences.py#L71))

(thanks to @glemaitre for your help)

### DIFF
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -66,10 +66,10 @@ class NameFinder(ast.NodeVisitor):
         # ensure that max is at least one
         max_import_splits = len(max(imported_names_split + [''], key=len))
         for name in self.accessed_names:
-            # ensures self.imported_names with '.' are not missed by adding
-            # consecutive '.' split names to local_name
             for split_level in range(max_import_splits):
                 local_name_split = name.split('.')
+                # add the next '.' split to local_name
+                # ensures self.imported names with '.' are not missed
                 local_name = '.'.join(local_name_split[:split_level+1])
                 remainder = name[len(local_name):]
                 class_attr = False
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -62,8 +62,12 @@ class NameFinder(ast.NodeVisitor):
     def get_mapping(self):
         imported_names_split = \
             [key.split('.') for key in self.imported_names.keys()]
+        if imported_names_split:
+            max_import_splits = len(max(imported_names_split, key=len))
+        else:
+            max_import_splits = 1
         for name in self.accessed_names:
-            for i in range(len(max(imported_names_split, key=len))):
+            for i in range(max_import_splits):
                 local_name_split = name.split('.')
                 local_name = '.'.join(local_name_split[:i+1])
                 remainder = name[len(local_name):]
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -111,7 +111,6 @@ class NameFinder(ast.NodeVisitor):
                             full_name = '.'.join(
                                 module[:depth] + [class_name] + method)
                             yield name, full_name, class_attr, is_class
-                    break
 
 
 def _from_import(a, b):
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -62,9 +62,6 @@ class NameFinder(ast.NodeVisitor):
     def get_mapping(self):
         imported_names_split = \
             [key.split('.') for key in self.imported_names.keys()]
-        print('accessed names')
-        print(self.accessed_names)
-        print(self.global_variables)
         for name in self.accessed_names:
             for i in range(len(max(imported_names_split, key=len))):
                 local_name_split = name.split('.')
@@ -195,7 +192,6 @@ def identify_names(script_blocks, global_variables=None, node=''):
     for key, value in fill_guess.items():
         if key not in example_code_obj:
             example_code_obj[key] = value
-    print(example_code_obj)
     return example_code_obj
 
 
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -63,9 +63,11 @@ class NameFinder(ast.NodeVisitor):
     def get_mapping(self):
         imported_names_split = \
             [key.split('.') for key in self.imported_names]
-        # ensure that max is at least one here
+        # ensure that max is at least one
         max_import_splits = len(max(imported_names_split + [''], key=len))
         for name in self.accessed_names:
+            # ensures self.imported_names with '.' are not missed by adding
+            # consecutive '.' split names to local_name
             for split_level in range(max_import_splits):
                 local_name_split = name.split('.')
                 local_name = '.'.join(local_name_split[:split_level+1])
@@ -109,7 +111,6 @@ class NameFinder(ast.NodeVisitor):
                             full_name = '.'.join(
                                 module[:depth] + [class_name] + method)
                             yield name, full_name, class_attr, is_class
-                            break
 
 
 def _from_import(a, b):
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -60,38 +60,46 @@ class NameFinder(ast.NodeVisitor):
             self.visit(node)
 
     def get_mapping(self):
+        imported_names_split = \
+            [key.split('.') for key in self.imported_names.keys()]
+        print('accessed names')
+        print(self.accessed_names)
+        print(self.global_variables)
         for name in self.accessed_names:
-            local_name = name.split('.', 1)[0]
-            remainder = name[len(local_name):]
-            class_attr = False
-            if local_name in self.imported_names:
-                # Join import path to relative path
-                full_name = self.imported_names[local_name] + remainder
-                yield name, full_name, class_attr
-            elif local_name in self.global_variables:
-                obj = self.global_variables[local_name]
-                if remainder and remainder[0] == '.':  # maybe meth or attr
-                    method = [remainder[1:]]
-                    class_attr = True
-                else:
-                    method = []
-                # Recurse through all levels of bases
-                classes = [obj.__class__]
-                offset = 0
-                while offset < len(classes):
-                    for base in classes[offset].__bases__:
-                        if base not in classes:
-                            classes.append(base)
-                    offset += 1
-                for cc in classes:
-                    module = cc.__module__.split('.')
-                    class_name = cc.__name__
-                    # a.b.C.meth could be documented as a.C.meth,
-                    # so go down the list
-                    for depth in range(len(module), 0, -1):
-                        full_name = '.'.join(
-                            module[:depth] + [class_name] + method)
-                        yield name, full_name, class_attr
+            for i in range(len(max(imported_names_split, key=len))):
+                local_name_split = name.split('.')
+                local_name = '.'.join(local_name_split[:i+1])
+                remainder = name[len(local_name):]
+                class_attr = False
+                if local_name in self.imported_names:
+                    # Join import path to relative path
+                    full_name = self.imported_names[local_name] + remainder
+                    yield name, full_name, class_attr
+                    break
+                elif local_name in self.global_variables:
+                    obj = self.global_variables[local_name]
+                    if remainder and remainder[0] == '.':  # maybe meth or attr
+                        method = [remainder[1:]]
+                        class_attr = True
+                    else:
+                        method = []
+                    # Recurse through all levels of bases
+                    classes = [obj.__class__]
+                    offset = 0
+                    while offset < len(classes):
+                        for base in classes[offset].__bases__:
+                            if base not in classes:
+                                classes.append(base)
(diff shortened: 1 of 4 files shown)


lucyleeow:
Does the failures have something to do with pillow?

larsoner:
This is going to conflict with #581. Can you see if #581 fixes the problem? If it doesn't, would it be difficult to make a PR after #581 is merged?

lucyleeow:
@larsoner #581 doesn't fix the problem but yes I can do the PR post #581 

Would the `get_mapping` function become too unwieldy? Do you suggest refactoring or to just add my for loop in?

larsoner:
> Would the get_mapping function become too unwieldy? Do you suggest refactoring or to just add my for loop in?

Either way is fine with me. Whatever you think helps keep the code readable. I trust the regression tests to make sure nothing is broken, `master` already covers a few good cases and with #581 we cover even more.

And if you can fix it so that `np.random.RandomState` actually links properly, that would be amazing! (See [here](https://github.com/sphinx-gallery/sphinx-gallery/pull/581/files#diff-fcd3c885788a47a95a7fc6c389c176deR38-R40) we still have some deficiencies)

sphinx_gallery/backreferences.py
```diff
@@ -61,47 +61,56 @@ def visit_Attribute(self, node):
             self.visit(node)
 
     def get_mapping(self):
+        imported_names_split = \
+            [key.split('.') for key in self.imported_names.keys()]
+        if imported_names_split:
+            max_import_splits = len(max(imported_names_split, key=len))
+        else:
+            max_import_splits = 1
```
larsoner: to avoid the conditional you could do:
```
# ensure that the max is at least one here
max_import_splits = len(max(imported_names_split + [''], key=len))
```

sphinx_gallery/backreferences.py
```diff
@@ -61,47 +61,56 @@ def visit_Attribute(self, node):
             self.visit(node)
 
     def get_mapping(self):
+        imported_names_split = \
+            [key.split('.') for key in self.imported_names.keys()]
```
larsoner:
```suggestion
            [key.split('.') for key in self.imported_names]
```

sphinx_gallery/backreferences.py
```diff
@@ -61,47 +61,56 @@ def visit_Attribute(self, node):
...
+            max_import_splits = 1
         for name in self.accessed_names:
-            local_name = name.split('.', 1)[0]
-            remainder = name[len(local_name):]
-            class_attr = False
-            if local_name in self.imported_names:
-                # Join import path to relative path
-                full_name = self.imported_names[local_name] + remainder
-                if local_name in self.global_variables:
+            for i in range(max_import_splits):
```
larsoner: `i` is a difficult variable name to read, maybe better to use `ii` or something more explicit like `split_level` (or even `si` as an abbreviation). If you don't use it then `_` is a standard "will not be used" name.

sphinx_gallery/tests/test_backreferences.py
```diff
@@ -137,10 +148,10 @@ def test_identify_names2(tmpdir):
 Title
 -----
 
-This example uses :func:`h.i`.
+This example uses :func:`k.l`.
 '''
 """ + code_str.split(b"'''")[-1]
-    expected['h.i'] = {u'module': u'h', u'module_short': u'h', u'name': u'i',
+    expected['k.l'] = {u'module': u'k', u'module_short': u'k', u'name': u'l',
```
larsoner: Can you also check the `tinybuild` and add tests that fail in `master` and work now? It ensures that the real-world / end use cases improve in addition to our lower-level functions.

There might already be examples that didn't work before but now work (i.e., just change `test_full.py` but not anything in the `tinybuild/` directory itself), which would be great. But if you need to also change things in `tinybuild/` to get something that fails in `master` but works here it's fine, too.

lucyleeow: There isn't an appropriate import example in `tinybuild/` - I can add a `import a.b` to `/plot_numpy_matplotlib` ('Link to other packages') example and add an assertion somewhere in `test_full`. Does that work?

sphinx_gallery/backreferences.py
```diff
@@ -111,6 +109,7 @@ def get_mapping(self):
                             full_name = '.'.join(
                                 module[:depth] + [class_name] + method)
                             yield name, full_name, class_attr, is_class
+                            break
```
larsoner: This break will always occur in the first iteration of the for loop, so do we even need the loop? It seems likely that this loop is now equivalent to just doing:
```
if len(module) >= 1:  # probably don't even need this conditional
    full_name = ...
    yield ...
```
Is the loop no longer necessary because somewhere else we're already traversing the `full_name.split('.')` levels backward?

lucyleeow:
I have added a import case to `plot_numpy_matplotlib.py` and checked that it generates the appropriate backreferences mini-gallery rst (and that it doesn't work on master).

I know it doesn't really fit in with the example `plot_numpy_matplotlib.py` so I am happy to put this in it's own example `.py` file.

I also noticed that incorrect mini-gallery rst files are generated, which reference incorrect example `.py` files. Neither of these functions are used at all in the tinybuild examples:

* `sphinx_gallery.backreferences.NameFinder.examples` - references `plot_future_imports.py`
* `sphinx_gallery.sorting.FileNameSortKey.examples` - references `plot_numpy_matplotlib.py`

I am not sure if this error is related to the `np.random.RandomState(0)` problem and the inconsistent variable name linking mentioned here: https://github.com/mne-tools/mne-python/pull/7174#issuecomment-571302740. I can investigate these in a separate issue/PR.

sphinx_gallery/tests/test_backreferences.py
larsoner: Yes that works, but in addition to what you've done you should add something like the following after line ~198 in `test_full.py` as that's where we test that the code blocks have the appropriate outgoing links:
```
assert 'sphinx_gallery.backreferences.identify_names.html' in lines
```
It's fine to have all this in `plot_numpy_matplotlib.py` of `tinybuild`, no need for a separate file.

sphinx_gallery/backreferences.py
lucyleeow: Yes you are right. The loop is needed for the conditional `if local_name in self.imported_names:` but is not required for the condition `elif local_name in self.global_variables:`.

In the case where you are calling an indirectly imported function (or class), e.g.,
```
import h.i
h.i.j()
```
the `local_name`, `h`, will not be in `self.imported_names` or `self.global_variables`. It needs to traverse the loop for the first conditional `if local_name in self.imported_names:`.

Not sure how to fix...

larsoner: > It needs to traverse the loop for the first conditional if local_name in self.imported_names:

I'm confused :( Which loop are you talking about with this statement? The one up on line 69, or 80? Those are the two that satisfy `if local_name in self.imported_names`. Maybe the best thing to do is to add a commit with some code comments to help explain why some of the conditionals / loops are needed in the places that you know they are (and even `# XXX don't know why this is needed` in some places).

I'm just saying that the change you made in the code is equivalent to something simpler. So we should either simplify the loop starting on line 108 accordingly or revert the addition of the `break`.

Are you saying that (regardless of this equivalence) the code as it's written does not work properly?

lucyleeow: I added a comment but not sure if it helps.

The 69 for loop `for split_level in range(max_import_splits):`

And yes I realise now that the `break` was on the wrong level. It was meant to be similar to/on the same level as the `break` here:
```
            for split_level in range(max_import_splits):
                local_name_split = name.split('.')
                local_name = '.'.join(local_name_split[:split_level+1])
                remainder = name[len(local_name):]
                class_attr = False
                if local_name in self.imported_names:
                    # Join import path to relative path
                    full_name = self.imported_names[local_name] + remainder
                    if local_name in self.global_variables:
                        obj = self.global_variables[local_name]
                        if remainder:
                            for level in remainder[1:].split('.'):
                                obj = getattr(obj, level)
                        is_class = inspect.isclass(obj)
                    else:
                        is_class = False
                    yield name, full_name, class_attr, is_class
                    break   # <----this break!!
```
but you made me realise that the `for split_level in range(max_import_splits):` is only required for the first loop. i.e.

```
for split_level in range(max_import_splits):
    if local_name in self.imported_names:
    <the for loop is needed in here>
    elif local_name in self.global_variables:
    <the for loop isn't needed ehre>
```

larsoner: This suggests that the code should be refactored to:
```
if local_name in self.imported_names:
    <the for loop is needed in here>
    for split_level in range(max_import_splits):
elif local_name in self.global_variables:
    <the for loop isn't needed here>
```
Right?

larsoner: Never mind, this wouldn't work as `local_name` is formed using the `split_level`. So maybe you were talking about a different for loop...

lucyleeow: Different loop?

For case of `import h.i; h.i.j()`:

```
for split_level in range(max_import_splits):
    if local_name in self.imported_names:
        # the for loop is needed in here because self.imported_names is `h.i`
        # so the loop would check for `h` in self.imported_names first time round, then `h.i` in
        # self.imported_names
    elif local_name in self.global_variables:
        # you don't need the loop here as you are checking for variable names, which
        # won't have a '.' - you just have to worry about attirbutes
```
But I don't have a great understanding of what goes on after the `elif local_name in self.global_variables:` so please correct me.

lucyleeow: I've put the `break` at the correct level now - which may or may not make things clearer.

larsoner:
That break appears to have broken things

sphinx_gallery/backreferences.py
lucyleeow: The `break` broke things because it seems that you need to traverse the loop again if in the first loop you have this case:

```
for split_level in range(max_import_splits):
    if local_name in self.imported_names:
        # condition not met on first loop
    elif local_name in self.global_variables:
        # condition met on first loop
```

Because otherwise `example_code_obj` returned from `identify_names` is incorrect. I don't really understand why. 
e.g., for tinybuild `sphinx_gallery.backreferences.identify_names([('text', 'Text block', 1)])`, `example_code_obj` is:

```
# with break
('sphinx_gallery.backreferences.identify_names', {'name': 'backreferences.identify_names', 'module': 'builtins.module', 'module_short': 'builtins.module', 'is_class': False})])
# without break
('sphinx_gallery.backreferences.identify_names', {'name': 'identify_names', 'module': 'sphinx_gallery.backreferences', 'module_short': 'sphinx_gallery.backreferences', 'is_class': False}),
```

I think that without the break both are yielded by `get_mapping` but somehow `identify_names` knows that `'name': 'backreferences.identify_names'` is the good one so only returns this? I'm a bit confused by how `identify_names` works but does that make sense?

larsoner: > but somehow identify_names knows that 'name': 'backreferences.identify_names' is the good one

It looks for valid intersphinx references, so yeah it should subselect the correct one (wherever it's documented in the docs)

lucyleeow:
This failure seems to be a problem with installation with Python Nightly

lucyleeow:
So this works and fixes this specific bug. Going forward and fixing the other 3 problems (RandomState, extra / incorrect backrefs, incomplete method parsing) we might want to think about refactoring the `get_mapping` function. 

I will tackle this when I feel brave enough (also have some other work to do), unless someone beats me to it.

larsoner:
Thanks @lucyleeow !

PR #586, opened 2020-01-05
MAINT: Update checks for PIL/JPEG

larsoner:


larsoner:
Added a `--tb=short` on the CI runs as it should help make the tracebacks more readable when there are errors.

Also determined that the problem is likely due to https://github.com/matplotlib/matplotlib/issues/16083. If everything comes back green I'll merge and re-run the PRs that are currently red.

PR #589, opened 2020-01-08
Remove duplicate doc_solver entry in the API reference structure

kanderso-nrel:
I just started looking at this project today and noticed that `docs_resolv` is in the API Reference list twice.  This is just a 1-line PR to remove the second entry. 

image


lucyleeow:
Thanks for the fix @kanderso-nrel! (test failure is unrelated)

PR #590, opened 2020-01-09
[MRG] [ENH] Add css for pandas df

### LINKED ISSUE #544: ENH Add CSS for pandas dataframe
With #541, Sphinx-Gallery can capture raw HTML (using the `_repr_html_()` of an object) and embed in rst. In some cases, where the captured raw HTML does not include CSS (e.g. pandas dataframes), this can be added to make the output look better. For reference [this](https://1052-25860190-gh.circle-artifacts.com/0/rtd_html/auto_examples/plot_capture_repr.html) (scroll down) is what a pandas dataframe looks like with just the raw HTML.

For example, Jupyter notebook seems to take CSS from their own [`renderedhtml.less`](https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/less/renderedhtml.less) (see @larsoner's [post](https://github.com/sphinx-gallery/sphinx-gallery/pull/541#issuecomment-539196065)). For reference this [SO](https://stackoverflow.com/questions/49324569/replicating-jupyter-notebook-pandas-dataframe-html-printout) question describes how to get pandas dataframes to look like those in Jupyter notebook. 

One solution is that we can include our own CSS for common objects like pandas dataframes (I think this might be beyond the scope of SG).

~Another solution is to obtain CSS from the objects we capture `_repr_html_` from, if possible.~

Another solution would be to add a  `add_css` configuration to allow the user add their own CSS file (e.g. pandas dataframes have specific HTML tags that can be used to write CSS specifically for dataframes), a reference to which would be added at the end of raw HTML embedded.

lucyleeow:
closes #544 

Added CSS for pandas df, copied from nbsphinx (suggested here: https://github.com/sphinx-gallery/sphinx-gallery/issues/544#issuecomment-539992747).

I can include a border:
bord

Or remove the black border:
nobord

The latter makes it look exactly like in jupyter notebook:
image

Note that rows become highlighted in light blue on hover, just like in notebooks.

### DIFF
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -206,6 +206,7 @@ table.dataframe table{
   border: none;
   border-collapse: collapse;
   border-spacing: 0;
+  border-color: transparent;
   color: black;
   font-size: 12px;
   table-layout: fixed;
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -202,7 +202,7 @@ a.sphx-glr-backref-instance {
 /* Pandas dataframe css */
 /* Taken from: https://github.com/spatialaudio/nbsphinx/blob/fb3ba670fc1ba5f54d4c487573dbc1b4ecf7e9ff/src/nbsphinx.py#L587-L619 */
 
-table.dataframe table{
+table.dataframe {
   border: none !important;
   border-collapse: collapse;
   border-spacing: 0;
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -203,7 +203,7 @@ a.sphx-glr-backref-instance {
 /* Taken from: https://github.com/spatialaudio/nbsphinx/blob/fb3ba670fc1ba5f54d4c487573dbc1b4ecf7e9ff/src/nbsphinx.py#L587-L619 */
 
 table.dataframe table{
-  border: none;
+  border: none !important;
   border-collapse: collapse;
   border-spacing: 0;
   border-color: transparent;
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -198,3 +198,39 @@ p.sphx-glr-signature a.reference.external {
 a.sphx-glr-backref-instance {
   text-decoration: none;
 }
+
+/* Pandas dataframe css */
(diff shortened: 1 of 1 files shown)


larsoner:
I like the second one (no border)

lucyleeow:
I'm having trouble removing the border and I think it is because of the inline `border="1"` of the `_repr_html_`:

```
<table border="1" class="dataframe">      <-----here
  <thead>
    <tr style="text-align: right;">
      <th></th>
      <th>a</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th>0</th>
      <td>1</td>
    </tr>
    <tr>
      <th>1</th>
      <td>2</td>
    </tr>
    <tr>
      <th>2</th>
      <td>3</td>
    </tr>
  </tbody>
</table>
```

I can make it go away by adding `border: none` in inspect element.
Screenshot_2020-01-09_16-23-44

But that is already in the CSS I added...

lucyleeow:
Manually adding `border: none` to `doc/_build/html/_static/css/theme.css` file works - but how is this `theme.css` file created?

larsoner:
> doc/_build/html/_static/css/theme.css

This probably comes from `sphinx_rtd_theme` so we can't easily modify it.

One we can modify is `doc/_static/theme_override.css`, which gets copied to `doc/_build/html/_static/theme_override.css`, but this will only affect our (SG's) particular doc build, not the doc build of projects that use SG, so I wouldn't look at that one.

If you want to modify it for all downstream projects (and SG's doc build, too), change `sphinx_gallery/_static/gallery.css` which ends up being copied to `doc/_build/html/_static/gallery.css`. So for testing you can see if changing `doc/_build/html/_static/gallery.css` (maybe with a `!important` if absolutely necessary) fixes things, and if it does, modify `sphinx_gallery/_static/gallery.css` the same way.

lucyleeow:
I was adding the CSS to `sphinx_gallery/_static/gallery.css` but I guess it was over-ridden for some reason? Anyway adding `!important` to the `border: none` tag in `sphinx_gallery/_static/gallery.css` worked.

larsoner:
@lucyleeow it looks like you did not commit the `!important` and accordingly there is still a border here:

https://1221-25860190-gh.circle-artifacts.com/0/rtd_html/auto_examples/plot_3_capture_repr.html

choldgraf:
FWIW I really like the change, +1 once the border is fixed :-)

lucyleeow:
Got it working, see [here](https://1240-25860190-gh.circle-artifacts.com/0/rtd_html/auto_examples/plot_3_capture_repr.html).

larsoner:
Thanks @lucyleeow !

PR #594, opened 2020-01-13
[DOC] Minor amendments to CSS config part

lucyleeow:
Minor clarifications because I was a bit confused when re-reading the doc. Suggestions/changes welcome.


larsoner:
Thanks @lucyleeow

PR #595, opened 2020-01-13
CSS: Restrict thumbnail height to 112 px

mgeier:
When using an SVG plot in "portrait" mode (i.e. height larger than width), the thumbnail can collide with the title text.

It looks like SVG thumbnails are currently assumed to be in "landscape" mode, this PR avoids this assumption.

I'm not very good at CSS, so this is very likely to have unintended side effects!

### DIFF
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -44,8 +44,8 @@ thumbnail with its default link Background color */
 }
 .sphx-glr-thumbcontainer img {
   display: inline;
-  max-height: 160px;
-  width: 160px;
+  max-height: 112px;
+  max-width: 160px;
 }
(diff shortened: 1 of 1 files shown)


larsoner:
FWIW this seems to be restricted to SVGs, probably because the thumbnailing rescaler takes care of cropping PNGs for us. Confirmed that on `master` using the [matplotlib SVG scraping](https://sphinx-gallery.github.io/stable/advanced.html#example-3-matplotlib-with-svg-format) with `plot_6_function_identifier` having `figsize=(4, 8)` the title does collide with the image, and on this PR it does not.

I'm not sure whether we should have `max-width` vs `width` (and same for `height`), but it does seems sensible to set both the same way. Eventually we might need `object-fit: contain` or something, too. But for now this seems to at least be a step in the right direction so let's get it in. Thanks @mgeier !

Issue #587, opened 2020-01-07
BUG: Bugs with backref links


larsoner:
Here are bugs I've seen with the backref links:

- [x] Class instances after creation sometimes [only get](https://github.com/mne-tools/mne-python/pull/7181#issuecomment-571685516) the `sphx-glr-backref-module-mne` class, not `sphx-glr-backref-type-class` and `sphx-glr-backref-instance` -- see also #588 (addressed in #591)
- [x] `np.random.RandomState` does not link properly (#593)
- [x] Method parsing is incomplete, in MNE things like `ica.plot_properties` do not link even though they should. This could be because of greedy replacement of `ica` occurring before `ica.plot_properties` can be replaced, or could be a problem with `get_mapping` (or downstream of there) (#596).
- [ ] There are [extra / incorrect backrefs](https://github.com/sphinx-gallery/sphinx-gallery/pull/584#issuecomment-571601897)

Hopefully to be fixed by follow-up PRs.

lucyleeow:
@larsoner number 2 `np.random.RandomState` is because it doesn't exist in the intersphinx named inventory.

I tried this on the template sample project and wrote `intersphinx_inv` -
```
intersphinx_inv = getattr(app.env, 'intersphinx_named_inventory', dict())
```
from [`_embed_code_links()`](https://github.com/sphinx-gallery/sphinx-gallery/blob/4cee8a0e82dc41ee58ce94214fd79eb808107dd4/sphinx_gallery/docs_resolv.py#L306) to a file - [intersphinx.txt](https://github.com/sphinx-gallery/sphinx-gallery/files/4047148/intersphinx.txt).

The intersphinx part of `conf.py` file looks like this:
```
intersphinx_mapping = {
    'python': ('https://docs.python.org/{.major}'.format(sys.version_info), None),
    'matplotlib': ('https://matplotlib.org/', None),
    'pandas': ('https://pandas.pydata.org/', None),
    'numpy': ('https://docs.scipy.org/doc/numpy', None),
}
```

Interestingly `pandas.Dataframe()` also does not exist - accordingly it is not linked in our [docs](https://sphinx-gallery.github.io/dev/auto_examples/plot_3_capture_repr.html) either.

Not familiar with intersphinx so not sure how to pursue this. What do you suggest?



(will look at your PR soon)

larsoner:
I have a fix almost complete for that too

lucyleeow:
[...]

PR #596, opened 2020-01-13
MRG, ENH: Link to methods and properties properly

larsoner:
Link to methods and properties properly (part of #587).

Locally I can confirm that building MNE docs now produces links to things like `ica.plot_properties`, which is great.

### DIFF
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -23,6 +23,19 @@ from .scrapers import _find_image_ext
 from .utils import _replace_md5
 
 
+class DummyClass(object):
+    """Dummy class for testing method resolution."""
+
+    def run(self):
+        """Do nothing."""
+        pass
+
+    @property
+    def prop(self):
+        """Property."""
+        return 'Property'
+
+
 class NameFinder(ast.NodeVisitor):
     """Finds the longest form of variable names and their imports in code.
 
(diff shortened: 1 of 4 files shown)


sphinx_gallery/backreferences.py
```diff
@@ -23,6 +23,19 @@
...
+    """Dummy class for testing method resolution."""
+
+    def run(self):
+        """Do nothing."""
+        pass
+
+    @property
+    def prop(self):
+        """Property."""
+        return 'Property'
```
larsoner: This is used for `tinybuild` testing. The cost is a useless entry in our actual official docs but I don't think it's too bad.

sphinx_gallery/backreferences.py
```diff
@@ -121,12 +125,20 @@ def get_mapping(self):
...
+                remainder = name[len(local_name):]
+                if local_name in self.imported_names:
                     full_name = self.imported_names[local_name] + remainder
-                    is_class = False
-                    yield name, full_name, class_attr, is_class
-                    break
+                    is_class = class_attr = False  # can't tell without import
+                    options.append(
+                        (name, full_name, class_attr, is_class))
+        return options
```
larsoner: @lucyleeow I simplified this a great deal. The diff is a bit ugly, but if you look at the actual file now hopefully it reads much more easily than it did before:

https://github.com/sphinx-gallery/sphinx-gallery/blob/58df0e6d472fab7e6745db6b01b9ee24333b3ab3/sphinx_gallery/backreferences.py#L80-L140

sphinx_gallery/tests/test_full.py
```diff
@@ -191,7 +191,7 @@ def test_embed_links_and_styles(sphinx_app):
     assert 'pyplot.html' in lines
     assert 'matplotlib.figure.Figure.html#matplotlib.figure.Figure.tight_layout' in lines  # noqa: E501
     assert 'matplotlib.axes.Axes.plot.html#matplotlib.axes.Axes.plot' in lines
-    assert 'matplotlib_configuration_api.html#matplotlib.rcParams' in lines
+    assert 'matplotlib_configuration_api.html#matplotlib.RcParams' in lines
```
larsoner: The changes above modified how this was resolved, which I think is okay. The `rcParams` is an instance of `RcParams`, so this is still a legitimate documentation link to provide. (Whether it should choose one or the other will depend on the package itself, and we probably don't have a good way to know. Plus this should be rare, as it should be rare that packages have case-based naming differentiation in the same submodules, both of which get documented.)

lucyleeow:
LGTM thanks @larsoner

PR #598, opened 2020-01-18
MRG, MAINT: Update deps

### LINKED ISSUE #597: Reduce the amount of hard dependencies?
In `setup.py`, there seem to be a lot of requirements in `install_requires` (which is populated from `requirements.txt`).

* [x] `setuptools` seems to be a strange requirement, since `setup.py` already imports it (before reading the list of requirements).

* [x] `matplotlib` is of course an important use case for SG, but is it really required? AFAICT, it's not used in the code, only in tests. There are use cases for SG without Matplotlib, right?

* [x] `pillow` seems to be needed for `scale_image()`. But what if no scaling is needed (for example when using only SVG images)?
   Also, why is this a hard requirement but in the code it looks like a soft requirement:

   https://github.com/sphinx-gallery/sphinx-gallery/blob/dff998dc51eb1c718be900e12aaa6394f61e5218/sphinx_gallery/utils.py#L23-L27

* [x] `sphinx>=1.8.3` I guess is justified because of the specific version. But I'm not quite sure.

* [x] `seaborn` is only included on `READTHEDOCS`, but it still seems strange. It would probably be better to use a `.readthedocs.yml` with probably a few "extra" requirements. Or probably a separate `requirements.txt` only for the docs?

larsoner:
Closes #597.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,6 +39,8 @@ Who uses Sphinx-Gallery
 * `PlasmaPy <http://docs.plasmapy.org/en/latest/auto_examples/>`_
 * `Fury <http://fury.gl/latest/auto_examples/index.html>`_
 
+.. installation-begin-content
+
 Installation
 ============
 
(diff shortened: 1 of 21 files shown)


README.rst
```diff
@@ -52,40 +52,27 @@ You can do a direct install via pip by using:
...
-    $ pip install -r requirements.txt
-    $ pip install pytest pytest-coverage
-    $ pip install -e .
-
-In addition, you will need the following dependencies to build the
-``sphinx-gallery`` documentation:
-
-* Scipy
-* Seaborn
+Additional to the dependencies listed above you will need to install `pytest`,
```
choldgraf: Could we just have a `dev-requirements.txt`?

larsoner: Done

larsoner:
@choldgraf comments addressed, feel free to merge if you're happy

choldgraf:
Thanks, it LGTM - clever use of "start-after" as well :-)

Issue #317, opened 2017-11-21
A place to put CSS for different gallery stylings


choldgraf:
One thing that's nice about sphinx-gallery is that you can use whatever CSS you want in order to get the gallery look you want. E.g., with a few small tweaks you can get seaborn-like gallery images (grey box misalignment notwithstanding...this only took me like 5 minutes)

![image](https://user-images.githubusercontent.com/1839645/33047947-43dc3db0-ce0d-11e7-9c45-2851afacdf93.png)

I wonder if it'd be worth having a `css` folder somewhere that we could store sample CSS files to achieve different kinds of galleries. e.g., in there could be a `seaborn.css` file that would define the necessary changes to get seaborn-like galleries. What do folks think about this?

choldgraf:
or e.g., the scikit-image styling that includes rounded borders

http://scikit-image.org/docs/stable/auto_examples/

GaelVaroquaux:
I like the general idea. How do you propose to expose this to users? 

⁣Sent from my phone. Please forgive typos and briefness.​

Titan-C:
We already have our css in sphinx-gallery/_static/gallery.css
We may put more there if there are radical changes needed for the
defaults on different themes.

I would nevertheless prefer to use css overloading. That is the user
puts his custom css in his doc/_static/mytheme.css and then lets Sphinx
use that. The same way we do to overload the readthedocs theme in our
conf. We could document that, although is more a Sphinx thing documented
by them.

https://github.com/sphinx-gallery/sphinx-gallery/blob/master/doc/conf.py#L138

Scikit-learn and scikit-image do the overloading directly in their css files.

Titan-C:
@choldgraf We have rounded borders as default. But the container has no border by default until you hover on it.

choldgraf:
@Titan-C that's what I'm suggesting too, I'm just saying we could have the CSS needed for different look and feel as a reference for others who want to overload on their own themes. Maybe this would be better as a separate repo or something

GaelVaroquaux:
I do think that it answers a desire: it's a theming. Maybe we could have
an example folder of themes and tell users to copy the corresponding css
file to something that they include.

Titan-C:
> I do think that it answers a desire: it's a theming.

I understand that people would be interested in using different styles
for their gallery. We have a reasonably pretty default, which "survives"
all the default Sphinx themes (we already experienced some collisions
with some themes).

[...]

PR #599, opened 2020-01-18
MRG, MAINT: Split CSS and add control

larsoner:
Implements ideas from https://github.com/sphinx-gallery/sphinx-gallery/issues/317#issuecomment-575719139.

If we decide to go this route (which seems like a useful refactoring) then I can add config instructions.

### DIFF
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -31,6 +31,9 @@ from .downloads import generate_zipfiles
 from .sorting import NumberOfCodeLinesSortKey
 from .binder import copy_binder_files
 
+
+_KNOWN_CSS = ('gallery', 'gallery-binder', 'gallery-dataframe')
+
 DEFAULT_GALLERY_CONF = {
     'filename_pattern': re.escape(os.sep) + 'plot',
     'ignore_pattern': r'__init__\.py',
@@ -65,6 +68,7 @@ DEFAULT_GALLERY_CONF = {
     'junit': '',
     'log_level': {'backreference_missing': 'warning'},
     'inspect_global_variables': True,
+    'css': _KNOWN_CSS,
 }
 
 logger = sphinx_compatibility.getLogger('sphinx-gallery')
@@ -188,6 +192,17 @@ def _complete_gallery_conf(sphinx_gallery_conf, src_dir, plot_gallery,
     if (not isinstance(backref, str)) and (backref is not None):
         raise ValueError("The 'backreferences_dir' parameter must be of type"
                          "str or None, found type %s" % type(backref))
+
+    if not isinstance(gallery_conf['css'], (list, tuple)):
+        raise TypeError('gallery_conf["css"] must be list or tuple, got %r'
+                        % (gallery_conf['css'],))
+    for css in gallery_conf['css']:
+        if css not in _KNOWN_CSS:
+            raise ValueError('Unknown css %r, must be one of %r'
+                             % (css, known_css))
+        if gallery_conf['app'] is not None:  # can be None in testing
+            gallery_conf['app'].add_css_file(css + '.css')
+
     return gallery_conf
 
 
@@ -594,8 +609,6 @@ def setup(app):
     app.add_config_value('sphinx_gallery_conf', DEFAULT_GALLERY_CONF, 'html')
     for key in ['plot_gallery', 'abort_on_example_error']:
         app.add_config_value(key, get_default_config_value(key), 'html')
-
-    app.add_css_file('gallery.css')
 
     if 'sphinx.ext.autodoc' in app.extensions:
         app.connect('autodoc-process-docstring', touch_empty_backreferences)
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -199,7 +199,7 @@ def _complete_gallery_conf(sphinx_gallery_conf, src_dir, plot_gallery,
     for css in gallery_conf['css']:
         if css not in _KNOWN_CSS:
             raise ValueError('Unknown css %r, must be one of %r'
-                             % (css, known_css))
+                             % (css, _KNOWN_CSS))
         if gallery_conf['app'] is not None:  # can be None in testing
             gallery_conf['app'].add_css_file(css + '.css')
 
--- a/sphinx_gallery/_static/gallery-binder.css
+++ b/sphinx_gallery/_static/gallery-binder.css
@@ -1,0 +1,6 @@
+/* CSS for binder integration */
+
+div.binder-badge {
+  margin: 1em auto;
+  vertical-align: middle;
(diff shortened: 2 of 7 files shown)


larsoner:
cc @mgeier see if this would help with what you're trying to do

mgeier:
Thanks, this is definitely a step in the right direction!

It works for my use case when I use this config:

```python
extensions = [
    'sphinx_gallery.gen_gallery',
]
sphinx_gallery_conf = {
    'css_styles': ['gallery.css'],
    'examples_dirs': [],
}
```

However, this does still quite a few things that I don't need.
Also, ideally I would like to have users do this (instead of my extension fiddling with the config behind their backs), but for that it would be great if this could be even simpler.

An arguably less complicated way (that also works for me with this PR) would be this:

```python
import sphinx_gallery
html_static_path = [sphinx_gallery.glr_path_static()]
html_css_files = ['gallery.css']
```

I think it would be great if we could reduce this even more.
For example, by only having to write something like this:

```python
extensions = [
    'sphinx_gallery.load_style',
]
```

This could be implemented with a reasonably simple sub-module `load_style.py` (there might be a better name?):

```python
from . import glr_path_static, __version__


def config_inited(app, config):
    path = glr_path_static()
    if path not in config.html_static_path:
        config.html_static_path.append(path)
    app.add_css_file('gallery.css')


def setup(app):
    app.require_sphinx('1.8')
    app.connect('config-inited', config_inited)
    return {
        'parallel_read_safe': True,
        'version': __version__,
    }
```

This can of course also be done in a follow-up PR.

larsoner:
> I think it would be great if we could reduce this even more.
> For example, by only having to write something like this:
>
> extensions = [
>     'sphinx_gallery.load_style',
> ]

This could be simpler in some cases for people who just want to load the default set of SG styles (or whatever set we make `load_style` load by default), but it's less simple for main users of SG (those who want to generate a gallery):

1. It is not backward compatible for current users of SG. They need to add a new line to `extensions` to get the old behavior.
2. It also makes it so that they "configure SG" in two ways, via adding to `extensions` another line or not, and by changing `gallery_conf`.

Even in the case where you want to load SG style(s), we'd still want a way to say which styles to load. To configure that, you'll need to set something in `sphinx_gallery_conf` anyway. So now the only work we will have saved is people don't need to write `examples_dirs=()` in the `sphinx_gallery_conf`. And in doing so, we've increased the amount of testing and support we need to do (multiple extensions, entry points for us as a sphinx extension, etc.) and so does not seem worth it.

@lucyleeow I'll mark this as MRG and we can get it in. Even if we do eventually want to go a new-sub-extension route at some point, we should do it in another PR since this one already adds stuff that should be useful for folks (and does useful refactoring).

sphinx_gallery/gen_gallery.py
```diff
@@ -188,6 +192,17 @@ def _complete_gallery_conf(sphinx_gallery_conf, src_dir, plot_gallery,
...
         raise ValueError("The 'backreferences_dir' parameter must be of type"
                          "str or None, found type %s" % type(backref))
+
+    if not isinstance(gallery_conf['css'], (list, tuple)):
+        raise TypeError('gallery_conf["css"] must be list or tuple, got %r'
+                        % (gallery_conf['css'],))
+    for css in gallery_conf['css']:
+        if css not in _KNOWN_CSS:
+            raise ValueError('Unknown css %r, must be one of %r'
+                             % (css, known_css))
```
mgeier: I guess this should be `_KNOWN_CSS`.

BTW, those exceptions don't look very nice. If you would use `sphinx.errors.ExtensionError`, the traceback would be omitted and it would displayed as an "Extension error".

mgeier:
> Even if we do eventually want to go a new-sub-extension route at some point, we should do it in another PR since this one already adds stuff that should be useful for folks (and does useful refactoring).

Yes please! I don't want to delay this PR (apart from my comment above).

Once this PR is merged, I will create a new PR with my suggested changes.
For now, you can have a look at https://github.com/sphinx-gallery/sphinx-gallery/commit/2d654451ca8ff1fec8dd705c9dfb827dd6f41c38.

> > I think it would be great if we could reduce this even more.
> > For example, by only having to write something like this:
> > ```python
> > extensions = [
> >     'sphinx_gallery.load_style',
> > ]
> > ```
> 
> This could be simpler in some cases for people who just want to load the default set of SG styles (or whatever set we make `load_style` load by default), but it's less simple for main users of SG (those who want to generate a gallery)

Sorry, I haven't been clear.
I meant `gen_gallery` and `load_style` to be mutually exclusive (see https://github.com/sphinx-gallery/sphinx-gallery/commit/2d654451ca8ff1fec8dd705c9dfb827dd6f41c38)!

> It is not backward compatible for current users of SG. They need to add a new line to `extensions` to get the old behavior.

This would be indeed bad.

> It also makes it so that they "configure SG" in two ways, via adding to `extensions` another line or not, and by changing `gallery_conf`.

This would also be quite bad.

> Even in the case where you want to load SG style(s), we'd still want a way to say which styles to load. To configure that, you'll need to set something in `sphinx_gallery_conf` anyway.

Sure, if you want that, we can do that, see my not-yet-PR https://github.com/sphinx-gallery/sphinx-gallery/commit/2d654451ca8ff1fec8dd705c9dfb827dd6f41c38.

> So now the only work we will have saved is people don't need to write `examples_dirs=()` in the `sphinx_gallery_conf`.

Yes, and that's great!

To be more exact, we will have saved people from having to write this:

```python
sphinx_gallery_conf = {
    'css': ['gallery'],
    'examples_dirs': [],
}
```

... and from loading `gen_gallery`, which would be a confusing thing to load if you don't actually want to generate a gallery.

And on top of that, we would have avoided to run quite a huge amount of unneeded code including setting up several unneeded Sphinx hooks (which might not be harmful, but anyway).

> And in doing so, we've increased the amount of testing and support we need to do (multiple extensions, entry points for us as a sphinx extension, etc.)

Well, yes, that would be the price you as maintainers would have to pay for a great user experience!

A taste of the additional code can be seen at https://github.com/sphinx-gallery/sphinx-gallery/commit/2d654451ca8ff1fec8dd705c9dfb827dd6f41c38, but this can probably be reduced by some refactoring (which we can discuss at a later point).

> and so does not seem worth it.

That's of course your (the project maintainers') decision.
But please wait with this decision until we have discussed my upcoming PR.

sphinx_gallery/gen_gallery.py
larsoner: I'll make a separate issue for that since we have lots of errors in the codebase, we should probably fix them all at once

larsoner:
@mgeier indeed 2d65445 looks like it might not be as painful as I thought, feel free to make a PR

mgeier:
Thanks @larsoner, I've created #601, which suggests a minimalistic implementation, but if you want, I can add a configuration option.

PR #601, opened 2020-01-24
MRG, ENH: Provide sub-extension sphinx_gallery.load_style

mgeier:
As discussed in https://github.com/sphinx-gallery/sphinx-gallery/pull/599#issuecomment-576200916, this enables following usage in `conf.py`:

```python
extensions = [
    'sphinx_gallery.load_style',
]
```

... which only loads the CSS style and sets the corresponding static directory. Nothing else.

For now, this provides a minimal solution without any configuration options.

@larsoner suggested in https://github.com/sphinx-gallery/sphinx-gallery/pull/599#issuecomment-576750591 to add configurability:

> Even in the case where you want to load SG style(s), we'd still want a way to say which styles to load.

I think we don't need that yet, because for now there aren't even multiple styles, there's only one.
I think configurability can be added later (when needed), an example for a possible implementation is in 2d654451ca8ff1fec8dd705c9dfb827dd6f41c38.

In case anyone is wondering what's the use case for this, here's some explanations: https://github.com/sphinx-gallery/sphinx-gallery/issues/317#issuecomment-575692356.

The "normal" use of SG should not be affected by this in any way.

### DIFF
--- a/sphinx_gallery/tests/conftest.py
+++ b/sphinx_gallery/tests/conftest.py
@@ -179,7 +179,7 @@ def conf_file(request):
 class SphinxAppWrapper(object):
     """Wrapper for sphinx.application.Application.
 
-    This allows to control when the sphinx application is initialized, since
+    This allows control over when the sphinx application is initialized, since
     part of the sphinx-gallery build is done in
     sphinx.application.Application.__init__ and the remainder is done in
     sphinx.application.Application.build.
--- a/sphinx_gallery/tests/conftest.py
+++ b/sphinx_gallery/tests/conftest.py
@@ -5,10 +5,16 @@ Pytest fixtures
 from __future__ import division, absolute_import, print_function
 
 import collections
+from contextlib import contextmanager
+from io import StringIO
+import os
+import shutil
 
 import pytest
 
 import sphinx
+from sphinx.application import Sphinx
+from sphinx.util.docutils import docutils_namespace
 from sphinx_gallery import (docs_resolv, gen_gallery, gen_rst, utils,
                             sphinx_compatibility, py_source_parser)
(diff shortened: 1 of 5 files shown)


mgeier:
In case somebody is interested in a concrete use case: https://github.com/spatialaudio/nbsphinx/pull/392.

larsoner:
Looks reasonable to me. It would be nice not to take a coverage hit, though. Can you add a test? If it's not clear how to do it with the `sphinx` machinery we have in place, I can take a look and push

mgeier:
@larsoner Thanks! If you already know how a test could be added and if you have a bit of time to spare, it would be great if you could create a test.

If not, I guess I could start experimenting with creating a test file similar to `sphinx_gallery/tests/test_gen_gallery.py` (but much shorter).
However, to re-use some of the fixtures from there, they would have to be factored out (e.g. into `sphinx_gallery/tests/conftest.py`?), right?

mgeier:
Thanks a lot @larsoner for writing the test!
You don't seem to have added the test file to version control though?

larsoner:

larsoner:
Or @choldgraf do you want to look?

sphinx_gallery/tests/conftest.py
```diff
@@ -151,6 +157,89 @@ def req_pil():
...
+    }
+    result.update(kwargs)
+
+    return result
+
+
+class SphinxAppWrapper(object):
+    """Wrapper for sphinx.application.Application.
+
+    This allows to control when the sphinx application is initialized, since
```
lucyleeow:
```suggestion
    This allows control over when the sphinx application is initialized, since
```

larsoner:
Thanks @mgeier !

mgeier:
Thanks for merging!

Issue #607, opened 2020-02-12
thumbnails from loaded figures


prisae:
I have a couple of examples where I compare one code to other, similar codes. These examples produce all very similar figures. I would therefore prefer the thumbs to be more informative, e.g. the logo of the code I am comparing to.

In other cases each example presents a method. Again, the plot itself is not very informative, but a little figure/sketch/drawing of the method would be great.

I can load figures easily in the examples with the `.. figure ::`-directive. However, these figures are not chosen even if they are the first figure in the example.

Is there any way how one can specify a, per-example-basis, figure for the thumbnail?

Or do you think it would be possible to extend the following directive
```# sphinx_gallery_thumbnail_number = 2```
 to also accept strings? E.g.,
```# sphinx_gallery_thumbnail_number = 'path/to/figure.png'```

These are some related issues/PRs I found, but none of them resolves it. #5 actually already suggested similar things, but it was closed then referring to the then introduced scraper API. However, I don't think the scraper API addresses this particular use-case.
- https://github.com/sphinx-gallery/sphinx-gallery/issues/5
- https://github.com/sphinx-gallery/sphinx-gallery/issues/110
- https://github.com/sphinx-gallery/sphinx-gallery/pull/114

larsoner:
I don't see why we couldn't at least try to support this. @prisae do you want to try implementing it and see how difficult it is?

prisae:
I had a quick try @larsoner, but without success. It is the first time that I looked at the sphinx-gallery code, and I am lost in the jungle of all the directories, where are the static files, where should what be stored, etc. The adjustment has to be in the function `save_thumbnail` of the file `gen_rst.py`. However, I couldn't figure it out.

I guess the image should be stored in `mymodule/docs/_static/`, right?

larsoner:
> It is the first time that I looked at the sphinx-gallery code, and I am lost in the jungle of all the directories, where are the static files, where should what be stored, etc.

It is part our mess, part Sphinx's mess (understanding and making use of how Sphinx organizes things). The combination yields something that is a pain to understand, at least at first :)

So feel free to try as long as you want and I can take over at the point you lose interest insuffering through it...

> I guess the image should be stored in mymodule/docs/_static/, right?

[...]

PR #609, opened 2020-02-13
ENH: New file-based conf-parameter thumbnail_path

prisae:
Adds the possibility to use an arbitrary image as thumbnail on a per-file basis by providing the path, e.g.,
```
# sphinx_gallery_thumbnail_path = '_static/yourimage.png'
```

Addresses #607


larsoner:
To test this I would make some change to an example in `sphinx_gallery/tests/tinybuild` and then assert that the `<img` HTML tag shows up properly somewhere in a `sphinx_gallery/tests/test_full.py` test (maybe a new one reusing a fixture or just in an existing one)

prisae:
I added it to the docs and as additional example. I also replicated the test for `thumbnail_number`. However, I am not sure if this is enough. But I am not sure I understood your input regarding the tests...

Also, the tests seem to fail, but not sure if it is actually related to this PR.

larsoner:
@prisae pushed a commit with a variant of the test I was talking about (plus some fixes to get some other tests happy again). See if you're happy. If so I'll take a look at the documentation and hopefully merge!

prisae:
I just tested it on my local gallery, and it works all fine with your changes, so LGTM if you think this implementation is stable enough. The only thing I was unsure is about the path, relative to what. I chose relative to `gallery_conf['src_dir']`, I guess that is OK.

Thanks for your help @larsoner !

larsoner:
Yes I think that's the sensible choice. And given that things are green here and not on `master` let's go ahead and merge. @lucyleeow feel free to look and comment if you have time and we can always do a follow-up PR

PR #625, opened 2020-03-19
ENH Add last cell config

### LINKED ISSUE #605: last_notebook_cell
Could a `last_notebook_cell` option be added much like the `first_notebook_cell` option for adding snippets that we want executed at the end of the downloaded notebook? This would be used to perhaps report on the user's environment with a tool like [`scooby`](https://github.com/banesullivan/scooby):

```py
sphinx_gallery_conf = {
    ...
    'last_notebook_cell': ("# Report on the environment specs\n"
                            "import scooby\n",
                            "scooby.Report()",)
}
```

With something like this, a user could download the example, run it, and encounter an error then immediately share the environment report... I imagine this will be useful for PyVista (`pv.Report`) and `mne-python` (`mne.sys_info()`)

lucyleeow:
closes #605

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -37,10 +37,10 @@ file:
 - ``min_reported_time`` (:ref:`min_reported_time`)
 - ``show_memory`` (:ref:`show_memory`)
 - ``binder`` (:ref:`binder_links`)
-- ``first_notebook_cell`` (:ref:`first_notebook_cell`)
+- ``first_notebook_cell`` and ``last_notebook_cell`` (:ref:`own_notebook_cell`)
 - ``junit`` (:ref:`junit_xml`)
 - ``log_level`` (:ref:`log_level`)
-- ``capture_repr``, ``ignore_repr_types`` (:ref:`capture_repr`)
+- ``capture_repr`` and ``ignore_repr_types`` (:ref:`capture_repr`)
 
 Some options can also be set or overridden on a file-by-file basis:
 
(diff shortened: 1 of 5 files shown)


lucyleeow:
@larsoner could you explain to me why `gallery_conf['first_notebook_cell'] = first_cell` is needed at the end of:

```
    # Ensure the first cell text is a string if we have it
    first_cell = gallery_conf.get("first_notebook_cell")
    if (not isinstance(first_cell, str)) and (first_cell is not None):
        raise ValueError("The 'first_notebook_cell' parameter must be type str"
                         "or None, found type %s" % type(first_cell))
    gallery_conf['first_notebook_cell'] = first_cell
```

larsoner:
Thanks @lucyleeow

lucyleeow:
Thanks @larsoner !

I'm still confused about `gallery_conf['first_notebook_cell'] = first_cell` after `first_cell = gallery_conf.get("first_notebook_cell")`...? Is it needed?

larsoner:
I looked to respond but didn't see the comment here in the PR so just skipped it (force push maybe?)

I would keep those so that `'first_notebook_cell'` is a valid key, but then later in the code you shouldn't need `.get` on the dict access.

Also we might not even need the `.get` in the first place if the default-filling code does the job of ensuring the key is in the dict.

lucyleeow:
>  would keep those so that `'first_notebook_cell'` is a valid key,

but how would it ensure that it is a valid key? `first_cell` isn't changed. 
`first_cell = gallery_conf.get("first_notebook_cell")` followed by `gallery_conf['first_notebook_cell'] = first_cell` seems circular to me. What am I missing?
```
    # Ensure the first cell text is a string if we have it
    first_cell = gallery_conf.get("first_notebook_cell")
    if (not isinstance(first_cell, str)) and (first_cell is not None):
        raise ValueError("The 'first_notebook_cell' parameter must be type str"
                         "or None, found type %s" % type(first_cell))
    gallery_conf['first_notebook_cell'] = first_cell
```

> Also we might not even need the `.get` in the first place if the default-filling code does the job of ensuring the key is in the dict.

Yes I thought that was redundant/overly cautious. I can remove the `get`?

lucyleeow:
(I mean the `raise ValueError` doesn't change `first_cell`)

larsoner:
You can hopefully remove it both places

lucyleeow:
new PR?

larsoner:
Sure if you want, or we can tackle it later as part of another PR

lucyleeow:
I'll make a new PR! Then we can isolate if it breaks things haha!

PR #632, opened 2020-03-28
Fix missing space in error message.

anntzer:
Currently the error render as "... of typestr or None..."


lucyleeow:
Thanks @anntzer!

Issue #634, opened 2020-03-31
TypeError when creating links from gallery to documentation


david-zwicker:
I'm using the excellent sphinx-gallery extension in my python project [`py-pde`](https://github.com/zwicker-group/py-pde). The [documentation is hosted on readthedocs](https://py-pde.readthedocs.io) and things worked perfectly until very recently. Since a few days every documentation build fails with the following traceback:

```
[...]
copying static files... done
copying readthedocs static files... done
copying searchtools... done
copying extra files... done
dumping search index in English (code: en) ... done
dumping object inventory... done
build succeeded, 3 warnings.

The HTML pages are in _build/html.

Sphinx-gallery successfully executed 0 out of 0 files subselected by:

    gallery_conf["filename_pattern"] = '/'
    gallery_conf["ignore_pattern"]   = '.*movie.*'

after excluding 16 files that had previously been run (based on MD5).

embedding documentation hyperlinks...
embedding documentation hyperlinks for examples_gallery... [  5%] simple.html

Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/py-pde/envs/latest/lib/python3.7/site-packages/sphinx/cmd/build.py", line 304, in build_main
    app.build(args.force_all, filenames)
  File "/home/docs/checkouts/readthedocs.org/user_builds/py-pde/envs/latest/lib/python3.7/site-packages/sphinx/application.py", line 369, in build
    self.emit('build-finished', None)
  File "/home/docs/checkouts/readthedocs.org/user_builds/py-pde/envs/latest/lib/python3.7/site-packages/sphinx/application.py", line 510, in emit
    return self.events.emit(event, self, *args)
  File "/home/docs/checkouts/readthedocs.org/user_builds/py-pde/envs/latest/lib/python3.7/site-packages/sphinx/events.py", line 80, in emit
    results.append(callback(*args))
  File "/home/docs/checkouts/readthedocs.org/user_builds/py-pde/envs/latest/lib/python3.7/site-packages/sphinx_gallery/docs_resolv.py", line 442, in embed_code_links
    _embed_code_links(app, gallery_conf, gallery_dir)
  File "/home/docs/checkouts/readthedocs.org/user_builds/py-pde/envs/latest/lib/python3.7/site-packages/sphinx_gallery/docs_resolv.py", line 340, in _embed_code_links
    for modname in (cobj['module_short'], cobj['module']):
TypeError: string indices must be integers

[...]

PR #637, opened 2020-03-31
WIP: Double check types

larsoner:
Let's check our types for #634


larsoner:
Closing as things are green here and we don't really want this code, but will reopen if there is more to test

PR #645, opened 2020-04-17
BUG Add CSS files to setup.py

lucyleeow:
Closes #644 

Adds split CSS files (#599) to `setup.py`

Should we do a patch release @larsoner...?

PR #648, opened 2020-04-18
BUG: Fix CSS includes and add manifest check in CI

### LINKED ISSUE #644: css not included in wheels?
Hi, 

I'm seeing a `gallery-dataframe.css` and `gallery-binder.css` not found error on my docs. It looks like `_static` only contains:
```
broken_example.png gallery.css        no_image.png
```
while on master the above css files are included. 

I think this is due to setup.py not listing all the expected static files to be included in the wheels.

larsoner:
Feel free to review/merge @lucyleeow 

EDIT: And yes we should do a patch release after this, ensuring that the CSS files are actually in the wheels and `sdist` (but they really should be after this)...

Closes #644
Closes #645


lucyleeow:
Thanks @larsoner, good pick up on the flake8! :joy:

PR #653, opened 2020-04-24
DOC: Fix PNGScraper example

denkii:
The `PNGScraper` example has some syntax errors that are fixed here. Also, the searched `pngs` should be checked within `path_current_example` instead of `os.getcwd()` for most use-cases.


choldgraf:
good catch, thanks very much 👍

denkii:
Also, is there any reason why `sphinx-gallery` doesn't include this within the package? This is particularly useful for us when dealing with `graphviz` visualisations and any other more difficult visualisations that just get written to disk.

choldgraf:
I think we basically started off by saying "let's see what community folks create with this extension point, and provide documentation for how to use it, and later on assess whether to support some scrapers natively". If you think it should make it into core sphinx-gallery, perhaps open an issue to discuss?

PR #656, opened 2020-04-25
ENH: Use optipng when requested

### LINKED ISSUE #655: thumbnails cause heavy gallery pages and long loading time
I would like to start a discussion regarding the loading time of galleries.

I opened here a [PR for scikit-image](https://github.com/scikit-image/scikit-image/pull/4613) showing that thumbnails are responsible for long loading time and large bandwidth consumption. Comments show that scikit-learn and matplotlib are also affected by this issue.

I made some tests and the mitigation I found on the user side is to reduce slightly the default resolution of the thumbnails. As our galleries are expected to grow, I think that a longer term solution would be appreciated.

I read the sphinx-gallery code and saw that you are using your own thumbnail generator, relaying on pillow. So far, thumbnails are saved in png (because our plots with matplotlib are so), which is particularly good for rendering edges and fonts. The drawback is the file size as explained above. I made some tests with pillow and I couldn't reach better performances while keeping a png format.

Have you ever considered to switch to jpg? Jpg has some advantages such as a good compression level (with loss) and also [progressive loading](https://www.infoq.com/news/2019/09/lazy-loading-http-range-requests/). I would like also to know whether svg could provide better performances for plots. 

I'm not a master in web optimization, so I'm open to other opinions.
Thank you!

CC @tacaswell @emmanuelle

larsoner:
With this we ~~automatically~~ use `optipng` if it's available EDIT: and the user requests it, and emit a `logger.warning` if it's not.

Intentionally installed on some CI runs but not others to ensure we don't break things for people. @GaelVaroquaux WDYT?

Closes #655

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -68,7 +68,9 @@ Sphinx-Gallery has also support for packages like:
 * Seaborn
 * Mayavi
 
-For much of this functionality, you will need `pillow`.
+For much of this functionality, you will need ``pillow``. We also recommend
+installing system ``optipng`` binaries to reduce the file sizes of the
+generated PNG files.
 
 Install as a Sphinx-gallery developer
 -------------------------------------
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -31,6 +31,7 @@ file:
 - ``plot_gallery`` (:ref:`without_execution`)
 - ``image_scrapers`` (and the deprecated ``find_mayavi_figures``)
   (:ref:`image_scrapers`)
+- ``optimize_images`` (:ref:`optimize_images`)
 - ``reset_modules`` (:ref:`reset_modules`)
 - ``abort_on_example_error`` (:ref:`abort_on_first`)
 - ``expected_failing_examples`` (:ref:`dont_fail_exit`)
@@ -854,6 +855,35 @@ a default::
 
 The highest precedence is always given to the `-D` flag of the
 ``sphinx-build`` command.
+
+
+.. _optimize_images:
+
+Optimizing images
+=================
+
+When writing PNG files (the default scraper format), sphinx-gallery can be
+configured to use ``optipng`` to optimize the PNG file sizes. Typically this
+yields roughly a 50% reduction in file sizes. However, it can increase build
+time. The allowed values are ``'images'`` and ``'thumbnails'``, or a
+tuple/list, such as (to optimize both)::
+
+    sphinx_gallery_conf = {
+        ...
+        'optimize_images': ('images', 'thumbnails'),
+    }
+
+The default is ``()`` (no optimization) and a warning will be emitted if
+optimization is requested but not available. You can also pass additional
+command-line options (starting with ``'-'``), for example to optimize less
+but speed up the build time you could do::
+
+    sphinx_gallery_conf = {
+        ...
+        'optimize_images': ('images', 'thumbnails', '-o1'),
+    }
+
+See ``$ optipng --help`` for a complete list of options.
 
 
 .. _image_scrapers:
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -31,7 +31,7 @@ file:
 - ``plot_gallery`` (:ref:`without_execution`)
 - ``image_scrapers`` (and the deprecated ``find_mayavi_figures``)
   (:ref:`image_scrapers`)
-- ``optimize_images`` (:ref:`optimize_images`)
+- ``compress_images`` (:ref:`compress_images`)
 - ``reset_modules`` (:ref:`reset_modules`)
 - ``abort_on_example_error`` (:ref:`abort_on_first`)
 - ``expected_failing_examples`` (:ref:`dont_fail_exit`)
@@ -857,10 +857,10 @@ The highest precedence is always given to the `-D` flag of the
 ``sphinx-build`` command.
 
 
-.. _optimize_images:
+.. _compress_images:
 
-Optimizing images
-=================
+Compressing images
+==================
 
 When writing PNG files (the default scraper format), sphinx-gallery can be
 configured to use ``optipng`` to optimize the PNG file sizes. Typically this
@@ -870,7 +870,7 @@ tuple/list, such as (to optimize both)::
 
     sphinx_gallery_conf = {
         ...
-        'optimize_images': ('images', 'thumbnails'),
+        'compress_images': ('images', 'thumbnails'),
     }
 
 The default is ``()`` (no optimization) and a warning will be emitted if
@@ -880,7 +880,7 @@ but speed up the build time you could do::
 
     sphinx_gallery_conf = {
         ...
-        'optimize_images': ('images', 'thumbnails', '-o1'),
+        'compress_images': ('images', 'thumbnails', '-o1'),
     }
 
 See ``$ optipng --help`` for a complete list of options.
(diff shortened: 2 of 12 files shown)


doc/configuration.rst
```diff
@@ -858,6 +858,13 @@ The default value is ``'image_scrapers': ('matplotlib',)`` which only scrapes
 Matplotlib images. Note that this includes any images produced by packages that
 are based on Matplotlib, for example Seaborn or Yellowbrick.
 
+.. note::
+   When writing PNG files (the default scraper format), sphinx-gallery will
+   attempt to use system calls to ``optipng`` to optimize the PNG file sizes,
+   so it's recommended to ensure it's callable (on the PATH) wher building
```
lucyleeow, line 864:
```suggestion
   so it's recommended to ensure it's callable (on the PATH) when building
```

sciunto:
Many thanks @larsoner for this PR. I opened an issue on the scikit-image repository that provides a test on our gallery I just performed.

In short: a gain of about a factor 2 on the thumbnail size! but a very long built time. I'm wondering if it wouldn't be better to optimize only thumbnails, or at least to provide options to set either None, Thumbnails, All.

larsoner:
@sciunto on the latest commit you can do `'optimize_images': ('images', 'thumbnails'),` or just one string or the other. Given that there are performance considerations, I made the default `()` (no optimization) and now emit a warning if optimization is requested but optipng is now available.

sciunto:
Cool stuffs. Have you played with the optimization level proposed by optipng? I have not so far. I tried pngquant, which has no level to tune, it is not as good avec optipng, but does fairly well (see the file size on skimage's issue), and the runtime is significantly improved. I guess we can reach this balance as well with optipng

From optipng, for information (default is -o2):
```Optimization levels:
    -o0         <=>     -o1 -nx -nz                             (0 or 1 trials)
    -o1         <=>     -zc9 -zm8 -zs0 -f0                      (1 trial)
                (or...) -zc9 -zm8 -zs1 -f5                      (1 trial)
    -o2         <=>     -zc9 -zm8 -zs0-3 -f0,5                  (8 trials)
    -o3         <=>     -zc9 -zm8-9 -zs0-3 -f0,5                (16 trials)
    -o4         <=>     -zc9 -zm8 -zs0-3 -f0-5                  (24 trials)
    -o5         <=>     -zc9 -zm8-9 -zs0-3 -f0-5                (48 trials)
    -o6         <=>     -zc1-9 -zm8 -zs0-3 -f0-5                (120 trials)
    -o7         <=>     -zc1-9 -zm8-9 -zs0-3 -f0-5              (240 trials)
    -o7 -zm1-9  <=>     -zc1-9 -zm1-9 -zs0-3 -f0-5              (1080 trials)
Notes:
    The combination for -o1 is chosen heuristically.
    Exhaustive combinations such as "-o7 -zm1-9" are not generally recommended.
Examples:
    optipng file.png                                            (default speed)
    optipng -o5 file.png                                        (slow)
    optipng -o7 file.png                                        (very slow)
```

larsoner:
No I just picked `-o7` because this is what CPython used for their static images. Feel free to test recommend a default. We can also allow people to pass a list of arguments to pass at the end of the tuple, like `'optimize_images': ('thumbnails', 'images', '-zc9', '-zm8', '-zs0', '-f0')` or so

GaelVaroquaux:
One comment on performance: can we make it so that only the thumbnails
that are generated in a specific run are optimized: sphinx-gallery
rebuilds only small number of thumbnails when it is reran.

I had a quick look at the code, and I wasn't sure if the PR was already
clever on this or not.

larsoner:
In theory it should already be that clever since it should only regenerate the images and thumbnails it needs to, and they're optimized at the time of creation

larsoner:
Testing with `time make` plus `du -hs _build/html/_images/` for `tinybuild`:

- off: 4.5 sec, 240 kB
- `-o0`: 4.5 sec, 240 kB
- `-o1 `: 4.7 sec, 216 kB
- `-o2 `: 5.4 sec, 212 kB
- `-o7 `: 14.9 sec, 212 kB

Testing with Sphinx-gallery docs:

- off: 16 sec, 1.5 MB
- `-o0`: 16 sec, 1.5 MB
- `-o1`: 18 sec, 1.3 MB
- `-o2`: 22 sec, 1.2 MB
- `-o7`: 24 sec, 1.2 MB

Given that the optipng default (`-o2`) works pretty well, I'd say we should just not explicitly change it. All optimization is off by default anyway (it's opt-in), and I've pushed a commit to let people configure with their own command-line options like `'optimize_images': ('images', 'thumbnails', '-o1')` if they want. This should be flexible enough (hopefully) that each org can tailor things to their own needs.

I think this should be good to go.

larsoner:
Docs here:

https://1544-25860190-gh.circle-artifacts.com/0/rtd_html/configuration.html#optimizing-images

larsoner:
I'm fine with `compress_images`.

Eventually if we support other modes, `str/list` can mean "for png" and to get other formats we can do `dict(png=..., svg=...)`, so the current design is forward compatible I think. But the critical one to get is PNG since it's the one we use natively and by default, so should have broadest impact.

larsoner:
Looking into the sphinx dev failure now...

larsoner:
I think we can safely ignore the sphinx-dev error and merge this if you're happy with it otherwise @lucyleeow

README.rst
```diff
@@ -68,7 +68,9 @@ Sphinx-Gallery has also support for packages like:
 * Seaborn
 * Mayavi
 
-For much of this functionality, you will need `pillow`.
+For much of this functionality, you will need ``pillow``. We also recommend
+installing system ``optipng`` binaries to reduce the file sizes of the
+generated PNG files.
```
lucyleeow, line 73: nitpick - 'This reduces the loading time of large galleries'.

doc/configuration.rst
```diff
@@ -856,6 +857,35 @@ The highest precedence is always given to the `-D` flag of the
...
+.. _compress_images:
+
+Compressing images
+==================
+
+When writing PNG files (the default scraper format), sphinx-gallery can be
+configured to use ``optipng`` to optimize the PNG file sizes. Typically this
+yields roughly a 50% reduction in file sizes. However, it can increase build
+time. The allowed values are ``'images'`` and ``'thumbnails'``, or a
+tuple/list, such as (to optimize both)::
```
lucyleeow, line 869:
```suggestion
tuple/list (to optimize both), such as::
```

doc/configuration.rst
```diff
@@ -856,6 +857,35 @@ The highest precedence is always given to the `-D` flag of the
...
+
+    sphinx_gallery_conf = {
+        ...
+        'compress_images': ('images', 'thumbnails'),
+    }
+
+The default is ``()`` (no optimization) and a warning will be emitted if
+optimization is requested but not available. You can also pass additional
```
lucyleeow, line 877: 'but ``optipng`` is not available' ?

doc/configuration.rst
```diff
@@ -856,6 +857,35 @@ The highest precedence is always given to the `-D` flag of the
...
 
 
+.. _compress_images:
+
+Compressing images
+==================
+
+When writing PNG files (the default scraper format), sphinx-gallery can be
+configured to use ``optipng`` to optimize the PNG file sizes. Typically this
+yields roughly a 50% reduction in file sizes. However, it can increase build
```
lucyleeow, line 867:
```suggestion
yields roughly a 50% reduction in file sizes, thus reducing the loading time of galleries. However, it can increase build
```

sciunto:
Thank you!

PR #661, opened 2020-04-28
ENH Support _repr_mimebundle_

### LINKED ISSUE #660: Support for _repr_mimebundle_
The Jupyter Python kernel now supports a new repr function called `_repr_mimebundle_` which can be used to provide several representations of the same object in a dictionary. Unlike `_repr_[html|json|svg|png...]_` which are only a limited list of supported mimetype.

As an increasing number of packages use `_repr_mimebundle_`  it may be interesting for jupyter-sphinx to support it.

We could start with checking if the mimebundle includes a `text/html` entry and use that in the same way as `_repr_html_` is used...

Note: jupyter-sphinx's support of `_repr_mimebundle_` was already mentioned [here](https://github.com/scikit-learn/scikit-learn/pull/14180) and [here](https://github.com/sphinx-gallery/sphinx-gallery/issues/396).

lucyleeow:
closes #660 

Note CI's will be red for Sphinx dev until https://github.com/sphinx-doc/sphinx/issues/7570 fixed.

### DIFF
--- a/sphinx_gallery/tests/test_gen_rst.py
+++ b/sphinx_gallery/tests/test_gen_rst.py
@@ -604,8 +604,8 @@ def test_empty_output_box(gallery_conf, script_vars):
     assert output.isspace()
 
 
-code_repr_only = """
-class repr_only_class():
+CODE_REPR = """
+class ReprOnlyClass():
 
     def __init__(self):
         pass
@@ -613,12 +613,12 @@ class repr_only_class():
     def __repr__(self):
         return "This is the __repr__"
 
-class_inst = repr_only_class()
+class_inst = ReprOnlyClass()
 class_inst
 """
 
-code_repr_and_html = """
-class repr_and_html_class():
+CODE_REPR_HTML = """
+class ReprAndHtmlClass():
     def __init__(self):
         pass
 
@@ -628,14 +628,14 @@ class repr_and_html_class():
     def _repr_html_(self):
         return "<div> This is the _repr_html_ div </div>"
 
-class_inst = repr_and_html_class()
+class_inst = ReprAndHtmlClass()
 class_inst
 """
 
-code_print_and_repr_and_html = """
+CODE_PRINT_REPR_HTML = """
 print("print statement")
 
-class repr_and_html_class():
+class ReprAndHtmlClass():
     def __init__(self):
         pass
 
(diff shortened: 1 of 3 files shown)


lucyleeow:
Ready for review @SylvainCorlay. Note CI's will be red for Sphinx dev build until sphinx-doc/sphinx#7570 fixed. Thanks!

Just adding some documentation now.

It would be more difficult but possible to support image file formats if there is a desire to in future. On that note @larsoner do you think `execute_code_block` needs refactoring?

lucyleeow:
cc @GaelVaroquaux

lucyleeow:
Currently only 'text/html' and 'text/plain' are supported. 'text/html' is used in preference to 'text/plain'.

If only unsupported file formats are present in the mime-bundle SG will try to capture other reprs in the config `capture_repr`. If there are not other reprs, nothing is captured.

SylvainCorlay:
Thanks. I had a glance at the code and it looks fine. I think that one thing to check is that it handles precedence between `_repr_html_` and `_repr_mimebundle_`  in the same way as in the notebook (or VSCode) when both are defined.

lucyleeow:
How does the notebook handle the precedence? It depends on the config here. Whichever appears first has precendece, e.g., `capture_repr: ('_repr_mimebundle_', '__repr__')` would give `_repr_mimebundle_` over `_repr_html_`

SylvainCorlay:
> How does the notebook handle the precedence?

I think you got it right. I just meant that it might be able to double check that both sphinx-gallery and e.g. notebook show the same thing when both are defined.

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -649,6 +649,52 @@ def _repr_html_(self):
 class_inst
 """
 
+code_mime_plain = """
+class mime_plain():
```
GaelVaroquaux, line 653: It's a nitpick, but I'm finding it ugly that there is a class that is not CamelCass.

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -649,6 +649,52 @@ def _repr_html_(self):
...
+        return {
+            "text/plain": "This is mime plain"
+        }
+
+class_inst = mime_plain()
+class_inst
+"""
+
+code_mime_html_plain = """
+class mime_html_plain():
```
GaelVaroquaux, line 667: Likewise: non CamelCase class

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -649,6 +649,52 @@ def _repr_html_(self):
...
+            "text/html": "<div> This is the _repr_html_ div </div>",
+            "text/plain": "This is mime plain"
+        }
+
+class_inst = mime_html_plain()
+class_inst
+"""
+
+code_mime_unsupported = """
+class mime_unsupported():
```
GaelVaroquaux, line 682: Likewise: non CamelCase class

PR #663, opened 2020-04-28
ENH More informative 'alt' attribute for images

### LINKED ISSUE #538: More informative "alt" attribute for image tags
The alternative text for generated HTML `img`  tags could be more informative. Currently it's the filename of the png file.

Instead one could use the title of the example or even better, the "title" attribute of the underlying matplotlib figure when available (and not empty).

lucyleeow:
closes #538 

For matplotlib figures, checks for 'suptitle', then tries `get_title()` for all axes of a figure.
If no text found, uses the file name. File name also used by default, for all non-matplotlib figures.

Currently only works for single figures. Will rework `matplotlib_scraper` so `figure_rst` is run on every loop so it works for 'HLIST_IMAGE_TEMPLATE'. Suggestions welcome

cc @ogrisel

### DIFF
--- a/sphinx_gallery/scrapers.py
+++ b/sphinx_gallery/scrapers.py
@@ -14,6 +14,7 @@ live in modules that will support them (e.g., PyVista, Plotly).
 import os
 import sys
 import re
+from textwrap import indent
 
 from .utils import scale_image
 
@@ -90,8 +91,8 @@ def matplotlib_scraper(block, block_vars, gallery_conf, **kwargs):
     """
     matplotlib, plt = _import_matplotlib()
     image_path_iterator = block_vars['image_path_iterator']
-    image_paths = list()
     fig_titles = ''
+    image_rsts = []
     for fig_num, image_path in zip(plt.get_fignums(), image_path_iterator):
         if 'format' in kwargs:
             image_path = '%s.%s' % (os.path.splitext(image_path)[0],
@@ -112,9 +113,19 @@ def matplotlib_scraper(block, block_vars, gallery_conf, **kwargs):
                     attr not in kwargs:
                 these_kwargs[attr] = fig_attr
         fig.savefig(image_path, **these_kwargs)
-        image_paths.append(image_path)
+        image_rsts.append(
+            figure_rst([image_path], gallery_conf['src_dir'], fig_titles))
     plt.close('all')
-    return figure_rst(image_paths, gallery_conf['src_dir'], fig_titles)
+    if len(image_rsts) == 1:
+        rst = image_rsts[0]
+    else:
+        image_rsts = [re.sub(r'sphx-glr-single-img',
+                             'sphx-glr-multi-img',
+                             image) for image in image_rsts]
+        image_rsts = [HLIST_IMAGE_MATPLOTLIB + indent(image, u' ' * 6)
+                      for image in image_rsts]
+        rst = HLIST_HEADER + ''.join(image_rsts)
+    return rst
 
 
 def mayavi_scraper(block, block_vars, gallery_conf):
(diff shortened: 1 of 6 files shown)


lucyleeow:
Note matplotlib's `plt.title` can be in different locs (and does not overwrite if different locs are used: https://matplotlib.org/3.1.1/gallery/text_labels_and_annotations/titles_demo.html#sphx-glr-gallery-text-labels-and-annotations-titles-demo-py). Thus we cycle through different locs.

I realised that we must keep `figure_rst` backwards compatible for user generated scrapers, thus the complicated workaround for multi-figures in `matplotlib_scraper`

Read for review @larsoner

sphinx_gallery/scrapers.py
```diff
@@ -47,6 +49,22 @@ def _import_matplotlib():
     return matplotlib, plt
 
 
+def _matplotlib_fig_titles(fig):
+    titles = []
+    # get supertitle if exists
+    if fig._suptitle:
+        titles.append(fig._suptitle.get_text())
```
ogrisel, line 56: Since this is private API, it might be safe to do:

```suggestion
    suptitle = getattr(fig, "_suptitle", None)
    if suptitle is not None:
        titles.append(suptitle.get_text())
```

sphinx_gallery/scrapers.py
```diff
@@ -47,6 +49,22 @@ def _import_matplotlib():
...
+    # get titles from all axes, for all locs
+    title_locs = ['left', 'center', 'right']
+    for ax in fig.axes:
+        for loc in title_locs:
+            text = ax.get_title(loc=loc)
+            if text:
+                titles.append(text)
```
larsoner, line 63: and `break` right?

sphinx_gallery/scrapers.py
```diff
@@ -73,14 +91,17 @@ def matplotlib_scraper(block, block_vars, gallery_conf, **kwargs):
     """
     matplotlib, plt = _import_matplotlib()
     image_path_iterator = block_vars['image_path_iterator']
-    image_paths = list()
+    fig_titles = ''
```
larsoner, line 94: Shouldn't need this line, `fig_titles` is only used inside the loop and it's set on line 104

sphinx_gallery/scrapers.py
larsoner: Ignore, meant to delete this comment!

sphinx_gallery/scrapers.py
lucyleeow: thanks! I couldn't find another way to get the suptitle.

lucyleeow:
@ogrisel added titles here: https://1453-25860190-gh.circle-artifacts.com/0/rtd_html/auto_examples/plot_1_exp.html#sphx-glr-auto-examples-plot-1-exp-py

choldgraf:
got a travis failure here: https://travis-ci.org/github/sphinx-gallery/sphinx-gallery/jobs/680921561#L1596

but other than that, this is a good idea!

ogrisel:
> @ogrisel added titles here: https://1453-25860190-gh.circle-artifacts.com/0/rtd_html/auto_examples/plot_1_exp.html#sphx-glr-auto-examples-plot-1-exp-py

This is a great improvement @lucyleeow ! It would be even greater to similarly change the alt text of the thumbnails of the index page of the gallery:

https://1453-25860190-gh.circle-artifacts.com/0/rtd_html/auto_examples/index.html

lucyleeow:
@choldgraf unfortunately that is due to https://github.com/sphinx-doc/sphinx/issues/7570

@ogrisel sure. I think in a different PR. Does the title of the example sound reasonable?

ogrisel:
> @ogrisel sure. I think in a different PR. Does the title of the example sound reasonable?

Yes on both points.

PR #668, opened 2020-04-30
ENH Add alt attribute for thumbnails

### LINKED ISSUE #664: ENH More informative 'alt' attribute for thumbnails in index
Following from #663, make the alternative text ('alt' attribute of image in rst) for thumbnail images in the index.rst of generated galleries to be the title of the example file.

cc @ogrisel

lucyleeow:
closes #664

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -336,7 +336,6 @@ def generate_dir_rst(src_dir, target_dir, gallery_conf, seen_backrefs):
         'generating gallery for %s... ' % build_target_dir,
         length=len(sorted_listdir))
     for fname in iterator:
-        print('for {} generate file run'.format(fname))
         intro, title, cost = generate_file_rst(
             fname, target_dir, src_dir, gallery_conf, seen_backrefs)
         src_file = os.path.normpath(os.path.join(src_dir, fname))
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -336,12 +336,12 @@ def generate_dir_rst(src_dir, target_dir, gallery_conf, seen_backrefs):
         'generating gallery for %s... ' % build_target_dir,
         length=len(sorted_listdir))
     for fname in iterator:
-        intro, cost = generate_file_rst(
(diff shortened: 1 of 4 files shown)


lucyleeow:
@larsoner I don't understand the cause of the test failures. I seems that `test_junit` and `test_rebuild` are failing during setup but I don't understand why. tinybuild builds fine when I check.
The error is
```
ValueError: not enough values to unpack (expected 3, got 2)
```
but there all 3 output variables (intro, title, cost) are present.

larsoner:
https://github.com/sphinx-gallery/sphinx-gallery/pull/668/files#diff-0cc5ccb98f5854a18c184ad4644a3509R759

beware the multiple `return` code path...

lucyleeow:
@larsoner the title cannot contain `\n` so we won't have the same problem as in #669. It can contain non-word characters but will they be a problem?
Also do you think the test is adequate?
And thanks for the tip!

sphinx_gallery/gen_rst.py
```diff
@@ -336,12 +336,13 @@ def generate_dir_rst(src_dir, target_dir, gallery_conf, seen_backrefs):
         'generating gallery for %s... ' % build_target_dir,
         length=len(sorted_listdir))
     for fname in iterator:
-        intro, cost = generate_file_rst(
+        print('for {} generate file run'.format(fname))
```
larsoner, line 339: debugging cruft?

larsoner:
>  It can contain non-word characters but will they be a problem?

Maybe but we can wait until someone has a problem :)

sphinx_gallery/gen_rst.py
lucyleeow: :joy: oh no, my print statement debugging exposed!

lucyleeow:
@larsoner added a check in html and rst for index and backref thumbnail alt. Also added a html test to image alts.

larsoner:
Thanks @lucyleeow

lucyleeow:
cc @ogrisel see https://1473-25860190-gh.circle-artifacts.com/0/rtd_html/auto_examples/index.html for thumbnail alt example :)

PR #676, opened 2020-05-04
Update manifest for changes to check-manifest

### LINKED ISSUE #675: check-manifest causing master to fail
Master failing due to:

```
lists of files in version control and sdist do not match!

missing from VCS:

  sphinx_gallery/tests/testconfs/src/__pycache__/plot_1.cpython-37-pytest-5.3.5.pyc

  sphinx_gallery/tests/testconfs/src/__pycache__/plot_2.cpython-37-pytest-5.3.5.pyc

  sphinx_gallery/tests/testconfs/src/__pycache__/plot_3.cpython-37-pytest-5.3.5.pyc
```
I didn't realise these files are in sdist...??

lucyleeow:
closes #675

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 # Byte-compiled / optimized / DLL files
 __pycache__/
-sphinx_gallery/tests/testconfs/src/__pycache__/*
 *.py[cod]
 .coverage*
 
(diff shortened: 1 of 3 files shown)


lucyleeow:
I don't think the update to check-manifest should have changed the behaviour of the `MANIFEST.in` commands, only the `--ignore` so I don't see why it's failing now

lucyleeow:
Whoops the message is missing from VCS!

PR #678, opened 2020-05-04
FIX: Fix sys.stdout patching

### LINKED ISSUE #672: Output of logger is not captured if the logger is created in a different cell
Logger output is not captured if the logger is not created within the same cell. Consider the following example:

```python
"""
Logging test
============
"""

import logging
import sys


def create_logger(name=None):
    logger = logging.getLogger(name=name)
    logger.setLevel(logging.INFO)

    sh = logging.StreamHandler(sys.stdout)
    sh.setLevel(logging.INFO)
    logger.addHandler(sh)

    return logger


logger = create_logger("first_logger")
logger.info("I'm in the same cell that I was created in.")


logger.info("I'm **not** in the same cell that I was created in.")

logger = create_logger("second_logger")
logger.info("I'm again in the same cell that I was created in.")
```

Running `make html` generates the following extract in the corresponding`.rst` file

```rst
.. code-block:: default
   :lineno-start: 26


    logger = create_logger("first_logger")
    logger.info("I'm in the same cell that I was created in.")

.. rst-class:: sphx-glr-script-out

 Out:

 .. code-block:: none

    I'm in the same cell that I was created in.

.. code-block:: default
   :lineno-start: 32


    logger.info("I'm **not** in the same cell that I was created in.")

.. code-block:: default
   :lineno-start: 36


    logger = create_logger("second_logger")
    logger.info("I'm again in the same cell that I was created in.")

.. rst-class:: sphx-glr-script-out

 Out:

 .. code-block:: none

    I'm again in the same cell that I was created in.
```
(Whitespace removed to enhance readability).

As you can see, the second info is missing. If I run this file manually with python all three lines are printed on `stdout`.

larsoner:
This restructures our patching of `sys.stdout`/`stderr` to occur on a per-script basis, with some gymnastics to deal with our testing infrastructure calling things like `execute_code_block` outside a standard context where `with _LoggingTee` would otherwise have already set `sys.stdout`.

Closes #672.

@pmeier can you see if this fixes things for you?

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -53,18 +53,30 @@ logger = sphinx_compatibility.getLogger('sphinx-gallery')
 ###############################################################################
 
 
-class LoggingTee(object):
-    """A tee object to redirect streams to the logger"""
+class _LoggingTee(object):
+    """A tee object to redirect streams to the logger."""
 
-    def __init__(self, output_file, logger, src_filename):
-        self.output_file = output_file
+    def __init__(self, src_filename):
         self.logger = logger
         self.src_filename = src_filename
-        self.first_write = True
         self.logger_buffer = ''
+        self.set_std_and_reset_position()
+
+    def set_std_and_reset_position(self):
+        if not isinstance(sys.stdout, _LoggingTee):
+            self.origs = (sys.stdout, sys.stderr)
+        sys.stdout = sys.stderr = self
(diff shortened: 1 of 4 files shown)


pmeier:
Indeed, this works for my toy example as well as my real-world gallery. Thanks a lot for the quick fix!

larsoner:
I'll go ahead and merge since this fixes a user problem, and in theory should be more robust

PR #680, opened 2020-05-07
ENH Add cross ref label to template module.rst

### LINKED ISSUE #496: ENH: easy linking to "examples using my_function"
Is there a way to cross-link to the "Examples using..." mini-galleries in the API docs?  I know it auto-generates a URL anchor 
`module.name_of_function.html#examples-using-module-name-of-function` so I can add a normal URL link, but it would be cleaner if I could just do something like
```
:func:`module.name_of_function#minigallery`
```
or similar. Any suggestions (either how to do it in user code, or general approach / possible pitfalls to implementing it in sphinx-gallery)?

lucyleeow:
closes #496 

Add cross referencing label to `module.rst` to easily link to mini-galleries of module fun/classes.
Add test in tinybuild

### DIFF
--- a/doc/_templates/module.rst
+++ b/doc/_templates/module.rst
@@ -16,6 +16,8 @@
 
    .. autofunction:: {{ item }}
 
+   .. _sphx_glr_backref_{{fullname}}.{{item}}:
+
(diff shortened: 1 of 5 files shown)


larsoner:
Awesome, thanks @lucyleeow

PR #693, opened 2020-05-21
ENH: Use broader def of Animation

larsoner:
Matplotlib also has `ArtistAnimation`. It would be nice to have this for matplotlib (https://github.com/matplotlib/matplotlib/pull/17477), maybe in a 0.7.1 release sometime? Let's wait to see what else breaks in the meantime before cutting one, though.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ jobs:
     steps:
       # Get our data and merge with upstream
       - run: sudo apt-get update
-      - run: sudo apt-get --no-install-recommends install -y texlive texlive-latex-extra latexmk libxkbcommon-x11-0
+      - run: sudo apt-get --no-install-recommends install -y texlive texlive-latex-extra latexmk libxkbcommon-x11-0 optipng
       - checkout
       - run: echo $(git log -1 --pretty=%B) | tee gitlog.txt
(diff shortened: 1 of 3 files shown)


lucyleeow:
@larsoner how did you know the error:

> Exception occurred:
  File "/home/circleci/.local/lib/python3.7/site-packages/sphinx/setup_command.py", line 196, in run
    'caused by %s builder.' % app.builder.name)
distutils.errors.DistutilsExecError: caused by html builder.
The full traceback has been saved in /tmp/sphinx-err-q_gjzrqx.log, if you want to report the issue to the developers.

was caused by optipng? That error message is a bit cryptic.

larsoner:
Warnings are treated as errors, I ctrl-F'ed "warning" in the doc build log and found the warning

lucyleeow:
I see the `-W` is doing that in `python setup.py build_sphinx -nW --keep-going` but what does the exception above mean? Does it just mean that there was a warning somewhere?

larsoner:
> Does it just mean that there was a warning somewhere?

Yeah. Newer sphinx is supposed to tell you that there was a warning that was treated as an error that killed the build but apparently it didn't work in this case

lucyleeow:
Ah right. Thanks!

PR #696, opened 2020-05-26
Remove matplotlib agg warning

### LINKED ISSUE #694: Calling "plt.show()" raises an ugly warning
When a script has "plot.show()" in it, it results in the following ugly warning:
<pre>
UserWarning: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.
</pre>
Can we capture it and hide it?

lucyleeow:
closes #694

Did not realise it would be that simple, seems magical.

I didn't add anything to doc assuming no one would miss this warning.

### DIFF
--- a/sphinx_gallery/tests/test_full.py
+++ b/sphinx_gallery/tests/test_full.py
@@ -748,3 +748,14 @@ def test_minigallery_directive(sphinx_app):
             # Check for examples
             assert explicitorder_example.search(text) is not None
             assert filenamesortkey_example.search(text) is not None
+
+def test_matplotlib_warning_filter(sphinx_app):
+    """Test Matplotlib agg warning is removed."""
+    out_dir = sphinx_app.outdir
+    example_html = op.join(out_dir, 'auto_examples',
+                           'plot_matplotlib_alt.html')
+    with codecs.open(example_html, 'r', 'utf-8') as fid:
+        html = fid.read()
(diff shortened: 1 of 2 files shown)


lucyleeow:
Maybe if anyone raises this we can document how to remove warning filters...

sphinx_gallery/tests/test_full.py
```diff
@@ -748,3 +748,14 @@ def test_minigallery_directive(sphinx_app):
             # Check for examples
             assert explicitorder_example.search(text) is not None
             assert filenamesortkey_example.search(text) is not None
+
```
GaelVaroquaux, line 751: I think that PEP8 wants an extra line return here

sphinx_gallery/tests/test_full.py
```diff
@@ -748,3 +748,14 @@ def test_minigallery_directive(sphinx_app):
...
+def test_matplotlib_warning_filter(sphinx_app):
```
larsoner, line 752: PEP8, two newlines before func def

GaelVaroquaux:
This is green. Merging.

lucyleeow:
Hmm for some reason travis didn't seem to activate, or was that just me?

larsoner:
Looks like the status update just didn't come through, this is green thankfully

https://travis-ci.org/github/sphinx-gallery/sphinx-gallery/builds/691349478

PR #700, opened 2020-06-04
Enable html to be rendered on readthedocs

sdhiscocks:
Per comment: https://github.com/readthedocs/readthedocs.org/issues/4765#issuecomment-431927720

I was able to test this by monkey patching the change, and confirm that HTML now renders both on local build and on Read the Docs.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -163,11 +163,9 @@ SPHX_GLR_SIG = """\n
 """
 
 # Header used to include raw html
-html_header = """.. only:: builder_html or readthedocs
+html_header = """.. raw:: html
 
-    .. raw:: html
-
-{0}\n        <br />\n        <br />"""
+{0}\n    <br />\n    <br />"""
 
 
 def codestr2rst(codestr, lang='python', lineno=None):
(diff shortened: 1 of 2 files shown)


sphinx_gallery/gen_rst.py
```diff
@@ -163,7 +163,7 @@ def __exit__(self, type_, value, tb):
 """
 
 # Header used to include raw html
-html_header = """.. only:: builder_html
+html_header = """.. only:: builder_html or readthedocs
```
larsoner, line 166: I'm not sure if we even need this line given that it's wrapped in a `.. raw:: html`. It seems redundant, and thus simpler just to remove this `.. only::` line altogether

larsoner:
Some tests need to be updated to reflect this change:

https://travis-ci.org/github/sphinx-gallery/sphinx-gallery/jobs/694782511#L989

sdhiscocks:
> Some tests need to be updated to reflect this change:
> 
> https://travis-ci.org/github/sphinx-gallery/sphinx-gallery/jobs/694782511#L989

Thanks. Sorry, I should really have checked the tests first. Should be fixed now. Looks like remaining failures on TravisCI and CircleCI are unrelated to this change.

larsoner:
Thanks @sdhiscocks !

PR #702, opened 2020-06-05
DOC Update warning section in configuration.rst

lucyleeow:
Add ref label and update example since we remove the matplotlib warning by default.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -74,12 +74,12 @@ warning', you can add::
     import warnings
 
     warnings.filterwarnings("ignore", category=UserWarning,
-                            message='I don't want this warning.')
+                            message='Matplotlib is currently using agg, which is a'
+                                    ' non-GUI backend, so cannot show the figure.')
 
 to your ``conf.py`` file.
 
-Note that the Matplotlib warning: 'Matplotlib is currently using agg, which is
-a non-GUI backend, so cannot show the figure' is removed by default.
+Note that the above Matplotlib warning is removed by default.
 
 .. _multiple_galleries_config:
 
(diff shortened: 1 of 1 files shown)


doc/configuration.rst
```diff
@@ -61,18 +61,26 @@ And some things can be tweaked directly in CSS:
...
+documentation, you can use the package ``warnings`` in the ``conf.py`` file.
+For example, to remove the specific ``UserWarning``: 'I don't want this
+warning', you can add::
 
     import warnings
 
     warnings.filterwarnings("ignore", category=UserWarning,
-                            message='Matplotlib is currently using agg, which is a'
-                                    ' non-GUI backend, so cannot show the figure.')
+                            message='I don't want this warning.')
```
larsoner, line 77: This is less clear than before -- before it was clear that the `message` arg is somehow the thing you want to suppress, and it was a good example of one that you'd want to. I would either leave this, and say below that this is actually one that we already remove, or choose a more suitable `message` that people will recognize / have probably seen.

lucyleeow: Good point, I wasn't sure about that as well. Amended back.

doc/configuration.rst
```diff
@@ -61,9 +61,15 @@ And some things can be tweaked directly in CSS:
...
-example, to remove the specific Matplotlib agg warning, you can add::
+.. _removing_warnings:
+
+Removing warnings
+=================
+
+To prevent warnings from being captured and included in your built
+documentation, you can use the package ``warnings`` in the ``conf.py`` file.
+For example, to remove the specific ``UserWarning``: 'I don't want this
+warning', you can add::
```
larsoner, line 72: this text/explanation no longer matches what's done below

lucyleeow: Sorry, it is not my day today. Thanks!

larsoner:
Thanks @lucyleeow

PR #705, opened 2020-06-11
ENH Support pypandoc to convert rst to md for ipynb

### LINKED ISSUE #219: Limited rst to md support in notebooks
Hi,

Current parser for rst to md in notebook.py is quite limited. For example this [page](http://pytorch.org/tutorials/advanced/neural_style_tutorial.html#)'s notebooks looks like this: 
screen shot 2017-04-06 at 12 13 02 pm

Here's what I propose: use [pypandoc](https://github.com/bebraw/pypandoc) if available else fallback to current implementation. If you like this, I can raise a PR.

lucyleeow:
closes #219

Use pypandoc (https://github.com/bebraw/pypandoc), if available, to convert rst to md for ipynbs.

Considerations:
* Should we add a configuration option for this to allow users to set pandoc arguments or filters? Not sure how often these are used. The downside is that some settings may break our ipynb conversion.

Todo:
Will add a test in `test_full.py`

### DIFF
--- a/continuous_integration/travis/install.sh
+++ b/continuous_integration/travis/install.sh
@@ -12,7 +12,7 @@ if [ "$DISTRIB" == "conda" ]; then
     export CONDA_DEPENDENCIES="pip numpy setuptools matplotlib pillow pytest pytest-cov coverage seaborn joblib flake8 check-manifest ${CONDA_PKGS}"
     export PIP_DEPENDENCIES="sphinx_rtd_theme"
     if [ "$PYTHON_VERSION" != "3.5" ] && [ "$PYTHON_VERSION" != "3.6" -o "$LOCALE" != "C" ]; then
-        export PIP_DEPENDENCIES="${PIP_DEPENDENCIES} memory_profiler vtk mayavi ipython"
+        export PIP_DEPENDENCIES="${PIP_DEPENDENCIES} memory_profiler vtk mayavi ipython pypandoc"
     fi
     if [ "$SPHINX_VERSION" != "dev" ]; then
         export PIP_DEPENDENCIES="${PIP_DEPENDENCIES} sphinx${SPHINX_VERSION}"
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -266,7 +266,7 @@ def _complete_gallery_conf(sphinx_gallery_conf, src_dir, plot_gallery,
     for key in pypandoc:
         if key not in accepted_keys:
             raise ConfigError("'pypandoc' only accepts the following key "
-                                "values: %s, got: %s." % accepted_keys, key)
+                                "values: %s, got: %s." % (accepted_keys, key))
 
     # Make it easy to know which builder we're in
     gallery_conf['builder_name'] = builder_name
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -256,17 +256,18 @@ def _complete_gallery_conf(sphinx_gallery_conf, src_dir, plot_gallery,
         raise ConfigError("The 'last_notebook_cell' parameter must be type str"
                           " or None, found type %s" % type(last_cell))
     # Check pypandoc
-    gallery_conf['pypandoc'] = dict() if gallery_conf['pypandoc'] is True \
-        else gallery_conf['pypandoc']
     pypandoc = gallery_conf['pypandoc']
-    if not isinstance(pypandoc, dict):
+    if not isinstance(pypandoc, (dict, bool)):
         raise ConfigError("'pypandoc' must be a dict or bool, got: "
                           "%s" % (type(pypandoc),))
-    accepted_keys = ('extra_args', 'filters')
-    for key in pypandoc:
-        if key not in accepted_keys:
-            raise ConfigError("'pypandoc' only accepts the following key "
-                              "values: %s, got: %s." % (accepted_keys, key))
+    if isinstance(pypandoc, dict):
+        accepted_keys = ('extra_args', 'filters')
+        for key in pypandoc:
+            if key not in accepted_keys:
+                raise ConfigError("'pypandoc' only accepts the following key "
+                                  "values: %s, got: %s."
+                                  % (accepted_keys, key))
+    gallery_conf['pypandoc'] = dict() if pypandoc is True else pypandoc
 
     # Make it easy to know which builder we're in
     gallery_conf['builder_name'] = builder_name
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -260,14 +260,14 @@ def _complete_gallery_conf(sphinx_gallery_conf, src_dir, plot_gallery,
         pypandoc = gallery_conf['pypandoc']
         if not isinstance(pypandoc, (bool, dict)):
             raise ConfigError("'pypandoc' must be a dict or bool, got: "
-                              f"{type(pypandoc)}.")
+                              "%s" % type(pypandoc))
         if isinstance(pypandoc, dict):
             accepted_keys = ('extra_args', 'filters')
             for key in pypandoc.keys():
                 if key not in accepted_keys:
                     raise ConfigError("'pypandoc' only accepts the following "
-                                      f"key values: {accepted_keys}, "
-                                      f"got: {key}.")
+                                      "key values: %s, got: %s."
+                                      % accepted_keys, key)
 
     # Make it easy to know which builder we're in
     gallery_conf['builder_name'] = builder_name
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -266,7 +266,7 @@ def _complete_gallery_conf(sphinx_gallery_conf, src_dir, plot_gallery,
                                   "values: %s, got: %s." % accepted_keys, key)
     else:
         raise ConfigError("'pypandoc' must be a dict or bool, got: "
-                          "%s" % type(pypandoc))
+                          "%s" % type(gallery_conf['pypandoc']))
 
     # Make it easy to know which builder we're in
     gallery_conf['builder_name'] = builder_name
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -256,18 +256,17 @@ def _complete_gallery_conf(sphinx_gallery_conf, src_dir, plot_gallery,
         raise ConfigError("The 'last_notebook_cell' parameter must be type str"
                           " or None, found type %s" % type(last_cell))
     # Check pypandoc
-    if gallery_conf['pypandoc']:
-        pypandoc = gallery_conf['pypandoc']
-        if not isinstance(pypandoc, (bool, dict)):
-            raise ConfigError("'pypandoc' must be a dict or bool, got: "
-                              "%s" % type(pypandoc))
-        if isinstance(pypandoc, dict):
-            accepted_keys = ('extra_args', 'filters')
-            for key in pypandoc.keys():
-                if key not in accepted_keys:
-                    raise ConfigError("'pypandoc' only accepts the following "
-                                      "key values: %s, got: %s."
-                                      % accepted_keys, key)
+    gallery_conf['pypandoc'] = dict() if gallery_conf['pypandoc'] is True \
+        else gallery_conf['pypandoc']
+    if isinstance(gallery_conf['pypandoc'], dict):
+        accepted_keys = ('extra_args', 'filters')
+        for key in gallery_conf['pypandoc']:
+            if key not in accepted_keys:
+                raise ConfigError("'pypandoc' only accepts the following key "
+                                  "values: %s, got: %s." % accepted_keys, key)
+    else:
+        raise ConfigError("'pypandoc' must be a dict or bool, got: "
+                          "%s" % type(pypandoc))
 
     # Make it easy to know which builder we're in
     gallery_conf['builder_name'] = builder_name
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -23,7 +23,7 @@ from xml.sax.saxutils import quoteattr, escape
 from sphinx.errors import ConfigError, ExtensionError
 from sphinx.util.console import red
 from . import sphinx_compatibility, glr_path_static, __version__ as _sg_version
-from .utils import _replace_md5, _has_optipng
+from .utils import _replace_md5, _has_optipng, _has_pypandoc
 from .backreferences import _finalize_backreferences
 from .gen_rst import (generate_dir_rst, SPHX_GLR_SIG, _get_memory_base,
                       _get_readme)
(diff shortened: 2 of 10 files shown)


sphinx_gallery/notebook.py
```diff
@@ -171,7 +183,17 @@ def fill_notebook(work_notebook, script_blocks):
         if blabel == 'code':
             add_code_cell(work_notebook, bcontent)
         else:
-            add_markdown_cell(work_notebook, bcontent + '\n')
+            try:
+                import pypandoc  # noqa
+            except ImportError:
+                logger.warning("pypandoc not available. Using Sphinx-Gallery "
```
choldgraf, line 189: I think we'd want to check for a user option here to explicitly use pypandoc, else we'll start throwing warnings to users all over the place

lucyleeow: Yeah I realised this and have made those changes. Come back in 5! :)

lucyleeow: Also putting the warning in a for loop was not a good idea!

choldgraf: What do you mean not a good idea?
What do you mean not a good idea?
What do you mean not a good idea?
What do you mean not a good idea?
What do you mean not a good idea?
What do you mean not a good idea?
What do you mean not a good idea?
What do you mean not a good idea?
What do you mean not a good idea?
What do you mean not a good idea?
What do you mean not a good idea?
What do you mean not a good idea?

sphinx_gallery/notebook.py
```diff
@@ -142,36 +145,65 @@ def add_code_cell(work_notebook, code):
...
+    If available, uses pypandoc to convert rst to markdown.
+
     Parameters
     ----------
     script_blocks : list
         Each list element should be a tuple of (label, content, lineno).
     """
+    pandoc = False
+    pandoc_kwargs = {}
+    if gallery_conf["pypandoc"] or isinstance(gallery_conf["pypandoc"], dict):
```
lucyleeow, line 183: I did this so it uses pypandoc if config is `True` or if empty dict `{}`. Happy to change if unnecessarily complex

lucyleeow:
I'm not sure why travis is failing. For some reason the 'pypandoc' config is set to `True` even though the default is `False`. Even so, importing pypandoc is in a `try/except` so even without it installed, it should work fine.

lucyleeow:
I've restarted the build.

ping @larsoner

larsoner:
Travis failures appear real

continuous_integration/travis/install.sh
```diff
@@ -12,7 +12,7 @@ if [ "$DISTRIB" == "conda" ]; then
     export CONDA_DEPENDENCIES="pip numpy setuptools matplotlib pillow pytest pytest-cov coverage seaborn joblib flake8 check-manifest ${CONDA_PKGS}"
     export PIP_DEPENDENCIES="sphinx_rtd_theme"
     if [ "$PYTHON_VERSION" != "3.5" ] && [ "$PYTHON_VERSION" != "3.6" -o "$LOCALE" != "C" ]; then
-        export PIP_DEPENDENCIES="${PIP_DEPENDENCIES} memory_profiler vtk mayavi ipython"
+        export PIP_DEPENDENCIES="${PIP_DEPENDENCIES} memory_profiler vtk mayavi ipython pypandoc"
```
larsoner, line 15: This is only for some builds, the others will need to `pytest.mark.skipif` pandoc is not available

sphinx_gallery/notebook.py
```diff
@@ -142,36 +145,65 @@ def add_code_cell(work_notebook, code):
...
+        try:
+            import pypandoc  # noqa
+        except ImportError:
```
larsoner, line 186: Also it appears as though `pypandoc` can raise an `OSError` if `pandoc` is not found, so this should be at least:
```
except (ImportError, OSError) as exc:
```
And then below it might be nice to do:
```
"files:\n%s" % (exc,))
```

sphinx_gallery/tests/test_full.py
```diff
@@ -760,3 +760,22 @@ def test_matplotlib_warning_filter(sphinx_app):
...
+    with codecs.open(fname, 'r', 'utf-8') as fid:
+        md = fid.read()
+
+    print(md)
+    md_sg = r"Use :mod:`sphinx_gallery` to link to other packages, like\n:mod:`numpy`, :mod:`matplotlib.colors`, and :mod:`matplotlib.pyplot`."  # noqa
+    md_pandoc = r'Use `sphinx_gallery`{.interpreted-text role=\"mod\"} to link to other\npackages, like `numpy`{.interpreted-text role=\"mod\"},\n`matplotlib.colors`{.interpreted-text role=\"mod\"}, and\n`matplotlib.pyplot`{.interpreted-text role=\"mod\"}.'  # noqa
+
+    try:
+        import pypandoc  # noqa
+    except ImportError:
```
larsoner, line 778: ahh okay, no need for `skipif`, just need the better `except (ImportError, OSError)` here as well

lucyleeow: Thanks!

lucyleeow:
I suspect that the `import pypandoc` isn't generating any error, even when it is not available, and it's actually when a function gets called that it raises the error.

Trying a bad fix.

lucyleeow:
So it seems that import error is only raised when a function is called. I use `pypandoc.get_pandoc_version()` and `logger.info()` it.

sphinx_gallery/notebook.py
```diff
@@ -142,36 +145,70 @@ def add_code_cell(work_notebook, code):
...
+            "cell_type": "markdown",
+            "metadata": {},
+            "source": [text]
+        }
+    else:
+        markdown_cell = {
+            "cell_type": "markdown",
+            "metadata": {},
+            "source": [rst2md(text)]
+        }
```
larsoner, line 167:
```suggestion
    if not pandoc:
        text = rst2md(text)
    markdown_cell = {
        "cell_type": "markdown",
        "metadata": {},
        "source": [text]
    }
```

larsoner: And actually `pandoc` as a kwarg is not so informative here. Maybe `convert=True` should be the default, meaning "convert RST to MD" and in whatever code calls this sets `convert=not pandoc`

sphinx_gallery/notebook.py
```diff
@@ -142,36 +145,70 @@ def add_code_cell(work_notebook, code):
...
+        try:
+            import pypandoc  # noqa
+            # Import error raised only when function called
+            logger.info("pandoc version: %s"
+                        % (pypandoc.get_pandoc_version(),))
+        except (ImportError, OSError) as e:
+            logger.warning("'pypandoc' not available. Using Sphinx-Gallery to "
+                           "convert rst text blocks to markdown for .ipynb "
+                           "files.")
+            if e:
```
larsoner, line 193: Is this conditional necessary? Why not just always output it with the line above?

sphinx_gallery/notebook.py
```diff
@@ -142,36 +145,70 @@ def add_code_cell(work_notebook, code):
...
         else:
-            add_markdown_cell(work_notebook, bcontent + '\n')
+            if pandoc:
+                md = pypandoc.convert_text(
+                    bcontent, to='md', format='rst', **pandoc_kwargs
+                )
+                # pandoc automatically adds '\n' at end
+                add_markdown_cell(work_notebook, md, pandoc=True)
+            else:
+                add_markdown_cell(work_notebook, bcontent + '\n')
```
larsoner, line 212: Alternatively, maybe you could do above:
```
converter = partial(pypandoc.convert_text, to='md', format='rst', **pandoc_kwargs)  # if pandoc
...
converter = None  # if not pandoc
```
And then this call can just be:
```
add_markdown_cell(work_notebook, bcontent + '\n', converter=converter)
```
And `add_markdown_cell` has `converter = rst2md if converter is None else converter`.

To me it seems to have a nicer logic/flow compared to the current implementation, WDYT?

sphinx_gallery/tests/test_full.py
```diff
@@ -760,3 +760,23 @@ def test_matplotlib_warning_filter(sphinx_app):
...
+    with codecs.open(fname, 'r', 'utf-8') as fid:
+        md = fid.read()
+
+    md_sg = r"Use :mod:`sphinx_gallery` to link to other packages, like\n:mod:`numpy`, :mod:`matplotlib.colors`, and :mod:`matplotlib.pyplot`."  # noqa
+    md_pandoc = r'Use `sphinx_gallery`{.interpreted-text role=\"mod\"} to link to other\npackages, like `numpy`{.interpreted-text role=\"mod\"},\n`matplotlib.colors`{.interpreted-text role=\"mod\"}, and\n`matplotlib.pyplot`{.interpreted-text role=\"mod\"}.'  # noqa
+
+    try:
+        import pypandoc  # noqa
+        # Import error raised only when function called
+        pypandoc.get_pandoc_version()
```
larsoner, line 778: Seeing this more than one place makes it seem like we should have a `_get_pandoc()` for example that either returns `pypandoc, version` or `None, None` (DRY)

sphinx_gallery/notebook.py
larsoner: (and maybe we just kill `add_markdown_cell` entirely and keep the logic all in this function, since now `add_markdown_cell` does almost nothing)

larsoner: ahh but I see there is a `add_code_cell`, so probably worth keeping `add_markdown_cell`. But maybe we should give it content already in MD format, rather than asking it to do the conversion

sphinx_gallery/notebook.py
lucyleeow: Not familiar with python errors and wasn't sure if there would always be a message. Thanks!

lucyleeow:
Thanks for the review @larsoner. Added a `_has_pypandoc()` fun, which optionally outputs logger warning/info and `add_markdown_cell()` now takes the markdown as input.

sphinx_gallery/notebook.py
```diff
@@ -142,36 +145,50 @@ def add_code_cell(work_notebook, code):
...
 
+    If available, uses pypandoc to convert rst to markdown.
+
     Parameters
     ----------
     script_blocks : list
         Each list element should be a tuple of (label, content, lineno).
     """
+    use_pypandoc = \
+        gallery_conf["pypandoc"] or isinstance(gallery_conf["pypandoc"], dict)
```
larsoner, line 175: Logic here could be simplified if ...

sphinx_gallery/gen_gallery.py
```diff
@@ -254,6 +255,20 @@ def call_memory(func):
...
                           " or None, found type %s" % type(last_cell))
+    # Check pypandoc
+    if gallery_conf['pypandoc']:
+        pypandoc = gallery_conf['pypandoc']
+        if not isinstance(pypandoc, (bool, dict)):
+            raise ConfigError("'pypandoc' must be a dict or bool, got: "
+                              "%s" % type(pypandoc))
+        if isinstance(pypandoc, dict):
+            accepted_keys = ('extra_args', 'filters')
+            for key in pypandoc.keys():
```
larsoner, line 266:
```suggestion
            for key in pypandoc:
```

sphinx_gallery/gen_gallery.py
```diff
@@ -254,6 +255,20 @@ def call_memory(func):
...
+                if key not in accepted_keys:
+                    raise ConfigError("'pypandoc' only accepts the following "
+                                      "key values: %s, got: %s."
+                                      % accepted_keys, key)
```
larsoner, line 270: ... here you do the triaging to make `pypandoc` False or `dict` (only allowing the latter if it's avaliable). So something like:
```
pypandoc = dict() if pypandoc is True else pypandoc
if isinstance(pypandoc, dict):
    if not _has_pypandoc():
        warn(...)
        pypandoc = False
    else:
        accepted_keys ...
        ...
gallery_conf['pypandoc'] = pypandoc  # now either False or a dict of options to pass
```

lucyleeow: Thanks that's nicer!

lucyleeow:
ping @larsoner, the error is codecov. Thanks!

sphinx_gallery/utils.py
```diff
@@ -151,3 +154,21 @@ class Bunch(dict):
     def __init__(self, **kwargs):  # noqa: D102
         dict.__init__(self, kwargs)
         self.__dict__ = self
+
+
+def _has_pypandoc(raise_error=False):
```
larsoner, line 159: `raise_error` seems like an inappropriate name, it implies that a `raise ...Error` will happen if something goes wrong. maybe `log_result=True`?

But really if you do the refactoring I suggest, you will only ever use this once so you can eliminate the `kwarg` altogether.

sphinx_gallery/gen_gallery.py
```diff
@@ -254,6 +255,20 @@ def call_memory(func):
...
+        raise ConfigError("'pypandoc' must be a dict or bool, got: "
+                          "%s" % (type(pypandoc),))
+    if isinstance(pypandoc, dict):
+        accepted_keys = ('extra_args', 'filters')
+        for key in pypandoc:
+            if key not in accepted_keys:
+                raise ConfigError("'pypandoc' only accepts the following key "
+                                  "values: %s, got: %s."
+                                  % (accepted_keys, key))
+    gallery_conf['pypandoc'] = dict() if pypandoc is True else pypandoc
```
larsoner, line 270: The place you want to check-and-log/convert is here (once, at the beginning)
```suggestion
    gallery_conf['pypandoc'] = dict() if pypandoc is True else pypandoc
    if (isinstance(gallery_conf['pypandoc'], dict) and \
            not _has_pypandoc(log_result=True)):
        gallery_conf['pypandoc'] = False
```
        
```

File: sphinx_gallery/notebook.py [hunk 15]
```diff
@@ -142,36 +145,47 @@ def add_code_cell(work_notebook, code):
...
         Each list element should be a tuple of (label, content, lineno).
     """
-
     for blabel, bcontent, lineno in script_blocks:
         if blabel == 'code':
             add_code_cell(work_notebook, bcontent)
         else:
-            add_markdown_cell(work_notebook, bcontent + '\n')
+            if (isinstance(gallery_conf["pypandoc"], dict) and
+                    _has_pypandoc(raise_error=True)):
```
larsoner, line 179:
```suggestion
            if isinstance(gallery_conf["pypandoc"], dict):
```
Then here you don't even need to check. (The way you have it coded now, it would emit a warning/log for every block of every example I think.)

sphinx_gallery/utils.py
lucyleeow: Changed to:

> Seeing this more than one place makes it seem like we should have a _get_pandoc() for example that either returns pypandoc, version or None, None (DRY)

sphinx_gallery/notebook.py
lucyleeow: Eurgh yes I did it again. Was trying to avoid the logic at the top.

sphinx_gallery/notebook.py
```diff
@@ -142,36 +145,46 @@ def add_code_cell(work_notebook, code):
...
+            if gallery_conf["pypandoc"] == False:
+                markdown = rst2md(bcontent + '\n')
+            else:
+                import pypandoc
+                # pandoc automatically addds \n to the end
+                markdown = pypandoc.convert_text(
+                    bcontent, to='md', format='rst', **gallery_conf["pypandoc"]
+                )
+
+        add_markdown_cell(work_notebook, markdown)
```
larsoner, line 187: This still needs to be indented one level deeper (to match old behavior at least)

lucyleeow: whoops, thanks

lucyleeow:
Thanks for your patience @larsoner

PR #706, opened 2020-06-11
BUG: Use relative paths for zip downloads

### LINKED ISSUE #704: Working with pre-built galleries
I'm using `sphinx-gallery` to build example galleries. The documentation is hosted on [Read the Docs (RTD)](https://pystiche.readthedocs.io/en/latest/galleries/examples/index.html). My examples rely on a GPU, so I can't let them be built by RTD.

I came up with the following to overcome this: I've set up a separate build server that spins up every few hours and runs a `sphinx-gallery` build for the current `master`. Afterwards the generated galleries are uploaded. Within the `conf.py` I detect if a build is performed by RTD and download the pre-built galleries instead. 

I've ran into the following problems:

1. I need to set `plot_gallery=False` since the pre-built galleries might not match the examples in the current commit. If I don't RTD tries to rebuild the changed examples, which will fail. Still, sometimes it recompiles the galleries without running them. I've yet to figure out how to reproduce this. I think it detects non-matching checksums, but I'm not sure yet. Unfortunately this removes the output from the pre-built galleries, defeating their purpose.
2. In response to 1. I've tried to remove the `"sphinx_gallery.gen_gallery"` from the `extensions` if run by RTD with moderate success. The output is preserved, but I think some "style files" (not sure if this is the correct term) are missing. For example the background color of the output blocks is no longer yellow, the thumbnails in the gallery are no longer clickable, and so on.

Is there any way to overcome these issues? Can I for example force `sphinx-gallery` to accept the pre-built galleries or copy the "style files" without executing the scripts?

pmeier:
Closes #704, Cc @larsoner

### TL;DR

The zip files comprising all examples were included with absolute file links. If one wants to use galleries built on another machine this results in an error if the paths do not match.

### DIFF
--- a/sphinx_gallery/downloads.py
+++ b/sphinx_gallery/downloads.py
@@ -117,14 +117,14 @@ def generate_zipfiles(gallery_dir):
             target_dir = os.path.join(gallery_dir, directory)
             listdir.extend(list_downloadable_sources(target_dir))
 
-    py_zipfile = python_zip(listdir, gallery_dir)
-    jy_zipfile = python_zip(listdir, gallery_dir, ".ipynb")
+    py_zipfile = os.path.basename(python_zip(listdir, gallery_dir))
+    jy_zipfile = os.path.basename(python_zip(listdir, gallery_dir, ".ipynb"))
 
     def rst_path(filepath):
         return filepath.replace(os.sep, '/')
 
-    dw_rst = CODE_ZIP_DOWNLOAD.format(os.path.basename(py_zipfile),
+    dw_rst = CODE_ZIP_DOWNLOAD.format(py_zipfile,
                                       rst_path(py_zipfile),
-                                      os.path.basename(jy_zipfile),
+                                      jy_zipfile,
                                       rst_path(jy_zipfile))
     return dw_rst
--- a/sphinx_gallery/downloads.py
+++ b/sphinx_gallery/downloads.py
@@ -95,7 +95,7 @@ def list_downloadable_sources(target_dir):
             if fname.endswith('.py')]
 
 
-def generate_zipfiles(gallery_dir):
+def generate_zipfiles(gallery_dir, src_dir):
     """
     Collects all Python source files and Jupyter notebooks in
     gallery_dir and makes zipfiles of them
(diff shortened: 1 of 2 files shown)


pmeier:
CircleCI failure seems unrelated. TravisCI failure seems related, but I can't figure out why it would only fail for one environment. Can someone help me here?

sphinx_gallery/downloads.py
```diff
@@ -117,14 +117,14 @@ def generate_zipfiles(gallery_dir):
             target_dir = os.path.join(gallery_dir, directory)
             listdir.extend(list_downloadable_sources(target_dir))
 
-    py_zipfile = python_zip(listdir, gallery_dir)
-    jy_zipfile = python_zip(listdir, gallery_dir, ".ipynb")
+    py_zipfile = os.path.basename(python_zip(listdir, gallery_dir))
+    jy_zipfile = os.path.basename(python_zip(listdir, gallery_dir, ".ipynb"))
```
larsoner, line 121: Probably better to just modify `python_zip` to do the right thing. I'll see if I can fix it, as `os.path.basename` is linking to the wrong place currently (just `doc/` / the root, which won't be correct). Probably needs `gallery_dir/` prepended...

larsoner:
@pmeier I pushed a commit, can you see if it works for you? Basically the idea is that the RST download link should be relative to the source root, hopefully this change does it. It's possible that there are other places in the codebase that could benefit from this relative path business, but maybe this is already enough for your use case.

pmeier:
@larsoner Works as expected. Thanks for the follow-up.

larsoner:
@lucyleeow feel free to review and merge if you're happy

lucyleeow:
LGTM, thanks @pmeier!

pmeier:
@larsoner deserves the credit since no line of my initial "fix" is left. I merely discovered the problem.

larsoner:
> no line of my initial "fix" is left

Still deserve credit for digging into the code to figure out it needed to be a relative path, I just fixed the relative-path-ization

lucyleeow:
@pmeier you did the tough part of the job! And I always rely on @larsoner to clean up my mess :sweat_smile:

PR #714, opened 2020-06-19
BUG: Always do linking

larsoner:
I was doing a `make html_dev-noplot` run and I was confused why links weren't showing up in the code. There isn't really a good reason not to put them in nowadays since we use `intersphinx` to find names and such. It's quick relative to the entire build time, and useful even in noplot runs.

Hopefully the "inexplicable fix" workaround is no longer needed. If it causes problems, we should fix the underlying bug.

### DIFF
--- a/sphinx_gallery/docs_resolv.py
+++ b/sphinx_gallery/docs_resolv.py
@@ -423,12 +423,7 @@ def embed_code_links(app, exception):
     if exception is not None:
         return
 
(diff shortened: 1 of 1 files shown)


lucyleeow:
The appyveyor failure is due to known bug: https://github.com/matplotlib/matplotlib/issues/16965

GaelVaroquaux:
Looks good. Merging.

Issue #715, opened 2020-06-21
Gallery example using plotly


emmanuelle:
Hi, I was planning to submit a PR showing how to use plotly figures in sphinx-gallery examples and I'm opening this issue to prepare this PR and to be sure whether it would be well received.

There are two ways to use plotly in sphinx-gallery:
- either by using the recently introduced `_repr_html_` of plotly figure objects. This is the easiest way to use plotly figures, but then you don't get png files saved for the thumbnails
- or use a custom scraper provided by the plotly package, which saves both png and html files. For this scraper to work you need to install the `orca` package which is easy to install with conda but with pip you need to install other dependencies (basically because you need a browser to run to save a figure as png). We're planning to make this process easier (ie pip-installable) in the coming months.

My plan was to write an example for the sphinx-gallery website using just `_repr_html_` since it's easy to do, and to mention that the scraper exists if you want to get the png thumbnails as well. How does that sound? Also where should I put the example?  In the main gallery or create a separate gallery like the mayavi one?

emmanuelle:
Hi @lucyleeow thanks a lot, I'll do as you suggest!

PR #716, opened 2020-06-22
DOC Specify matplotlib in animation example

lucyleeow:
I thought it would be useful to specify 'Maplotlib' in the animation example, especially if we will add a plotly example (#715).


lucyleeow:
Thanks @larsoner

PR #718, opened 2020-06-22
Plotly example for the gallery

### LINKED ISSUE #715: Gallery example using plotly

emmanuelle:
Closes #715.  I hope this is the right level of details.


emmanuelle:
ok, so CI is passing now on CircleCI. About Appveyor maybe the failure is not related to this PR since #713 fails in the same way.

lucyleeow:
> About Appveyor maybe the failure is not related to this PR since #713 fails in the same way.

Yes, sorry I should have mentioned, appveyor is failing due to known matplotlib bug (https://github.com/matplotlib/matplotlib/issues/16965).

appveyor.yml
```diff
@@ -6,7 +6,7 @@ environment:
 
   global:
     PYTHON: "C:\\conda"
-    CONDA_DEPENDENCIES: "numpy seaborn matplotlib sphinx pillow pytest pytest-cov sphinx_rtd_theme joblib"
+    CONDA_DEPENDENCIES: "numpy seaborn matplotlib sphinx pillow pytest pytest-cov sphinx_rtd_theme joblib plotly pandas"
```
lucyleeow, line 9: I don't think these are needed as pandas was already being used in an example and it wasn't installed. Though I see we build with no-plot but also do 'make html' ? Maybe @larsoner can clarify?

continuous_integration/travis/install.sh
```diff
@@ -9,7 +9,7 @@
 set -e
 
 if [ "$DISTRIB" == "conda" ]; then
-    export CONDA_DEPENDENCIES="pip numpy setuptools matplotlib pillow pytest pytest-cov coverage seaborn joblib flake8 check-manifest ${CONDA_PKGS}"
+    export CONDA_DEPENDENCIES="pip numpy setuptools matplotlib pillow pytest pytest-cov coverage seaborn plotly pandas joblib flake8 check-manifest ${CONDA_PKGS}"
```
lucyleeow, line 12: Not sure about here too for same reason...?

examples/plot_9_plotly.py
```diff
@@ -0,0 +1,64 @@
...
+``_repr_html_`` of plotly figure objects. Therefore you should return a figure
+object to display the figure.
+
+In order to use plotly, the ``conf.py`` of the project should include the
+following lines to select the appropriate plotly renderer::
+
+    import plotly.io as pio
+    pio.renderers.default = 'sphinx_gallery'
+
+**Optional**: the sphinx_gallery renderer of plotly will not generate png
```
lucyleeow, line 17:
```suggestion
**Optional**: the ``sphinx_gallery`` renderer of plotly will not generate png
```
(to be consistent with formatting of sphinx_gallery_png below)

examples/plot_9_plotly.py
```diff
@@ -0,0 +1,64 @@
...
+thumbnails. For png thumbnails, you can use instead the ``sphinx_gallery_png``
+renderer, and add ``plotly.io._sg_scraper.plotly_sg_scraper`` to the list of
+:ref:`image_scrapers`. The scraper requires to
```
lucyleeow, line 20:
```suggestion
:ref:`image_scrapers`. The scraper requires you to
```

examples/plot_9_plotly.py
```diff
@@ -0,0 +1,64 @@
+"""
+========================================
+Example with the plotly graphing library
+========================================
+
+sphinx-gallery supports examples made with the
+`plotly library <https://plotly.com/python/>`_. sphinx-gallery captures the
+``_repr_html_`` of plotly figure objects. Therefore you should return a figure
+object to display the figure.
```
lucyleeow, line 9: 'sphinx-gallery is able to capture the ``_repr_html_`` of plotly figure objects (see :ref:`capture_repr`). To display your figure you should return the plotly figure object as the last line in your code block.' 

(not sure about my wording either)

lucyleeow: I am not sure 'return' is the right word here but not sure what would be better. 'The last line in your code block should be the plotly figure object' ?

examples/plot_9_plotly.py
lucyleeow: Maybe you could mention here that a sphinx-gallery image scraper is included by the plotly package?

appveyor.yml
emmanuelle: I think it is still needed for the appveyor build. However I can remove pandas since it is a dependency of seaborn, hence the example with pandas was working before (thanks to seaborn)

emmanuelle:
Thanks a lot for the review @lucyleeow ! I had a closer look at the different CI files and I think I need to add plotly to all the files since all builds do `make html` but I removed pandas since seaborn is already taking care of pandas.

lucyleeow:
> seaborn is already taking care of pandas.

Ah I didn't realise this! Thank you.

And yes I think you are right, all the CIs build the docs.

lucyleeow:
@larsoner feel free to merge if you are happy!

larsoner:
Thanks @emmanuelle !

PR #720, opened 2020-06-25
MAINT: Deal with mayavi

larsoner:


larsoner:
Matplotlib bug will only show up on 3.2.2 and when using animations and when imagemagick is not available which is a pretty deep corner case (upstream bug https://github.com/matplotlib/matplotlib/pull/17759) so I think we can ignore it for now

PR #722, opened 2020-06-30
Make md5 hash independent of platform line endings

sdhiscocks:
Currently md5 hash for Python files depends on what platform file is on, as file endings will differ. This change reads the Python files in the hashing function with universal line endings, such md5 is independent of platform being used.

Surrogate escapes are used, to ensure that it can handle any errors without data loss.

This is useful when adding a complex or long running examples into a repository that may be used across platforms.

### DIFF
--- a/sphinx_gallery/utils.py
+++ b/sphinx_gallery/utils.py
@@ -124,20 +124,32 @@ def replace_py_ipynb(fname):
     return '{}{}'.format(fname_prefix, new_extension)
 
 
-def get_md5sum(src_file):
-    """Returns md5sum of file"""
-    with open(src_file, 'rt', errors='surrogateescape') as src_data:
-        src_content = src_data.read().encode(errors='surrogateescape')
+def get_md5sum(src_file, mode='b'):
+    """Returns md5sum of file
+
+    Parameters
+    ----------
+    src_file : str
+        Filename to get md5sum for.
+    mode : 't' or 'b'
+        File mode to open file with. When in text mode, universal line endings
+        are used to ensure consitency in hashes between platforms.
+    """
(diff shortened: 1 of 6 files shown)


lucyleeow:
Thanks @sdhiscocks. Maybe we could also add a test in `test_full.py` that checks value of the md5 hash of an example? Since it will be run on Windows and Linux in our CI, it would check this. WDYT @larsoner ?

larsoner:
Sure, seems like that would work

lucyleeow:
Thanks @sdhiscocks, @larsoner merge when you're happy/CI's green!

sphinx_gallery/utils.py
```diff
@@ -126,8 +126,8 @@ def replace_py_ipynb(fname):
 
 def get_md5sum(src_file):
     """Returns md5sum of file"""
-    with open(src_file, 'rb') as src_data:
-        src_content = src_data.read()
+    with open(src_file, 'rt', errors='surrogateescape') as src_data:
```
larsoner, line 129: Some of the things we md5sum are binary pickle objects. It seems like it would be better to add a `get_md5sum(src_file, mode='b')` default, and if `mode='t'` then you use this escape-replacement version, and the default binary mode stays the same (plain `'rb'` read).

sdhiscocks: How about checking _src_file_ extension is ".py"?

larsoner: There are other things we check, too, like `ipynb` that can be text mode. I'd prefer to keep the decision about which mode to use with the places we make calls to `_replace_md5` (which calls `get_md5sum` internally) or `get_md5sum` directly

larsoner:
Thanks @sdhiscocks !

PR #724, opened 2020-07-04
Add support for image links and data URIs for notebooks

sdhiscocks:
Currently, when notebooks are generated, the links to images found in rST remain unmodified, meaning images expected to be either local or found on web server hosting documentation are broken.

This change allows inclusion of images via data URIs, or prefixing the image links in the notebooks, for example to server hosting documentation.

This also changes the markdown output to use HTML `img` tag, and such also supports the width and height options.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -39,6 +39,7 @@ file:
 - ``show_memory`` (:ref:`show_memory`)
 - ``binder`` (:ref:`binder_links`)
 - ``first_notebook_cell`` and ``last_notebook_cell`` (:ref:`own_notebook_cell`)
+- ``notebook_images`` (:ref:`notebook_images`)
 - ``pypandoc`` (:ref:`use_pypandoc`)
 - ``junit`` (:ref:`junit_xml`)
 - ``log_level`` (:ref:`log_level`)
@@ -621,6 +622,62 @@ parameter::
 
 If the value of ``first_notebook_cell`` or ``last_notebook_cell`` is set to
 ``None``, then no extra first or last cell will be added to the notebook.
+
+.. _notebook_images:
+
+Adding images to notebooks
+==========================
+
+When notebooks are produced, by default (``notebook_images = False``) image
+paths from the `image` directive in rST documentation blocks (not images
+generated fom code) are included in markdown in their original form. This
+includes paths to images expected to be present on the local filesystem which
+is unlikely to be the case for those downloading the notebook.
+
+By setting ``notebook_images = True``, images will be included in the generated
+notebooks via Base64-encoded `data URIs <https://en.wikipedia.org/wiki/Data_URI_scheme>`_.
+As inclusion of images via data URIs can significantly increase size of the
+notebook, it's suggested this only be used when small images are used throughout
+galleries.
+
+An alternative is to instead provide a prefix string that'll be used for images
+e.g. the root URL of where your documentation is hosted. So for example the
+following configuration::
+
+    sphinx_gallery_conf = {
+        ...
+        'notebook_images': 'https://project.example.com/en/latest/',
+        ...
+    }
+
+with an example `image` directive in an rST being output to gallery dir
+`auto_examples` (relative and absolute paths are supported):
+
+.. code-block:: rst
+
+    .. image:: ../_static/example.jpg
+        :alt: An example image
+
+The image will be added to the generated notebook pointing to the source URL
+``https://project.example.com/en/latest/_static/example.jpg``
+
+Note that the prefix is applied directly, so a trailing ``/`` should be
+included in the prefix if it's required.
+
+.. tip::
+
+    If building multiple versions of your documentation on a hosted service and
+    using prefix, consider using `sphinx build -D <https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-D>`_
+    command line option to ensure links point to the correct version. For
+    example:
+
+    .. code-block:: sh
+
+        sphinx-build \
+            -b html \
+            -D sphinx_gallery_conf.notebook_images="https://project.example.com/docs/${VERSION}/" \
+            source_dir build_dir
+
 
 .. _use_pypandoc:
 
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -646,12 +646,14 @@ following configuration::
 
     sphinx_gallery_conf = {
         ...
+        'examples_dirs': ['../examples'],
+        'gallery_dirs': ['auto_examples'],
+        ...
         'notebook_images': 'https://project.example.com/en/latest/',
         ...
     }
 
-with an example `image` directive in an rST being output to gallery dir
-`auto_examples` (relative and absolute paths are supported):
+with an example `image` directive in an rST documentation block being:
 
 .. code-block:: rst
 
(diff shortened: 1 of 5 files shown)


larsoner:
@sdhiscocks have you tried the [pandoc mode](https://sphinx-gallery.github.io/dev/configuration.html#using-pypandoc-to-convert-rst-to-markdown) for any of this stuff? I'm wondering if it might already be good enough for you, rather than extending our own implementation it might be more sustainable just to use that...

sdhiscocks:
> @sdhiscocks have you tried the [pandoc mode](https://sphinx-gallery.github.io/dev/configuration.html#using-pypandoc-to-convert-rst-to-markdown) for any of this stuff? I'm wondering if it might already be good enough for you, rather than extending our own implementation it might be more sustainable just to use that...

I did take a look, but didn't appear to support anything like this. There was a way to use Pandoc to create data URIs, by first converting to standalone HTML, and then back to markdown, which isn't really suitable solution. Plus, I'd prefer to use links rather than data URIs as even for small images they are very verbose. I couldn't see anyway to do that with Pandoc.

larsoner:
@sdhiscocks this one needs a rebase or merge now

sphinx_gallery/notebook.py
```diff
@@ -125,16 +134,53 @@ def rst2md(text, heading_levels):
...
+        return image_path
+
+    if not gallery_conf['notebook_images']:
+        return "file://" + image_path.lstrip('/')
+
+    if image_path.startswith('/'):
+        # Absolute path from source directory
+        target_dir = gallery_conf['src_dir']
+        image_path = image_path.lstrip('/')
+    full_path = os.path.join(target_dir, image_path.replace('/', os.sep))
```
lucyleeow, line 162: I might be confused here but is it possible to refer to image file but not start with '/', with rST `image` directive? 
From [sphinx docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#images), you can refer to image using:
```rst
.. image:: gnu.png
or
.. image:: ../images/gnu.png
```

sdhiscocks: Yes. So if it's absolute (starting with `/`), full path is relative from the documentation source directory. If it's relative (not starting with `/`), full path will be from `target_dir` (location of RST).

lucyleeow: Ah yes, thank you, I was confused.

For other poor readers could we move the comment:
```
    # If absolute path from source directory given
    if image_path.startswith('/'):
        
```

sphinx_gallery/tests/test_notebook.py
```diff
@@ -158,11 +161,88 @@ def test_headings():
     assert "# White space above\n" in text
 
 
+def test_notebook_images(gallery_conf):
```
lucyleeow, line 164: Would it be manageable to use pytest parameterize for the first 4 tests and put the last 2 'tests' (the data URI and the missing file) in a separate test? I think it would make this easier to maintain.

doc/configuration.rst
```diff
@@ -622,6 +623,62 @@ parameter::
...
+e.g. the root URL of where your documentation is hosted. So for example the
+following configuration::
+
+    sphinx_gallery_conf = {
+        ...
+        'notebook_images': 'https://project.example.com/en/latest/',
+        ...
+    }
+
+with an example `image` directive in an rST being output to gallery dir
```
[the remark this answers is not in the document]
sdhiscocks: Thanks for the review. I've updated the documentation slightly.

sphinx_gallery/tests/test_notebook.py
sdhiscocks: Done

larsoner:
@lucyleeow let me know when you're happy and I'll take a look

sphinx_gallery/notebook.py
```diff
@@ -64,14 +68,19 @@ def directive_fun(match, directive):
                     match.group(1).strip()))
 
 
-def rst2md(text, heading_levels):
-    """Converts the RST text from the examples docstrings and comments
+def rst2md(text, gallery_conf, target_dir, heading_levels):
+    """Converts the RST text from the examples docstrigs and comments
```
lucyleeow, line 72: Hmm how did we go back to 'docstrigs'? Bad merge?

lucyleeow:
Sorry for the delay. Thanks for making the changes @sdhiscocks.

@larsoner I'm happy (though CI Python nightly is red again due to numpy install).

sphinx_gallery/notebook.py
sdhiscocks: Bad conflict resolution on rebase. Fixed now.

doc/configuration.rst
```diff
@@ -622,6 +623,67 @@ parameter::
...
 ``None``, then no extra first or last cell will be added to the notebook.
 
+.. _notebook_images:
+
+Adding images to notebooks
+==========================
+
+When notebooks are produced, by default (``notebook_images = False``) image
+paths from the `image` directive in rST documentation blocks (not images
+generated fom code) are included in markdown in their original form. This
```
larsoner, line 633: "in their original form" could mean "using RST syntax" (very bad) or "using their original paths". I think you mean the latter, but can you update the description to clarify?

doc/configuration.rst
```diff
@@ -622,6 +623,67 @@ parameter::
...
+includes paths to images expected to be present on the local filesystem which
+is unlikely to be the case for those downloading the notebook.
+
+By setting ``notebook_images = True``, images will be included in the generated
```
larsoner, line 637:
```suggestion
By setting ``notebook_images = True``, images will be embedded in the generated
```

doc/configuration.rst
```diff
@@ -622,6 +623,67 @@ parameter::
...
+        ...
+        'notebook_images': 'https://project.example.com/en/latest/',
+        ...
+    }
+
+with an example `image` directive in an rST documentation block being:
+
+.. code-block:: rst
+
+    .. image:: ../_static/example.jpg
```
larsoner, line 660: These docs should make it clear that it does a `relpath`-like computation relative to the docs root when building, so the `..` here takes us out of essentially `auto_examples` path where the SG-generated RST file lives. Without the `..`, the link would go to `https://project.example.com/en/latest/auto_examples/_static/example.jpg` (right?), which would be wrong.

sphinx_gallery/notebook.py
```diff
@@ -125,16 +134,54 @@ def rst2md(text, heading_levels):
...
+        prefix = gallery_conf['notebook_images']
+        rel_path = os.path.relpath(full_path, gallery_conf['src_dir'])
+        return prefix + rel_path.replace(os.sep, '/')
+    else:
+        # True, but not string. Embed as data URI.
+        try:
+            with open(full_path, 'rb') as image_file:
+                data = base64.b64encode(image_file.read())
+        except OSError:
+            warnings.warn(
```
larsoner, line 176: Better to raise an `ExtensionError` here? Are there cases where people would actually want this just to warn?

If not going to raise an error, this should be `logger.warning` instead of `warnings.warn`

sdhiscocks:
Thanks for the review @larsoner. I've updated the documentation and also changed to `ExtensionError` being raised for missing images (which I agree makes sense, as HTML output as well as notebooks will have missing images).

larsoner:
Thanks @sdhiscocks !

PR #729, opened 2020-07-24
ENH: Dont allow input

larsoner:
Prevent users from having their builds hang because of `input()` calls. Instead of getting a hang, they now get:
```
Extension error:
Here is a summary of the problems encountered when running the examples

Unexpected failing examples:
/home/larsoner/python/sphinx-gallery/sphinx_gallery/tests/tinybuild/examples/plot_animation.py failed leaving traceback:
  File "/home/larsoner/python/sphinx-gallery/sphinx_gallery/tests/tinybuild/examples/plot_animation.py", line 15, in <module>
    input()
sphinx.errors.ExtensionError: Cannot use input() builtin function in Sphinx-gallery examples
```
Note that I added a bit of code to `_handle_exception` to trim these unhelpful lines from the traceback we print:
```
  File "/home/larsoner/python/sphinx-gallery/sphinx_gallery/gen_gallery.py", line 164, in call_memory
    mem, out = memory_usage(func, max_usage=True, retval=True,
  File "/home/larsoner/.local/lib/python3.8/site-packages/memory_profiler.py", line 343, in memory_usage
    returned = f(*args, **kw)
  File "/home/larsoner/python/sphinx-gallery/sphinx_gallery/gen_rst.py", line 464, in __call__
    exec(self.code, self.fake_main.__dict__)
sphinx.errors.ExtensionError: Cannot use input() builtin function in Sphinx-gallery examples
```
Users don't need to care about our internal call stack.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -382,15 +382,26 @@ def generate_dir_rst(src_dir, target_dir, gallery_conf, seen_backrefs):
 
 
 def handle_exception(exc_info, src_file, script_vars, gallery_conf):
+    """Trim and format exception, maybe raise error, etc."""
     from .gen_gallery import _expected_failing_examples
     etype, exc, tb = exc_info
     stack = traceback.extract_tb(tb)
-    # Remove our code from traceback:
-    if isinstance(exc, SyntaxError):
-        # Remove one extra level through ast.parse.
-        stack = stack[2:]
-    else:
-        stack = stack[1:]
+    # The full traceback will look something like:
+    #
+    #   File "/home/larsoner/python/sphinx-gallery/sphinx_gallery/gen_rst.py...
+    #     mem_max, _ = gallery_conf['call_memory'](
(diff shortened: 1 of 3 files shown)


larsoner:
I don't see any suggestion for some reason

sphinx_gallery/gen_rst.py
```diff
@@ -391,6 +391,23 @@ def handle_exception(exc_info, src_file, script_vars, gallery_conf):
         stack = stack[2:]
     else:
         stack = stack[1:]
+    start = 0
```
GaelVaroquaux, line 394: I would add here a big-picture comment on the purpose of this code block.

GaelVaroquaux:
I messed up. So I added it back.

sphinx_gallery/gen_rst.py
larsoner: Done

GaelVaroquaux:
Thanks!!

Issue #728, opened 2020-07-21
How to scrape for images without executing example scripts


mschmidt87:
Hi, 
I am trying to create a gallery of my examples *without* executing the scripts but *with* images. The use case is that I have already executed the scripts before building the documentation and the images were saved to disk. Now I want to build the documentation and scrape the folder for images (png files) to add them to the gallery in the documentation.

As far as I understood, this example is supposed to deal with this case: https://sphinx-gallery.github.io/stable/advanced.html#example-2-detecting-image-files-on-disk
(at least that's how I interpret its documentation).

However, when I include this in my `conf.py`, the scraper is only executed for scripts that are run during `make html`, i.e., during building the docs.

For instance, if my files are not covered by the `filename_pattern` in the `sphinx_gallery_conf`, i.e., not executed, the scraper is not run. 
The same happens if I include them in the pattern, but set `plot_gallery: False`.

What am I doing wrong here?
Thanks, and apologies if this has been asked before, I couldn't find any precedent.

larsoner:
> I am trying to create a gallery of my examples without executing the scripts but with images. The use case is that I have already executed the scripts before building the documentation and the images were saved to disk. Now I want to build the documentation and scrape the folder for images (png files) to add them to the gallery in the documentation.

Keep in mind that when running an example, sphinx-gallery generates several other files in addition to images, for example in addition to creating `images/sphx_glr_plot_0_sin_001.png`, `examples/plot_0_sin.py` produces in the `auto_examples/` dir:
```
-rw-rw-r-- 1 larsoner larsoner     853 Jul 21 13:36 plot_0_sin_codeobj.pickle
-rw-rw-r-- 1 larsoner larsoner    4704 Jul 21 13:36 plot_0_sin.ipynb
-rw-rw-r-- 1 larsoner larsoner    3346 Jul 21 13:35 plot_0_sin.py
-rw-rw-r-- 1 larsoner larsoner      32 Jul 21 13:35 plot_0_sin.py.md5
-rw-rw-r-- 1 larsoner larsoner    4858 Jul 21 13:35 plot_0_sin.rst
```
I'm not sure how many of these you officially need to replace, but at the very least you need to make sure that `plot_0_sin.py.md5` contains what running the script would produce, which should be [some variant of what we do](https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/gen_rst.py#L723-L724):
```
[...]

PR #730, opened 2020-07-27
Improve documentation of example for custom image scraper

mschmidt87:
This PR adds a sentence to the documentation of Example 2 for custom image scraper to make it clearer and prevent misunderstandings like the one documented in #728 .


larsoner:
Thanks @mschmidt87

Issue #731, opened 2020-07-30
Pass command line arguments to examples


mschmidt87:
I have a bunch of examples included in my sphinx-gallery and these examples accept comand-line arguments.
Is it possible to pass command-line arguments through sphinx-gallery? I couldn't find that option in the documentation.

Two use cases would be (in my case):
- control the duration of the execution of examples: It would be nice to control how long the examples take via a command-line arguments (e.g. controlling the number of iterations in some optimization scheme), such that I can build my docs with full functionality in continuous integration, but I'm able to control the runtime so that CI doesn't take too long.
- control the creation of figures. In one example, I am using OpenAI gym, and this causes a problem when I render their environment (rendered with `pyglet`). I receive an error when I execute the example via sphinx-gallery, which is caused by some problem in `pyglet` (as far as I understand). It would be nice to be able to switch off the rendering via the command line.

larsoner:
There is some relevant discussion in #252, and after reading it I can think of two solutions:

1. Allow a `# sphinx_gallery_argv = ..." inline argument to pass the command-line arguments.
2. Allow a `sphinx_gallery_conf['reset_argv']` callable that takes the `script_vars` and `gallery_conf` and does whatever it needs to do to `sys.argv`. By default in sphinx-gallery it would do what's done currentyl, which is equivalent to:

    ```
    import sys
    sys.argv[0] = script_vars['src_file']
    sys.argv[1:] = []
    ```

I prefer (2) since it allows people to control things however they want during doc builds (e.g., via some class), and is also more lightweight from the sphinx-gallery end because we don't need to do any parsing or passing -- just call a callable at the right time (bugs with setting `sys.argv` are at the user's end at that point, not ours).

Any interest in trying to implement this @mschmidt87 ?

mschmidt87:
Sure, I followed your suggestion and opened #732 .

PR #732, opened 2020-07-31
Implement option to pass command line args to example scripts

mschmidt87:
This PR implements an option to pass command line args to example scripts. It closely follows the suggestion by @larsoner made in #731 :

- I added a `reset_argv` option to the configuration which needs to be a callable that accepts `gallery_conf` and `script_vars` as input arguments and return a list of strings which will be the additional `sys.argv` (in addition to the src_file itself).
- As a default value, I set `reset_argv` to be a callable that just returns an empty list, which is equivalent to the current behavior.
- I also added documentation of this option.

The PR is missing a test for this function. I was not sure how to add a test for this and would need some advice since I'm not familiar with the test suite.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -190,9 +190,12 @@ As the patterns are parsed as `regular expressions`_, users are advised to consu
 Passing command line arguments to example scripts
 =================================================
 
-By default, Sphinx-Gallery will not pass any command line arguments to example scripts.
-By setting the ``reset_argv`` option, it is possible to change this behavior and pass command line arguments to example scripts.
-``reset_argv`` needs to be a Callable that accepts the ``gallery_conf`` and ``script_vars`` dictionaries and returns a list of strings that are passed as additional command line arguments to the interpreter.
+By default, Sphinx-Gallery will not pass any command line arguments to example
+scripts.  By setting the ``reset_argv`` option, it is possible to change this
+behavior and pass command line arguments to example scripts.  ``reset_argv``
+needs to be a Callable that accepts the ``gallery_conf`` and ``script_vars``
+dictionaries as input and returns a list of strings that are passed as
+additional command line arguments to the interpreter.
 
 An example could be::
 
@@ -201,6 +204,13 @@ An example could be::
 	    return ['-a', '1']
 	elif script_vars['src_file'] == 'example2.py':
 	    return ['-a', '2']
+
+which is included in the configuration dictionary as::
+
+    sphinx_gallery_conf = {
+        ...
+        'reset_argv': reset_argv,
+    }
 
 which is then used by Sphinx-Gallery as::
 
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -203,7 +203,7 @@ An example could be::
         def __repr__(self):
 	    return 'ResetArgv'
 
-	def __call__(self):
+	def __call__(self, sphinx_gallery_conf, script_vars):
             if script_vars['src_file'] == 'example1.py':
 	        return ['-a', '1']
             elif script_vars['src_file'] == 'example2.py':
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -199,17 +199,21 @@ additional command line arguments to the interpreter.
 
 An example could be::
 
-    def reset_argv(gallery_confg, script_vars):
-        if script_vars['src_file'] == 'example1.py':
-	    return ['-a', '1']
-	elif script_vars['src_file'] == 'example2.py':
-	    return ['-a', '2']
+    class ResetArgv:
+        def __repr__(self):
+	    return 'ResetArgv'
+
+	def __call__(self):
+            if script_vars['src_file'] == 'example1.py':
+	        return ['-a', '1']
+            elif script_vars['src_file'] == 'example2.py':
+	        return ['-a', '2']
 
 which is included in the configuration dictionary as::
 
     sphinx_gallery_conf = {
         ...
-        'reset_argv': reset_argv,
+        'reset_argv': ResetArgv(),
     }
 
 which is then used by Sphinx-Gallery as::
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -19,6 +19,7 @@ file:
 
 - ``examples_dirs`` and ``gallery_dirs`` (:ref:`multiple_galleries_config`)
 - ``filename_pattern`` and ``ignore_pattern`` (:ref:`build_pattern`)
+- ``reset_argv`` (:ref:`reset_argv`)
 - ``subsection_order`` (:ref:`sub_gallery_order`)
 - ``within_subsection_order`` (:ref:`within_gallery_order`)
 - ``reference_url`` (:ref:`link_to_documentation`)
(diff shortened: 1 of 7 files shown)


lucyleeow:
We have a 'tinybuild' with a few examples that are built during testing. I would suggest adding an example to `sphinx_gallery/tests/tinybuild/examples/` that takes command line args (update the `conf.py` of the tinybuild) and check that the resulting output html file is correct in `test_full.py`.

We should also update our sys.argv example: https://sphinx-gallery.github.io/stable/auto_examples/plot_7_sys_argv.html
to provide an example of passing arguments. We can show the function used for the conf as well.

doc/configuration.rst
```diff
@@ -184,6 +185,31 @@ As the patterns are parsed as `regular expressions`_, users are advised to consu
...
         $ sphinx-build -D sphinx_gallery_conf.filename_pattern=plot_specific_example\.py ...
 
+.. _reset_argv:
+
+Passing command line arguments to example scripts
+=================================================
+
+By default, Sphinx-Gallery will not pass any command line arguments to example scripts.
+By setting the ``reset_argv`` option, it is possible to change this behavior and pass command line arguments to example scripts.
+``reset_argv`` needs to be a Callable that accepts the ``gallery_conf`` and ``script_vars`` dictionaries and returns a list of strings that are passed as additional command line arguments to the interpreter.
```
lucyleeow, line 195: Nitpick:
```suggestion
``reset_argv`` needs to be a Callable that accepts the ``gallery_conf`` and ``script_vars`` dictionaries as input and returns a list of strings that are passed as additional command line arguments to the interpreter.
```

also this line seems longer than 80 characters?

doc/configuration.rst
```diff
@@ -184,6 +185,31 @@ As the patterns are parsed as `regular expressions`_, users are advised to consu
...
+
+An example could be::
+
+    def reset_argv(gallery_confg, script_vars):
+        if script_vars['src_file'] == 'example1.py':
+	    return ['-a', '1']
+	elif script_vars['src_file'] == 'example2.py':
+	    return ['-a', '2']
+
```
lucyleeow, line 204: Just to be explicit could we also add an example of how to set the conf, underneath defining the function, e.g.,:

```
    sphinx_gallery_conf = {
        ...
        'reset_argv': reset_argv(gallery_confg, script_vars),
    }
```
(not sure that I've done it right there)

larsoner: It would just be:
```
    sphinx_gallery_conf = {
        ...
        'reset_argv': reset_argv,
    }
```

lucyleeow: > It would just be:

:sweat: that's what I thought initially

doc/configuration.rst
mschmidt87: Thanks, updated in fcf0db3ba3253cf0e0b793a2dd4f395296d9a968

doc/configuration.rst
mschmidt87: Thanks, updated in fcf0db3ba3253cf0e0b793a2dd4f395296d9a968 and a674033f414ccb15434b4a52b6af0ad171ec15d5

mschmidt87:
Thank you for your feedback @lucyleeow and @larsoner . I modified the example on usage of `sys.argv` with a simple cross-reference to the documentation on `reset_argv`.

I also added a simple test. However, at least on my system now the `test_full::test_rebuild` test is failing and I can't figure out why. Perhaps someone with more knowledge of the test suite could take a look at that?

doc/configuration.rst
mschmidt87: I also changed the example from a function to a class. However, @larsoner , could you explain why a stable `__repr__` is important? Is this used by sphinx-gallery internally to determine if there were any changes to the code?

lucyleeow:
I think it is due to the stable `__repr__` @larsoner mentioned:

> the given example is problematic because it does not have a stable __repr__ (it will change from sphinx build to sphinx build) and hence will cause rebuilds. 

We use `hashlib.md5` to check if an example file has changed and thus if it needs to be rebuilt. Apparently you need a stable `__repr__` to ensure the hash is the same across builds. I think changing the `reset_argv` function for tinybuild to have a `__repr__`, like in the example, should fix the test failure.

Though @larsoner maybe you could expand on the stable `__repr__`. I thought we read the `.py` files as text and hash? How does `__repr__` feature?

larsoner:
Same idea as [here](https://sphinx-gallery.github.io/stable/configuration.html#sorting-gallery-subsections), it's a Sphinx issue not a Sphinx-gallery one:

> If you create your own class for 'subsection_order', ensure that the __str__ of your class is stable across runs. Sphinx determines if the build environment has changed (and thus if all documents should be rewritten) by examining the config values using md5(str(obj).encode()).hexdigest() in sphinx/builders/html.py. Default class instances in Python have their memory address in their __repr__ which will in general change for each build.

lucyleeow:
Thanks @larsoner. Since `str(obj)` is called, would seting a `__str__()` method instead work as well?

larsoner:
I think it would

mschmidt87:
Do you have an idea why the `test_rebuild` fails, @lucyleeow , @larsoner ?

larsoner:
We explicitly test that not all examples are rebuilt on a second sphinx run. The error is:
```
E   AssertionError: ['\x1b[01mbuilding [html]\x1b[39;49;00m: targets for 25 source files that are out of date']
```
because it's looking for:
```
'.*targets for [2|3|4] source files that are out of date$.*
```
suggesting that all source files are being rebuilt. This suggests to me that the `config` is probably being seen by Sphinx as "changed". 

Looking at the code, you just use a function instead of a class with a stable `str(...)`. You need to do what we discussed above and recommend in the docs (use a class with a `__repr__` or `__str__`) to avoid this.

sphinx_gallery/gen_gallery.py
```diff
@@ -41,6 +41,7 @@
     'filename_pattern': re.escape(os.sep) + 'plot',
     'ignore_pattern': r'__init__\.py',
     'examples_dirs': os.path.join('..', 'examples'),
+    'reset_argv': lambda gallery_conf, script_vars: [],
```
larsoner, line 44: Unstable repr:
```
(base) bunker:~ larsoner$ python -c "print(lambda x: None)"
<function <lambda> at 0x7ffdf80cd160>
(base) bunker:~ larsoner$ python -c "print(lambda x: None)"
<function <lambda> at 0x7f95b80cd160>
```
This needs to be a class with a stable `repr` to avoid always rebuilding docs for people who otherwise don't specify a value.

sphinx_gallery/tests/tinybuild/conf.py
```diff
@@ -16,6 +16,13 @@ def __call__(self, block, block_vars, gallery_conf):
         return matplotlib_scraper(block, block_vars, gallery_conf, **kwargs)
 
 
+def reset_argv(sphinx_gallery_conf, script_vars):
```
larsoner, line 19: This one, too

sphinx_gallery/gen_gallery.py
mschmidt87: Thank you, I missed that. Fixed it not now 8f1f10c0e6d16b25d5f73b9dbc4948485ae242f7

sphinx_gallery/tests/tinybuild/conf.py
mschmidt87: Thanks, fixed it in aea11787f7d8e452ffe03169975810c79ae466aa

mschmidt87:
Thank you @larsoner , I replaced the functions with classes and now the testsuite builds (locally at least).

larsoner:
[Style](https://travis-ci.org/github/sphinx-gallery/sphinx-gallery/jobs/714447774#L1150):
```
sphinx_gallery/tests/test_full.py:186:1: W293 blank line contains whitespace
sphinx_gallery/tests/tinybuild/examples/plot_command_line_args.py:15:1: W293 blank line contains whitespace
sphinx_gallery/tests/tinybuild/examples/plot_command_line_args.py:17:5: E303 too many blank lines (2)
```

larsoner:
Thanks @mschmidt87

PR #733, opened 2020-08-02
Enable ffmpeg for animations for newer matplotlib

dopplershift:
Since ffmpeg GIF writing was fixed in matplotlib/matplotlib#18093, which will be included in matplotlib 3.3.1, enable using that for capturing animations (when running a recent enough version).

### DIFF
--- a/sphinx_gallery/scrapers.py
+++ b/sphinx_gallery/scrapers.py
@@ -14,7 +14,7 @@ live in modules that will support them (e.g., PyVista, Plotly).
 import os
 import sys
 import re
-from distutils import LooseVersion
+from distutils.version import LooseVersion
 from textwrap import indent
 from warnings import filterwarnings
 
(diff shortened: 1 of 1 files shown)


larsoner:
Pushed a couple of trivial commits to fix SyntaxError and NameError. I can test on a machine that has matplotlib `master` on it tomorrow on SG `master` and this PR and make sure it works.

larsoner:
Whoops, wrong button...

dopplershift:
Every time I think I can write correct code the first time without running it, I get proven wrong.

larsoner:
`master` produces a 126k gif:

sphx_glr_plot_8_animations_thumb

Hacking the version check to use 3.3.0 (latest `master` still installs on my version as a `3.3.0.something`) it produces a 16 kB one:

sphx_glr_plot_8_animations_thumb

Awesome! Thanks @dopplershift

dopplershift:
🎉 Glad it had a good impact.

PR #745, opened 2020-09-05
BUG Copy Binder logo to avoid Window drive rel path error

### LINKED ISSUE #744: BUG Windows relative path error with _static Binder logo
In Windows, when the package _static files are in a different drive to the `gallery_conf['src_dir']`, `os.path.relpath` gives an error:
```
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\site-packages\sphinx_gallery\binder.py", line 112, in gen_binder_rst
    binder_logo_path = os.path.relpath(binder_logo, gallery_conf['src_dir'])
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\ntpath.py", line 584, in relpath
    path_drive, start_drive))
ValueError: path is on mount 'C:', start on mount 'D:
```

See: https://github.com/scikit-image/scikit-image/issues/4958 and https://github.com/scikit-image/scikit-image/pull/4959

I can copy the binder logo to `target_dir`/images (or just `target_dir` ?) like I was doing in the first iter of #738 .

Also what would be the best way to add a test for this. Can we change `appveyor.yml` such that Sphinx-Gallery is installed in a different drive to where the testing occurs?

WDYT @larsoner ?

(I'll do a bug fix release when fixed)

lucyleeow:
Fixes #744

Copy Binder logo from package `_static` folder to `target_dir`/images.

@larsoner happy to explore another solution.

Test: would changing appveyor setup such that `_static` is in a different drive be a good test? I am not so familiar with appveyor but if I add a `cd D:` here:
https://github.com/sphinx-gallery/sphinx-gallery/blob/839c02b0b076f20ea7cc0ca57a619ac7f00467f5/appveyor.yml#L28-L29

would that work?


larsoner:
I suspect you shouldn't need to manually copy any static file, just link to it properly. The question is how to do that. Usually / hopefully you can do it by doing a relpath between the example file and the `src_dir` or so (?), or maybe get the static path from Sphinx. I can dig into it Tuesday if we can wait that long

lucyleeow:
The problem is that relative paths between 2 different drives on Windows doesn't exist. It may be possible with symbolic links but we might not want to mess with that.

No hurry, take a look on Tuesday.

larsoner:
@lucyleeow pushed a commit. The critical point is that, regardless of what `html_static_path`s users enter, they always end up in `_static/`. So we just need to link directly to `OUT_DIR/_static/image.svg`. See if it makes sense and if so, feel free to merge and cut a 0.8.1

lucyleeow:
I think I've just missed something important. So all files in the package `_static` dir (i.e. `<where you installed SG>/sphinx-gallery/sphinx_gallery/_static`) are always copied over to the built html (e.g., `<where your package is>/doc/_build/html/_static/`)? 

Can you show me where this is done in the code?

larsoner:
> Can you show me where this is done in the code?

It's done by `sphinx` itself somewhere (feel free to dig through their code to figure this out). If you look at `tinybuild` I modified it to use `html_static_paths = ['_static_nonstandard']`, but when you `make html`, the static files from that dir and from `glr_path_static` (because we add it to the static paths) get copied to OUT_DIR/_static. AFAIK this is a hard-coded output path

lucyleeow:
Thanks, I had no idea Sphinx did this but it makes sense.

larsoner:
Feel free to fix the path error in CIs, or I can push a fix in a couple of hours. Have meetings now

larsoner:
Actually I think that all of those build warnings are bogus, they say files don't exist but it renders fine:

https://1811-25860190-gh.circle-artifacts.com/0/rtd_html/auto_examples/plot_0_sin.html#sphx-glr-auto-examples-plot-0-sin-py

I'm not sure how to avoid this warning. Maybe worth asking on the Sphinx listserv or GitHub. But we do need to make these warnings go away somehow because this will cause `-nW` builds fail. :(

larsoner:
This might also be relevant

https://github.com/sphinx-gallery/sphinx-gallery/issues/120

And maybe

https://stackoverflow.com/questions/52138336/sphinx-reference-to-an-image-from-different-locations#answer-52141199

lucyleeow:
I'll do this tomorrow

larsoner:
Actually I wonder if the correct fix is to give the path to the `sphx_glr_static()` without any `relpath` business. Sphinx in theory should recognize that this is a path to a static file and substitute OUT_DIR/static. I'll give it a shot...

larsoner:
Okay @lucyleeow I ended up giving up on the static idea and ended up with something that I think is pretty close to what you had. Can you take a look and merge if you're happy?

sphinx_gallery/binder.py
```diff
@@ -105,18 +104,30 @@ def gen_binder_rst(fpath, binder_conf, gallery_conf):
...
+    #     https://github.com/sphinx-doc/sphinx/issues/7772
+    # And in any case, it does not work on Windows, so here we just for each
+    # gallery copy the SVG to `images` and link to it there. This will make
+    # a few copies, and there will be an extra in `_static` at the end of the
+    # build, but it at least works...
+    physical_path = os.path.join(
+        os.path.dirname(fpath), 'images', 'binder_badge_logo.svg')
+    os.makedirs(os.path.dirname(physical_path), exist_ok=True)
+    if not os.path.isfile(physical_path):
+        shutil.copyfile(
```
lucyleeow, line 121: I think for the test `test_gen_binder_rst` we shouldn't copy the file over

lucyleeow: Nevermind, `isfile` is better than `os.path.exists` that I used.

larsoner: I'll try some `tmpdir` gymnastics...

lucyleeow:
Sorry can you clarify what `glr_path_static` returns? Is it the packaged `_static`, where the package files are or the path to the the `_static` folder where the built html files are?

sphinx_gallery/tests/test_binder.py
```diff
@@ -115,9 +116,10 @@ def apptmp():
     assert url == expected
 
 
-def test_gen_binder_rst():
+def test_gen_binder_rst(tmpdir):
     """Check binder rst generated correctly."""
     gallery_conf_base = {'gallery_dirs': None, 'src_dir': 'blahblah'}
```
lucyleeow, line 121: I set `'gallery_dirs': None` to avoid copying files over when testing, but we can change this back to what it was before

larsoner: Pushed a commit to be more like how it should in actually building

larsoner:
`glr_path_static` is the path to sphinx-gallery's `sphinx_gallery/_static` folder. The contents will be copied to `OUT_DIR/_static` on build. The problem is that we can't actually use it (see code comments)

lucyleeow:
Okay thanks. But in Windows, if the path has to be relative to the build root, and the build root is in a different drive to `glr_path_static`, I thought there was not relative path between different drives?
Sorry maybe I missed something?

larsoner:
> Sorry maybe I missed something?

`relpath` is no longer used

sphinx_gallery/tests/test_full.py
```diff
@@ -445,7 +444,8 @@ def test_rebuild(tmpdir_factory, sphinx_app):
         new_app.build(False, [])
     status = new_app._status.getvalue()
     lines = [line for line in status.split('\n') if '0 removed' in line]
-    assert re.match('.*[0|1] added, [1-9] changed, 0 removed$.*',
+    # XXX on AppVeyor this can be 13
```
lucyleeow, line 447: What can be 13?

larsoner: The number of changed files in the regex below

lucyleeow: Ah okay. DI]id this change?

larsoner: It did in previous runs because I added `binder` to the tinybuild conf. Not sure if it's still there or not but I'm okay living with it

lucyleeow: Ahh I forgot binder changed. Can we change to:
`assert re.match('.*[0|1] added, [1-9][0-3]? changed, 0 removed$.*',`

larsoner: Sure, feel free to push. You could even make the `[1-9][0-3]?` only if on windows and use `[1-9]` otherwise

PR #748, opened 2020-09-11
Enable animations HTML to be rendered on readthedocs

sdhiscocks:
This removes the `only` directive, leaving `raw` directive to allow builder to decide how/if to render raw HTML.

This is similar to issue in #700


larsoner (approved):
LGTM +1 for merge once CIs come back happy

larsoner:
`git grep` says this is the only instance of `builder_html` in our codebase, hopefully we (I) can stop adding new instances :)

PR #750, opened 2020-09-28
ENH: Add RST comments, read-only

### LINKED ISSUE #725: Relate warnings and errors on generated rst file back to source Python file / prevent accidental writing of generated files
When there is a problem in the generated RST file from a Gallery Python file, Sphinx reports the line error in the generated rst file. There's a few problems with this:

1. If you're not paying close attention, you might try editing the RST file to fix the error. But this is the wrong thing to do: you should have edited the source Python file. Some ways to make it harder to make this mistake include (1) removing the write bit from generated files, and (2) inserting a "@generated" comment which is respected by many editors and will warn people when editing
2. It would have been much better to report the error in the source Python file in the first place.

Compilers with the C preprocessor have solved this problem by generating comment lines that let you correspond lines in the generated source back to the original source, and then having the eventual frontend report errors by first translating them back to the original source. sphinx-gallery should do this too. It might be difficult to fix this without also getting Sphinx to play ball (e.g., a logical separation would be for Sphinx to add support for file/line number comments, and then teaching sphinx-gallery to respect them).

Is a feature like this something that would be accepted by sphinx-gallery?

larsoner:
There are parts of #725 that have to do with better identifying what the reported RST errors *mean*, but that's really an upstream docutils bug/enhancement, thus:

Closes #725.


sphinx_gallery/gen_rst.py
```diff
@@ -138,8 +138,7 @@ def __exit__(self, type_, value, tb):
     :class: sphx-glr-single-img
 """
 
-# This one could contain unicode
-CODE_OUTPUT = u""".. rst-class:: sphx-glr-script-out
+CODE_OUTPUT = """.. rst-class:: sphx-glr-script-out
```
lucyleeow, line 141: Why did we allow unicode in the first place...?

sphinx_gallery/gen_rst.py
```diff
@@ -902,9 +927,14 @@ def rst_blocks(script_blocks, output_blocks, file_conf, gallery_conf):
     # A simple example has two blocks: one for the
     # example introduction/explanation and one for the code
     is_example_notebook_like = len(script_blocks) > 2
-    example_rst = u""  # there can be unicode content
-    for (blabel, bcontent, lineno), code_output in \
-            zip(script_blocks, output_blocks):
+    example_rst = ""
+    for bi, ((blabel, bcontent, lineno), code_output) in \
```
lucyleeow, line 931: why `bi` .. ? block index?

lucyleeow (commented):
LGTM! Thanks! I just had some questions.

sphinx_gallery/gen_rst.py
larsoner: yes I just like to avoid `i` and `j` so I usually prefix with something at least somewhat meaningful but still short

sphinx_gallery/gen_rst.py
larsoner: We still do implicitly because you just don't need `u` in Python3

lucyleeow:
Thanks, feel free to merge!

PR #752, opened 2020-10-06
ENH: Style HTML output like jupyter

larsoner:
I dug into the Jupyter notebook default formatting a bit, and this allows us to hopefully style our HTML outputs similarly to theirs. See for example:

https://github.com/mne-tools/mne-nirs/pull/167#issuecomment-698277516

On my machine on this PR I now get:

Screenshot from 2020-10-06 09-31-38

### DIFF
--- a/examples/plot_3_capture_repr.py
+++ b/examples/plot_3_capture_repr.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """
-.. _capture-repr-examples:
+.. _capture_repr_examples:
 
 Capturing output representations
 ================================
--- a/examples/plot_3_capture_repr.py
+++ b/examples/plot_3_capture_repr.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
-r"""
+"""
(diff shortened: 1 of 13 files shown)


examples/plot_3_capture_repr.py
```diff
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
-r"""
+"""
+.. _capture-repr-examples:
```
lucyleeow, line 3: Super nitpick, can we change to `_` like with other ref labels?

lucyleeow (approved):
LGTM. I am not very knowledgeable about CSS though - I am not sure what the div classes e.g., `output_subarea` mean as I don't see these classes in the CSS, just `rendered_html`.

larsoner:
> I am not sure what the div classes e.g., output_subarea mean as I don't see these classes in the CSS, just rendered_html.

The reason for this is just for future compat / configurability for users -- these are the classes I saw when running in a Jupyter notebook so it seemed to make sense to copy them all.

PR #756, opened 2020-10-09
Add show_signature option

jschueller:
Allows to disable the "Generated by Sphinx-gallery" notice


larsoner:
Looks straightforward enough to me. We can add some tiny test for this at some point maybe, but seems overkill for now. Thanks @jschueller !

Issue #757, opened 2020-10-14
Option to disable note in example header


josephmure:
Hello! Is there an option to disable the note linking to the download buttons in the example header?

larsoner:
You can override the CSS in your own static CSS file like:
```
div.sphx-glr-download-link-note {
  height: 0px;
  visibility: hidden;
}
```

larsoner:
(solution from https://github.com/mne-tools/mne-python/blob/master/doc/_static/style.css#L240-L244)

josephmure:
Thanks @larsoner! That is exactly what I was looking for.

PR #758, opened 2020-10-15
Add show_download_note option

jschueller:


larsoner:
I would rather not fix this using a config var but rather have people use CSS:

https://github.com/sphinx-gallery/sphinx-gallery/issues/757

Can we just document this CSS-based solution instead?

GaelVaroquaux:
> Can we just document this CSS-based solution instead?

+1

jschueller:
ok

PR #759, opened 2020-10-16
ENH: Add run_stale_examples config var

larsoner:
Makes it much easier to have a Makefile target to force re-running of a specific example/examples. See extended explanation in `configuration.rst`.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -19,6 +19,7 @@ file:
 
 - ``examples_dirs`` and ``gallery_dirs`` (:ref:`multiple_galleries_config`)
 - ``filename_pattern`` and ``ignore_pattern`` (:ref:`build_pattern`)
+- ``run_stale_examples`` (:ref:`run_stale_examples`)
 - ``reset_argv`` (:ref:`reset_argv`)
 - ``subsection_order`` (:ref:`sub_gallery_order`)
 - ``within_subsection_order`` (:ref:`within_gallery_order`)
@@ -155,10 +156,7 @@ one is targeting a specific file, it would match the dot in the filename even wi
 
 .. note::
     Sphinx-gallery only re-runs examples that have changed (according to their
-    md5 hash). You can delete the associated MD5 files (e.g.,
-    ``./auto_examples/plot_awesome_example.py.md5``) to force a rebuild if
-    you have not changed the example itself between subsequent ``sphinx``
-    calls.
+    md5 hash). See :ref:`run_stale_examples` below for information.
 
 Similarly, to build only examples in a specific directory, you can do::
 
@@ -185,6 +183,46 @@ As the patterns are parsed as `regular expressions`_, users are advised to consu
     .. code-block:: console
 
         $ sphinx-build -D sphinx_gallery_conf.filename_pattern=plot_specific_example\.py ...
+
+.. _run_stale_examples::
+
+Rerunning stale examples
+========================
+By default, sphinx-gallery only rebuilds examples that have changed.
+For example, starting from a clean ``doc/`` directory, running your HTML build
+once will result in Sphinx-gallery executing all examples that match your given
+filename / ignore patterns. Then, running
+the exact same command a second time *should not run any examples*, because the
+MD5 hash of each example will be checked against the MD5 hash (saved to disk
+as ``<filename>.md5`` in the generated directory) that the example file had
+during the first build, they will match and thus the example will be determined
+to be "stale", and it will not be rebuilt by Sphinx-gallery.
+This design feature allows for more rapid documentation iteration by only
+rebuilding examples when they change.
+
+However, this presents a problem during some modes of debugging and
+iteration. Let's say that you have one particular
+example that you want to rebuild repeatedly while modifying some function in
+your underlying library, while leaving the example file contents themselves
+fixed. To do this, you'd either need to make some change (e.g., add/delete a
+newline) to your example or delete the ``.md5`` file to force Sphinx-gallery
+to rebuild the example. Instead, you can use the configuration value::
+
+    sphinx_gallery_conf = = {
+        ...
+        'run_stale_examples': True,
+    }
+
+This can be combined with other options to repeatedly rerun a single example
+from the command line for example by doing some variant of:
+
+.. code-block:: console
+
+    $ make html SPHINXOPTS='-D sphinx_gallery_conf.run_stale_examples=True' -Dsphinx_gallery_conf.filename_pattern='my_example_name'``
+
+This will cause any examples matching the filename/ignore patterns to be
+rebuilt, regardless of their MD5 hashes.
+
 
 .. _reset_argv:
 
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -191,7 +191,7 @@ Rerunning stale examples
 By default, sphinx-gallery only rebuilds examples that have changed.
 For example, when starting from a clean ``doc/`` directory, running your HTML
 build once will result in Sphinx-gallery executing all examples that match your
-given :ref:`filename/ignore patterns <build pattern>`. Then, running
+given :ref:`filename/ignore patterns <build_pattern>`. Then, running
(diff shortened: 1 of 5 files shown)


larsoner:
Needed to move nightly to allowed failures until NumPy has 3.10 pre wheels. Added a 3.8 build (overdue) but no 3.9 until NumPy, matplotlib, etc. have time to get wheels out.

larsoner:
I think codecov is okay (96.66%). @lucyleeow feel free to review/merge

doc/configuration.rst
```diff
@@ -186,6 +184,46 @@ As the patterns are parsed as `regular expressions`_, users are advised to consu
 
         $ sphinx-build -D sphinx_gallery_conf.filename_pattern=plot_specific_example\.py ...
 
+.. _run_stale_examples:
+
+Rerunning stale examples
+========================
+By default, sphinx-gallery only rebuilds examples that have changed.
+For example, starting from a clean ``doc/`` directory, running your HTML build
```
lucyleeow, line 192:
```suggestion
For example, when starting from a clean ``doc/`` directory, running your HTML build
```
?

doc/configuration.rst
```diff
@@ -186,6 +184,46 @@ As the patterns are parsed as `regular expressions`_, users are advised to consu
...
+once will result in Sphinx-gallery executing all examples that match your given
+filename / ignore patterns. Then, running
```
lucyleeow, line 194: :ref:`<filename / ignore patterns> build_pattern`

doc/configuration.rst
```diff
@@ -186,6 +184,46 @@ As the patterns are parsed as `regular expressions`_, users are advised to consu
...
+the exact same command a second time *should not run any examples*, because the
+MD5 hash of each example will be checked against the MD5 hash (saved to disk
+as ``<filename>.md5`` in the generated directory) that the example file had
+during the first build, they will match and thus the example will be determined
```
lucyleeow, line 198: Sentence is a little long. Maybe:
```suggestion
during the first build. These will match and thus the example will be determined
```

doc/configuration.rst
```diff
@@ -186,6 +184,46 @@ As the patterns are parsed as `regular expressions`_, users are advised to consu
...
+to be "stale", and it will not be rebuilt by Sphinx-gallery.
+This design feature allows for more rapid documentation iteration by only
+rebuilding examples when they change.
+
+However, this presents a problem during some modes of debugging and
+iteration. Let's say that you have one particular
+example that you want to rebuild repeatedly while modifying some function in
+your underlying library, while leaving the example file contents themselves
+fixed. To do this, you'd either need to make some change (e.g., add/delete a
```
lucyleeow, line 207:
```suggestion
your underlying library but do not want to change the example file contents
themselves. To do this, you'd either need to make some change (e.g., add/delete a
```

doc/configuration.rst
```diff
@@ -186,6 +184,46 @@ As the patterns are parsed as `regular expressions`_, users are advised to consu
...
+newline) to your example or delete the ``.md5`` file to force Sphinx-gallery
+to rebuild the example. Instead, you can use the configuration value::
+
+    sphinx_gallery_conf = = {
+        ...
+        'run_stale_examples': True,
+    }
+
```
lucyleeow, line 215: With this configuration all examples will be rebuilt, even if their MD5 hash show that the examples have not changed.

doc/configuration.rst
```diff
@@ -186,6 +184,46 @@ As the patterns are parsed as `regular expressions`_, users are advised to consu
...
+This can be combined with other options to repeatedly rerun a single example
```
lucyleeow, line 216:
```suggestion
You can combine this configuration with :ref:<filename / ignore patterns> build_pattern to repeatedly rerun a single example.
```

doc/configuration.rst
```diff
@@ -186,6 +184,46 @@ As the patterns are parsed as `regular expressions`_, users are advised to consu
...
+from the command line for example by doing some variant of:
```
lucyleeow, line 217:
```suggestion
This could be done from the command line, for example:
```

doc/configuration.rst
```diff
@@ -186,6 +184,46 @@ As the patterns are parsed as `regular expressions`_, users are advised to consu
...
+
+.. code-block:: console
+
+    $ make html SPHINXOPTS='-D sphinx_gallery_conf.run_stale_examples=True' -Dsphinx_gallery_conf.filename_pattern='my_example_name'``
+
+This will cause any examples matching the filename/ignore patterns to be
```
lucyleeow, line 223:
```suggestion
This command will cause any examples matching the filename pattern 'my_example_name' to be
```

lucyleeow (commented):
Some nits and questions - sorry `test_rebuild` was confusing for me.

Also could we add to the description of `test_rebuild`:
```
# Make sure that examples that haven't been changed aren't run twice.
# Also checks `run_stale_examples` config works.
```
(also sorry for delay, I have been busy).

doc/configuration.rst
```diff
@@ -186,6 +184,46 @@ As the patterns are parsed as `regular expressions`_, users are advised to consu
 
         $ sphinx-build -D sphinx_gallery_conf.filename_pattern=plot_specific_example\.py ...
 
+.. _run_stale_examples:
+
+Rerunning stale examples
+========================
+By default, sphinx-gallery only rebuilds examples that have changed.
+For example, starting from a clean ``doc/`` directory, running your HTML build
+once will result in Sphinx-gallery executing all examples that match your given
```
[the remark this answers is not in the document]
larsoner: There appear to be 15 of each kind in `configuration.rst`, plus 30 for Sphinx-Gallery. Not sure if we have a preference. Let's fix all of these in another PR

larsoner:
Thanks for the nice wording changes, pushed @lucyleeow

lucyleeow:
Thanks for the changes. I am still confused by `test_rebuild`. The status I get for both 'run_stale' and 'modify' rebuilds (when I test locally) are both:
```
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 27 source files that are out of date
```
Is it for some CI that the 'run_stale' and 'modify' give different values? Also why do I have 'po' after the number?

Finally, are we getting so many 'changed' files because we added binder to the tinybuild config (https://github.com/sphinx-gallery/sphinx-gallery/pull/745#discussion_r485744614) ? (I vaguely remember this but forgot details)

sphinx_gallery/tests/test_full.py
```diff
@@ -522,32 +527,67 @@ def test_rebuild(tmpdir_factory, sphinx_app):
...
+    lines = '\n'.join([how] + lines)
+    flags = re.MULTILINE | re.DOTALL
+    # for some reason, setting "confoverrides" above causes Sphinx to show
+    # all targets out of date, even though they haven't been modified...
+    if how == 'run_stale':
+        want = '.*targets for %s source files that are out of date$.*' % N_RST
+        assert re.match(want, status, flags) is not None, lines
+    # ... but then later detects that only some have actually changed
+    # (sometimes this is 8, other times 9, depending on Windows/Linux for some
+    # unknown reason)
```
lucyleeow, line 580: For me 'run_stale' gives '8 changed' and 'modify' gives '9 changed' - which seems to make sense as we modify a file in 'modify'...?

larsoner: The reason I had to put `[8|9]` was from 5164de3 -- Travis (Linux) passed with 9 but AppVeyor (Windows) wanted 8.

larsoner:
> Is it for some CI that the 'run_stale' and 'modify' give different values?

If you change the order IIRC (as I had it originally, `modify` was before `run_stale`) yes you get different values. I'll get rid of the conditional now, though, since the order has to be `modify` last (which means both runs get the all-files-out-of-date version).

>  Also why do I have 'po' after the number?

I have no idea what this means.

I think the bottom line is that we just need to get these to pass

lucyleeow:
Thanks @larsoner !

PR #763, opened 2020-10-27
DOC Amend run_stale_examples command in configuration.rst

lucyleeow:
I might have gotten this wrong as I am not familiar with SPHINXOPTS syntax but removed backticks and added space after `-D`

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -221,7 +221,7 @@ This could be done from the command line, for example:
 
 .. code-block:: console
 
-    $ make html SPHINXOPTS='-D sphinx_gallery_conf.run_stale_examples=True' -D sphinx_gallery_conf.filename_pattern='my_example_name'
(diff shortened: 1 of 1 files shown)


doc/configuration.rst
```diff
@@ -221,7 +221,7 @@ This could be done from the command line, for example:
 
 .. code-block:: console
 
-    $ make html SPHINXOPTS='-D sphinx_gallery_conf.run_stale_examples=True' -Dsphinx_gallery_conf.filename_pattern='my_example_name'``
+    $ make html SPHINXOPTS='-D sphinx_gallery_conf.run_stale_examples=True' -D sphinx_gallery_conf.filename_pattern='my_example_name'
```
lucyleeow, line 224: Also why is the first opt in quotes but not the second...?

larsoner: the whole thing should probably be in the single quotes. And space after the D or not shouldn't matter so if it looks better with a space that's fine

lucyleeow:
@larsoner I tried this and it works. Had to use double quotes as we use single quotes for the file name pattern.

larsoner:
Thanks @lucyleeow

PR #768, opened 2020-11-05
FIX: Restore whitespace

larsoner:
@lucyleeow feel free to merge if you're happy with this solution


lucyleeow:
I moved the white space to directly above the heading (not sure if it will make a difference, it's just where it was before) - it looks different on the git diff but the white space character is still the same (and doesn't show up on my editor as white space)

lucyleeow:
Thanks @larsoner

PR #778, opened 2020-12-23
Point to up-to-date re documentation

dstansby:
I noticed this link points to python 2; I've updated it so it will now point to python 3.


choldgraf:
thanks very much!!

Issue #777, opened 2020-12-20
test_rebuild and test_error_messages failing on travis


lucyleeow:
`test_rebuild` is failing here:
https://github.com/sphinx-gallery/sphinx-gallery/blob/ce5a58cfde1c030364c0bd34b21f2ba00b83896f/sphinx_gallery/tests/test_full.py#L636

when checking that the example `plot_numpy_matplotlib` which we modify above, has a different mtime:
```
sphinx_gallery/tests/test_full.py:387: in _assert_mtimes

    assert np.abs(op.getmtime(orig) - op.getmtime(new)) > 0.1

E   AssertionError: assert 0.0 > 0.1

E    +  where 0.0 = <ufunc 'absolute'>((1607961214.223356 - 1607961214.223356))
```
`test_error_messages` failing here:
```
_ test_error_messages[plot_scraper_broken-.*ValueError.*zero-size array to reduction.*] _

sphinx_gallery/tests/test_full.py:661: in test_error_messages

    assert re.match(want, rst) is not None

E   AssertionError: assert None is not None
```

@larsoner you probably know more about this but otherwise I can take a look next week.

larsoner:
I did not observe this on https://github.com/sphinx-gallery/sphinx-gallery/pull/775 . Sometimes those tests are a bit unreliable, I would try restarting the one that failed

lucyleeow:
Gah it didn't run on #775 because we didn't have enough credits: https://travis-ci.com/github/sphinx-gallery/sphinx-gallery/requests

lucyleeow:
I think we could apply for more credits, or switch e.g., to github actions?

larsoner:
I would just switch to GitHub actions, or maybe easier make a more complete set of Azure runs that encompass what Travis does. Do you want to try @lucyleeow ?

lucyleeow:
I can have a go this weekend. What do you think would be best, GitHub actions or Azure? Both would be of similar difficulty for me (I've fiddled a bit with GitHub actions but never really touched Azure, though we have existing Azure yml to build on).

larsoner:
Azure has better single job restart control and we already have some YAML so let's just do that.

lucyleeow:
Closing as it seems it's due to this test not being reliable in CI

PR #779, opened 2020-12-26
MAINT: Move travis CI jobs to Azure

lucyleeow:
Move Travis CI jobs to Azure due to cap on credits in Travis.
see: https://github.com/sphinx-gallery/sphinx-gallery/issues/777#issuecomment-749195163


lucyleeow:
Lots of things are not working:

* ubuntu_python36 - getting this error during `install.sh`: `[Errno 13] Permission denied: '/usr/local/lib/python3.6/dist-packages/test-easy-install-3620.write-test'`
* conda_python35 / conda_python35_localeC - getting `/home/vsts/work/1/s/continuous_integration/azure/test_script.sh: line 9: pytest: command not found` when trying to run the test script. Not sure why as `conda list` in the prev step showed pytest.
* I think the builds with mayavi are failing because I haven't used xvfb

and bit stuck on these.

Also not sure about best solution for python nightly - I've seen that we can deadsnakes but not sure what best solution is.

azure-pipelines.yml
```diff
@@ -49,3 +49,71 @@ jobs:
...
+      conda_python38_sphinxDev:
+        DISTRIB: 'conda'
+        PYTHON_VERSION: '3.8'
+        SPHINX_VERSION: 'dev'
+      python37_minimal:
+        DISTRIB: 'minimal'
+        PYTHON_VERSION: '3.7'
+      # pythonNightly:
+      #   DISTRIB: 'Nightly'
+  condition: not(contains(variables['Build.SourceVersionMessage'], '[doc only]'))
```
lucyleeow, line 88: This also doesn't work

larsoner: I wouldn't worry about `[doc only]` but instead use standard `[ci skip]` and what should someday be supported `[skip azp]` and even then it probably won't be used too often since our CIs are fast. You can use this code:

https://github.com/mne-tools/mne-python/blob/master/azure-pipelines.yml#L16-L34

larsoner:
> getting this error during install.sh

You probably need to use `--user`, on Linux standard python installs the user doesn't have permissions to `dist-packages` (you could use `sudo` but this is uglier)

> conda_python35

Let's just remove 3.5 in `setup.py` and require 3.6+

> I think the builds with mayavi are failing because I haven't used xvfb

Yes you'll want to spin up xvfb with something like

https://github.com/mne-tools/mne-python/blob/master/azure-pipelines.yml#L107-L109

after installing with apt

azure-pipelines.yml
lucyleeow: If I got the right page, supposedly [ci skip] is meant to work already: https://github.com/Microsoft/azure-pipelines-agent/issues/858#issuecomment-475768046
I'll test it out.

lucyleeow:
Not sure if I have done it correctly but adding:
```
  - bash: |
      sudo apt-get install xvfb
      /sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -screen 0 1400x900x24 -ac +extension GLX +render -noreset;
    displayName: Install and spin up xvfb
    condition: eq(variables['DISTRIB'], 'conda')
```
still gives me error:
```
    QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-vsts'
    qt.qpa.screen: QXcbConnection: Could not connect to display
    Could not connect to any X display.
```
For the non-mayavi jobs I get:
```
/home/vsts/work/1/s/continuous_integration/azure/test_script.sh: line 9: pytest: command not found
```

I also can't work out how 'minimal' worked in travis. It doesn't install a lot of packages required for the tests e.g., numpy

https://github.com/sphinx-gallery/sphinx-gallery/blob/d1425b498bd34be19abd21c7b758545ad668663b/continuous_integration/travis/install.sh#L30-L31

lucyleeow:
Also what was the purpose of `SPHINXOPTS=""` here:

https://github.com/sphinx-gallery/sphinx-gallery/blob/d1425b498bd34be19abd21c7b758545ad668663b/.travis.yml#L11

lucyleeow:
Well '[skip ci]' seems to work for circle ci but not azure.

azure-pipelines.yml
lucyleeow: Okay I see Guillaume made it work in mne: https://github.com/mne-tools/mne-python/pull/8514/

but what does $RET mean and how is it used downstream?
```
- stage: Check
  jobs:
    - job: Skip
      pool:
        vmImage: 'ubuntu-18.04'
      variables:
        RET: 'true'
      steps:
      - bash: |
          git_log=`git log --max-count=1 --skip=1 --pretty=format:"%s"`
          echo "##vso[task.setvariable variable=log]$git_log"
      - bash: echo "##vso[task.setvariable variable=RET]false"
        condition: or(contains(variables.log, '[skip azp]'), contains(variables.log, '[azp skip]'), contains(variables.log, '[skip ci]'), contains(variables.log, '[ci skip]'))
      - bash: echo "##vso[task.setvariable variable=start_main;isOutput=true]$RET"
        name: result
```

larsoner: RET defaults to `true`, but on the second `bash` line gets turned to `false` under some conditions. Then this `RET` env var gets set to an Azure runtime variable `result` in the last step via a magic `echo` command, and this `result` Azure env var is available to subsequent steps.

azure-pipelines.yml
```diff
@@ -49,3 +49,77 @@ jobs:
...
+    # Make sure that things work even if the locale is set to C (which
+    # effectively means ASCII). Some of the input rst files have unicode
+    # characters and we need to deal with this gracefully.
+  - bash: |
+      sudo apt-get install python3-numpy python3-matplotlib python3-pip python3-coverage optipng
+    displayName: Install ubuntu python for ubuntu
+    condition: eq(variables['DISTRIB'], 'ubuntu')
+  - bash: |
+      sudo apt-get install xvfb
+      /sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -screen 0 1400x900x24 -ac +extension GLX +render -noreset;
```
larsoner, line 111: You need to set the `DISPLAY` env var for all runs

https://github.com/mne-tools/mne-python/blob/master/azure-pipelines.yml#L81

larsoner:
> I also can't work out how 'minimal' worked in travis. It doesn't install a lot of packages required for the tests e.g., numpy

You can always look at what happened:

https://travis-ci.com/github/sphinx-gallery/sphinx-gallery/jobs/459566968#L325

Looks like tests that require NumPy should be skipped

larsoner:
> Well '[skip ci]' seems to work for circle ci but not azure.

Yes they don't actually support `skip ci`, `ci skip`, or `skip azp` this is what Guillaume's code is for

> 

It is passed to the `make` command here

https://github.com/sphinx-gallery/sphinx-gallery/blob/master/continuous_integration/travis/test_script.sh#L13

At some point on the old Ubuntu + Sphinx run (I think that's the one) we couldn't use `-nWT --keep-going` because those options didn't exist or there were warnings due to old Sphinx, but on the newer runs we can and should. We might be able to get rid of this at some point but please don't do it in this PR, ideally this one would just add to `azure-pipelines.yml`, delete `.travis.yml`, and not touch `continuous_integration/travis/*` other than maybe to rename the files to `continuous_integration/azure/*`. It makes it clear that all the same old install commands and such are still being used (as much as this is possible).

Also keep in mind one important difference between Azure (and GitHub actions) versus Travis is that if you `export X=Y` in Travis it persists across steps and in Azure it does not, you need to use a magic `vso` echo command to get it to persist.

> Edit: pytest is probably missing because pip install from dev-requirements did not work:

You probably need something like `pip install --upgrade pip setuptools`:

https://github.com/mne-tools/mne-python/blob/master/azure-pipelines.yml#L52

azure-pipelines.yml
lucyleeow: Sorry I was more asking what does RET stand for? Does this variable name have a special meaning?

sphinx_gallery/tests/test_scrapers.py
```diff
@@ -15,6 +14,10 @@
 @pytest.fixture(scope='function')
 def gallery_conf(tmpdir):
     """Sets up a test sphinx-gallery configuration"""
+    # Skip if numpy not installed
+    pytest.importorskip("numpy")
+    import numpy as np
```
larsoner, line 19: You can just do:
```
    np = pytest.importorskip("numpy")
```
but really `np` isn't even used so this is fine
```suggestion
    pytest.importorskip("numpy")
```

sphinx_gallery/tests/test_full.py
```diff
@@ -36,6 +33,12 @@
 
 @pytest.fixture(scope='module')
 def sphinx_app(tmpdir_factory, req_mpl, req_pil):
+    # Skip if numpy not installed
+    pytest.importorskip("numpy")
+
+    import numpy as np
```
larsoner, line 39:
```suggestion
```
Not used

continuous_integration/azure/install.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
```
larsoner, line 1: It would have been nice to start with `git mv continuous_integration/travis/install.sh continuous_integration/azure/install.sh` before modifying the file, then the diff would be meaningful. From this file I can't tell what's been changed. Maybe it's not too late if you do something like:
```
git mv continuous_integration/travis/install.sh continuous_integration/azure/install.sh
git checkout continuous_integration/azure/install.sh
git reset
```
Maybe that would locally show the `git mv` operation plus a diff after the move?

Either way, the `continuous_integration/travis/*` files should go away

azure-pipelines.yml
larsoner: It's just an all-upper (BASH typical naming) shorthand for "return" I think, could be some other name (but I'd keep it this one so that copy-paste comparisons to MNE make sense)

continuous_integration/azure/install.sh
lucyleeow: Will do. I was worried I wouldn't be able to replicate the travis jobs perfectly because the astropy CI helper files were so detailed, so I was thinking I should keep both travis and azure for a bit to make sure the jobs were giving the same results. But I guess having the exact same miniconda version etc probably isn't that important...

sphinx_gallery/tests/test_full.py
```diff
@@ -377,6 +378,9 @@ def test_logging_std_nested(sphinx_app):
 
 def _assert_mtimes(list_orig, list_new, different=(), ignore=()):
     """Assert that the correct set of files were changed based on mtime."""
+    import numpy as np
+    from numpy.testing import assert_allclose
```
lucyleeow, line 382: Not sure if this is best solution but putting the imports in `setup` didn't work

continuous_integration/azure/install.sh
lucyleeow: Not sure why the git rename tracking worked for the other 2 files but not for this one. I tried to get git to track it by renaming this file back to travis and then renaming to azure, with `git log` being:
```
commit 2488bfe0f9941dee4dfbcadafd4182772694c7f4 (HEAD -> azure, lucy/azure)
Author: Lucy Liu <jliu176@gmail.com>
Date:   Mon Jan 18 21:47:21 2021 +1100

    mv back to azure

 continuous_integration/{travis => azure}/install.sh | 0
 1 file changed, 0 insertions(+), 0 deletions(-)
```
but it doesn't seem to register here....?

lucyleeow: I renamed the files again and added the changes in several commits. Github was showing the file as being renamed until the last commit. Would this be due to the proportion of lines changed..? Sorry but not sure what I can do about this.

lucyleeow:
Some progress.
conda_python36_localeC: is failing because `test_rebuild` fails (I think same as #777)
ubuntu_python36: I think if I set the default python I might get this to work

The other 3 setups that are failing have this error:
```
/usr/share/miniconda/lib/python3.6/site-packages/joblib/externals/loky/backend/resource_tracker.py:120: UserWarning: resource_tracker: process died unexpectedly, relaunching.  Some folders/sempahores might leak.
  warnings.warn('resource_tracker: process died unexpectedly, '
  File "/usr/share/miniconda/lib/python3.6/site-packages/joblib/externals/loky/backend/resource_tracker.py", line 281, in main
    del registry[rtype][name]
KeyError: '/dev/shm/joblib_memmapping_folder_3689_217583a6ccb9436c9c5097b8cd916e64_524f36e107544cafa166a08fe97952fc'
  File "/usr/share/miniconda/lib/python3.6/site-packages/joblib/externals/loky/backend/resource_tracker.py", line 281, in main
    del registry[rtype][name]
KeyError: '/dev/shm/joblib_memmapping_folder_3689_601b0e8390e04b74a35ed66b550644f6_7860eb38be1146febe1de7ac9af75129'
  File "/usr/share/miniconda/lib/python3.6/site-packages/joblib/externals/loky/backend/resource_tracker.py", line 281, in main
    del registry[rtype][name]
KeyError: '/loky-3689-8zqiv7_s'
  File "/usr/share/miniconda/lib/python3.6/site-packages/joblib/externals/loky/backend/resource_tracker.py", line 281, in main
    del registry[rtype][name]
KeyError: '/loky-3689-oz7iobvd'
  File "/usr/share/miniconda/lib/python3.6/site-packages/joblib/externals/loky/backend/resource_tracker.py", line 281, in main
    del registry[rtype][name]
KeyError: '/loky-3689-3ih7lccf'
  File "/usr/share/miniconda/lib/python3.6/site-packages/joblib/externals/loky/backend/resource_tracker.py", line 281, in main
    del registry[rtype][name]
KeyError: '/loky-3689-mbox5vn4'
  File "/usr/share/miniconda/lib/python3.6/site-packages/joblib/externals/loky/backend/resource_tracker.py", line 281, in main
    del registry[rtype][name]
KeyError: '/loky-3689-_9l1w3i2'
loading intersphinx inventory from https://pandas.pydata.org/pandas-docs/stable/objects.inv...
libgcc_s.so.1 must be installed for pthread_cancel to work
Makefile:64: recipe for target 'html-noplot' failed
make: *** [html-noplot] Aborted (core dumped)
```
does `libgcc_s.so.1` need to be installed?

lucyleeow:
To do:

- fix failure during doc build
- decide what to do about python nightly since azure doesn't have a nightly version like travis does. Deadsnakes?

larsoner:
> fix failure during doc build

pthread_cancel is probably a missing lib and probably fixed by something like

https://github.com/mne-tools/mne-python/blob/master/azure-pipelines.yml#L87

> decide what to do about python nightly since azure doesn't have a nightly version like travis does. Deadsnakes?

Deadsnakes would be great

lucyleeow:
The env 'ubuntu_python36' is now failing at the 'Install Ubuntu Python packages' step with:

> E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/p/pillow/python3-pil_5.1.0-1ubuntu0.3_amd64.deb  404  Not Found [IP: 40.81.13.82 80]

Could this be a temporary thing?

larsoner:
> The env 'ubuntu_python36' is now failing at the 'Install Ubuntu Python packages' step with:

> Nevermind, it's decided to work. Honestly

Usually this sort of error can be due to a package being updated on the server, but `sudo apt-get update` not being run to sync with it. If you're already doing that, I guess you might have gotten really unlucky where you (or the container image maintainers?) did `sudo apt-get update` and between then and when `sudo apt-get install` ran, the package was updated remotely.

lucyleeow:
I do not do `sudo apt-get update`, I will add this.

Not having any luck with the 'libgcc_s.so.1 must be installed for pthread_cancel to work' error. The MNE line didn't work. And googling is not fruitful so far. Any ideas?

lucyleeow:
I added:
```
which gcc
gcc --version
```
which gave this:
> /usr/bin/gcc
gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0

larsoner:
Let's just change to using ubuntu-20.04, it's what we use on MNE-Python and it works

larsoner:
... or you could try a horrible `apt-get` install line like this one, which I think was introduced when similar errors came up in MNE

https://github.com/mne-tools/mne-python/blob/master/.circleci/config.yml#L57-L61

lucyleeow:
Wait I think you are using 18.04 too: https://github.com/mne-tools/mne-python/blob/master/azure-pipelines.yml

larsoner:
Oh sorry I was looking at github actions not Azure. Maybe you need that annoying line, or some other variant like https://github.com/ContinuumIO/anaconda-issues/issues/9190#issuecomment-386508136 (from upstream issue)

larsoner:
If you really want to debug it, you can do something like:
```
LD_DEBUG=libs python -c "import matplotlib.pyplot as plt; from mayavi import mlab; plt.figure(); mlab.test_plot3d()"
```
This should cause the error, but print a ton of library information. It will tell you the first library that was missing. If you iterate you'll find them all until it succeeds. This is how I wrote the CircleCI line, but it was less tedious because you can SSH into a CircleCI session...

lucyleeow:
Getting too late, will try tomorrow. Thanks @larsoner

larsoner:
Okay to merge if it works? Want me to push a commit or two if it doesn't work on this last one?

lucyleeow:
Yes sure, please go for it!

lucyleeow:
Oh we should/could add deadsnakes before merging though...

larsoner:
I can't even get `pip` to run on Python 3.10, so we'll just go with 3.9 for now and test 3.10 sometime later

larsoner:
Thanks @lucyleeow !

PR #791, opened 2021-02-13
BUG Remove if final block empty

### LINKED ISSUE #765: BUG: Empty code block output
From @lucyleeow in https://github.com/matplotlib/matplotlib/issues/17479#issuecomment-717011368:

> Just looking into this, the problem with # sphinx_gallery_thumbnail_path is that it needs to be outside of the docstring, which makes it complicated for examples like annotation.py which is all docstring. Adding a code block under the docstring will work but will result in an empty code block at the end of the example.

@lucyleeow it seems like it would make sense to detect this case and not render it

lucyleeow:
closes #765


larsoner:
Thanks @lucyleeow !

PR #792, opened 2021-02-15
Add a configuration to warn on error not fail

### LINKED ISSUE #789: Add a mode which "skips" an example if it fails
We have the situation where some of our examples are dependant on remote servers or other things which are a little flakey. For our tagged releases we want these to pass, but for our regular builds on our development branch, it would be better if they rendered the same way as if they were skipped.

Would people be interested in a config option which didn't cause sphinx to exit with a non-zero code if an example errored, and just renders the output of that example with no output at all?

Cadair:
With this configuration option set sphinx will only exit with a non-zero code if the -W flag is set.

I would find this useful on sunpy's CI because then sphinx-gallery will respect the fact that some of our builds have -W set and some (ones where we want to generate previews) don't.

Fixes #789

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -36,7 +36,7 @@ file:
 - ``compress_images`` (:ref:`compress_images`)
 - ``reset_modules`` (:ref:`reset_modules`)
 - ``abort_on_example_error`` (:ref:`abort_on_first`)
-- ``only_warning_on_example_error`` (:ref:`warning_on_error`)
+- ``only_warn_on_example_error`` (:ref:`warning_on_error`)
 - ``expected_failing_examples`` (:ref:`dont_fail_exit`)
 - ``min_reported_time`` (:ref:`min_reported_time`)
 - ``show_memory`` (:ref:`show_memory`)
@@ -1257,7 +1257,7 @@ the example script.
 
 .. _warning_on_error:
 
-Never Fail the build on error
+Never fail the build on error
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Sphinx-Gallery can be configured to only log warnings when examples fail. This means that sphinx will only exit with a non-zero exit code if the ``-W`` flag is passed to ``sphinx-build``. This can be enabled by setting::
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -36,6 +36,7 @@ file:
 - ``compress_images`` (:ref:`compress_images`)
 - ``reset_modules`` (:ref:`reset_modules`)
 - ``abort_on_example_error`` (:ref:`abort_on_first`)
+- ``only_warning_on_example_error`` (:ref:`warning_on_error`)
 - ``expected_failing_examples`` (:ref:`dont_fail_exit`)
 - ``min_reported_time`` (:ref:`min_reported_time`)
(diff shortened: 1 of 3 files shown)


Cadair:
I tested this out with sunpy here: https://github.com/sunpy/sunpy/pull/4993 note how the Azure build_docs job fails because the gallery fails to build, but RTD is green so you can see the preview and some of the examples show "broken" with the errors.

doc/configuration.rst
```diff
@@ -36,6 +36,7 @@ file:
 - ``compress_images`` (:ref:`compress_images`)
 - ``reset_modules`` (:ref:`reset_modules`)
 - ``abort_on_example_error`` (:ref:`abort_on_first`)
+- ``only_warning_on_example_error`` (:ref:`warning_on_error`)
```
larsoner, line 39:
```suggestion
- ``only_warn_on_example_error`` (:ref:`warning_on_error`)
```

doc/configuration.rst
```diff
@@ -1248,6 +1249,18 @@ the example script.
 .. note:: If an example is expected to fail, Sphinx-Gallery will error if
           the example runs without error.
 
+.. _warning_on_error:
+
+Never Fail the build on error
```
larsoner, line 1254:
```suggestion
Never fail the build on error
```

larsoner (commented):
I can see the use case and the changes look backward compatible, so seems reasonable to me!

Could you add a test, maybe to `test_gen_gallery.py`? I'd expect you to need a `sphinx_app_wrapper` with a `build_sphinx_app` call after putting a failing example in the `sphinx_app.srcdir` somewhere. I would `pytest.mark.parametrize` over `only_warn_on_example_error`, and when it's `True` the build should exit with 0 and when it's `False` it should exit with 1 / raise an error (not sure which).

Cadair:
Thanks for the review @larsoner.

I added two tests, one where it asserts the error is logged and the other that asserts it is raised if the equivalent of `-W` is set. I didn't think the `False` behaviour needed an explicit test as it's already covered by existing tests? I am not sure the `-W` test is really needed either, given you could probably argue that's just testing sphinx. :shrug:

sphinx_gallery/tests/test_gen_gallery.py
```diff
@@ -321,6 +319,45 @@ def test_expected_failing_examples_were_executed(sphinx_app_wrapper):
...
+    Test behaviour of only_warn_on_example_error flag.
+    """
+    example_dir = Path(sphinx_app_wrapper.srcdir) / 'src'
+    with codecs.open(example_dir / 'plot_3.py', 'a', encoding='utf-8') as fid:
+        fid.write('raise ValueError')
+    sphinx_app = sphinx_app_wrapper.build_sphinx_app()
+
+    build_warn = sphinx_app._warning.getvalue()
+    assert 'plot_3.py failed to execute correctly' in build_warn
+    assert 'WARNING: Here is a summary of the problems encountered when running the examples' in build_warn
```
larsoner, line 339:
```
sphinx_gallery/tests/test_gen_gallery.py:339:80: E501 line too long (107 > 79 characters)
```
```suggestion
    assert 'WARNING: Here is a summary of the problems' in build_warn
```

larsoner:
Thanks @Cadair !

Cadair:
Thanks!

Issue #805, opened 2021-04-07
How to link from one example to another?


mfeurer:
Hi,

first of all, I'm a huge fan of this package and I'm using for packages I'm working on such as [OpenML-Python](https://github.com/openml/openml-python/) and [Auto-sklearn](https://github.com/automl/auto-sklearn).

I'm currently trying to create links between examples by using rst references (to allow for automatic link checking by sphinx). However, I cannot find any mention of this in the documentation. The only thing I could find was https://github.com/sphinx-gallery/sphinx-gallery/pull/83 but the resulting example doesn't render the link: 

![image](https://user-images.githubusercontent.com/5320498/113837763-73ddd480-978e-11eb-852d-778244c3795a.png)

Therefore I was wondering what if it is possible at all to link between examples using rst references or whether this not being possible as suggested by https://github.com/sphinx-gallery/sphinx-gallery/pull/83 is simply a bug?

lucyleeow:
What do you mean the resulting example doesn't render the link?
Inside the 'see also' box, we link to the example 'Syntax error' using the syntax described above, and I checked that the link does work.
I can see how the documentation is confusing though

mfeurer:
Please excuse this issue, after you answered I tried again and now it works. I must have made a mistake before. Thanks a lot for your answer.

Regarding confusing documentation: I just didn't find this section which explains everything I needed to know: https://sphinx-gallery.github.io/stable/advanced.html#know-your-gallery-files but I'm not sure if this information could be put somewhere else or whether the heading could get a better name.

lucyleeow:
No worries. Yes I agree about the documentation, if you have a suggestion feel free to put in a PR!

PR #806, opened 2021-04-08
DOC: clarify advanced doc wrt referencing examples

mfeurer:
As discussed via https://github.com/sphinx-gallery/sphinx-gallery/issues/805, the documentation could be structured a bit better to make it easier to find how to link examples. This PR suggests adding a new sub-heading *Referencing Examples* so that the term *reference* explicitly shows up in the table of contents. 

Two things to discuss:
* I'm not sure yet whether the heading should be *referencing* or *linking*. 
* Is it possible to link to heading in examples, too? Is there a special syntax for that? If yes, I could extend this PR to include it.


doc/advanced.rst
```diff
@@ -57,12 +57,15 @@ into the HTML version of your documentation have a particular
...
-So for example if we want to reference the example
-:ref:`sphx_glr_auto_examples_plot_0_sin.py`, we just call
-its reference
+Referencing Examples
+--------------------
+
+The namespace convention mentioned in the paragraph above is valid for
+cross-references labels, and images. So for example if we want to 
+reference the example :ref:`sphx_glr_auto_examples_plot_0_sin.py`, 
+we just call its reference
```
lucyleeow, line 68: I think I would put the cross referencing part a bit separate because the cross referencing of examples is enabled because we put a reference label at the start of each generated `.rst` file, which is a slightly different topic than how we name generated files.

Maybe:

"Our namespace convention for generated files is to prefix everything with ``sphx_glr`` and end with the name of the file. For example the first image generated by the example ``plot_0_sin.py`` is ``sphx_glr_plot_0_sin_001.png`` and its thumbnail is ``sphx_glr_plot_0_sin_thumb.png``.

I would have the literal include paragraphs, starting with "You can also include part of", follow.

At the end of the section (line 97 in the new commit), before the line ".. _warning_errors:" I would add your new referencing header.

I would call it "cross referencing", to make it consistent with https://sphinx-gallery.github.io/stable/auto_examples/plot_0_sin.html#cross-referencing

```rst
Cross referencing
-----------------

You can also cross reference an example using similar naming convention. For example
if we want to reference the example :ref:`sphx_glr_auto_examples_plot_0_sin.py`,
we just call its reference ``:ref:`sphx_glr_auto_examples_plot_0_sin.py```.
Note that we have included the path to the example file (relative to
the ``conf.py`` file) after ``sphx_glr``. Path separators are replaced with
underscores.
```

lucyleeow (commented):
Some re-jiggling, mostly related to how our doc was poorly structured before.

lucyleeow:
Small push to get the test fixes. Tests should pass now.

doc/advanced.rst
mfeurer: Hey, I'm not sure if I understood your suggestion correctly, but I added the suggested paragraph and reverted the other changes. I think having a separate paragraph makes total sense, but I'm not sure whether you suggested that I should edit the section before.

lucyleeow:
Thanks @mfeurer !

Issue #803, opened 2021-03-28
Lazy loading of gallery thumbnails


timhoffm:
Loading all tumbnails can take a significant amount of time/bandwidth (c.f. Matplotlib galleries).

It would be great if the loading could be deferred to the point when the user actually scrolls to the thumbnail.

Unfortunately, from a quick look this seems to not be easy. Lazy loading requires to change the `<img>` tag to either have an attribute `loading='lazy'` or use `data-src` instead of `src`(https://www.sitepoint.com/five-techniques-lazy-load-images-website-performance/) The thumbnail is generated via a `.. figure::` directive (see `THUMBNAIL_TEMPLATE`), which does not expose any of those.

Ideas are welcome how this could be solved.

larsoner:
In lieu of modifying docutils to add this sort of support, which they might not even go for, we could consider adding our own `.. lazyfigure::` directive that subclasses the standard figure one but overrides the HTML formatting to add this extra `loading='lazy'` (or JS with data-src or whatever).

Maybe even easier would be to add a JS library that modifies the HTML at load time to have this loading=lazy attribute, but I'm not sure if that would even work (the browser might already be trying to load the images by the time the element is updated).

timhoffm:
I've posted a feature request to docutils.
https://sourceforge.net/p/docutils/feature-requests/78/

Let's see how that goes.

tacaswell:
It seems the upstream feature request was implemented!

> Implemented in Docutils 0.18.
Thanks again for the suggestion.

lucyleeow:
Great. Should this be closed now then?

timhoffm:
My original idea was that this setting was applied to the gallery thumbnails only.

This would require an image option
```
.. image::
   :loading: lazy
```

AFAICS, docutils has not implemented that. docutils 0.18 only added a global config parameter https://docutils.sourceforge.io/docs/user/config.html#image-loading but not a per-image setting.

Does this handle this issue?

I think that depends. The core issue here is that we want incremental loading of gallery images. Is it ok to tell users to globally switch to lazy loading for all content? If yes, then we can close. If no, we'd need to go back to docutils and request the per-image option, and when that's done add a sphinx-gallery option to set that option for all thumbnails.

tacaswell:
[...]

PR #808, opened 2021-04-10
ENH: add hidpi option to matplotlib_scraper and directive

jklymak:
### Motivation

Right now, at least at Matplotlib.org, all images are 100 dpi, and 200 dpi images look bad on hi-dpi monitors.  We could save 200 dpi and scale, but that was deemed problematic.  The "proper" solution is to serve images with the `srcset` option and let the client browser decide what to download: https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images

This PR implements this for the matplotlib scraper.  Of course this requires a new directive as well.  

### Solution

The PR adds a `image_srcset=["", "2x"]` configuration option that causes the matplotlib scraper to save two (or more) files, the usual image, and a dpi=200 (or other dpi, i.e. `image_srcset=["", "2x", "3x"]`) version that gets named `filename_2_0.png`.  It then adds a new directive to the gallery rst:

```rst
.. image-srcset:: /gallery/subdir/filename.png
   :alt: filename
   :srcset: /gallery/subdir/filename.png, /gallery/subdir/filename_2_0.png 2.0x, 
   :class: sphx-glr-single-img
```

The second part implements this new directive to create an html as:

```html
<img srcset="/gallery/subdir/filename.png, /gallery/subdir/filename_2_0.png 2.0x", src="/gallery/subdir/filename.png", 
  class="sphx-glr-single-img", alt="filename">
```

This needs tests and documentation, but I thought I'd share now to get feedback.  

- [x] tests
- [x] docs

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1107,17 +1107,18 @@ Multi-resolution images
 Web browsers allow a ``srcset`` parameter to the ``<img>`` tag that 
 allows the browser to support `responsive resolution images 
 <https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images>`__
-for hi-dpi/retina displays. Sphinx Gallery supports this via the ``image_srcset`` parameter::
+for hi-dpi/retina displays. Sphinx Gallery supports this via the 
+``image_srcset`` parameter::
 
     sphinx_gallery_conf = {
         ...
-        'image_srcset': ["", "2x"],
+        'image_srcset': ["2x"],
     }
 
 that saves a 1x image at the normal figure dpi (usually 100 dpi) and a 2x 
 version at twice the density (e.g. 200 dpi).  The default is no extra images 
-(``'image_srcset': [""]``), and you can specify other resolutions if desired (e.g.
-``"1.5x"``).
+(``'image_srcset': []``), and you can specify other resolutions if desired as a 
+list: ``["2x", "1.5x"]``.
 
 The matplotlib scraper creates a custom image directive, ``image-sg`` in the 
 rst file::
@@ -1127,13 +1128,13 @@ rst file::
         :srcset: /examples/images/sphx_glr_test_001.png, /examples/images/sphx_glr_test_001_2_0x.png 2.0x
         :class: sphx-glr-single-img
 
-Which is converted to html by the custom directive as::
+This is converted to html by the custom directive as::
 
     .. <img src="../_images/sphx_glr_test_001.png" alt="test", class="sphx-glr-single-img", 
         srcset="../_images/sphx_glr_test_001.png, ../_images/sphx_glr_test_001_2_0x.png 2.0x>
 
-This leads to a larger website, but responsive images will only serve the appropriate-sized 
-images to website visitors.
+This leads to a larger website, but clients that support the ``srcset`` tag will only 
+download the appropriate-sized images.
 
 Note that the ``.. image-sg`` directive currently ignores other ``.. image`` directive 
 tags like ``width``, ``height``, and ``align``.  It also only works with the *html* and 
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -34,6 +34,7 @@ file:
 - ``image_scrapers`` (and the deprecated ``find_mayavi_figures``)
   (:ref:`image_scrapers`)
 - ``compress_images`` (:ref:`compress_images`)
+- ``image_srcset`` (:ref:`image_srcset`)
 - ``reset_modules`` (:ref:`reset_modules`)
 - ``abort_on_example_error`` (:ref:`abort_on_first`)
 - ``only_warn_on_example_error`` (:ref:`warning_on_error`)
@@ -1097,6 +1098,46 @@ optimize less but speed up the build time you could do::
     }
 
 See ``$ optipng --help`` for a complete list of options.
+
+.. _image_srcset:
+
+Multi-resolution images
+=======================
+
+Web browsers allow a ``srcset`` parameter to the ``<img>`` tag that 
+allows the browser to support `responsive resolution images 
+<https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images>`__
+for hi-dpi/retina displays. Sphinx Gallery supports this via the ``image_srcset`` parameter::
+
+    sphinx_gallery_conf = {
+        ...
+        'image_srcset': ["", "2x"],
+    }
+
+that saves a 1x image at the normal figure dpi (usually 100 dpi) and a 2x 
+version at twice the density (e.g. 200 dpi).  The default is no extra images 
+(``'image_srcset': [""]``), and you can specify other resolutions if desired (e.g.
+``"1.5x"``).
+
+The matplotlib scraper creates a custom image directive, ``image-sg`` in the 
+rst file::
+
+    .. image-sg:: /examples/images/sphx_glr_test_001.png
+        :alt: test
+        :srcset: /examples/images/sphx_glr_test_001.png, /examples/images/sphx_glr_test_001_2_0x.png 2.0x
+        :class: sphx-glr-single-img
+
+Which is converted to html by the custom directive as::
+
+    .. <img src="../_images/sphx_glr_test_001.png" alt="test", class="sphx-glr-single-img", 
+        srcset="../_images/sphx_glr_test_001.png, ../_images/sphx_glr_test_001_2_0x.png 2.0x>
+
+This leads to a larger website, but responsive images will only serve the appropriate-sized 
+images to website visitors.
+
+Note that the ``.. image-sg`` directive currently ignores other ``.. image`` directive 
+tags like ``width``, ``height``, and ``align``.  It also only works with the *html* and 
+*latex*  builders.  
 
 .. _image_scrapers:
 
--- a/sphinx_gallery/directives.py
+++ b/sphinx_gallery/directives.py
@@ -2,11 +2,14 @@
 Custom Sphinx directives
 ========================
 """
-
 import os
+from pathlib import PurePosixPath
+import shutil
 
+from docutils import nodes
 from docutils import statemachine
 from docutils.parsers.rst import Directive, directives
+from docutils.parsers.rst.directives import images
 
 
 class MiniGallery(Directive):
@@ -77,3 +80,198 @@ class MiniGallery(Directive):
         self.state_machine.insert_input(include_lines, path)
 
         return []
+
+
+"""
+Image sg for responsive images
+"""
+
+
+class imgsgnode(nodes.General, nodes.Element):
+    pass
+
+
+def directive_boolean(value):
+    if not value.strip():
+        raise ValueError("No argument provided but required")
+    if value.lower().strip() in ["yes", "1", 1, "true", "ok"]:
+        return True
+    elif value.lower().strip() in ['no', '0', 0, 'false', 'none']:
+        return False
+    else:
+        raise ValueError(u"Please use one of: yes, true, no, false. "
+                         u"Do not use `{}` as boolean.".format(value))
+
+
+class ImageSg(images.Image):
+    """
+    Implements a directive to allow an optional hidpi image.  Meant to be
+    used with the `image_srcset` configuration option.
+
+    e.g.::
+
+        .. image-sg:: /plot_types/basic/images/sphx_glr_bar_001.png
+            :alt: bar
+            :srcset: /plot_types/basic/images/sphx_glr_bar_001.png,
+                     /plot_types/basic/images/sphx_glr_bar_001_2_0x.png 2.0x
+            :class: sphx-glr-single-img
+
+    The resulting html is::
+
+        <img src="sphx_glr_bar_001_hidpi.png"
+            srcset="_images/sphx_glr_bar_001.png,
+                    _images/sphx_glr_bar_001_2_0x.png 2x",
+            alt="bar"
+            class="sphx-glr-single-img" />
+
+    """
+
+    has_content = False
+    required_arguments = 1
+    optional_arguments = 3
+    final_argument_whitespace = False
+    option_spec = {
+        'srcset': directives.unchanged,
+        'class': directives.class_option,
+        'alt': directives.unchanged,
+    }
+
+    def run(self):
+        image_node = imgsgnode()
+
+        imagenm = self.arguments[0]
+        image_node['alt'] = self.options.get('alt', '')
+        image_node['class'] = self.options.get('class', None)
+
+        # we would like uri to be the highest dpi version so that
+        # latex etc will use that.  But for now, lets just make
+        # imagenm
+
+        image_node['uri'] = imagenm
+        image_node['srcset'] = self.options.get('srcset', None)
+
+        return [image_node]
+
+
+def _parse_srcset(st):
+    """ parse st"""
+    entries = st.split(',')
+    srcset = {}
+    for entry in entries:
+        spl = entry.strip().split(' ')
+        if len(spl) == 1:
+            srcset[0] = spl[0]
+        elif len(spl) == 2:
+            mult = spl[1][:-1]
+            srcset[float(mult)] = spl[0]
+    return srcset
+
+
+def visit_imgsg_html(self, node):
+
+    if node['srcset'] is None:
+        self.visit_image(node)
+        return
+
+    imagedir, srcset = _copy_images(self, node)
+
+    # /doc/examples/subd/plot_1.rst
+    docsource = self.document['source']
+    # /doc/
+    # make sure to add the trailing slash:
+    srctop = os.path.join(self.builder.srcdir, '')
+    # examples/subd/plot_1.rst
+    relsource = os.path.relpath(docsource, srctop)
+    # /doc/build/html
+    desttop = os.path.join(self.builder.outdir, '')
+    # /doc/build/html/examples/subd
+    dest = os.path.join(desttop, relsource)
+
+    # ../../_images/
+    imagerel = os.path.relpath(imagedir, os.path.dirname(dest))
+    imagerel = os.path.join(imagerel, '')
+    if '\\' in imagerel:
+        imagerel = imagerel.replace('\\', '/')
+    # make srcset str.  Need to change all the prefixes!
+    srcsetst = ''
+    for mult in srcset:
+        nm = os.path.basename(srcset[mult][1:])
+        # ../../_images/plot_1_2_0x.png
+        relpath = imagerel+nm
+        srcsetst += f'{relpath}'
+        if mult == 0:
+            srcsetst += ', '
+        else:
+            srcsetst += f' {mult:1.1f}x, '
+    # trim trailing comma and space...
+    srcsetst = srcsetst[:-2]
+
+    # make uri also be relative...
+    nm = os.path.basename(node['uri'][1:])
+    uri = imagerel + nm
+
+    alt = node['alt']
+    if node['class'] is not None:
+        classst = node['class'][0]
+        classst = f'class = "{classst}"'
+    else:
+        classst = ''
+
+    html_block = (f'<img src="{uri}" srcset="{srcsetst}" alt="{alt}"' +
+                  f' {classst}/>')
+    self.body.append(html_block)
+
+
+def visit_imgsg_latex(self, node):
+
+    if node['srcset'] is not None:
+        imagedir, srcset = _copy_images(self, node)
+        maxmult = -1
+        # choose the highest res version for latex:
+        for key in srcset.keys():
+            maxmult = max(maxmult, key)
+        node['uri'] = str(PurePosixPath(srcset[maxmult]).name)
+
+    self.visit_image(node)
+
+
+def _copy_images(self, node):
+    srcset = _parse_srcset(node['srcset'])
+
+    # where the sources are.  i.e. myproj/source
+    srctop = self.builder.srcdir
+
+    # copy image from source to imagedir.  This is
+    # *probably* supposed to be done by a builder but...
+    # ie myproj/build/html/_images
+    imagedir = os.path.join(self.builder.imagedir, '')
+    imagedir = PurePosixPath(self.builder.outdir, imagedir)
+
+    os.makedirs(imagedir, exist_ok=True)
+
+    # copy all the sources to the imagedir:
+    for mult in srcset:
+        abspath = PurePosixPath(srctop, srcset[mult][1:])
+        shutil.copyfile(abspath, imagedir / abspath.name)
+
+    return imagedir, srcset
+
+
+def depart_imgsg_html(self, node):
+    pass
+
+
+def visit_sg_other(self, node):
+    if node['uri'][0] == '/':
+        node['uri'] = node['uri'][1:]
+    self.visit_image(node)
+
+
+def depart_imgsg_latex(self, node):
+    self.depart_image(node)
+
+
+def imagesg_addnode(app):
+    app.add_node(imgsgnode,
+                 html=(visit_imgsg_html, depart_imgsg_html),
+                 latex=(visit_imgsg_latex, depart_imgsg_latex))
--- a/sphinx_gallery/directives.py
+++ b/sphinx_gallery/directives.py
@@ -11,6 +11,8 @@ from docutils import statemachine
 from docutils.parsers.rst import Directive, directives
 from docutils.parsers.rst.directives import images
 
+from sphinx.errors import ExtensionError
+
 
 class MiniGallery(Directive):
     """
@@ -164,6 +166,8 @@ def _parse_srcset(st):
         elif len(spl) == 2:
             mult = spl[1][:-1]
             srcset[float(mult)] = spl[0]
+        else:
+            raise ExtensionError('srcset argument "{entry}" is invalid.')
     return srcset
 
 
--- a/sphinx_gallery/scrapers.py
+++ b/sphinx_gallery/scrapers.py
@@ -114,15 +114,19 @@ def matplotlib_scraper(block, block_vars, gallery_conf, **kwargs):
     image_path_iterator = block_vars['image_path_iterator']
     image_rsts = []
     # Check for srcset hidpi images
-    srcset = gallery_conf.get('image_srcset', [""])
-    srcset_mult_facs = []
+    srcset = gallery_conf.get('image_srcset', [])
+    srcset_mult_facs = [1]  # one is always supplied...
     for st in srcset:
         if (len(st) > 0) and (st[-1] == 'x'):
             # "2x" = "2.0"
             srcset_mult_facs += [float(st[:-1])]
+        elif st == "":
+            pass
         else:
-            # "" = "1"
-            srcset_mult_facs += [1]
+            raise ExtensionError(
+                f'Invalid value for image_srcset parameter: "{st}". '
+                'Must be a list of strings with the multiplicative '
+                'factor followed by an "x".  e.g. ["2.0x", "1.5x"]')
 
     # Check for animations
     anims = list()
@@ -524,6 +528,13 @@ SG_IMAGE = """
    :alt: %s
    :srcset: %s
    :class: sphx-glr-single-img
+"""
+
+# keep around for back-compat:
+SINGLE_IMAGE = """
+ .. image:: /%s
+     :alt: %s
+     :class: sphx-glr-single-img
 """
 
 ###############################################################################
--- a/sphinx_gallery/scrapers.py
+++ b/sphinx_gallery/scrapers.py
@@ -8,7 +8,9 @@ Scrapers for embedding images
 Collect images that have been produced by code blocks.
 
 The only scrapers we support are Matplotlib and Mayavi, others should
-live in modules that will support them (e.g., PyVista, Plotly).
+live in modules that will support them (e.g., PyVista, Plotly).  Scraped
+images are injected as rst ``image-sg`` directives into the ``.rst``
+file generated for each example script.
 """
 
 import os
@@ -16,6 +18,7 @@ import sys
 import re
 from distutils.version import LooseVersion
 from textwrap import indent
+from pathlib import PurePosixPath
 from warnings import filterwarnings
 
 from sphinx.errors import ExtensionError
@@ -110,6 +113,17 @@ def matplotlib_scraper(block, block_vars, gallery_conf, **kwargs):
     from matplotlib.animation import Animation
     image_path_iterator = block_vars['image_path_iterator']
     image_rsts = []
+    # Check for srcset hidpi images
+    srcset = gallery_conf.get('image_srcset', [""])
+    srcset_mult_facs = []
+    for st in srcset:
+        if (len(st) > 0) and (st[-1] == 'x'):
+            # "2x" = "2.0"
+            srcset_mult_facs += [float(st[:-1])]
+        else:
+            # "" = "1"
+            srcset_mult_facs += [1]
+
     # Check for animations
     anims = list()
     if gallery_conf.get('matplotlib_animations', False):
@@ -118,9 +132,9 @@ def matplotlib_scraper(block, block_vars, gallery_conf, **kwargs):
                 anims.append(ani)
     # Then standard images
     for fig_num, image_path in zip(plt.get_fignums(), image_path_iterator):
+        image_path = PurePosixPath(image_path)
         if 'format' in kwargs:
-            image_path = '%s.%s' % (os.path.splitext(image_path)[0],
-                                    kwargs['format'])
+            image_path = image_path.with_suffix('.' + kwargs['format'])
         # Set the fig_num figure as the current figure as we can't
         # save a figure that's not the current figure.
         fig = plt.figure(fig_num)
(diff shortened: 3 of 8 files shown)


timhoffm:
I've not looked into the details, but it might be worth considering making this a bit more general. The image `srcset` attribute supports more values than A`2x`. Instead of `matplotlib_HiDPI` you could go e.g. with a config variable `image_srcset = ["", "1.5x", "2x"]`. Having the option for multiple pixel density desciptors seems a trivial and reasonable extension.

Note that one can alternatively specify the width in srcset,  but I don't see a real benefit from that and one could always add that later if needed.

jklymak:
For the name of the config variable, I think it needs to be prefixed by `matplotlib` since I don't think we want to claim this will work for other scrapers. `matplotlib_srcset` is maybe good?  I agree its a good idea to point to the img tag argument. 

For resolution options, that is definitely doable.  I was hesitant because I don't quite know how to get "1x" and how that interacts with how users may specify base dpi.  The simplest was just to hardcode a "2x" as 200 dpi.  

I agree that specifying pixel width would be more confusing than not.  By definition it would be inconsistent with our specification of figure size and dpi, unless the user calculated the exact width they wanted themselves.  2x, 3x etc are at least consistent scaling factors. 

I'll take a crack at implementing `matplotlib_srcset=["", "2x", "3x"]` etc...

BTW, if anyone has ideas for how to test the sphinx directive, that would be appreciated!  I don't see that we have any tests of that.  I've already added a test for the scraper, and that works fine.

jklymak:
OK, did `matplotlib_srcset=["", "2x"]`  Not too much more complicated, though dicts have to be passed around, and there is some parsing and unparsing.

sphinx_gallery/scrapers.py
```diff
@@ -110,6 +111,17 @@ def matplotlib_scraper(block, block_vars, gallery_conf, **kwargs):
...
+    # Check for srcset hidpi images
+    srcset = gallery_conf.get('matplotlib_srcset', None)
+    hidpi = srcset is not None
+    if hidpi:
+        hidpi_mult_facs = []
+        for st in srcset:
+            if (len(st) > 0) and (st[-1] == 'x'):
+                hidpi_mult_facs += [float(st[:-1])]
+            else:
+                hidpi_mult_facs += [1]
```
larsoner, line 123: It is better to do this conversion in `_complete_gallery_conf` so that if it fails, it does so immediately before trying to build any examples. For example you can make `matplotlib_srcset` always a list of float in there, thereby translating `None` to `[1.]` or so.

sphinx_gallery/scrapers.py
```diff
@@ -388,14 +429,43 @@ def figure_rst(figure_list, sources_dir, fig_titles=''):
     images_rst = ""
     if len(figure_paths) == 1:
         figure_name = figure_paths[0]
-        images_rst = SINGLE_IMAGE % (figure_name, alt)
+        if hipaths is None:
+            images_rst = SINGLE_IMAGE % (figure_name, alt)
+        else:
+            hinames = hipaths[0]
+            srcset = _get_srcset_st(sources_dir, hinames)
+            images_rst = SRCSET_IMAGE % (figure_name, alt, srcset)
```
larsoner, line 437: Can we simplify this to always use `img-srcset` instead?

larsoner (commented):
My high level comment is that this PR introduces a lot of branching of the form `f hipaths is None:`. Is it possible to always use `img-srcset`, treat `'matplotlib_srcset': None,` the same way as `'matplotlib_srcset': [1.],` and remove a bunch of conditionals in the code?

jklymak:
We could, but I'd assume we would rather not use a custom directive if we didn't need to?

larsoner:
> For the name of the config variable, I think it needs to be prefixed by matplotlib since I don't think we want to claim this will work for other scrapers. matplotlib_srcset is maybe good? I agree its a good idea to point to the img tag argument.

So far the only internal scrapers are really `matplotlib` and `mayavi`. I think in principle we could do it for Mayavi, too, but I'm not sure how many people would use it. I wouldn't mind a more general name like `srcset`, we say the builtin Matplotlib scraper supports it, Mayavi doesn't, and other custom scrapers can poll this attribute and expect a list of float. I have a custom PyVista-based scraper that I would probably try to convince to support the `sphinx_gallery_conf['srcset']` parameter if this were the case. 

> BTW, if anyone has ideas for how to test the sphinx directive, that would be appreciated! I don't see that we have any tests of that. I've already added a test for the scraper, and that works fine.

The other PR that added support for a directive was https://github.com/sphinx-gallery/sphinx-gallery/pull/601, and you can see directive tests in `test_load_style.py`: https://github.com/sphinx-gallery/sphinx-gallery/pull/601/files#diff-bafd845f88b524c7185062ce480e1e15b1222faa678dbe0c3e90efc2a6572288 . Maybe something like that could work?

> We could, but I'd assume we would rather not use a custom directive if we didn't need to?

I'd prefer that we switch to and make `img-srcset` work as fully as possible to keep the code branching down (maintenance cost). We're going to have to fix bugs with the new directive as they come up, I'd rather just commit to making it work well from the start rather than go halfway.

Did you base the code off of the standard image directive in Sphinx? I noticed it only had `latex` and `html` builder support -- if there are more than that for the base `image` directive, can we add them by just wrapping to the base `image` directive somehow?

jklymak:
> Did you base the code off of the standard image directive in Sphinx? I noticed it only had latex and html builder support -- if there are more than that for the base image directive, can we add them by just wrapping to the base image directive somehow?

Not really, and in fact it doesn't support a good chunk of what the image directive does - for instance, width, height, align are all ignored currently.  That can all be improved.  

Also, I don't think the copying of the files really belongs in `visit_imgsrcset`; I think `sphinx/docutils` would do this in the `builder`?  But that seemed a huge pain to parse everything in both the visit and build stages?  

I think adding other writers is as simple as just passing them to `visit_srcset_other`, and having `visit_image` called on the node.  That *does* lead to one question: Currently, `node['uri']` is the base-res version of the image.  This is the version that will be passed to latex, for instance.  An earlier version I had it send the "hidpi" version, but that is harder to identify when we have more than one hidpi version.  

> I'd prefer that we switch to and make img-srcset work as fully as possible to keep the code branching down (maintenance cost). We're going to have to fix bugs with the new directive as they come up, I'd rather just commit to making it work well from the start rather than go halfway.

I think thats really easy to change, if you are happy with the gallery rst always having the `img-srcset` directive instead of the `img` directive.  

Currently the directive module has to be included as a sphinx optional module.  Are you thinking it would just be part of the base `sphinx_gallery` module?

larsoner:
>  for instance, width, height, align are all ignored currently. That can all be improved.

Currently for SG's internal use we don't set any of these (right?) so we don't have to care too much for now. People could implement these as enhancements. We just have to clearly state in the docs that the `sphinx_gallery.img_srcset` extension only supports specific tags, and only HTML and LaTeX builders.

> Also, I don't think the copying of the files really belongs in visit_imgsrcset; I think sphinx/docutils would do this in the builder? But that seemed a huge pain to parse everything in both the visit and build stages?

My hope is that we can get away with just copy-paste-modify'ing the upstream image directive code for this somehow. If it ends up being 10 lines that's great, 1000 is too many (maybe we can reuse public functions somehow?), and in between we'd have to make a go-no-go call. Do you want to look to see how much work it would be to make `img-srcset` fully-featured like `image`? If it's too much then HTML and LaTeX support is probably good enough.

Really HTML support covers most use cases, we have a number of `.. raw:: html` / `.. only:: html`s in the codebase already. So HTML + LaTeX might even cover all existing users.

> Currently the directive module has to be included as a sphinx optional module. Are you thinking it would just be part of the base sphinx_gallery module?

Yes, or to put it another way (I think), adding `sphinx_gallery.gen_gallery` to the `extensions` list in `conf.py` should automatically also add the `sphinx_gallery.img_srcset` directive functionality for people in their doc builds. This is implicitly how the `sphinx_gallery.load_style` works -- when using `gen_gallery` we add the same CSS that `load_style` will add, so at the end of the day the documenter/user has access to the same stuff either way.

jklymak:
Right just to be clearer.  The img-srcset directive loads a node with the correct  info.  If the writer is html we do our thing that looks at the srcset node contents, copies the files, etc.  If it is latex, we just pass though self.visit_node where self is the latex translator.  The same code can be added to the other writers easily enough 

I did note an issue with adding or stopping leading slashes on the file names.  Does the latex translator even work with sphinx_gallery?

larsoner:
> I did note an issue with adding or stopping leading slashes on the file names. Does the latex translator even work with sphinx_gallery?

We do a latex build of our own docs:

https://github.com/sphinx-gallery/sphinx-gallery/blob/master/.circleci/config.yml#L73-L75

but admittedly I have never actually looked at it. Looking at it now, it looks like examples and code blocks do get rendered, albeit not completely correctly (page 31+):

https://2118-25860190-gh.circle-artifacts.com/0/latex/python.pdf

jklymak:
> not completely correctly

I'd say many of them do not get rendered at all?  In fact I'm not sure any of them do?

larsoner:
Argh sorry I misspoke, we don't build our core docs but rather the `tinybuild` build:

https://github.com/sphinx-gallery/sphinx-gallery/tree/master/sphinx_gallery/tests/tinybuild

Things like this are actually correct-ish, for example [this example](https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/tests/tinybuild/examples/plot_command_line_args.py) produces:

Screenshot from 2021-04-12 15-28-40

Locally you can build `tinybuild` very easily with `make` inside the `sphinx_gallery/tests/tinybuild` directory.

(FWIW I think we used to build our own docs on CircleCI, too, but got rid of it since we never checked the output and it just added to build time, so switching to tinybuild was just a way to make sure things didn't totally break.)

jklymak:
Thats super helpful for testing, and seeing what the current behaviour is.  Thanks!

jklymak:
Whew, I had to chop most of the cone stuff out to get the tiny build to build on my machine with any sort of speed.  Maybe you need a simpler "microbuild"  ;-)

I think, however, that latex and html both work.

larsoner:
FYI there are some failures in `master` that have been fixed by https://github.com/sphinx-gallery/sphinx-gallery/pull/811, feel free to rebase or merge with `upstream/master` to get the fixes

jklymak:
I've rebased on master, but still get two test failures.  They pass locally, so I'm not sure what the issue is.   :shrug:

jklymak:
OK, down to windows errors.  Its a little hard for me to guess what those might be.

sphinx_gallery/directives.py
```diff
@@ -2,11 +2,14 @@
 Custom Sphinx directives
 ========================
 """
-
 import os
+from pathlib import Path
```
larsoner, line 6: Looking at your test, the failure is that some `path/to/file.ext` fails. It's almost certainly related to this -- somewhere you need to ensure that the URI that ends up in the `img src="..."` HTML tag is constructed using the Posix convention rather than the system native convention (which will be wrong on Windows because it will use `\` instead of `/`). I'd expect you to need to use `PosixPath` instead of `Path` somewhere to fix it.

sphinx_gallery/directives.py
```diff
@@ -77,3 +80,192 @@ def run(self):
...
+
+def visit_imgsrcset_latex(self, node):
+
+    if node['srcset'] is not None:
+        imagedir, srcset = _copy_images(self, node)
+        maxmult = -1
+        # choose the highest res version for latex:
+        for key in srcset.keys():
+            maxmult = max(maxmult, key)
+        node['uri'] = str(Path(srcset[maxmult]).name)
```
larsoner, line 229: This is probably one place

sphinx_gallery/directives.py
```diff
@@ -77,3 +80,192 @@ def run(self):
...
+        srcsetst += f'{relpath}'
+        if mult == 0:
+            srcsetst += ', '
+        else:
+            srcsetst += f' {mult:1.1f}x, '
+    # trim trailing comma and space...
+    srcsetst = srcsetst[:-2]
+
+    # make uri also be relative...
+    uri = Path(imagerel, Path(node['uri'][1:]).name)
```
larsoner, line 207: This is probably another one

sphinx_gallery/directives.py
```diff
@@ -77,3 +80,192 @@ def run(self):
...
+
+        imagenm = self.arguments[0]
+        image_node['alt'] = self.options.get('alt', '')
+        image_node['class'] = self.options.get('class', None)
+
+        # we would like uri to be the highest dpi version so that
+        # latex etc will use that.  But for now, lets just make
+        # imagenm
+
+        image_node['uri'] = imagenm
```
larsoner, line 149: This might be another (but I'm not usre where `self.arguments[0]` gets created)

sphinx_gallery/directives.py
```diff
@@ -77,3 +80,192 @@ def run(self):
...
+
+    return imagedir, srcset
+
+
+def depart_imgsrcset_html(self, node):
+    pass
+
+
+def visit_srcset_other(self, node):
+    if node['uri'][0] == '/':
```
larsoner, line 259: and this will only work if all `uri`s are Posix flavored (which they should / probably need to be in principle I think)

jklymak:
`cannot instantiate 'PosixPath' on your system`  ha ha...

jklymak:
`ValueError: 'C:\\Users\\VssAdministrator\\AppData\\Local\\Temp\\pytest-of-VssAdministrator\\pytest-0\\root\\auto_examples\\plot_animation.rst' does not start with 'C:\\Users\\VssAdministrator\\AppData\\Local\\Temp\\pytest-of-VssAdministrator\\pytest-0\\root'`  

But I _really_ feel that the first string *does* start with the second....

jklymak:
OMG I finally fixed it.  Had to manually convert `..\..\_images` to `../../_images`, but I think its safe to say that will never have valid backslashes in a posix name.  Its a bit of a sad state of affairs that converting file types is such a PITA in python; you would think `as_posix` would do something useful, but `Path('\a\b\').as_posix()` is `'\a\b'` because that is apparently a valid filename.  Very very hard to debug when you don't have a windows machine. 

This still needs some more tests of the directive, but those will be substantially easier at this point.

jklymak:
I'll make ready for review.  I think all the pieces are there.

larsoner (approved):
Awesome bits of code! Tests seem reasonable, CircleCI output looks good. LGTM +1 for merge.

@lucyleeow did you want to look?

sphinx_gallery/scrapers.py
```diff
@@ -8,14 +8,17 @@
 Collect images that have been produced by code blocks.
 
 The only scrapers we support are Matplotlib and Mayavi, others should
-live in modules that will support them (e.g., PyVista, Plotly).
+live in modules that will support them (e.g., PyVista, Plotly).  These are
```
lucyleeow, line 11:
```suggestion
live in modules that will support them (e.g., PyVista, Plotly).  Scraped images are
```

lucyleeow (commented):
This is just a flyby. Will probably have time to have a more thorough look later this weekend

doc/configuration.rst
```diff
@@ -1098,6 +1099,41 @@ optimize less but speed up the build time you could do::
...
+        :srcset: /examples/images/sphx_glr_test_001.png, /examples/images/sphx_glr_test_001_2_0x.png 2.0x
+        :class: sphx-glr-single-img
+
+Which is converted to html by the custom directive as::
+
+    .. <img src="../_images/sphx_glr_test_001.png" alt="test", class="sphx-glr-single-img", 
+        srcset="../_images/sphx_glr_test_001.png, ../_images/sphx_glr_test_001_2_0x.png 2.0x>
+
+Note of course that this leads to a larger website, but responsive images will only serve the 
+appropriate-sized images to website visitors.
```
lucyleeow, line 1135: Should maybe mention the things Eric said in https://github.com/sphinx-gallery/sphinx-gallery/pull/808#issuecomment-818043417

> We just have to clearly state in the docs that the sphinx_gallery.img_srcset extension only supports specific tags, and only HTML and LaTeX builders.

sphinx_gallery/scrapers.py
```diff
@@ -362,16 +402,36 @@ def figure_rst(figure_list, sources_dir, fig_titles=''):
...
         Titles of figures, empty string if no titles found. Currently
         only supported for matplotlib figures, default = ''.
+    srcsetpaths : list or None
+        List of dictionaries containing absolute paths.  If
+        empty, then srcset field is populated with the figure path.
+        (see ``image_srcset`` configuration option).  Otherwise,
+        each dict is of the form
+        {0: /images/image.png, 2.0: /images/image_2_0x.png}
+        where the key is the multiplication factor and the contents
+        the image created above.
```
lucyleeow, line 412:
```suggestion
        the path to the image created above.
```
?

lucyleeow (approved):
LGTM. Just small note

lucyleeow:
The test failures appear to just be lint:
```
sphinx_gallery/scrapers.py:11:80: E501 line too long (83 > 79 characters)
```

doc/configuration.rst
jklymak: Added:
```
Note that the ``.. image-srcset`` directive currently ignores other ``.. image`` directive 
tags like ``width``, ``height``, and ``align``.  It also only works with the *html* and 
*latex*  builders.  
```

timhoffm (commented):
I think `ImageSrcset` \ `.. image-srcset:` is a too specific name for the directive. Given that we might want to add additional functionality later, like lazy loading #803, I propose to choose a more general name, e.g. `ImageExt` / `.. image-ext:` for *extended* or any other name not focusing on a single feature.

larsoner:
+1 for image-ext

jklymak:
how about `image-sg` since it is a sphinx-gallery extension?

jklymak:
... though FWIW, I don't think this extension is being used for the gallery thumbnails, so I'm not sure lazy loading is that imperative.

timhoffm:
`image-sg` is also fine.

> ... though FWIW, I don't think this extension is being used for the gallery thumbnails

Not yet... 😄

jklymak:
These are really a pretty trivial extension of the `:image:` directive.   I wonder if docutils would entertain adding the tags.  OTOH they are still on source forge, so their development path is not trivial

timhoffm:
I've posted a request to docutils for the almost trivial case of lazy loading (it's one parameter on the directive that directly translates to a parameter on the `` tag): https://sourceforge.net/p/docutils/feature-requests/78/

That's sitting for three weeks without any response. So I would not hold my breath. If you want to move something there, maybe you'll have to post to the mailinglist.

jklymak:
@larsoner @lucyleeow before I spool up my search and replace job, which would you prefer? `image-sg` or `image-ext`?  

WRT to docutils, we can look to bump it all up a level if it works for a while.  Its probably easier to convince them that its useful if we have it working.

larsoner:
Either is fine with me, no real opinion either way

jklymak:
OK, I think I did the rename correctly.  Just the directive (and its `ImageSg` class) changed names.  The config option is still `image_srcset`.  Thanks!

sphinx_gallery/scrapers.py
```diff
@@ -110,6 +113,17 @@ def matplotlib_scraper(block, block_vars, gallery_conf, **kwargs):
     from matplotlib.animation import Animation
     image_path_iterator = block_vars['image_path_iterator']
     image_rsts = []
+    # Check for srcset hidpi images
+    srcset = gallery_conf.get('image_srcset', [""])
```
timhoffm, line 117:
```suggestion
    srcset = gallery_conf.get('image_srcset')
    if not srcset:
        srcset = [""]
```
This also catches the case of `gallery_conf['image_srcset'] = []`. AFAICT an emtpy  `srcset` is not valid in HTML, so we should handle this. The alternative would be to remove the `srcset` attribute from ``, but that's more work and I don't think it can do harm if it's present with the 1x data.

jklymak: Just to be clear: `""` translates simply to  `:srcset: images/myimage.png`

timhoffm: I think that's ok https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-srcset.

jklymak: I guess we could remove the explicit "" and just assume it is implied.  And sure we could not include the srcset tag if it's empty.

timhoffm: Whichever works for you,  as long as we don't insert an empty parameter in HTML.

jklymak:
So, this version changed a little bit.  Now one would say `'image_srcset': ["2x"]` for a factor of 2, instead of `["", "2x"]`.  If you don't want the feature, an empty list is fine now `[]`.  

OTOH, it still writes `srcset` into the directive and the html, it just puts the 1-x version only in the list of files.

sphinx_gallery/scrapers.py
```diff
@@ -110,6 +113,21 @@ def matplotlib_scraper(block, block_vars, gallery_conf, **kwargs):
...
+        if (len(st) > 0) and (st[-1] == 'x'):
+            # "2x" = "2.0"
+            srcset_mult_facs += [float(st[:-1])]
+        elif st == "":
+            pass
+        else:
+            raise ExtensionError(
+                f'Invalid value for image_srcset parameter: "{st}". '
+                'Must be a list of strings with the multiplicative '
+                'factor followed by an "x".  e.g. ["2.0x", "1.5x"]')
```
larsoner, line 129:
```suggestion
                'factor followed by an "x".  e.g. ["2.0x", "1.5x"]')
```

doc/configuration.rst
```diff
@@ -1098,6 +1099,46 @@ optimize less but speed up the build time you could do::
...
+=======================
+
+Web browsers allow a ``srcset`` parameter to the ``<img>`` tag that 
+allows the browser to support `responsive resolution images 
+<https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images>`__
+for hi-dpi/retina displays. Sphinx Gallery supports this via the ``image_srcset`` parameter::
+
+    sphinx_gallery_conf = {
+        ...
+        'image_srcset': ["", "2x"],
```
larsoner, line 1114: Outdated, right?
```suggestion
        'image_srcset': ["2x"],
```

doc/configuration.rst
```diff
@@ -1098,6 +1099,46 @@ optimize less but speed up the build time you could do::
...
+    }
+
+that saves a 1x image at the normal figure dpi (usually 100 dpi) and a 2x 
+version at twice the density (e.g. 200 dpi).  The default is no extra images 
+(``'image_srcset': [""]``), and you can specify other resolutions if desired (e.g.
```
larsoner, line 1119:
```suggestion
(``'image_srcset': []``), and you can specify other resolutions if desired (e.g.
```

doc/configuration.rst
```diff
@@ -1098,6 +1099,46 @@ optimize less but speed up the build time you could do::
...
+        :alt: test
+        :srcset: /examples/images/sphx_glr_test_001.png, /examples/images/sphx_glr_test_001_2_0x.png 2.0x
+        :class: sphx-glr-single-img
+
+Which is converted to html by the custom directive as::
+
+    .. <img src="../_images/sphx_glr_test_001.png" alt="test", class="sphx-glr-single-img", 
+        srcset="../_images/sphx_glr_test_001.png, ../_images/sphx_glr_test_001_2_0x.png 2.0x>
+
+This leads to a larger website, but responsive images will only serve the appropriate-sized
```
larsoner, line 1135: or "clients"?
```suggestion
This leads to a larger website, but responsive browsers will only serve the appropriate-sized 
```

sphinx_gallery/directives.py
```diff
@@ -77,3 +80,198 @@ def run(self):
...
+    """ parse st"""
+    entries = st.split(',')
+    srcset = {}
+    for entry in entries:
+        spl = entry.strip().split(' ')
+        if len(spl) == 1:
+            srcset[0] = spl[0]
+        elif len(spl) == 2:
+            mult = spl[1][:-1]
+            srcset[float(mult)] = spl[0]
```
larsoner, line 166: Safer to `else: raise ...`?

sphinx_gallery/directives.py
```diff
@@ -77,3 +80,198 @@ def run(self):
...
+    srcsetst = ''
+    for mult in srcset:
+        nm = os.path.basename(srcset[mult][1:])
+        # ../../_images/plot_1_2_0x.png
+        relpath = imagerel+nm
+        srcsetst += f'{relpath}'
+        if mult == 0:
+            srcsetst += ', '
+        else:
+            srcsetst += f' {mult:1.1f}x, '
```
larsoner, line 205:
```suggestion
        if mult != 0:
            srcsetst += f' {mult:1.1f}x'
        srcsetst += ', '
```

sphinx_gallery/tests/test_full.py
```diff
@@ -236,11 +236,20 @@ def test_image_formats(sphinx_app):
...
+            want_html = 'src="%s"' % (img_fname0,)
             assert want_html in html
+            if ext in ('png', 'jpg'):  # check 2.0x file.  tests the directive
+                img_fname2 = ('../_images/sphx_glr_%s_%03d_2_0x.%s' %
+                              (ex, num, ext))
+                file_fname2 = op.join(generated_examples_dir, img_fname2)
+                assert op.isfile(file_fname2), file_fname2
+
+                want_html = 'srcset="%s, %s 2.0x"' % (img_fname0, img_fname2)
+                assert want_html in html
```
larsoner, line 251: Maybe more complete as
```suggestion
            img_fname2 = ('../_images/sphx_glr_%s_%03d_2_0x.%s' %
                          (ex, num, ext))
            file_fname2 = op.join(generated_examples_dir, img_fname2)
            want_html = 'srcset="%s, %s 2.0x"' % (img_fname0, img_fname2)
            if ext in ('png', 'jpg'):  # check 2.0x file.  tests the directive
                assert op.isfile(file_fname2), file_fname2
                assert want_html in html
            else:
                assert ext == 'svg'
                assert not op.isfile(file_fname2), file_fname2
                assert want_html not in html
```

sphinx_gallery/scrapers.py
```diff
@@ -415,24 +514,26 @@ def _single_line_sanitize(s):
...
+      .. image-sg:: /%s
           :alt: %s
+          :srcset: %s
           :class: sphx-glr-multi-img
 """
 
-SINGLE_IMAGE = """
-.. image:: /%s
-    :alt: %s
-    :class: sphx-glr-single-img
```
larsoner, line 429: Let's keep this constant around since it's likely that custom scrapers in the wild have used it, and the maintenance cost of keeping this in the namespace is negligible.

larsoner (commented):
Other than some minor comments/suggestions, LGTM!

larsoner:
Also I pushed a tiny commit to make CIs happy again, feel free to force-push right over it with the same PEP8 fix + other fixes if you want

doc/configuration.rst
jklymak: Oops, yes, that change didn't get pushed...

doc/configuration.rst
jklymak: Changed to 
```
This leads to a larger website, but clients that support the ``srcset`` tag will only 
download the appropriate-sized images.
```

sphinx_gallery/directives.py
jklymak: Thats more elegant, but less clear to me.  I'm a very literal programmer ;-)

larsoner: To me it also matches better with the statement below about removing the trailing comma-space, but agreed it's really six of one half a dozen of the other so is fine as is

sphinx_gallery/tests/test_full.py
```diff
@@ -241,14 +241,17 @@ def test_image_formats(sphinx_app):
...
-                img_fname2 = ('../_images/sphx_glr_%s_%03d_2_0x.%s' %
-                              (ex, num, ext))
-                file_fname2 = op.join(generated_examples_dir, img_fname2)
                 assert op.isfile(file_fname2), file_fname2
-
-                want_html = 'srcset="%s, %s 2.0x"' % (img_fname0, img_fname2)
                 assert want_html in html
+            else:
+                assert ext == 'svg'
+                assert not op.isfile(file_fname2), file_fname2
```
jklymak, line 253: ... ha, this test caught a bug.  This file *was* being made...

jklymak: Well, hmm, I guess this is a bit of a question - dpi does mean something for rasterized artists in an svg.  So what I was doing was making a 100-dpi svg and a 200-dpi svg.  Of course if there were no rasterized artists they are the same.  But if there are rasterized artists, they can be different.  

So, is it worth making svg behave differently?

larsoner: I don't know enough about HTML/CSS to know for sure, but based on some quick reading just now it seems like SVG shouldn't need multiple versions. I would try with and without multiple versions and use different devices in Chrome's developer mode (F12) where you can set the pixel device ratio explicitly. In theory some of the builtin ones should have different ratios, but it looks like you can set it explicitly in a custom device so I'd go that route:

Screenshot from 2021-04-21 10-32-18

larsoner: Well I tried using different pixel ratios there and it didn't make any difference when I toggled

jklymak: Pngs change though?   

The svg in tiny doesn't have any rasterized artists so I wouldn't expect a difference in the svg.  But it is possible to create svg with rasterized artists like pcolormesh or image.

larsoner: Actually it looks like you can do it just by making the top-bar have the ratio option, and refreshing -- it correctly pulls the 2x source but looks identical to me (not surprising I guess):

Screenshot from 2021-04-21 10-38-00

At least this one SVG one looks identical as well:

Screenshot from 2021-04-21 10-40-11

larsoner: The PNGs do seem to change (even though I can't detect it by eye), at least acconding to the `currentSrc` that shows up.

> The svg in tiny doesn't have any rasterized artists so I wouldn't expect a difference in the svg. But it is possible to create svg with rasterized artists like pcolormesh or image.

Ahh that makes sense, hadn't considered the rasterized parts. Then I suppose it does make sense to have the additional sizes

jklymak: On my retina display there is a pretty clear difference in the websites before and after ;-)  Though I will admit the default matplotlib line widths end up looking a little spindly!

jklymak:
Is your release settled enough to merge this?   Then we can start eating our (my?) dogfood in Matplotlib...

larsoner:
Yes, thanks @jklymak !

jklymak:
Thanks for being willing to add this!

QuLogic:
We're coming close to our release, but need to depend on a git version of sphinx-gallery for this. Is it possible to make a release soon?

lucyleeow:
We're probably due for a release, there are a few small PRs I might get in before hand. Can do it this weekend, early next week - WDYT @larsoner ?

lucyleeow:
I've forgotten that my laptop just died on me (and I can't use the work laptop, vpn restrictions) - so the release would be pending on me fixing it.

larsoner:
@QuLogic what's your deadline?

Yes @lucyleeow feel free to release when you have time. In the meantime, if matplotlib needs it before your setup is working again, I can do it

lucyleeow:
Thanks @larsoner, I almost have it set up again (my partitions saved me!)

QuLogic:
The aim is for the end of the month, roughly speaking, depending on how things work out.

sphinx_gallery/directives.py
```diff
@@ -77,3 +82,200 @@ def run(self):
...
+        shutil.copyfile(abspath, imagedir / abspath.name)
+
+    return imagedir, srcset
+
+
+def depart_imgsg_html(self, node):
+    pass
+
+
+def visit_sg_other(self, node):
```
lucyleeow, line 268: Would anyone be able to explain this function? It does not seem to be used anywhere?
Saw while doing some maintenance dealing with our new linting. Thanks.
cc @jklymak @larsoner

sphinx_gallery/directives.py
```diff
@@ -77,3 +82,200 @@ def run(self):
...
+"""
+Image sg for responsive images
+"""
+
+
+class imgsgnode(nodes.General, nodes.Element):
+    pass
+
+
+def directive_boolean(value):
```
lucyleeow, line 96: Similar for this function, it does not seem to be used? Thanks!
cc @jklymak @larsoner

sphinx_gallery/directives.py
jklymak: This isn't being used by matplotlib, which is the only context where I use sphinx-gallery.  I imagine it was left over from hacking and didn't get cleaned up (so good lint material).

lucyleeow: Thanks, I'm not so familiar with Sphinx app stuff so wasn't sure if I missed something.

sphinx_gallery/directives.py
larsoner: I would try removing them and see if anything breaks. We've been decent at adding new tests for new functionality for the last few years so hopefully we can rely on them

PR #810, opened 2021-04-13
Fix SVG default thumbnail support

jacobolofsson:
In gen_rst.save_thumbnail() the file extension is from file_conf is
always used, regardless of the value of 'default_thumb_file'.  If
'default_thumb_file' is .svg (or .gif) it is still passed to pillow,
which does not handle them properly.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -310,11 +310,18 @@ def save_thumbnail(image_path_template, src_file, script_vars, file_conf,
         img = thumbnail_image_path
     elif not os.path.exists(thumb_file):
         # create something to replace the thumbnail
-        img = gallery_conf.get("default_thumb_file")
-        if img is None:
-            img = os.path.join(glr_path_static(), 'no_image.png')
+        default_thumb_path = gallery_conf.get("default_thumb_file")
+        if default_thumb_path is None:
+            default_thumb_path = os.path.join(
+                glr_path_static(),
+                'no_image.png',
+            )
+        img, ext = _find_image_ext(default_thumb_path)
     else:
         return
+    # update extension, since gallery_conf setting can be different
+    # from file_conf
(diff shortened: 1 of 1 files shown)


jacobolofsson:
Pytest is failing on test_embed_links_and_styles. It fails locally for me on current master also. I would be grateful to anyone a bit more experienced in this lib to help resolve
```
    def test_embed_links_and_styles(sphinx_app):
        """Test that links and styles are embedded properly in doc."""
        out_dir = sphinx_app.outdir
        src_dir = sphinx_app.srcdir
        examples_dir = op.join(out_dir, 'auto_examples')
        assert op.isdir(examples_dir)
        example_files = os.listdir(examples_dir)
        assert 'plot_numpy_matplotlib.html' in example_files
        example_file = op.join(examples_dir, 'plot_numpy_matplotlib.html')
        with codecs.open(example_file, 'r', 'utf-8') as fid:
            lines = fid.read()
        # ensure we've linked properly
>       assert '#module-matplotlib.colors' in lines
E       assert '#module-matplotlib.colors' in '\r\n<!DOCTYPE html>\r\n\r\n<html>\r\n  <head>\r\n    <meta charset="utf-8" />\r\n    <meta name="viewport" content="w...lib.rst.txt"\r\n          rel="nofollow">Page source</a>\r\n    </div>\r\n\r\n    \r\n\r\n    \r\n  </body>\r\n</html>'

larsoner:
I'll take a look in https://github.com/sphinx-gallery/sphinx-gallery/pull/811

jacobolofsson:
Thanks @larsoner that seemed to do the trick

larsoner:
In principle it would be nice to have a test for this but I think it would be too much of a pain. In the meantime this seems reasonable to me and CIs are happy at least with the PNG case, thanks @jacobolofsson

Issue #812, opened 2021-04-15
Minigalleries are generated even for objects without examples


NicolasHug:
For example in https://sphinx-gallery.github.io/stable/gen_modules/sphinx_gallery.scrapers.html#module-sphinx_gallery.scrapers the functions all have an empty minigallery saying "Examples using ...", but there's nothing to show.

As often minigalleries are automatically generated by adding the directive in a template, it'd be nice to only add it if there's actually a backreference to show.

larsoner:
Yes, this has bothered me for a while in repos where I use SG!

PR #813, opened 2021-04-15
ENH: Only add minigallery if there's something to show

NicolasHug:
Closes https://github.com/sphinx-gallery/sphinx-gallery/issues/812

This PR avoids showing the minigallery if there's no example using the object. It's probably too hacky for inclusion as-is but I'm not sure how to best proceed, so I'd be happy to address any feeback :)

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -444,7 +444,8 @@ The ``add-heading`` option adds a heading for the mini-gallery, which will be a
 default generated message if no string is provided as an argument.  The example
 mini-gallery shown above uses the default heading.  The level of the heading
 defaults to ``^``, but can be changed using the ``heading-level`` option, which
-accepts a single character (e.g., ``-``).
+accepts a single character (e.g., ``-``). The mini-gallery will only be shown
+if the item (here ``numpy.exp``) is actually used in an example.
 
 You can also list multiple items, separated by spaces, which will merge all
 examples into a single mini-gallery, e.g.:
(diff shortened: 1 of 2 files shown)


sphinx_gallery/directives.py
```diff
@@ -58,6 +58,14 @@ def run(self):
             heading_level = self.options.get('heading-level', '^')
             lines.append(heading_level * len(heading))
 
+        def should_show(obj):
+            # TODO: get the path in a cleaner and more robust way
+            path = os.path.abspath(os.path.join("source", backreferences_dir, '{}.examples'.format(obj)))
```
larsoner, line 63: We populate `sphinx_gallery_conf` with extra values:

https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/gen_gallery.py#L161

One of these is `src_dir`, so you should be able to do something like:
```suggestion
            src_dir = config.sphinx_gallery_conf['src_dir']
            path = os.path.abspath(os.path.join(src_dir, backreferences_dir, '{}.examples'.format(obj)))
```

lucyleeow (approved):
LGTM. Might be worth adding a note in `configuration.rst` under [Add mini-galleries](https://github.com/sphinx-gallery/sphinx-gallery/blob/master/doc/configuration.rst#add-mini-galleries-for-api-documentation) that a backreference rst file will only be generated for functions/classes that are used, so you will not be able include to mini galleries for those not used

doc/configuration.rst
```diff
@@ -444,7 +444,8 @@ The ``add-heading`` option adds a heading for the mini-gallery, which will be a
 default generated message if no string is provided as an argument.  The example
 mini-gallery shown above uses the default heading.  The level of the heading
 defaults to ``^``, but can be changed using the ``heading-level`` option, which
-accepts a single character (e.g., ``-``).
+accepts a single character (e.g., ``-``). The mini-gallery will only be shown
+if the item (here ``numpy.exp``) is actually used in an example.
```
lucyleeow, line 448:
```suggestion
if the item (here ``numpy.exp``) is actually used or referred to in an example.
```

lucyleeow:
Thanks @NicolasHug !

PR #814, opened 2021-04-16
DOC Include SOURCEDIR in Makefile

NicolasHug:
I think the `SOURCEDIR` variable is missing from the suggested Makefile in the docs. Running it as-is was raising an error for me:

```
usage: sphinx-build [OPTIONS] SOURCEDIR OUTPUTDIR [FILENAMES...]
sphinx-build: error: the following arguments are required: outputdir, filenames
```


lucyleeow:
Thanks!

PR #816, opened 2021-04-19
BUG: Fix CSS selector

### LINKED ISSUE #815: Custom CSS for space above title target conflicts with pydata-sphinx-theme
The following snippet from sphinx-gallery's custom css conflicts with the pydata-sphinx-theme (see https://github.com/pydata/pydata-sphinx-theme/issues/389, https://github.com/mne-tools/mne-python/pull/9288):

https://github.com/sphinx-gallery/sphinx-gallery/blob/ec22616661d3728f6fc962d622ddba5686f5356b/sphinx_gallery/_static/gallery.css#L148-L154

Because it adds some space above each header on the page, it confuses the "highlight TOC on scroll" feature of the right side page TOC. 

Now, I am not really able to understand what the above CSS is actually intended to be doing .. (but I am also no CSS expert ;)).

It was introduced in https://github.com/sphinx-gallery/sphinx-gallery/pull/362, and seems to be related to the "note" that is added on the top of a page (with the download link). 
But, the way the CSS selector is now implemented (`.sphx-glr-example-title > :target::before`), it doesn't actually select the top header that is directly below the note, but every other header lower on the page (and adding some space above that). So it might be that the CSS selector is actually selecting the wrong thing (but again, not fully understanding it). 
Changing the selector to `.sphx-glr-example-title:target::before` would have it select only the top header.

larsoner:
Closes #815

Thanks @jorisvandenbossche for the suggested fix, will merge once CIs come back happy. I've tested locally and everything looks good on MNE-Python.

Also happy to let you make the PR but I figured this would be quicker :)

### DIFF
--- a/sphinx_gallery/_static/gallery.css
+++ b/sphinx_gallery/_static/gallery.css
@@ -145,7 +145,7 @@ div.sphx-glr-download a:hover {
   background-color: #d5d57e;
 }
 
-.sphx-glr-example-title > :target::before {
+.sphx-glr-example-title:target::before {
   display: block;
   content: "";
   margin-top: -50px;


jorisvandenbossche:
Thanks!

> Also happy to let you make the PR but I figured this would be quicker :)

I am totally fine with not doing the PR ;)

PR #828, opened 2021-05-02
ENH Add config that generates dummy images to prevent missing image warnings

### LINKED ISSUE #751: Add a mode that fails only for rst warnings and does not run examples
Idea by @jeromedockes

Have a mode where the examples are built without running the code and the links to images are not inserted. This mode would be used to detect errors in the restructured text without waiting for examples to run, which can take a long time. It is important in such a mode that the links to images are not inserted, because when the examples are not run, the images are not present, and hence they lead to sphinx warnings that drown the important warnings.

@jeromedockes, did I explain well your ideas? @lucyleeow : does this make sense?

lucyleeow:
closes #751

It is possible for this configuration to work only when the gallery is NOT executable (e.g., when plot_gallery is False) but this complicates the code and makes testing (using tinybuild) difficult, and possibly unnecessarily restrains the config function. I have left it as always creating dummy images. The images are created after executing the code block (so after any 'real' figures are saved) - thus would not overwrite any generated 'real' figures. Open to changing.

TODO
Still need to add documentation.
Should amend to test in example that generates images, to ensure no real images are overwritten

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -54,7 +54,7 @@ Some options can also be set or overridden on a file-by-file basis:
 - ``# sphinx_gallery_line_numbers`` (:ref:`adding_line_numbers`)
 - ``# sphinx_gallery_thumbnail_number`` (:ref:`choosing_thumbnail`)
 - ``# sphinx_gallery_thumbnail_path`` (:ref:`providing_thumbnail`)
-- ``# sphinx_gallery_dummy_image`` (:ref:`dummy_images`)
+- ``# sphinx_gallery_dummy_images`` (:ref:`dummy_images`)
 
 Some options can be set on a per-code-block basis in a file:
 
@@ -1233,7 +1233,7 @@ which you then reference manually elsewhere, e.g.,:
 To prevent missing image file warnings when building without executing, you
 can add the following to the example file::
 
-    # sphinx_gallery_dummy_image=2
+    # sphinx_gallery_dummy_images=2
 
 This will cause Sphinx-Gallery to generate 2 dummy images with the same
 naming convention and stored in the same location as images that would be
@@ -1241,7 +1241,7 @@ generated when building with execution.
 
 .. note::
     This configuration **only** works when ``plot_gallery`` is False. This
-    means that you will not need to remove any ``sphinx_gallery_dummy_image``
+    means that you will not need to remove any ``sphinx_gallery_dummy_images``
     lines in your examples when you switch to building your gallery with
     execution.
 
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1211,11 +1211,12 @@ Generating dummy images
 =======================
 
 For quick visualization of your gallery, especially during the writing process,
-Sphinx-Gallery allows you to build your gallery without executing any of the
-code (see :ref:`without_execution`). This however, can cause warnings about
-missing image files if you have manually written links to automatically
-generated images. To prevent these warnings you can tell Sphinx-Gallery to
-create a number of dummy images for an example.
+Sphinx-Gallery allows you to build your gallery without executing the
+code (see :ref:`without_execution` and
+:ref:`filename/ignore patterns <build_pattern>`). This however,
+can cause warnings about missing image files if you have manually written
+links to automatically generated images. To prevent these warnings you can
+tell Sphinx-Gallery to create a number of dummy images for an example.
 
 For example, you may have an example ('my_example.py') that generates 2 figures,
 which you then reference manually elsewhere, e.g.,:
@@ -1240,10 +1241,12 @@ naming convention and stored in the same location as images that would be
 generated when building with execution.
 
 .. note::
-    This configuration **only** works when ``plot_gallery`` is False. This
-    means that you will not need to remove any ``sphinx_gallery_dummy_images``
-    lines in your examples when you switch to building your gallery with
-    execution.
+    This configuration **only** works when the example is set to not execute
+    (i.e., the ``plot_gallery`` is False, the example is in `ignore_pattern`
+    or the example is not in `filename_pattern - see
+    :ref:`filename/ignore patterns <build_pattern>`). This means that you will
+    not need to remove any ``sphinx_gallery_dummy_images`` lines in your
+    examples when you switch to building your gallery with execution.
 
 .. _reset_modules:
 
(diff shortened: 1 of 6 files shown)


GaelVaroquaux:
That's a pretty big deal!! Thanks a lot Lucy!

@jeromedockes @NicolasGensollen, do you want to have a look to see if
this is going in the direction that could help nilearn.

lucyleeow:
Thinking about it more, it may be useful for the dummy images to be generated ONLY if the example is not executable (e.g., plot_gallery is False). This way you would not have to remove the config once you are done doing local testing. Though, in the current design, accidentally leaving the config should not cause problems as the images are added after any generated plots, you would just have additional images.

Are there cases when the dummy image generating functionality would be useful when the examples are executable?

Note making it only work when not executable would make testing much more difficult.

jeromedockes:
> Though, in the current design, accidentally leaving the config should not cause problems as the images are added after any generated plots, you would just have additional images.

for the case of nilearn at least I don't think having some extra images
would be a problem, as long as the true images are the ones that show up
in the gallery when they are available.
but having to change this configuration parameter depending on whether
the examples are executed would not be a problem either.

thanks for working on this! let us know when it is ready to try it on
the nilearn example gallery

GaelVaroquaux:
> Thinking about it more, it may be useful for the dummy images to be generated ONLY if the example is not executable

+1

lucyleeow:
@larsoner is it possible to run tinybuild with different config (e.g., plot_gallery: False)? Alternatively we could create another tinierbuild that runs with plot_gallery False, but I worry about maintenance.

larsoner:
From the command line you can always change config values using something like `-D sphinx_gallery_conf.plot_gallery=False`, there is presumably some way to do something equivalent from our app testing infrastructure, too (temporarily setting `sys.argv`, maybe?). This would be my preferred solution.

Another way would be to `monkeypatch` our own `_complete_gallery_conf` to always set `plot_gallery=False`.

Another option (ugliest IMO) is to `shutil.copytree` to a `tmpdir`, then modify `conf.py`, then make the sphinx app.

lucyleeow:
@larsoner I added a new file `test_full_noexec.py` and in the sphinx app here added `confoverrides` argument.

Was not sure what best to do about temp dir:
```
temp_dir = (tmpdir_factory.getbasetemp() / 'root_nonexec').strpath
```

I had to name it something else, as 'root' was already created by `test_full.py`.

larsoner:
Personally I don't use `tmpdir_factory` in pytest but just the `tmpdir` fixture, which IIRC uses a different tempdir per test function so you shouldn't have collisions. But what you did sounds reasonable, too

doc/configuration.rst
```diff
@@ -1204,6 +1205,46 @@ further deferred, if desired).  The following produces only one plot::
...
+
+This will cause Sphinx-Gallery to generate 2 dummy images with the same
+naming convention and stored in the same location as images that would be
+generated when building with execution.
+
+.. note::
+    This configuration **only** works when ``plot_gallery`` is False. This
+    means that you will not need to remove any ``sphinx_gallery_dummy_images``
+    lines in your examples when you switch to building your gallery with
+    execution.
```
larsoner, line 1246: This actually makes it slightly less useful. For example, a common use case in MNE-Python is to do:
```
PATTERN=plot_something make html_dev-pattern
```
which translates to a `sphinx-build .. -D sphinx_gallery_conf.filename_pattern=$PATTERN ...`. Under this condition, according to the documentation here, even with this `dummy_images` line in place, we would see warnings. In other words, I think it should place dummy images 1) if the example isn't going to be executed for any reason (i.e., in exclude, or not matching filename pattern, or plot_gallery=False), and 2) the images don't exist. This I think would make it a "always avoid a missing image warning" pattern, which seems more useful than a "avoid a missing image warning only when plot_gallery is False". But maybe I'm missing something

sphinx_gallery/gen_rst.py
```diff
@@ -865,6 +865,20 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf,
 
     logger.debug("%s ran in : %.2g seconds\n", src_file, time_elapsed)
 
+    # Create dummy images
+    if not executable:
```
larsoner, line 869: Okay it looks like the documentation above that I just complained about might just be incorrect -- it's not just in the `not plot_gallery` condition but also excludes, etc.

sphinx_gallery/gen_rst.py
```diff
@@ -865,6 +865,20 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf,
...
+        if dummy_image is not None:
+            if type(dummy_image) is not int:
+                raise ExtensionError(
+                    'sphinx_gallery_dummy_images setting is not a number, '
+                    'got %r' % (dummy_image,))
+
+            image_path_iterator = script_vars['image_path_iterator']
+            stock_img = os.path.join(glr_path_static(), 'no_image.png')
+            for _, path in zip(range(dummy_image), image_path_iterator):
+                copyfile(stock_img, path)
```
larsoner, line 880: I would do a `if not os.path.isfile(path):` here. That way if someone incrementally rebuilds examples with `PATTERN=... make html_dev-pattern` we only overwrite files if we need to.

doc/configuration.rst
lucyleeow: Ah good point. It does behave as you suggest - if the example is not executed for any reason, the dummy images will be generated. I should update the doc.

With regard to 2 - good suggestion, will implement.

lucyleeow:
> Personally I don't use `tmpdir_factory` in pytest but just the `tmpdir` fixture, which IIRC uses a different tempdir per test function so you shouldn't have collisions. But what you did sounds reasonable, too

How about I amend to using `tmpdir` (instead of tmpdir_factory - maybe we don't even need this fun) in both `test_full.py` and `test_full_noexec.py` ?

lucyleeow:
> How about I amend to using tmpdir (instead of tmpdir_factory - maybe we don't even need this fun) in both test_full.py and test_full_noexec.py ?

So this does not work because `tmpdir` is a function scope fixture, and it can't be used in `sphinx_app` which is module scoped.

lucyleeow:
This is ready for review if you want to take a look at it again.

Can fix errors once #830 and #831 are merged in.

larsoner:
There is a merge conflict now

lucyleeow:
@jeromedockes feel free to give it a go...!

PR #830, opened 2021-05-15
BUG Fix test_minigallery_directive failing on Windows

lucyleeow:
Parameterise test_minigallery_directive to make debugging easier.

Fix `test_minigallery_directive` failing in Windows. HTML output in Windows has more (new) lines, so had to make `nlines` bigger to capture some sections completely. It should not interfer with the linux html output as have only increased `nlines` by 1-3, and this should max only capture the test name part of the next section, e.g.:

```html
</div>
</div><div class="sphx-glr-clear"></div><p>Test 2-D-D</p>
</div>
```
and not capture the heading lines.


lucyleeow:
This is good to go. Python nightly fixed in #831

larsoner:
Thanks @lucyleeow

PR #831, opened 2021-05-15
Bug Pin markupsafe version in Python nightly

lucyleeow:
Fix
> ../../../.local/lib/python3.10/site-packages/jinja2/__init__.py:6: in <module>
    from markupsafe import escape
/usr/lib/python3/dist-packages/markupsafe/__init__.py:13: in <module>
    from collections import Mapping
E   ImportError: cannot import name 'Mapping' from 'collections' (/usr/lib/python3.10/collections/__init__.py)

I think this is fixed here: https://github.com/pallets/markupsafe/commit/86d4146f0debc004b4a8b6aa69df818ef4ad9080

Pin markupsafe to version >=1.1

(Previously we had to pin jinja2 due to same eror, I wonder if adding all these pins is robust/a good fix?)


larsoner:
Thanks @lucyleeow

PR #838, opened 2021-06-25
Sanitize ANSI characters from generated RST

agramfort:
See the problem in the output block in https://braindecode.org/auto_examples/plot_sleep_staging.html#training

See all the `[36m` which are otherwise used to get colors in the terminal.

With this fix it looks much better !

cc @robintibor 

Screenshot 2021-06-25 at 18 03 35

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,14 @@
 Change Log
 ==========
 
+Latest
+------
+
+**Fixed bugs:**
+
+-  Remove `ANSI characters <https://en.wikipedia.org/wiki/ANSI_escape_code>`_ from HTML output `#838 <https://github.com/sphinx-gallery/sphinx-gallery/pull/838>`__ (`agramfort <https://github.com/agramfort>`__)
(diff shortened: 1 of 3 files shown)


agramfort:
@lucyleeow what do you think? Any guidance to add a test or anything you expect from me is very welcome

lucyleeow:
A test similar to `test_output_indentation` in `test_gen_rst.py` that checks that the output is what you expect would do.

(This is for my future reference: https://en.wikipedia.org/wiki/ANSI_escape_code)

agramfort:
let me know what you think

>

lucyleeow (approved):
LGTM! Just a note - we don't bother with the change log as we auto generate this from closed issues and merged PRs when doing the release.

Will wait a fews days to see if anyone else reviews, then merge!

agramfort:
I am sure @larsoner will be glad to review :)

larsoner:
Thanks @agramfort !

Issue #399, opened 2018-07-06
How to get the `.css` files to copy over on building the docs?


choldgraf:
This is only tangentially related to sphinx-gallery, but I have a quick question for @Titan-C (or whoever might know)

I'm having a heck of a time figuring out how sphinx-gallery is able to get its `.css` file copied over to the built `_static` directory (so that it can be linked in pages). I see that https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/gen_gallery.py#L116 seems to be adding the path to the static folder in this repo, but I demoed this on a different repo and none of the files were copied over.

Anyone know how we accomplish this in sphinx-gallery?

Titan-C:
This is one of Sphinx's dark magic(you need to do 2 things). Sphinx
should copy anything that is on its _static directory, or any directory
in the list html_static_path.

Nevertheless to actually use it you need to use a link to it. We do it
with

app.add_stylesheet('gallery.css')

https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/gen_gallery.py#L360

choldgraf:
so in my case, I have both `add_stylesheet` there (which adds the relevant line to the `<head>` section, and I've added to the static path with `app.html_static_path.append(path_to_my_files)`, but those files aren't actually copied over on build. It's super confusing to me :-P

In the end I just pushed those files to github and am using github as a CDN to load them, but this is quite hacky :-/

maartenbreddels:
I didn't understand this as well, and found a solution: https://github.com/sphinx-doc/sphinx/issues/2090
(also knowing sphinx-gallery could do it)

yjmantilla:
@maartenbreddels Thank you! I think your solution should be in the sphinx gallery documentation. It doesn't make sense for such a basic functionality to be so hidden away in an issue. 

Btw, there was an API change to your solution, instead of app.add_stylesheet() it should be app.add_css_file() 

https://github.com/sphinx-doc/sphinx/issues/7747

lucyleeow:
I recently added some documentation here: https://sphinx-gallery.github.io/stable/advanced.html#altering-sphinx-gallery-css but did not detail the `app.add_css_file()` part

Please feel free to put in a PR to make it more clear/accessible. This can be a bit tricky as I wasn't sure how much to document as it is sphinx functionality.

yjmantilla:
[...]

PR #845, opened 2021-07-06
Minor documentation about overwriting css files

yjmantilla:
@lucyleeow Here is the PR related to #399  . I mentioned that the css file had to be registered through the add_css_file() method.

I also included the recipe from @maartenbreddels , problem with including this is that:

- Is more directly related to sphinx
- I'm not even sure why it works (just some minor intuition of "it makes sense")
- I'm still unsure why just doing ``app.html_static_path.append(path_to_my_files);app.add_css_file('gallery.css');`` does not work. I have the same problem as @choldgraf . 
 
My understanding is that it should work and seems to be the expected behavior since after all it seems what sphinx-gallery does  by itself:

https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/gen_gallery.py#L350-L355

I'm yet too much of a sphinx newbie to understand this.

The problem with not including the recipe in the documentation is that is a headache for something seemingly simple. Another option is to link the issue rather than including the recipe directly on the documentation.

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -406,6 +406,20 @@ the path(s) listed will be copied over after the builtin Sphinx-Gallery files,
 so if you have a file named "gallery.css", it will overwrite the builtin
 "gallery.css".
 
+To complete the process though, you need to register the stylesheet
+as mentioned in the `sphinx documentation 
+<https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_css_file>`_ .
+
+A quick recipe you can do if you find that the files are not getting copied over 
+is adding the following to your ``conf.py`` file:
+
+.. code-block:: python
+
+    def setup(app):
+        app.connect('builder-inited', lambda app: app.config.html_static_path.append('_static'))
+        app.add_css_file('gallery.css')
+
+
 The appearance of :ref:`code links <stylizing_code_links>` and
 :ref:`thumbnail size <setting_thumbnail_size>` can be altered via addition
 of your own CSS. Can also do things like hide the download buttons in the


lucyleeow:
Just to clarify this is only needed for the case when you want to over-write existing .css files?

yjmantilla:
> Just to clarify this is only needed for the case when you want to over-write existing .css files?

My intuition is that that is correct. One needs to register a css file for it to be taken into account by sphinx. 

What I have seen is that  if I have the following "_static" directory tree in the source:

```yaml
_static:
  - gallery.css
  - whatever.txt
```

Then we have the following results given each case:

## 1
```python
html_static_path = ['_static']
```
Doesn't overwrite gallery.css but copies other files (like one i just named "whatever.txt") . So it seems that some files dont need registration as css does.

## 2
```python
html_static_path = ['_static']
def setup(app):
    app.add_css_file('gallery.css')
```

Same as above, copies whatever.txt but does not overwrite gallery.css . Although my intuition is that it SHOULD overwrite the css,  it just doesn't ever do.

## 3
```python
html_static_path = ['_static']
def setup(app):
    app.connect('builder-inited', lambda app: app.config.html_static_path.append('_static'))
    app.add_css_file('gallery.css')
```

Copies whatever.txt and overwrites gallery.css .

lucyleeow:
Sorry for the delay, I didn't realise it was only required to overwrite existing files. I'm in 2 minds about it...WDYT @larsoner ?
(I think I fixed the CI's, if you merge/rebase master, hopefully the tests will go green)

lucyleeow:
Have checked and can confirm https://github.com/sphinx-gallery/sphinx-gallery/pull/845#issuecomment-876102461 is all correct. I am inclined to add this to the documentation or at least amend our documentation as the following line is a bit misleading:

>  The static files in the path(s) listed will be copied over after the builtin Sphinx-Gallery files, so if you have a file named “gallery.css”, it will overwrite the builtin “gallery.css”.

@larsoner can you confirm if the above is true, and thus sphinx just doesn't seem to copy files over if they already exist?

larsoner:
I'm not sure whether or not it's true, I would confirm via trial and error using our `tinybuild` infrastructure

lucyleeow:
Sorry for the delay. From some digging, it seems that the sphinx gallery files are copied over AFTER paths listed in the user's project `conf.py` `html_static_path`.

We copy the sphinx gallery static path to `html_static_path` here:

https://github.com/sphinx-gallery/sphinx-gallery/blob/b176e22ff754ccf9407cb6e7347bf3b7eef72b2e/sphinx_gallery/gen_gallery.py#L127

Before this is run line `app.config.html_static_path` contains the paths listed in the `html_static_path` of the user's project `conf.py` file - i.e. the the user's `html_static_path` paths are added first.

I think `app.config.html_static_path` is updated with the user's `config.py` settings when Sphinx [`_init_env`](https://github.com/sphinx-doc/sphinx/blob/8fd4373d3aec07b8d9a4fb159b6236f4dc715b21/sphinx/application.py#L300) is run, which is before we copy `glr_path_static`.

Sphinx copies the static files [here](https://github.com/sphinx-doc/sphinx/blob/336605b8e4b14c5da9f4d872fb730dc6894edb77/sphinx/builders/html/__init__.py#L798), looping through the `self.config.html_static_path` list (thus files at the end of the list will over write files at the start of the list):

```Python
    def copy_html_static_files(self, context: Dict) -> None:
        def onerror(filename: str, error: Exception) -> None:
            logger.warning(__('Failed to copy a file in html_static_file: %s: %r'),
                           filename, error)

        excluded = Matcher(self.config.exclude_patterns + ["**/.*"])
        for entry in self.config.html_static_path:
            copy_asset(path.join(self.confdir, entry),
                       path.join(self.outdir, '_static'),
                       excluded, context=context, renderer=self.templates, onerror=onerror)
```

What the code below does is append `_static` to `app.config.html_static_path` at the end, i.e., the order of static files being added to the `app.config.html_static_path` list is as follows:

1. files included in the user's config.py `html_static_path`
2. SG adds `glr_path_static`
3. files added by the code below

```Python
def setup(app):
    app.connect('builder-inited', lambda app: app.config.html_static_path.append('_static'))
```

Note that the [`add_css_file`](https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_css_file) part is not required, and from the documentation, I think it is adding files to the default HTML template, and is not related to copying over static html file.

Further, the code below:

```Python
html_static_path = ['_static']
def setup(app):
    app.connect('builder-inited', lambda app: app.config.html_static_path.append('_static'))
```

will mean that `app.config.html_static_path` will have `_static` appear twice (i.e. `['_static', <sphx_glr_static>, '_static']`).

I think the 'trick' in the 'hack' is the 'builder-inited' part because while the `app.connect` seems to run early on in the build (before [`_init_env`](https://github.com/sphinx-doc/sphinx/blob/8fd4373d3aec07b8d9a4fb159b6236f4dc715b21/sphinx/application.py#L300) even), the appending to be done later in Sphinx [`Builder.read()`](https://github.com/sphinx-doc/sphinx/blob/336605b8e4b14c5da9f4d872fb730dc6894edb77/sphinx/builders/__init__.py#L365)

So it seems that the code above is a bit hacky (?) and maybe we are at fault because there may be a way to append SG's static path earlier - Sphinx [`html_static_path` documentation](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_static_path) says:

>  They (the user's list of static paths) are copied to the output’s _static directory after the theme’s static files, so a file named default.css will overwrite the theme’s default.css.

I guess we have 2 options:
* investigate how to copy (no idea how difficult this may be, Sphinx is magic)
* recommend that people not use gallery .css file names when adding their own .css and/or document the hack here

WDYT @larsoner ?

larsoner:
For now if possible let's just tell people not to use our names. Another low effort solution is to make our names more unique for example by adding _sg at the end of the filename before the extension

lucyleeow:
Thanks @larsoner 

Will make those changes in a new PR, probably this weekend.

In the mean time I think I will close this PR, it's nice that a solution is documented and searchable, but I think we won't include it in our documentation. (Hope this is okay with everyone, let me know if not)

Also I had a quick look and it seems like themes may be added by the config `add_html_theme`, so figuring out how to add the SG static files in earlier is probably harder than just 'copy how themes do it'.

lucyleeow:
Also thanks for your work figuring this out @yjmantilla !

PR #844, opened 2021-07-02
BUG URL quote branch names and filepaths in Binder URLs

sdhiscocks:
Fixes issue with special characters in Binder URLs for branch names and file paths. No characters that need quoting are allowed for GitHub _orgs_ and _repos_

PR #848, opened 2021-07-09
MAINT: Fix mayavi build hang

lucyleeow:
Try pinning vtk version <=9.0.1 as suggested in: https://github.com/sphinx-gallery/sphinx-gallery/pull/844#issuecomment-874800615

More details: https://github.com/enthought/mayavi/pull/1050


lucyleeow:
Not sure if we should pin vtk for fewer environments (e.g., only `conda_python38_sphinxDev`). I will also raise and issue to revise this pending https://github.com/enthought/mayavi/pull/1050.

larsoner:
Good enough! I don't think we have to be too fine-grained here

PR #852, opened 2021-08-06
Add unique output block references for literalinclude

### LINKED ISSUE #851: Referencing output from cells
I can reference code from a notebook using
```
.. literalinclude:: ../auto_examples/plot_myscript.py
    :start-after: [some-tag]
    :end-before: [another-tag]
```
and the resulting images using
```
.. figure:: ../auto_examples/images/sphx_glr_plot_myscript_001.png
    :target: ../auto_examples/plot_myscript.html
```
However, it doesn't seem to be possible to reference output from code that's part of a sphinx-gallery. This would be very useful.

I did find the following in the generated `.rst` file:
```
.. rst-class:: sphx-glr-script-out

 Out:

 .. code-block:: none

    <the output>
```
Perhaps if we had a dedicated output file (which we do for images) it would be possible? Thanks for considering!

romanlutz:
It's sometimes useful to reference code block output in addition to produced images (using `figure`) or code lines (using `literalinclude`), so this change adds comments to the produced RST file before and after the code output as

```
sphx-glr-output-block-001-start
<output>
sphx-glr-output-block-001-end
```
which can then be referenced in other RST files as
```
Out:

.. literalinclude:: ../auto_examples/plot_myscript.rst
    :start-after: .. sphx-glr-output-block-001-start
    :end-before: .. sphx-glr-output-block-001-end
    :lines: 2-
```

The `:lines: 2-` part is necessary to avoid including `.. code-block:`. Adding `Out:` is also optional, but otherwise it doesn't necessarily distinguish output rendering from normal code blocks.

I'm looking for feedback on this before proceeding with potentially adding tests and other required parts for merging this PR.

closes #851

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -142,9 +142,12 @@ CODE_OUTPUT = """.. rst-class:: sphx-glr-script-out
 
  Out:
 
+.. sphx-glr-output-block-{1:03}-start
  .. code-block:: none
 
-{0}\n"""
+{0}
+
+.. sphx-glr-output-block-{1:03}-end\n"""
 
 TIMING_CONTENT = """
 .. rst-class:: sphx-glr-timing
@@ -617,7 +620,7 @@ def _get_last_repr(gallery_conf, ___):
 
 
 def _get_code_output(is_last_expr, example_globals, gallery_conf, logging_tee,
-                     images_rst):
+                     images_rst, block_index):
     """Obtain standard output and html output in rST."""
     last_repr = None
     repr_meth = None
@@ -637,7 +640,8 @@ def _get_code_output(is_last_expr, example_globals, gallery_conf, logging_tee,
     if repr_meth in ['__repr__', '__str__'] and last_repr:
         captured_std = u"{0}\n{1}".format(captured_std, last_repr)
     if captured_std and not captured_std.isspace():
-        captured_std = CODE_OUTPUT.format(indent(captured_std, u' ' * 4))
+        captured_std = CODE_OUTPUT.format(indent(captured_std, u' ' * 4),
+                                          block_index)
     else:
         captured_std = ''
     # give html output its own header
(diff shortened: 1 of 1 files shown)


larsoner:
This seems reasonable and useful to me. @lucyleeow can you look to see if you agree?

Eventually we'd want tests (as mentioned above) and ideally it documented and shown somewhere in the config/use docs

lucyleeow:
Will take a look this weekend!

lucyleeow:
Agree seems like a easy to maintain, useful addition! Tests and documentation would be good.

(The CI's are too old to look at what's erroring but locally it seems to mostly needing to add the extra positional argument to tests where we use `execute_code_block`)

lucyleeow:
@romanlutz I'm happy to add the tests and docs for this if you would be happy with this?

romanlutz:
@lucyleeow definitely, please go ahead! I won't get to it anytime soon I'm afraid. Thank you!

larsoner:
That sounds like a clean solution to me!

lucyleeow:
I've realised this would not work for html output (just 'text' output). html always has it's own header and is added at the end if captured output has both text and html, e.g.:

```
.. rst-class:: sphx-glr-script-out

 Out:

 .. code-block:: none
    :name: sphx-glr-test_gen_rst.py-output-001

    print statement


.. sphx-glr-test_gen_rst.py-output-001-end

.. raw:: html

    <div class="output_subarea output_html rendered_html output_result">
    <div> This is the _repr_html_ div </div>
    </div>
    <br />
    <br />
```

We could ignore html output or use the comments around all output?

romanlutz:
> We could ignore html output or add the comments around all output?

How would we be able to add comments around all the output? If it's somehow possible (without too much extra work) that would be great, otherwise perhaps even without html is fine (?)

lucyleeow:
I think I could manage it and I agree, it's probably best to include html even though it's slightly uglier.

Don't have the bandwidth atm but will work on this as soon as I do!

PR #858, opened 2021-09-15
Add info on enabling animation support to example

dstansby:
When looking at how to enable animations in sphinx gallery, this example was the first thing I came upon. I thought it would be helpful to indicate how to enable support here in the example, so users don't have to go looking in the Configuration page.

### DIFF
--- a/examples/plot_8_animations.py
+++ b/examples/plot_8_animations.py
@@ -5,7 +5,8 @@ Matplotlib animation support
 Show a Matplotlib animation, which should end up nicely embedded below.
 
 In order to enable support for animations ``'matplotlib_animations'``
-must be set to ``True`` in the sphinx gallery configuration.
+must be set to ``True`` in the sphinx gallery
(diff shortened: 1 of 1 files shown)


examples/plot_8_animations.py
```diff
@@ -3,6 +3,9 @@
 ============================
 
 Show a Matplotlib animation, which should end up nicely embedded below.
+
+In order to enable support for animations ``'matplotlib_animations'``
+must be set to ``True`` in the sphinx gallery configuration.
```
larsoner, line 8: Let's add a nice link, too!
```suggestion
must be set to ``True`` in the sphinx gallery
:ref:`configuration <image_scrapers>`.
```

lucyleeow (approved):
Python nightly/our test_rebuild is a bit tempermental, let's see if it fixes itself.

lucyleeow:
Thanks @dstansby @larsoner

PR #863, opened 2021-10-07
Use bools for 'plot_gallery' in sphinx_gallery_conf

timhoffm:
The parser accepts both bools and strings. Strings occur when
passing the parameter in as a command line swith to sphinx-build.
However, in a python dict, we should still use the bool directly.
Having a boolean string is rather confusing.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1065,7 +1065,7 @@ a default::
 
     sphinx_gallery_conf = {
         ...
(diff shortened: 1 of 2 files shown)


timhoffm:
> Does 'command line switch' mean the -D flag of sphinx-build?

Yes.

larsoner:
Thanks @timhoffm !

PR #870, opened 2021-10-14
Enable `reset_modules` to run either before or after examples, or both

### LINKED ISSUE #866: Feature Request: `reset_modules` to be applied after each or all examples
My understanding is that the functions supplied to `reset_modules` will run _before_ each example.  But if the final example modifies a module, then the modified modules will leak out into other parts of the documentation building process.  This is particularly troublesome when regenerating a single failed example that modifies a module, which I've run into when trying to implement `reset_modules` to address https://github.com/pyvista/pyvista/issues/1711.  Right now, I need to manually force the build process to process one extra example to successful reset the module, which works, but is cumbersome.

It would be helpful to also provide a `reset_modules_post` which would work the same way, but runs _after_ each example or after all examples are run.  This would guard against this edge case.  

A counterargument to this request would be that we should reset the module in each build step, e.g. docstring examples.  But, I think it would be beneficial to have consistency in the end state after `sphinx-gallery` is completed that does not depend on the order of running the examples.

MatthewFlamm:
# Summary

As noted in #866, `reset_modules` functions are currently only called _before_ examples are run.  In the case that the last example modifies an import, `reset_modules` won't run subsequent to this example.  The modified module will leak out into other parts of the build process.

This PR adds another configuration option `reset_modules_order` that can be `after`, `before`, or `both`, which specifies when the callables in `reset_modules` are called.  The default is `before` which retains the current behavior.

This PR closes #866.

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -405,7 +405,7 @@ is required to be named ``when``::
             mpl.rcParams['lines.linestyle'] = '-'
 
 The value passed into ``when`` can be ``before`` or ``after``.
-If ``reset_modules_order`` in the :ref:`configuration <reset_modules_order>`is set to ``before`` or
+If ``reset_modules_order`` in the :ref:`configuration <reset_modules_order>` is set to ``before`` or
 ``after``, ``when`` will always be the same value.  This function signature
 is only useful when used in conjuction with ``reset_modules_order`` set to ``both``.
 
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -405,7 +405,7 @@ is required to be named ``when``::
             mpl.rcParams['lines.linestyle'] = '-'
 
 The value passed into ``when`` can be ``before`` or ``after``.
-If ``reset_modules_order`` in the configuration is set to ``before`` or 
+If ``reset_modules_order`` in the :ref:`configuration <reset_modules_order>`is set to ``before`` or
 ``after``, ``when`` will always be the same value.  This function signature
 is only useful when used in conjuction with ``reset_modules_order`` set to ``both``.
 
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -401,7 +401,7 @@ is required to be named ``when``::
 
         import matplotlib as mpl
         mpl.rcParams['lines.linewidth'] = 2
-        if when == 'after' and fname=='dashed_lines'
+        if when == 'after' and fname=='dashed_lines':
             mpl.rcParams['lines.linestyle'] = '-'
 
 The value passed into ``when`` can be ``before`` or ``after``.
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -356,7 +356,7 @@ it with Sphinx-Gallery. You can:
 Resetting before each example
 =============================
 
-Sphinx-Gallery supports 'resetting' function(s) that are run before and/or after 
+Sphinx-Gallery supports 'resetting' function(s) that are run before and/or after
 each example script is executed. This is used natively in Sphinx-Gallery to 'reset'
 the behavior of the visualization packages ``matplotlib`` and ``seaborn``
 (:ref:`reset_modules`). However, this functionality could be used to reset other
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -356,12 +356,12 @@ it with Sphinx-Gallery. You can:
 Resetting before each example
 =============================
 
-Sphinx-Gallery supports 'resetting' via a prolog that is run before each
-example script is executed. This is used natively in Sphinx-Gallery to 'reset'
+Sphinx-Gallery supports 'resetting' function(s) that are run before and/or after 
+each example script is executed. This is used natively in Sphinx-Gallery to 'reset'
 the behavior of the visualization packages ``matplotlib`` and ``seaborn``
 (:ref:`reset_modules`). However, this functionality could be used to reset other
 libraries, modify the resetting behavior for a natively-reset library or run
-an arbitrary custom function at the start of each script.
+an arbitrary custom function at the start and/or end of each script.
 
 This is done by adding a custom function to the resetting tuple defined in
 ``conf.py``. The function should take two variables: a dictionary called
@@ -392,6 +392,22 @@ function (see :ref:`reset_modules`).
           standard behavior that users will experience when manually running
           examples themselves is discouraged due to the inconsistency
           that results between the rendered examples and local outputs.
+
+If the custom function needs to be aware of whether it is being run before or after
+an example, a function signature with three parameters can be used, where the third parameter
+is required to be named ``when``::
+
+    def reset_mpl(gallery_conf, fname, when):
+
+        import matplotlib as mpl
+        mpl.rcParams['lines.linewidth'] = 2
+        if when == 'after' and fname=='dashed_lines'
+            mpl.rcParams['lines.linestyle'] = '-'
+
+The value passed into ``when`` can be ``before`` or ``after``.
+If ``reset_modules_order`` in the configuration is set to ``before`` or 
+``after``, ``when`` will always be the same value.  This function signature
+is only useful when used in conjuction with ``reset_modules_order`` set to ``both``.
 
 Altering Sphinx-Gallery CSS
 ===========================
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -36,6 +36,7 @@ file:
 - ``compress_images`` (:ref:`compress_images`)
 - ``image_srcset`` (:ref:`image_srcset`)
 - ``reset_modules`` (:ref:`reset_modules`)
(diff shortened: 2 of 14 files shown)


larsoner:
> This is a bit of magic for the user, but I like it. 

I also think of it as a way for us to add new information to callbacks in a backward-compatible way. I don't think the implementation is too painful and I don't anticipate any terrible maintenance from it, so why not :)

MatthewFlamm:
Currently, we aren't restricting users to have a function signature that has exactly the same parameters defined, only that there are 2 parameters.  I've taken the same approach and have inspected for function with 3 parameters, otherwise fallback to the old behavior.

Looks like python 3.8 passed by 3.6 all failed due to some difference in inspect like you mentioned.  There could also be a difference in `unittest.mock`, but I haven't checked yet.  The only possible relevant difference I can see in the documentation between 3.6 and 3.7 is the below, but it shouldn't be applicable here.  https://docs.python.org/3/library/inspect.html#inspect.Signature.parameters

> Changed in version 3.7: Python only explicitly guaranteed that it preserved the declaration order of keyword-only parameters as of version 3.7, although in practice this order had always been preserved in Python 3.

larsoner:
I would prefer to check for kwarg name rather than number, it's more explicit that people have when=None in their signatures than some_other_name and rely on that in the third position

I recall needing a shim until 3.7 for inspection, let's just bump...

MatthewFlamm:
Are you suggesting enforcing that `when` is a parameter?  Or that `when` additionally has a default (note that we always provide a value, so I'm unclear on why we would enforce a signature with a default)?

In `def fun(a, b, when=None)`, all parameters are `POSITIONAL_OR_KEYWORD`, but `when` has a default.

 I would more strongly not recommend that we enforce `when` to be a keyword only argument as the function signature becomes more complex, e.g. `def fun(a, b, *, when=None):`.

I'm leaning even more towards the current PR state.  It seems overly restrictive to require users to use the same variable names with no clear benefit to me.  I think it will be surprising to users that they cannot use any name as this isn't common practice.  And that we would enforce the name for only 1 out of 3 parameters will lead to even more confusion.

MatthewFlamm:
Sorry for all the pushes in a row, I know it can be a lot of notifications potentially.  I had to figure out how to run the local style check to match the CI, and then all the subtleties with the CI requirements.  I think I have it figured out now after poring over the different CI scripts, so it should be less going forward.

I could not find any examples of projects using `inspect` like this to require function signatures to have certain names.  The most prevalent examples I could find are for documentation purposes, for example to modify or style parameters that are shown in the documentation.   But, I'm OK in implementing it if it is this project's preference.  Either way, documentation is very important here, which still needs to be worked on.

larsoner:
> I could not find any examples of projects using inspect like this to require function signatures to have certain names

We use a `_get_args` [in MNE-Python](https://github.com/mne-tools/mne-python/blob/main/mne/fixes.py#L77-L88) all the time to look to see for example if `matplotlib` or [`pyvista` has a specific argument](https://github.com/mne-tools/mne-python/blob/75a13167e3d89fae36aeda7da2f34cd00cd3e995/mne/viz/backends/_pyvista.py#L585) as it's sometimes easier than relying on a version number. It has worked well over the years.

>  Either way, documentation is very important here, which still needs to be worked on.

I think we just need to say that in the callback, feel free to add a `when` after the first two positional arguments (should work as any of `a, b, when`, `a, b, when=None`, or `a, b, *, when`) and then the callback will get the information. I don't think it's too complicated...?

MatthewFlamm:
I think these examples introduce more flexibility to the users, which I would be in support of.  As I understand, one example checks number of arguments and then adjusts the code to work with varying numbers.  The other checks for a specific argument for specific functionality.  Here we are only limiting flexibility to users by requiring a name. 

One case where this naming requirement would add flexibility is if we allowed users to use any combination, order, number of parameter inputs like:

```
def mpl_reset(fname):
    ...
```

vs.

```
def mpl_reset(when, gallery_conf):
    ...
```

For these examples, it is required to know which parameter name is used and in what order. In the current PR state, it is determined by number of parameters only, we just restrict the third parameters name.

But, I agree that it isn't complex, and if it is wanted by this project, I am OK implementing.  In fact, it is already implemented.  `when` is now the required third parameter.  I am working on the documentation.

doc/advanced.rst
```diff
@@ -393,6 +393,22 @@ function (see :ref:`reset_modules`).
...
+
+    def reset_mpl(gallery_conf, fname, when):
+
+        import matplotlib as mpl
+        mpl.rcParams['lines.linewidth'] = 2
+        if when == 'after' and fname=='dashed_lines'
+            mpl.rcParams['lines.linestyle'] = '-'
+
+The value passed into ``when`` can be ``before`` or ``after``.
+If ``reset_modules_order`` in the configuration is set to ``before`` or
```
lucyleeow, line 408:
```suggestion
If ``reset_modules_order`` in the :ref:`configuration <reset_modules_order>`is set to ``before`` or
```
nitpick

doc/configuration.rst
```diff
@@ -1317,6 +1318,26 @@ this tuple in order to define resetting behavior for other visualization librari
 
 To do so, follow the instructions in :ref:`custom_reset`.
 
+.. _reset_modules_order:
+
+Order of resetting modules
+==========================
+
+By default, the Sphinx-Gallery will reset modules before each example is run.
```
lucyleeow, line 1326:
```suggestion
By default, Sphinx-Gallery will reset modules before each example is run.
```
nitpick

doc/configuration.rst
```diff
@@ -1317,6 +1318,26 @@ this tuple in order to define resetting behavior for other visualization librari
...
+The choices for ``reset_modules_order`` are ``before`` (default), ``after``, and
+``both``. If the last example run in Sphinx-Gallery modifies a module, it is
+recommended to use ``after`` or ``both`` to avoid leaking out a modified module to
+other parts of the Sphinx build process.  For example, to use both use the following
+configuration::
```
lucyleeow, line 1331:
```suggestion
other parts of the Sphinx build process.  For example, set ``reset_modules_order``
to ``both`` in the configuration::
```
nitpick

sphinx_gallery/scrapers.py
```diff
@@ -577,6 +578,20 @@ def clean_modules(gallery_conf, fname):
...
     """
     for reset_module in gallery_conf['reset_modules']:
-        reset_module(gallery_conf, fname)
+
+        sig = inspect.signature(reset_module)
+        if len(sig.parameters) == 3:
+            third_param = list(sig.parameters.keys())[2]
+            if third_param != 'when':
+                raise ValueError("3rd parameter in function signature must "
+                                 "be 'when', got %s" % third_param)
```
lucyleeow, line 594:
```suggestion
                raise ValueError(f"3rd parameter in {reset_module.__name__} function"
                                          signature must be 'when', got {third_param}")
```
nitpick

lucyleeow (commented):
Sorry for the delay, this just fell off my radar, not sure what happened!

Some nitpicks, happy to merge as is as well, thanks!

doc/advanced.rst
```diff
@@ -393,6 +393,22 @@ function (see :ref:`reset_modules`).
...
 
+If the custom function needs to be aware of whether it is being run before or after
+an example, a function signature with three parameters can be used, where the third parameter
+is required to be named ``when``::
+
+    def reset_mpl(gallery_conf, fname, when):
+
+        import matplotlib as mpl
+        mpl.rcParams['lines.linewidth'] = 2
+        if when == 'after' and fname=='dashed_lines'
```
MatthewFlamm, line 404:
```suggestion
        if when == 'after' and fname=='dashed_lines':
```

MatthewFlamm:
Most of the additional commits related to cleaning up the change in error message.  One was related to fixing the example using 3 parameters in the documentation.

I was able to get all of the CI green except for circleci, which I do not have access to.

doc/advanced.rst
```diff
@@ -393,6 +393,22 @@ function (see :ref:`reset_modules`).
...
+
+    def reset_mpl(gallery_conf, fname, when):
+
+        import matplotlib as mpl
+        mpl.rcParams['lines.linewidth'] = 2
+        if when == 'after' and fname=='dashed_lines':
+            mpl.rcParams['lines.linestyle'] = '-'
+
+The value passed into ``when`` can be ``before`` or ``after``.
+If ``reset_modules_order`` in the :ref:`configuration <reset_modules_order>`is set to ``before`` or
```
larsoner, line 408: The error is
```
/home/circleci/project/doc/advanced.rst:407: WARNING: undefined label: configuration <reset_modules_order>`is set to ``before`
```
So probably you need a space
```suggestion
If ``reset_modules_order`` in the :ref:`configuration <reset_modules_order>` is set to ``before`` or
```

larsoner:
@lucyleeow feel free to merge if you're happy!

lucyleeow:
Thank you @MatthewFlamm and @larsoner !

PR #871, opened 2021-10-16
DOC Fix typo in dummy images doc

lucyleeow:
Fix typo in configuration.rst


lucyleeow:
I feel like test_rebuild has a tendency to fail on the weekends. Merging as only doc changed and it renders fine.

Issue #707, opened 2020-06-11
Open Graph protocol and link sharing


hayesall:
It would be cool if there was a configuration option to support Open Graph protocol or Twitter cards when sharing examples:

- [Open Graph protocol](https://ogp.me/): "*enables any web page to become a rich object in a social graph.*"
- [Twitter Cards](https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/abouts-cards): "*... you can attach rich photos, videos and media experiences to Tweets.*"

I think this could be accomplished by adding relevant `meta` tags to the head when generating pages.

### Current version

Using `scikit-learn` as an example, sharing the [`linear_model`](https://scikit-learn.org/stable/modules/linear_model.html) and [`plot_comparison`](https://scikit-learn.org/stable/auto_examples/classification/plot_classifier_comparison.html) pages on Slack results in the following:

![image](https://user-images.githubusercontent.com/11916674/84408763-891fbc00-abda-11ea-982f-a07fc71ab25d.png)

But the [generated image](https://scikit-learn.org/stable/_images/sphx_glr_plot_classifier_comparison_001.png), description, and a few other things could also be part of this.

### Possible version

Now using [this web page](https://hayesall.com/blog/latex-automata/) as an example, sharing a link results in:

![image](https://user-images.githubusercontent.com/11916674/84409445-6d68e580-abdb-11ea-9cb4-52b2b00586c5.png)

... because of the following `meta` tags:

```html
<!-- Open Graph -->
<meta property="og:locale" content="en">
<meta property="og:type" content="article">
<meta property="og:title" content="LaTeX Finite Automata and State Diagrams with Tikz">
<meta property="og:description" content="This short guide collects some examples of drawing finite state machines using the Tikz LaTeX library.">
<meta property="og:url" content="https://hayesall.com/blog/latex-automata/">
```

### Other notes

- `OPG` support might be incorporated elsewhere (e.g. [`sphinx-contrib/ogp`](https://github.com/sphinx-contrib/ogp)).
- Here's a liquid template for generating some of these: https://github.com/hayesall/hayesall.github.io/blob/fee47f82f5cb489f14cbe8018cf812733670138c/_includes/open-graph.html

larsoner:
@hayesall sounds reasonable to me, seems like something we could add in a `.. raw:: html` in our generated RST files without too much work. Would you be up for trying to implement it?

How do you envision the API? It seems like:
```
[...]

PR #872, opened 2021-10-16
DOC Add sphinxext opengraph for our doc

lucyleeow:
Related to #707 

Also adds a 'related_projects' page in documentation.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -56,6 +56,7 @@ stages:
     - script: pytest --tb=short sphinx_gallery
       displayName: pytest
     - script: |
+        pip install sphinxext-opengraph
         cd doc
         sphinx-build -D plot_gallery=0 -b html -d _build\doctrees . _build\html
       displayName: make html-noplot


larsoner:
To me the important change is to make it so that individual rendered examples have OpenGraph information so that they embed nicely. Do you want to try that here, too?

lucyleeow:
> To me the important change is to make it so that individual rendered examples have OpenGraph information so that they embed nicely. 

Did not realise more work was required for that...I guess I will look into that.

larsoner:
Disclaimer: I don't know what is required. Ideally though I think the opengraph for example A would show the scaled or (probably better) original thumbnail image for example A, example B would show example B etc. So the image+link would be example specific and need to know the image URL to show on an example by example basis

larsoner:
Just saw your edits... Yes I mean that when MNE or scikit learn deploys docs a someone embeds a link to example A, the preview that is down should be specific to that example. When they link to example B, the image should be specific to that example, and so on

lucyleeow:
Ah okay I understand now - I was wondering about the image config of opengraph. Yes I agree their implementation could be improved but reading through the thread in #707 it sounded like some work would be required and I am not sure I have the time now. Maybe I should close this and re-think.

lucyleeow:
Of note, I have just seen the `ogp_use_first_image` config:
```
This is not required. Set to True to use each page's first image, if available. If set to True but no image is found, Sphinx will use ogp_image instead.
```

This is not necessarily the thumbnail though.

PR #873, opened 2021-10-16
Add Neuraxle to "Who uses Sphinx-Gallery"

guillaume-chevalier:


lucyleeow:
Thanks! Would you be able to add it to the readme as well: https://github.com/sphinx-gallery/sphinx-gallery/blob/master/README.rst ?

(we should probably automate this so we don't have add it twice...)

Issue #846, opened 2021-07-06
Using only the backlinks feature in existing RST prose


felix-hilden:
Hi!

Firstly, thanks for the seemingly great package 😄 I see lots of successful libraries are using it to a great effect. And I intend to be your next user, especially once I write a bunch of python examples. For now though, I'd like to only use the *awesome* backlink feature to enhance my examples that are written inside existing RST prose.

Essentially, what I'm after is being able to write ordinary Sphinx RST like

```rst
Document
========
This is a document with the following example:

.. code:: python

    import lib

    lib.foo()
    b = lib.Bar()
    b.meth()
```

and have the links be generated regardless of the file's location. It's not obvious whether or not this is even possible since the documentation is so focused on the example gallery (which is hardly surprising 😅), but I suspect it is some combination of `backreferences_dir`, `doc_module` and `reference_url` as laid out in the few tutorials starting [here](https://sphinx-gallery.github.io/stable/configuration.html#add-intersphinx-links-to-your-examples). How could this be achieved?

Apologies for essentially asking the library to do something that it wasn't necessarily supposed to do. So feel free to shut this down if you already know it won't be possible. But I'm over the moon that this is even a thing! Keep making documentation great ❤️

---

Eventually having mini-galleries or just plain lists of example locations in reference documentation would be great too, but I think that's too much for one user support question 😅 I might still be misunderstanding the features altogether.

lucyleeow:
This is slightly related to #330 (in particular @GaelVaroquaux's case where we convert .rst to .py so sphinx-gallery can convert back to rst), do you think this would solve your problem? With regards to the .rst file being anywhere - that's possibly a bit tricker, we could enable sphinx-gallery to look through all the `docs/` source dir and convert all rst to .py but I am not sure that is what you want? I am also hesitant about this I don't want to accidently mess up the `docs/` dir structure. @larsoner may have an opinion on this, I don't use Sphinx enough.

(also I see that I said I will take a look at implementing this 😓 - whoops, adding to my todo list)

felix-hilden:
Thanks for the fast reply!

[...]

PR #874, opened 2021-10-17
DOC Add reference to sphinx-codeautolink

felix-hilden:
From #846: I created a page for related projects (because the PR #872 was closed), and added `sphinx-codeautolink` there. I thought I'd change the message to be in line with the one in #872. Cheers!


lucyleeow:
Ah I get what you mean, maybe in this page where we talk about how to structure the .py files: https://sphinx-gallery.github.io/stable/syntax.html

And maybe add a reference here: https://sphinx-gallery.github.io/stable/configuration.html#auto-documenting-your-api-with-links-to-examples ?

PR #877, opened 2021-10-28
ENH: Parallel gallery generation

### LINKED ISSUE #25: Multiprocesor support?

jschueller:
Closes #25


jschueller:
maybe the sphinx folks (@samdoran or @tk0miya) could give hints, or sklearn (@Titan-C @GaelVaroquaux) would be interested in speeding up their builds too ?

tk0miya:
It seems this handler is invoked on the bootstrap process of Sphinx (on builder-inited event). So there is no special support from Sphinx framework.

jschueller:
indeed, we're looking at parallelizing the gallery jobs at the extension level

larsoner:
@jschueller there has been some renewed interest in this so I'm taking a stab at actually making it work. I think it's close but I need to fix a test and maybe a couple of examples, we'll see!

jschueller:
> @jschueller there has been some renewed interest in this so I'm taking a stab at actually making it work. I think it's close but I need to fix a test and maybe a couple of examples, we'll see!

this is great!

jschueller:
it passes! can I test it already ?

jschueller (commented):
It seems to work! Thanks a lot
Do you plan a new release with this ?

larsoner:
I still need to figure out how to test this. And also decide if `joblib` is the right thing to use for parallelization. I suspect if I just used `multiprocessing` with `spawn` then we maybe could keep memory measurements working, which would be nice. I have to think about it...

But yes I think once this is in and maybe @jschueller you and @lagru run some preliminary tests we could cut a release. I'll probably write in the docs that it's new/experimental, though, since I do expect there to be issues just because parallelization seems to always create them!

jschueller:
I already tested it on a real project (openturns) and it works

larsoner:
> I already tested it on a real project (openturns) and it just works

I mean also in terms of unit tests :)

I tried enabling it in `tinybuild` but it failed, need to look at why.

larsoner:
Yikes what a nightmare to track down -- turns out if you `pickle.dump` identical objects the outputs are not byte-by-byte identical 😠 🤦  And when you build with multiple spawned processes it becomes much more apparent. This could actually explain some of our spurious build failures we've seen in random jobs over the years now that I think about it!

This PR thus switches us away from `_codeobj.pickle` to `.codeobj.json`, which *is* stable over rebuilds (identical dict/str/list/bool combinations that we use will result in the same output json). As a nice side effect the files become nicely human readable, which is nice, e.g., from our doc build `plot_0_sin.codeobj.json` is ~100 lines of:
```json
{
 "plt.ylabel": [
  {
   "name": "function",
   "module": "builtins",
   "module_short": "builtins",
   "is_class": false,
   "is_explicit": false
  },
  {
   "name": "ylabel",
   "module": "matplotlib.pyplot",
   "module_short": "matplotlib.pyplot",
   "is_class": false,
   "is_explicit": false
  }
 ],
 "plt.plot": [
  {
   "name": "function",
   "module": "builtins",
   "module_short": "builtins",
   "is_class": false,
   "is_explicit": false
  },
...
```
I don't think the files produced are too big but if we decide they are we could minify them.

jschueller:
hi, do you know how to fix the last failure ?

larsoner:
It's not obvious to me so far, I need to figure out why an example is/isn't being rebuilt

jschueller:
seems close, just some tests failing because its sensitive to linux LF windows / CRLF comparisons

larsoner:
Yikes that parallel bug was hard to find. In one test, we move an example from one dir to another, then rerun the doc build. That examples has the line `import local_module`. This works in non-parallel mode only because of Python import magic (it's already in `sys.modules` so it doesn't actually try to re-import it). In parallel mode, it *might or might not* be in `sys.modules` depending on which Python interpreter was used to run the given example on the previous parallel run. That's a nasty heisenbug! But now that it's isolated, it's easily fixed by (correctly) copying the `local_module.py` over when moving the example file. I'm going to push an empty commit or two just so I can sleep well at night but I expect them to pass.

In any case, I think things should be working now @lucyleeow, ready for review/merge from my end. I also:

1. modified our requirements to be `Sphinx >= 5` because two years ago 5.0 came out (SPEC0).
2. made some progress toward preferring `pathlib.Path` operations over `os.path`
3. moving away from `codecs.open` to builtin `open`, as the docs [say](https://docs.python.org/3/library/codecs.html):

   > the builtin [open()](https://docs.python.org/3/library/functions.html#open) and the associated [io](https://docs.python.org/3/library/io.html#module-io) module are the recommended approach for working with encoded text files

   and it's easy enough with some unified keyword arguments on writing.

lesteve:
Just curious, has this actually been tested on some documentation out there, for example scikit-learn, MNE-Python, matplotlib ... to get some idea about potential speed-ups?

It does seem like additional complexity and I am wondering how much things can go wrong if the examples use BLAS or OpenMP, so that people may need to know about setting `OPENBLAS_NUM_THREADS`, `MKL_NUM_THREADS`, `OMP_NUM_THREADS`, to avoid oversubscription ...

larsoner:
> Just curious, has this been tested on some documentation out there to get some idea about potential speed-ups? For example on scikit-learn, MNE-Python, matplotlib etc ...

- [A few comments up](https://github.com/sphinx-gallery/sphinx-gallery/pull/877#issuecomment-2150274625) @jschueller used it on openturns and reported that it worked. Not sure how the timings chaged, though.
- I just tried `matplotlib` doing `make clean && time make html` (after adding a `raise RuntimeError` at the end of `gen_gallery_rst` to limit the sphinx-build primarily to the sphinx-gallery bits) and got:
   ```
   real	7m9.734s
   user	6m59.549s
   ```
   I then added `'parallel': 4` to the config and reran and got a decent speedup:
   ```
   real	2m46.106s
   user	7m15.104s
   ```
- I then tried MNE-Python but got an Exception in `ExecutorManagerThread`. But MNE does all sorts of stuff in the doc build like 3D plotting and such so it'll require a bit more investigation why it happens, but don't think it has to hold up this PR.

This error did happen on matplotlib on Linux until I set `batch_size=1`, though. And even then on macOS I still sometimes got the error: 

<details>
<summary>ExecutorManagerThread error</summary>

```
Exception in thread ExecutorManagerThread:
  File "/Applications/MNE-Python/1.7.0_0/.mne-python/lib/python3.12/threading.py", line 1073, in _bootstrap_inner
    self.run()
  File "/Applications/MNE-Python/1.7.0_0/.mne-python/lib/python3.12/site-packages/joblib/externals/loky/process_executor.py", line 602, in run
    self.process_result_item(result_item)
  File "/Applications/MNE-Python/1.7.0_0/.mne-python/lib/python3.12/site-packages/joblib/externals/loky/process_executor.py", line 752, in process_result_item
    warnings.warn(
UserWarning: A worker stopped while some jobs were given to the executor. This can be caused by a too short worker timeout or by a memory leak.
```

</details>

So still something to investigate here perhaps. Not sure if should hold the PR up or not, though, since `matplotlib` seems to work on Linux now at least 🤔 .

> It does seem like additional complexity

At least at the SG end it didn't turn out to be too bad (thanks `joblib`!), other than the difficult tracking down of an existing bug that was just hidden by the magic in Python's import mechanisms.

> I am wondering how much things can go wrong if the examples use BLAS or OpenMP, so that people may need to know about setting OPENBLAS_NUM_THREADS, MKL_NUM_THREADS, OMP_NUM_THREADS to avoid oversubscription ...

Sure but this will be true of any library that adds parallelization. I could add a note to the warning to consider reading the joblib docs carefully to understand the ramifications and limitations perhaps?

lesteve:
Thanks for the details, I don't think this should hold up the PR either, I was mostly curious about how this would improve things.

I tested quickly on the scikit-learn doc. On my computer (12 cores but a mix of "performance" and "efficient" cores), the full doc takes ~25 minutes (~19 minutes taken by running the examples), with parallel=4 ~15 minutes for the full doc, not much improvement after this by using bigger `parallel`. With parallel=12, this seems to hang (probably need to setup `OPENBLAS_NUM_THREADS` and other environment variables).

For completeness, I did also notice some weird example error that only seemed to happen in parallel somehow and that would need some investigation (doesn't make too much sense right now).
```
  ../examples/feature_selection/plot_rfe_with_cross_validation.py failed leaving traceback:

      File "/home/lesteve/dev/scikit-learn/examples/feature_selection/plot_rfe_with_cross_validation.py", line 57, in <module>
        rfecv.fit(X, y)
      File "/home/lesteve/dev/scikit-learn/sklearn/base.py", line 1514, in wrapper
        return fit_method(estimator, *args, **kwargs)
      File "/home/lesteve/dev/scikit-learn/sklearn/feature_selection/_rfe.py", line 781, in fit
        scores = np.array(scores)
    ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (5,) + inhomogeneous part.
```

> This error did happen on matplotlib on Linux until I set batch_size=1, though. And even then on macOS I still sometimes got the error:

I don't have a good suggestion off the top of my head on the `ExecutorManagerThread` error. I guess turning off auto-batching (`batch_size=1`) seems probably a good idea in general since joblib is trying to batch short tasks to have a minimum task time of 0.2s https://github.com/joblib/joblib/blob/f70939cd06416676be6e2c038dbb93bb7c2e2556/joblib/_parallel_backends.py#L299. I would expect that in most projects, there are not that many examples that take less than 0.2s?

jschueller:
its quite beneficial for us on openturns, where we have 270 examples some of which run for longer than 0.2s :)
```
computation time summary:
    - examples/data_analysis/distribution_fitting/plot_estimate_gpd_rain.py:                                     29.72 sec   0.0 MB
    - examples/meta_modeling/fields_metamodels/plot_fieldfunction_metamodel.py:                                  23.01 sec   0.0 MB
    - examples/meta_modeling/kriging_metamodel/plot_kriging_multioutput_firesatellite.py:                        22.76 sec   0.0 MB
    - examples/calibration/bayesian_calibration/plot_gibbs.py:                                                   20.12 sec   0.0 MB
    - examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_cv.py:                                        19.42 sec   0.0 MB
    - examples/reliability_sensitivity/sensitivity_analysis/plot_sensitivity_wingweight.py:                      17.69 sec   0.0 MB
    - examples/meta_modeling/polynomial_chaos_metamodel/plot_functional_chaos.py:                                17.32 sec   0.0 MB
    - examples/meta_modeling/kriging_metamodel/plot_kriging_categorical.py:                                      16.11 sec   0.0 MB
    - examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_flooding.py:                  14.00 sec   0.0 MB
    - examples/data_analysis/distribution_fitting/plot_estimate_multivariate_distribution.py:                    13.98 sec   0.0 MB
    - examples/numerical_methods/general_methods/plot_ifs.py:                                                    12.32 sec   0.0 MB
    - examples/calibration/least_squares_and_gaussian_calibration/plot_calibration_chaboche.py:                  10.62 sec   0.0 MB
    - examples/reliability_sensitivity/design_of_experiments/plot_smolyak_quadrature.py:                          9.90 sec   0.0 MB
    - examples/meta_modeling/polynomial_chaos_metamodel/plot_chaos_beam_sensitivity_degree.py:                    9.08 sec   0.0 MB
    - examples/calibration/bayesian_calibration/plot_rwmh_python_distribution.py:                                 8.92 sec   0.0 MB
    - examples/reliability_sensitivity/reliability/plot_estimate_probability_form_oscillator.py:                  8.27 sec   0.0 MB
    - examples/reliability_sensitivity/reliability_processes/plot_field_fca_sobol.py:                             7.88 sec   0.0 MB
    - examples/probabilistic_modeling/stochastic_processes/plot_userdefined_covariance_model.py:                  6.62 sec   0.0 MB
    - examples/meta_modeling/general_purpose_metamodels/plot_distribution_linear_regression.py:                   6.34 sec   0.0 MB
...
```

larsoner:
Okay pushed a doc update -- @lucyleeow feel free to review and merge if happy when you get a chance!

larsoner:
Okay I realized that for MNE-Python we treat all warnings in doc builds/examples as errors, so I needed to prevent raising the `UserWarning` that `joblib` emits as an error. Then I could test a few cases:

# MNE-Python

## No examples (noplot)

This build takes ~8 min.

## All examples, parallel=1 (serial)

Four slowest examples took:
```
    - ../examples/datasets/opm_data.py:                                       57.88 sec   0.0 MB
    - ../examples/time_frequency/source_power_spectrum_opm.py:                55.67 sec   0.0 MB
    - ../examples/datasets/spm_faces_dataset.py:                              49.65 sec   0.0 MB
    - ../tutorials/io/60_ctf_bst_auditory.py:                                 45.00 sec   0.0 MB
```
the full doc build takes about an hour:
```
real	59m47.935s
user	54m54.940s
```
So ~52 minutes to build the examples.

## All examples, parallel=4

My machine has 4 physical cores so I tried this next and saw a bit of CPU oversubscription by eye and looking at the four slowest examples:
```
    - ../examples/datasets/opm_data.py:                                       68.57 sec   0.0 MB
    - ../examples/inverse/multi_dipole_model.py:                              63.33 sec   0.0 MB
    - ../examples/datasets/spm_faces_dataset.py:                              57.83 sec   0.0 MB
    - ../tutorials/preprocessing/40_artifact_correction_ica.py:               50.90 sec   0.0 MB
```
but the overall build time was cut in half (hooray!):
```
real	27m9.492s
user	59m48.404s
```
And the example part of that is ~19min. 

So our speedup for the example-running portion is ~52min down to ~19min! A factor of ~2.7 which is pretty good I think!

There is one fancy example that has a custom scraper in MNE that I don't think I could actually get to work properly in parallel based on what it does (creates HTML files, sets them to be copied later, etc.). I think we'll want something like a `# sphinx_gallery_non_parallel` comment or something that forces it to be run in non-parallel mode, but I can do that in a follow-up PR I think. I don't think it will be very complicated but don't want to add more to this PR!

lucyleeow:
Sorry, I've missed this, I'll take a quick look today.

sphinx_gallery/gen_rst.py
```diff
@@ -1237,21 +1267,25 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf, seen_backrefs=No
...
         The introduction of the example.
+    title : str
+        The example title.
     cost : tuple
         A tuple containing the ``(time_elapsed, memory_used)`` required to run the
         script.
+    backrefs : set
+        The backrefs seen in this example.
+    out_vars : dict
+        Variables used to run the script.
```
lucyleeow, line 1283: Could this detail dict keys?

sphinx_gallery/gen_rst.py
```diff
@@ -570,11 +578,36 @@ def generate_dir_rst(
...
+        gallery_conf["titles"][src_file] = title
+        if "formatted_exception" in out_vars:
+            assert "passing" not in out_vars
+            assert "stale" not in out_vars
+            gallery_conf["failing_examples"][src_file] = out_vars["formatted_exception"]
+        elif "passing" in out_vars:
+            assert "stale" not in out_vars
+            gallery_conf["passing_examples"].append(src_file)
+        elif "stale" in out_vars:  # non-executable files have none of these three
+            gallery_conf["stale_examples"].append(out_vars["stale"])
```
lucyleeow, line 610: This is much nicer all together. Is it worth putting this in a helper fun?

sphinx_gallery/gen_rst.py
```diff
@@ -570,11 +578,36 @@ def generate_dir_rst(
...
         src_file = os.path.normpath(os.path.join(src_dir, fname))
+        gallery_conf["titles"][src_file] = title
+        if "formatted_exception" in out_vars:
+            assert "passing" not in out_vars
+            assert "stale" not in out_vars
+            gallery_conf["failing_examples"][src_file] = out_vars["formatted_exception"]
+        elif "passing" in out_vars:
+            assert "stale" not in out_vars
+            gallery_conf["passing_examples"].append(src_file)
+        elif "stale" in out_vars:  # non-executable files have none of these three
```
lucyleeow, line 609: what are "these three"?

Would it be clearer to say that 'stale' examples are 'not re-executed'? or something similar?

sphinx_gallery/gen_rst.py
```diff
@@ -338,16 +341,16 @@ def extract_intro_and_title(filename, docstring):
...
-            ref_md5 = file_checksum.read()
+    if not os.path.exists(src_md5_file):
+        return False
 
-        return src_md5 == ref_md5
+    with open(src_md5_file) as file_cs:
+        ref_md5 = file_cs.read()
 
-    return False
+    return src_md5 == ref_md5
```
lucyleeow, line 353: This reads much nicer!

sphinx_gallery/gen_rst.py
```diff
@@ -737,23 +782,11 @@ def __call__(self):
 
 
 def _get_memory_base():
-    """Get the base amount of memory used by running a Python process."""
+    """Get the base amount of memory used by the current Python process."""
     # There might be a cleaner way to do this at some point
```
lucyleeow, line 786: Probably can do away with this comment now that this is a 2 liner function :sweat_smile:

sphinx_gallery/gen_rst.py
```diff
@@ -737,23 +782,11 @@ def __call__(self):
...
-        sleep, timeout = (0.5, 1)
-    proc = subprocess.Popen(
-        [sys.executable, "-c", f"import time, sys; time.sleep({sleep}); sys.exit(0)"],
-        close_fds=True,
-    )
-    memories = memory_usage(proc, interval=1e-3, timeout=timeout)
-    proc.communicate(timeout=timeout)
-    # On OSX sometimes the last entry can be None
-    memories = [mem for mem in memories if mem is not None] + [0.0]
-    memory_base = max(memories)
```
lucyleeow, line 756: Just for my info - why are we deleting? :grimacing:

lucyleeow (commented):
Thanks! I pushed a small commit to note version support drop in `CHANGES.rst`

I know you've noted more fixes to do, happy to look again later.

sphinx_gallery/gen_rst.py
```diff
@@ -541,15 +543,21 @@ def generate_dir_rst(
...
-    if not os.path.exists(target_dir):
-        os.makedirs(target_dir)
+    # Make all dirs ahead of time to avoid collisions in parallel processing
+    os.makedirs(target_dir, exist_ok=True)
+    image_dir = os.path.join(target_dir, "images")
+    os.makedirs(image_dir, exist_ok=True)
+    thumb_dir = os.path.join(image_dir, "thumb")
+    os.makedirs(thumb_dir, exist_ok=True)
+    if gallery_conf["jupyterlite"] is not None:
+        _add_jupyterlite_badge_logo(image_dir)
```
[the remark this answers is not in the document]
larsoner: These dir-making functions? I would say no because it's a fairly short block of code with no conditionals so I don't think much would be gained by moving it

sphinx_gallery/gen_rst.py
larsoner: I would say no for the same reasons as above

sphinx_gallery/gen_rst.py
larsoner: Changing to at the top of the three conditionals:
```
        # n.b. non-executable files have none of these three variables defined,
        # so the last conditional must be "elif" not just "else"
```

sphinx_gallery/gen_rst.py
larsoner: It never really worked robustly. On macOS it almost always errored to open and measure the subprocess

sphinx_gallery/gen_rst.py
```diff
@@ -1651,21 +1681,28 @@ def _sg_call_memory_noop(func):
...
         if callable(show_memory):
             call_memory = show_memory
-        elif gallery_conf["plot_gallery"]:  # True-like
+        elif gallery_conf["parallel"]:
+            if update:
+                logger.warning(
+                    f"{gallery_conf['show_memory']=} disabled due to "
+                    f"{gallery_conf['parallel']=}."
+                )
+                gallery_conf["show_memory"] = False
```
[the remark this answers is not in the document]
larsoner: As a `dict, `gallery_conf` is mutable so the change does not need to be passed back, changing it here changes it everywhere

larsoner (commented):
Okay pushing a commit, let me know what you think @lucyleeow !

sphinx_gallery/gen_rst.py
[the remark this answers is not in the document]
lucyleeow: Welp I returned gallery_conf unnecessarily in my configuration clean up PR then. Will fix this later

larsoner:
@lucyleeow okay to merge this now?

lucyleeow:
Yes I'll merge. Sorry I totally missed that you said fixes can be in follow up PR!

> Okay I realized that for MNE-Python we treat all warnings in doc builds/examples as errors, so I needed to prevent raising the UserWarning that joblib emits as an error.

Just for info, what `UserWarning` ?

lucyleeow:
Should we release or should we wait for maybe #1313 ?

jschueller:
Dont wait! (kidding :)
Thanks a lot for pursuing this @larsoner @lucyleeow

larsoner:
Oh and forgot to say:

> Just for info, what UserWarning ?

Inside `joblib` it could emit a warning like "A worker stopped while some jobs were given to the executor", can't remember if it was because some examples executed too quickly or my `pre_dispatch` was set too high :shrug: But treating that as an error in the doc build (in MNE we treat all uncaught warnings as errors by default) was really problematic because it caused `joblib` to hang.

jschueller:
still any plans for a new release ? I see PR1313 and PR1344 are done

larsoner:
@lucyleeow you up for making a release soon? If not then I can do it

sphinx_gallery/gen_rst.py
```diff
@@ -1237,21 +1268,32 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf, seen_backrefs=No
...
+    title : str
+        The example title.
     cost : tuple
         A tuple containing the ``(time_elapsed, memory_used)`` required to run the
         script.
+    out_vars : dict
+        Variables used to run the script, possibly with entries:
+
+        "stale"
+            True if the example was stale.
```
QuLogic, line 1285: This doc looks incorrect? On line 1315, the 'stale' key is set to a string of something.

QuLogic:
PS, thanks for getting this done; it's going to [speed up our CI a bunch](https://github.com/matplotlib/matplotlib/pull/28617#issuecomment-2252108112) now that we can enable parallel builds.

QuLogic:
> Oh and forgot to say:
> 
> > Just for info, what UserWarning ?
> 
> Inside `joblib` it could emit a warning like "A worker stopped while some jobs were given to the executor", can't remember if it was because some examples executed too quickly or my `pre_dispatch` was set too high 🤷 But treating that as an error in the doc build (in MNE we treat all uncaught warnings as errors by default) was really problematic because it caused `joblib` to hang.

It's probably https://github.com/joblib/joblib/issues/883; it restarts the worker if it uses "too much" memory, which is a bit arbitrary. I also had to disable the warning-as-error to get Matplotlib to build; it might be a good idea to document that somewhere.

larsoner:
> I also had to disable the warning-as-error to get Matplotlib to build; it might be a good idea to document that somewhere.

I didn't realize other projects also had their own strict `filterwarnings("error")` in place! In that case yes it would be good to document it somewhere (making clear that it's different from the `-W` sphinx option).

PR #880, opened 2021-11-12
Add imagesg directive in load style

### LINKED ISSUE #878: `0.10.0` breaks `sphinx_gallery.load_style`
I'm using an exotic setup where the gallery built requires GPUs and thus cannot be run by RTD or the like (see #704) for details. I have a built server set up that builds the galleries and packages them, whereas the normal CI falls back to downloading these pre-built galleries. For that to work, the built server runs with `sphinx_gallery.gen_gallery` whereas the normal CI runs with [`sphinx_gallery.load_style`](https://sphinx-gallery.github.io/stable/advanced.html#using-only-sphinx-gallery-styles).

This worked fine until `0.9.0`. With `0.10.0` images scraped with the default `matplotlib` scraper now use the `.. image-sg::` introduced in #808. Unfortunately, this directive is not loaded when using `sphinx_gallery.load_style`. You end up with warnings like these:

```
.. image-sg:: /galleries/examples/beginner/images/sphx_glr_example_nst_without_pystiche_003.png
   :alt: Input image
   :srcset: /galleries/examples/beginner/images/sphx_glr_example_nst_without_pystiche_003.png
   :class: sphx-glr-single-img
/home/philip/git/pystiche/pystiche/docs/source/galleries/examples/beginner/example_nst_without_pystiche.rst:1305: WARNING: Unknown directive type "image-sg".
```

The build goes through, but the images are not shown in the HTML version.

lucyleeow:
closes #878

Hope you don't mind I gave this a go @pmeier, thought we'd try to get it in the next patch release.
Does this fix the problem?

### DIFF
--- a/sphinx_gallery/load_style.py
+++ b/sphinx_gallery/load_style.py
@@ -6,6 +6,7 @@ This should not be used at the same time as sphinx_gallery.gen_gallery.
 
 """
 from . import __version__, glr_path_static
+from .directives import ImageSg, imagesg_addnode
 
 
 def config_inited(app, config):
@@ -18,6 +19,8 @@ def config_inited(app, config):
 def setup(app):
     app.require_sphinx('1.8')
     app.connect('config-inited', config_inited)
+    app.add_directive("image-sg", ImageSg)
+    imagesg_addnode(app)
     return {
         'parallel_read_safe': True,
         'version': __version__,


pmeier:
> Hope you don't mind I gave this a go

Not at all. I'm pretty swamped at the moment so it would have taken quite while until I would have time to investigate it.

> Does this fix the problem?

Will have a look later today and get back to you.

pmeier:
Works like a charm. Thanks for the quick fix!

larsoner:
Thanks @lucyleeow !

PR #886, opened 2021-11-24
Fix "Out" layout for pydata-sphinx-theme

timhoffm:
Follow up to #484, which got misaligned again, likely through pydata-sphinx-theme.

Note: This is a purely empirical suggestion based on the matplotlib docs, e.g.:
https://matplotlib.org/devdocs/tutorials/colors/colormap-manipulation.html

I don't know how to find out whether this is generally ok.

Before:
grafik

grafik

After:
grafik

grafik

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1190,7 +1190,7 @@ than a single static image of the animation figure, you should add::
 
 HTML embedding options can be changed by setting ``rcParams['animation.html']``
 and related options in your
-:ref:`matplotlib rcParams <matplotlib:matplotlib-rcparams>`.
+`matplotlib rcParams <https://matplotlib.org/stable/tutorials/introductory/customizing.html>`_.
 It's also recommended to ensure that "imagemagick" is available as a
 ``writer``, which you can check with
 :class:`matplotlib.animation.ImageMagickWriter.isAvailable()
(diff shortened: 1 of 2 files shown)


larsoner:
> I don't know how to find out whether this is generally ok.

With stuff like this, I think if matplotlib (using pydata-sphinx-theme) looks good and sphinx-gallery's own doc build (using some RTD theme I think?) and `tinybuild` builds look good (e.g., `cd sphinx_gallery/tests/tinybuild && make`) then it suggests we won't break too many people's docs.

The failure is unrelated to your changes, but maybe you know the correct fix for this and want to implement it here? :)
```
/home/circleci/project/doc/configuration.rst:1191: WARNING: undefined label: matplotlib:matplotlib-rcparams
```

larsoner:
https://2433-25860190-gh.circle-artifacts.com/0/rtd_html/auto_examples/plot_0_sin.html#sphx-glr-auto-examples-plot-0-sin-py at least looks okay/unchanged

timhoffm:
> The failure is unrelated to your changes, but maybe you know the correct fix for this and want to implement it here? :)
> 
> ```
> /home/circleci/project/doc/configuration.rst:1191: WARNING: undefined label: matplotlib:matplotlib-rcparams
> ```

Uh, I didn't know sphinx can cross reference to other repos. That's of course very fragile as we have no way of knowing whether anybody uses our labels, so any change is potentially breaking to someone.

A URL is better than a sphinx ref. On the matplotlib side, we at least try to add redirects when we change URLs. And even if that's not the case, you'll get to the matplotlib main page. In contrast, broken sphinx are just dead.

larsoner:
Thanks @timhoffm !

PR #892, opened 2021-12-10
WIP Python nightly CI

lucyleeow:


StefRe:
The culprit appears to be the new version `7.0.0rc1` of `pytest`  in Python nightly. With this version I get the same failures for Python 3.8.2.

lucyleeow:
Thanks for that, I'm not so familiar with our logging part (and have been failing all day) but that sounds about right.

PR #894, opened 2021-12-10
FIX: Fix for latest pytest

### LINKED ISSUE #893: TestLoggingTee failing pytest v7.0.0rc1
`TestLoggingTee` fails with pytest v7.0.0rc1 (checked with python version 3.8 and 3.10)

When instantiating the `_LoggingTee` during setup:

https://github.com/sphinx-gallery/sphinx-gallery/blob/9a09fc1377de833db9f1d7f23686e37c3d7a02dc/sphinx_gallery/tests/test_gen_rst.py#L887-L890

`self.logger` here:

https://github.com/sphinx-gallery/sphinx-gallery/blob/9a09fc1377de833db9f1d7f23686e37c3d7a02dc/sphinx_gallery/gen_rst.py#L62-L63

should be a `FakeSphinxApp` object due to the fixture:

https://github.com/sphinx-gallery/sphinx-gallery/blob/9a09fc1377de833db9f1d7f23686e37c3d7a02dc/sphinx_gallery/tests/conftest.py#L80-L81

but it is actually a `SphinxLoggerAdapter` object, leading to the failing tests.

Adding a `print` at the end of `_LoggingTee` `setup` reveals a RecursionError, which seems to track to `sys.stdout = sys.stderr = self` in `set_std_and_reset_position` which is called during init of the `_LoggingTee` object.

Not sure if it is related to pytest-dev/pytest#9393 as mentioned in #892

larsoner:
The class-based test was making things a pain, let's just use standard functions and fixtures.

Diff looks worse than necessary because of the indentation change, but not that much actually changed.

Closes #893

### DIFF
--- a/sphinx_gallery/tests/conftest.py
+++ b/sphinx_gallery/tests/conftest.py
@@ -78,23 +78,13 @@ def fakesphinxapp():
 
 
 @pytest.fixture
-def log_collector():
-    orig_dr_logger = docs_resolv.logger
-    orig_gg_logger = gen_gallery.logger
-    orig_gr_logger = gen_rst.logger
-    orig_ps_logger = py_source_parser.logger
+def log_collector(monkeypatch):
     app = FakeSphinxApp()
-    docs_resolv.logger = app
-    gen_gallery.logger = app
-    py_source_parser.logger = app
-    gen_rst.logger = app
-    try:
-        yield app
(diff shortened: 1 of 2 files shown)


larsoner:

lucyleeow:
Thanks @larsoner !

PR #896, opened 2021-12-18
Use pseudo-elements for 'Out:' prefixing

QuLogic:
There's still an arbitrary padding, but now it goes on the added content instead of the existing `pre`. Without the negative margin, this prevents the `pre` from accidentally overlapping previous content. Some additional tweaks are necessary to work with the pydata sphinx theme, but it appears to work for the rtd theme as well.

For example, the error output [on this page](https://65575-1385122-gh.circle-artifacts.com/0/doc/build/html/gallery/style_sheets/style_sheets_reference.html) uses the same class, but is oddly placed due to the negative margins.

On the plot strings example (i.e., using rtd theme), this looks like this (at 500%):
image
and here's how it looks in a Matplotlib example (i.e., using mpl-sphinx-theme, which derives from pydata-sphinx-theme):
image
It's a tiny bit too low, but that's not noticeable at normal scale.

### DIFF
--- a/sphinx_gallery/_static/sg_gallery.css
+++ b/sphinx_gallery/_static/sg_gallery.css
@@ -74,15 +74,25 @@ thumbnail with its default link Background color */
 
 .sphx-glr-script-out {
   color: #888;
-  margin: 0;
+  display: flex;
+  gap: 0.5em;
 }
-p.sphx-glr-script-out {
-    padding-top: 0.7em;
+.sphx-glr-script-out::before {
+  content: "Out:";
+  /* These numbers come from the pre style in the pydata sphinx theme. This
+   * turns out to match perfectly on the rtd theme, but be a bit too low for
+   * the pydata sphinx theme. As I could not find a dimension to use that was
+   * scaled the same way, I just picked one option that worked pretty close for
+   * both. */
+  line-height: 1.4;
+  padding-top: 10px;
 }
 .sphx-glr-script-out .highlight {
   background-color: transparent;
-  margin-left: 2.5em;
-  margin-top: -2.9em;
+  /* These options make the div expand... */
+  flex-grow: 1;
+  /* ... but also keep it from overflowing its flex container. */
+  overflow: auto;
 }
 .sphx-glr-script-out .highlight pre {
   background-color: #fafae2;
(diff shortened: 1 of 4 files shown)


sphinx_gallery/_static/sg_gallery.css
```diff
@@ -74,23 +74,26 @@ thumbnail with its default link Background color */
...
+  content: "Out:";
+  line-height: 1.4;
+  padding-top: 10px;
 }
 .sphx-glr-script-out .highlight {
   background-color: transparent;
-  margin-left: 2.5em;
-  margin-top: -2.9em;
+  flex-grow: 1;
+  overflow: auto;
```
QuLogic, line 88: This keeps the `div` from overflowing its flex container.

sphinx_gallery/_static/sg_gallery.css
```diff
@@ -74,23 +74,26 @@ thumbnail with its default link Background color */
...
 }
 .sphx-glr-script-out .highlight pre {
   background-color: #fafae2;
   border: 0;
   max-height: 30em;
   overflow: auto;
   padding-left: 1ex;
-  margin: 0px;
+  margin: 2px;
```
QuLogic, line 96: This is necessary in the pydata sphinx theme because `pre` has a box shadow which would be clipped by the `overflow:auto` in the parent `div` above..

GaelVaroquaux (approved):
I think that this is overall a better solution than what we currently have, so +1 for my side.

However, we need a note in the human-readable changelog: this change will probably have a consequence on downstream project, and we need to point to what we changed specifically to help them adjust their CSS if needed.

QuLogic:
Sure, how is that?

sphinx_gallery/gen_rst.py
```diff
@@ -140,8 +140,6 @@ def __exit__(self, type_, value, tb):
 
 CODE_OUTPUT = """.. rst-class:: sphx-glr-script-out
 
- Out:
-
```
larsoner, line 144: The only disadvantage I can think of here is that this will render differently (without the `Out:`) in non-HTML builds like LaTeX for example. But I doubt that matters too much -- the gain for HTML builds is probably worth the small LaTeX loss.

larsoner (approved):

sphinx_gallery/gen_rst.py
QuLogic: I think that's okay, as a) the 'Out:' text is in the normal font and doesn't help to call out the output, and b) the output is in a frame already.

sphinx_gallery/_static/sg_gallery.css
lucyleeow: Nitpick - could you add this as a comment in the css for people not familiar with css (like me)?

lucyleeow (approved):
LGTM, just one nitpick

lucyleeow:
Thank you @QuLogic !

PR #903, opened 2022-01-07
DOC Fix example "Identifying function names in a script"

### LINKED ISSUE #900: DOC: Example "Identifying function names in a script"
While looking into #898 I came across the example  [Identifying function names in a script](https://sphinx-gallery.github.io/stable/auto_examples/plot_6_function_identifier.html). What is the point of retrieving the names of the `os` module here? Should the example not rather use the example file [plot_6_function_identifier.py](https://github.com/sphinx-gallery/sphinx-gallery/blob/master/examples/plot_6_function_identifier.py) itself to demonstrate this feature, or am I missing something here?

StefRe:
- use the example script itself instead of the os module for demonstration
- add invisible axes to figure so that the figure is being show in IPython
  using the inline backend (see https://stackoverflow.com/a/51764766/3944322)
- replace plt.draw with plt.show so that the figure is shown when running
  the plain py file

Closes #900.


StefRe:
I think the [error when importing `pyvista`](https://app.circleci.com/pipelines/github/sphinx-gallery/sphinx-gallery/979/workflows/0011d469-d133-40e1-bfd6-e49ad5ddb0cc/jobs/2489/parallel-runs/0/steps/0-108) is unrelated to the PR.

larsoner:
Can you rebase or merge with `master`? That should fix CircleCI and allow us to view the CircleCI artifacts

StefRe:
Rebased, now it works fine. Thanks for the hint :+1: .

examples/plot_6_function_identifier.py
```diff
@@ -34,7 +39,8 @@
 # Also note that global variables of the script have intersphinx references
 # added to them automatically (e.g., ``fig`` and ``fig.text`` below).
```
lucyleeow, line 40: This isn't from this PR but since we are cleaning up the example, I think this text block may be better suited at the bottom of the first text block (~line 7). Could we move this up?

In this text block could we explain that we are demonstrating this functionality by using the internal function `identify_names` to obtain the full resolved import path for all backreferences in the code block above (would be nice to add reference``` :ref:`references_to_examples` ```), and just explicitly say that the left side are functions as written in the code block and the right is the full resolved import path for that function (though I know it's fairly obvious already).

lucyleeow (commented):
This is much nicer, thanks @StefRe . I did just have one suggestion.

examples/plot_6_function_identifier.py
StefRe: Thanks for the comment, please see c7cf0116e427acff2f4b01d26b396e251ea17278. Hope I got it right, feel free to suggest any changes.

larsoner:
> should the code_obj in the example include ax.set_visible and fig.text since these are methods that are called

Not sure offhand, would have to look -- can't remember the key/value pairs and how they're generated. But even if so, it seems unrelated to this PR / an existing bug so I think we can ignore it here at least.

larsoner:
Thanks @StefRe !

PR #908, opened 2022-01-13
Possibility to exclude implicit backreferences

### LINKED ISSUE #898: Limit adding mini-galleries for API docs to explicitly referred to items only
When creating backreferences for mini-galleries for instance for matplotlib galleries, practically every example will include a call to `plt.plot` or `plt.subplots` or the like, leading to a huge amount of spurious references from API to gallery examples, see https://github.com/matplotlib/matplotlib/issues/22036 for details.  

Do you think it would be a good idea to have a config option to limit reference collection to explicitly mentioned examples only, e.g. an option `auto-doc` (by default `True`), so that
https://github.com/sphinx-gallery/sphinx-gallery/blob/fabc70a4afc0aa72db5d06ea5fd665e1cdcf17a6/sphinx_gallery/backreferences.py#L203-L206
could be put into an `if` block like
```
if auto_doc:
   finder = NameFinder(global_variables) 
   if node is not None: 
       finder.visit(node) 
   names = list(finder.get_mapping())
else:
   names = []
``` 
This gives the possibility to only include functions you really want to illustrate by the given example and the default behavior would be unchanged.

StefRe:
New configuration key 'exclude_implicit_doc' to prevent generation of
backreferences for the names given in this key. This only affects
implicite backreferences, i.e. names extracted from the code. Explicit
backreferences given the sphinx directives are not affected.

Closes #898.

I tried my best to write a readable and concise documentation section but this should certainly streamlined a bit by native speakers.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -24,7 +24,8 @@ file:
 - ``subsection_order`` (:ref:`sub_gallery_order`)
 - ``within_subsection_order`` (:ref:`within_gallery_order`)
 - ``reference_url`` (:ref:`link_to_documentation`)
-- ``backreferences_dir``, ``doc_module``, and ``inspect_global_variables`` (:ref:`references_to_examples`)
+- ``backreferences_dir``, ``doc_module``, ``exclude_implicit_doc``,
+  and ``inspect_global_variables`` (:ref:`references_to_examples`)
 - ``default_thumb_file`` (:ref:`custom_default_thumb`)
 - ``thumbnail_size`` (:ref:`setting_thumbnail_size`)
 - ``line_numbers`` (:ref:`adding_line_numbers`)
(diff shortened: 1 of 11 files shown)


doc/configuration.rst
```diff
@@ -432,7 +434,12 @@ your Sphinx-Gallery configuration ``conf.py`` file with::
 
         # Modules for which function/class level galleries are created. In
         # this case sphinx_gallery and numpy in a tuple of strings.
-        'doc_module'          : ('sphinx_gallery', 'numpy')}
+        'doc_module'          : ('sphinx_gallery', 'numpy')
```
larsoner, line 437:
```suggestion
        'doc_module'          : ('sphinx_gallery', 'numpy'),
```

doc/configuration.rst
```diff
@@ -432,7 +434,12 @@ your Sphinx-Gallery configuration ``conf.py`` file with::
...
+
+        # objects to exclude from implicit backreferences. The default option
+        # is an empty set, i.e. exclude nothing.
+        'exclude_implicit_doc': {}
```
larsoner, line 441: And if you want, so that the next person doesn't need to add one...
```suggestion
        'exclude_implicit_doc': {},
```

larsoner (commented):
Sorry for the slow review, this looks reasonable!

My only (minor) gripe is that `test_full.py` / `tinybuild` is not modified to make sure that this works in practice. Are you up for adding this new config var to `tests/tinybuild/conf.py`, maybe to exclude exactly one of the current implicit refs that that build creates (you can test locally by running `make` in that dir), then modify `sphinx_gallery/tests/test_full.py` to assert that the ref is *not* in there (and perhaps also check manually that it is there on `master`)?

StefRe:
Will certainly look into it but right now, I have neither time nor muse in the current situation.

larsoner:
Thanks @StefRe !

PR #916, opened 2022-02-16
MAINT: Remove LooseVersion

larsoner:
I don't like seeing in my Python 3.10 builds:
```
/home/larsoner/python/sphinx-gallery/sphinx_gallery/py_source_parser.py:105: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.           
```
So I removed one unnecessary comparison about python version (we require 3.7 already), and another involving matplotlib version that just triaged which GIF writer to use -- 3.3.1 is almost 2 years old, and the GIF-writing bug will only happen if someone actually uses GIFs in their docs *and* uses a quite old matplotlib version, which I think is rare enough not to worry about.

### DIFF
--- a/sphinx_gallery/py_source_parser.py
+++ b/sphinx_gallery/py_source_parser.py
@@ -10,7 +10,6 @@ from __future__ import division, absolute_import, print_function
 
 import codecs
 import ast
-from distutils.version import LooseVersion
 from io import BytesIO
 import re
 import sys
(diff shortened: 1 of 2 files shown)


larsoner:
Pushed a fix for this, too:
```
  File "/home/circleci/python_env/lib/python3.8/site-packages/sphinx_gallery/backreferences.py", line 99, in get_mapping
    obj = getattr(obj, level)
RuntimeError: wrapped C/C++ object of type PyQtGraphBrowser has been deleted
```
We just need to `except Exception` rather than `except AttributeError`

PR #917, opened 2022-02-18
MAINT: Clean up namespace

larsoner:
When running `reset_modules` in the `after` case, we need to `del script_vars, global_vars` in order to ensure a `gc.collect()` that follows will actually be able to collect the variables from the script that ran. A bit of a corner case, but it's easy to implement and hard to test, so I think this is probably good enough as is.

Okay for you @lucyleeow ?

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -1026,6 +1026,7 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf,
     _write_backreferences(backrefs, seen_backrefs, gallery_conf, target_dir,
                           fname, intro, title)
 
+    del script_vars, global_variables  # don't keep these during reset
     if executable and gallery_conf['reset_modules_order'] in ['after', 'both']:
         clean_modules(gallery_conf, fname, 'after')
 


larsoner:
I think this is pretty straightforward so I'll go ahead and merge, but @lucyleeow if you or others have comments I'm happy to follow up with another PR

lucyleeow:
Thanks @larsoner, sorry could you expand on:
> we need to del script_vars, global_vars in order to ensure a gc.collect() that follows will actually be able to collect the variables from the script that ran

You say it is an edge case, how was `gc.collect()` managing to collect global variables when they weren't being deleted?
Also why does `script_vars` need to be deleted?

Thanks!

larsoner:
> how was gc.collect() managing to collect global variables when they weren't being deleted?

It wasn't, that's precisely the problem! If `gen_rst` has `script_vars` (with `['global_vars']`) and the variable `global_vars` in its scope, when the post-run resetter is called, if *it* calls `gc.collect()`, the variables that the script used at runtime cannot be collected because they are still in the `gen_rst` frame's `locals()`. If we `del script_vars, global_vars` in `gen_rst` before calling the resetter, when the resetter function `gc.collect()`s, there are no more references anywhere in the stack frame (in particular, in sphinx-gallery's frames) to the variables that the given script uses.

Thus, after this PR, a resetter can call `gc.collect()` and check to make sure that their library *does* properly delete all references so that `gc.collect()` can do its job properly. This is important so that libraries can check to make sure memory consumption doesn't just go up over time, e.g., https://github.com/mne-tools/mne-qt-browser/pull/61#issue-1143583133 , https://github.com/mne-tools/mne-python/pull/9448#issuecomment-858740833 , etc. Basically in GUI operations and long builds, retained references that `gc.collect()` cannot resolve become problematic, and this PR makes sphinx-gallery's handling cleaner.

larsoner:
Also this PR, which involves the interactions of VTK, PyQt5, and MNE-Python you can see the struggles:

https://github.com/mne-tools/mne-python/pull/8399

If sphinx-gallery didn't clear its references, none of this debugging and fixing would be possible.

And for a particularly tricky corner case, see https://github.com/NeuralEnsemble/python-neo/pull/1075 !

lucyleeow:
Thanks for explaining! That makes more sense now.
Also I understand now that when you say 'edge case' you meant that the garbage collection does not always cause problems.

PR #919, opened 2022-02-22
Add Biotite to list of user projects

padix-key:
[Biotite](https://www.biotite-python.org/) is a bioinformatics library written in Python. It uses *sphinx-gallery* to show applications of the library on biological questions (https://www.biotite-python.org/examples/gallery/index.html).

If you see this library fit, this PR adds *Biotite* to the list of user projects.


larsoner:
Thanks @padix-key !

PR #922, opened 2022-03-03
Ensures right builder conifg

### LINKED ISSUE #921: Broken plot links if sphinx uses `htmldir`
If I build our documentation with the sphinx builder option `htmldir` the links to the in-site plots are broken. The thumbnails are working.

I already discussed this issue on [read the docs](https://github.com/readthedocs/readthedocs.org/issues/8982) because I thought the problems is there. But they figured out that this might be a flaw with sphinx-gallery.

ExtremOPS:
This should resolve #921. I am very unsure if that is the right place or the right style for a) fixing this issue and b) unittesting.

I am happy to refactor to meet your styles/philosophy.

### DIFF
--- a/doc/getting_started.rst
+++ b/doc/getting_started.rst
@@ -26,10 +26,10 @@ help you get started.
 
 .. note::
    Working `sphinx 
-   builders<https://www.sphinx-doc.org/en/master/man/sphinx-build.html#options>`_ 
-   for sphinx_gallery include `html` and `latex`.
-
-   E.g., `dirhtml` will cause broken image links.
+   builders <https://www.sphinx-doc.org/en/master/man/sphinx-build.html#options>`_ 
+   for sphinx_gallery include `html` and `latex`. Support for other builders
+   is not guaranteed (e.g., `dirhtml` is known to be broken and will cause
+   broken image links.).
 
 
 .. _set_up_your_project:
--- a/doc/getting_started.rst
+++ b/doc/getting_started.rst
@@ -24,6 +24,14 @@ A `template repository <https://github.com/sphinx-gallery/sample-project>`_,
 with sample example galleries and basic configurations is also available to
 help you get started.
 
+.. note::
+   Working `sphinx 
+   builders<https://www.sphinx-doc.org/en/master/man/sphinx-build.html#options>`_ 
+   for sphinx_gallery include `html` and `latex`.
+
+   E.g., `dirhtml` will cause broken image links.
+
+
 .. _set_up_your_project:
 
 Overview your project files and folders
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -425,7 +425,7 @@ def generate_gallery_rst(app):
     blocked_builder = ['dirhtml']
     if app.builder.name in blocked_builder:
         msg = (
-            f'Builder is so to be {app.builder.name}. sphinx_gallery does ',
+            f'Builder is set to be {app.builder.name}. sphinx_gallery does ',
             f'not work for any of the following:\n\t{[blocked_builder]}'
         )
         raise ConfigError(msg)
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -425,8 +425,8 @@ def generate_gallery_rst(app):
     blocked_builder = ['dirhtml']
     if app.builder.name in blocked_builder:
         msg = (
-            f'Builder is set to be {app.builder.name}. sphinx_gallery does ',
-            f'not work for any of the following:\n\t{[blocked_builder]}'
+            f'Builder is set to {repr(app.builder.name)}. sphinx_gallery does '
+            f'not work for any of the following: {[blocked_builder]}'
         )
         raise ConfigError(msg)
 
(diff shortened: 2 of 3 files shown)


sphinx_gallery/gen_gallery.py
```diff
@@ -422,6 +422,9 @@ def generate_gallery_rst(app):
     Start the Sphinx-Gallery configuration and recursively scan the examples
     directories in order to populate the examples gallery
     """
+    if app.builder.name != 'html':
+        raise ConfigError('Builder must be `html` for sphinx-gallery')
```
larsoner, line 426: I think technically we do support LaTeX as well. Maybe better to make this a blocklist (containing just `dirhtml` for now) rather than an allowlist?

ExtremOPS: Yes, probably a good idea. 

However, I have not checked other builders than `html` and `dirhtml` yet.

larsoner: I think that's okay. We can just add the broken ones as we come along them. We can take the easy way out and presume innocent until proven guilty here :)

ExtremOPS: Then it might be a good idea if I add this also in the docu. Could you please point me to the location where you would put this info.

larsoner: Maybe a `.. note::` at the bottom of this main/topmost section:

https://sphinx-gallery.github.io/stable/getting_started.html#creating-a-basic-gallery

Saying that the HTML and LaTeX builders are the only ones that have been tested/targeted so far? The corresponding place I think would be here:

https://github.com/sphinx-gallery/sphinx-gallery/blame/master/doc/getting_started.rst#L26

doc/getting_started.rst
```diff
@@ -24,6 +24,14 @@ A `template repository <https://github.com/sphinx-gallery/sample-project>`_,
 with sample example galleries and basic configurations is also available to
 help you get started.
 
+.. note::
+   Working `sphinx 
+   builders<https://www.sphinx-doc.org/en/master/man/sphinx-build.html#options>`_ 
+   for sphinx_gallery include `html` and `latex`.
+
+   E.g., `dirhtml` will cause broken image links.
```
larsoner, line 32: Code in RST is double-backticks. Single-backticks is "default_role", whatever we have that set to. Probably better just to be explicit with code mode here

Also, I think a missing space is causing the link rendering not to work

https://2561-25860190-gh.circle-artifacts.com/0/rtd_html/getting_started.html

```suggestion
   builders <https://www.sphinx-doc.org/en/master/man/sphinx-build.html#options>`_ 
   for sphinx_gallery include `html` and `latex`. Support for other builders
   is not guaranteed (e.g., `dirhtml` is known to be broken and will cause
   broken image links.).
```

sphinx_gallery/gen_gallery.py
```diff
@@ -422,6 +422,14 @@ def generate_gallery_rst(app):
     Start the Sphinx-Gallery configuration and recursively scan the examples
     directories in order to populate the examples gallery
     """
+    blocked_builder = ['dirhtml']
+    if app.builder.name in blocked_builder:
+        msg = (
+            f'Builder is so to be {app.builder.name}. sphinx_gallery does ',
```
larsoner, line 428:
```suggestion
            f'Builder is set to be {app.builder.name}. sphinx_gallery does ',
```

sphinx_gallery/tests/test_gen_gallery.py
```diff
@@ -22,6 +22,9 @@ def test_bad_config():
     sphinx_gallery_conf = dict(example_dir='')
     with pytest.raises(ConfigError, match="example_dir.*did you mean 'examples_dirs'?.*"):  # noqa: E501
         _complete_gallery_conf(sphinx_gallery_conf, '', True, False)
+    sphinx_gallery_conf = dict(builder='dirhtml')
+    with pytest.raises(ConfigError, match="Builder must be `html` for sphinx-gallery?.*"):  # noqa: E501
```
larsoner, line 26:
```suggestion
    with pytest.raises(ConfigError, match=".*dirhtml.*sphinx_gallery does not work.*"):  # noqa: E501
```

sphinx_gallery/tests/test_gen_gallery.py
```diff
@@ -22,6 +22,9 @@ def test_bad_config():
     sphinx_gallery_conf = dict(example_dir='')
     with pytest.raises(ConfigError, match="example_dir.*did you mean 'examples_dirs'?.*"):  # noqa: E501
         _complete_gallery_conf(sphinx_gallery_conf, '', True, False)
+    sphinx_gallery_conf = dict(builder='dirhtml')
+    with pytest.raises(ConfigError, match=".*dirhtml.*sphinx_gallery does not work.*"):  # noqa: E501
+        _complete_gallery_conf(sphinx_gallery_conf, '', True, False)
```
larsoner, line 27: This looks like it's not the right place to add a test actually. It's not a SG config thing, it's a sphinx builder thing. So I think this needs to be in a different test... let me look and push a quick commit to fix.

larsoner:
Thanks @ExtremOPS !

ExtremOPS:
You're welcome! I am happy about open source and helping each other with to make it better 🥇

PR #923, opened 2022-03-03
add GitHub URL for PyPi

andriyor:
Warehouse now uses the project_urls provided to display links in the sidebar on [this screen](https://pypi.org/project/requests/), as well as including them in API responses to help the automation tool find the source code for Requests.

Example: [pypi.org/pypi/requests/json](https://pypi.org/pypi/requests/json)
Docs: [packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,6 +46,9 @@ setup(
     ]},
     scripts=['bin/copy_sphinxgallery.sh', 'bin/sphx_glr_python_to_jupyter.py'],
     url="https://sphinx-gallery.github.io",
+    project_urls={
+        "Source": "https://github.com/sphinx-gallery/sphinx-gallery",
+    },
     author="Óscar Nájera",
     author_email='najera.oscar@gmail.com',
     install_requires=install_requires,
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -46,6 +46,9 @@
     ]},
     scripts=['bin/copy_sphinxgallery.sh', 'bin/sphx_glr_python_to_jupyter.py'],
     url="https://sphinx-gallery.github.io",
+    project_urls={
+        "Source": "https://github.com/sphinx-gallery/sphinx-gallery",
```
larsoner, line 50:
```suggestion
        "Source": "https://github.com/sphinx-gallery/sphinx-gallery",
        "Documentation": "https://sphinx-gallery.github.io",
```

andriyor: `url` already point to docs so I think it's duplication

larsoner: Indeed it's the same link -- but perhaps it's worth it to have both "Homepage" and "Documentation" show up in the sidebar?

PR #930, opened 2022-03-21
[WIP] sphinx_gallery_expected_error code-block directive

smarie:
Will fix #912

PR #932, opened 2022-03-22
MAINT: Fix CIs

larsoner:
Hopefully fix failures seen in https://github.com/sphinx-gallery/sphinx-gallery/pull/930 by combining the problematic job with another one (seems okay to do here)


larsoner:
This is a pyface/Qt issue similar to https://stackoverflow.com/questions/71006962/python-loading-mayavi-mlab-throws-attribute-error-probably-from-qt-loading-may -- I think testing Mayavi on one fewer run is reasonable here, so I'll merge if it comes back green

PR #940, opened 2022-04-11
Improve Jupyter notebook converter's handling of code blocks

guberti:
## What this PR does
This pull request improves the way `notebook.py` creates Jupyter notebooks by introducing two changes. First, when generating code blocks in Markdown, we now use the ```[language] format instead of indenting. This causes one-line code blocks to render correctly, and also makes syntax highlighting work correctly. For example, consider the lines:
```
    .. code-block:: bash

      flatc --version
```
Many Jupyter notebook editors (like [Google Colab](https://colab.research.google.com/) do not render this as code as all! With this change, it would instead be translated like the following, which both renders correctly and has syntax highlighting:
````
```bash
flatc --version
```
````
Additionally, this PR makes `notebook.py` recognize so-called [Jupyter magic commands](https://ipython.readthedocs.io/en/stable/interactive/magics.html). These commands (like `%%bash`) allow code in other languages to be executed in Jupyter. If the notebook converter sees a code block that starts with a Jupyter cell magic command, we will now turn it into a runnable code block.

## Why is this useful?

I'm a contributor on [Apache TVM](https://github.com/apache/tvm), which uses Sphinx Gallery for our [tutorials](https://tvm.apache.org/docs/how_to/work_with_microtvm/index.html). Having these features would make the notebooks generated from the "Download Jupyter notebook" button work "out of the box". 

## How is it tested?

This PR also adds both unit tests and integration tests for both of these features (for the integration tests, I added a few lines to `plot_parse.py`). I've also verified that `flake8` passes without errors.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -50,7 +50,7 @@ file:
 - ``show_memory`` (:ref:`show_memory`)
 - ``show_signature`` (:ref:`show_signature`)
 - ``binder`` (:ref:`binder_links`)
-- ``promote_jupyter_magic`` (:ref: `promote_jupyter_magic`)
+- ``promote_jupyter_magic`` (:ref:`promote_jupyter_magic`)
 - ``first_notebook_cell`` and ``last_notebook_cell`` (:ref:`own_notebook_cell`)
 - ``notebook_images`` (:ref:`notebook_images`)
 - ``pypandoc`` (:ref:`use_pypandoc`)
@@ -1090,8 +1090,8 @@ for an example that uses the `public Binder server <http://mybinder.org>`_.
 
 .. _promote_jupyter_magic:
 
-Make Jupyter cell magic compile as code
-=======================================
+Making cell magic executable in notebooks
+=========================================
 
 Often times, tutorials will include bash scripts for the user to copy/paste into
 their terminal. These scripts are not intended to be run when someone is
@@ -1139,7 +1139,7 @@ and only affects the creation of Jupyter notebooks.
 .. warning::
   It is good practice to ensure the ``.py`` and ``.html`` files match the ``.ipynb``
   files as closely as possible. This flag (and Jupyter cell magics) should only
-  be used when the relevant code is already intended to be executed by the end
+  be used when the relevant code is intended to be executed by the end
   user.
 
 .. _without_execution:
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -50,6 +50,7 @@ file:
 - ``show_memory`` (:ref:`show_memory`)
 - ``show_signature`` (:ref:`show_signature`)
 - ``binder`` (:ref:`binder_links`)
+- ``promote_jupyter_magic`` (:ref: `promote_jupyter_magic`)
 - ``first_notebook_cell`` and ``last_notebook_cell`` (:ref:`own_notebook_cell`)
 - ``notebook_images`` (:ref:`notebook_images`)
 - ``pypandoc`` (:ref:`use_pypandoc`)
@@ -1086,6 +1087,60 @@ Binder links will point to these notebooks.
 
 See the Sphinx-Gallery `Sphinx configuration file <https://github.com/sphinx-gallery/sphinx-gallery/blob/master/doc/conf.py>`_
 for an example that uses the `public Binder server <http://mybinder.org>`_.
+
+.. _promote_jupyter_magic:
+
+Make Jupyter cell magic compile as code
+=======================================
+
+Often times, tutorials will include bash scripts for the user to copy/paste into
+their terminal. These scripts are not intended to be run when someone is
+building the documentation, as they will already have those dependencies in
+their environment. Hence they are normally written as code blocks inside text::
+
+  #%%
+  # Installing dependencies
+  #
+  #     .. code-block:: bash
+  #
+  #       pip install -q tensorflow
+  #       apt-get -qq install curl
+
+This works fine for the ``.py`` and ``.html`` files, but causes problems when
+rendered as an Jupyter notebook. The downloaded ``.ipynb`` file will not have
+those dependencies installed, and thus will not work without running the script.
+
+To fix this, we can set the ``promote_jupyter_magic`` flag in ``conf.py``::
+
+  sphinx_gallery_conf = {
+      ...
+      'promote_jupyter_magic': True,
+  }
+
+If this flag is ``True``, then when a Jupyter notebook is being built, any code
+block starting with a `Jupyter cell magic <https://ipython.readthedocs.io/en/stable/interactive/magics.html>`_ (e.g. ``%%bash`` or ``%%writefile``)
+will be turned into a runnable code block.
+
+For our earlier example, we could change the Markdown text to::
+
+  #%%
+  # Installing dependencies
+  #
+  #     .. code-block:: bash
+  #
+  #       %%bash
+  #       pip install -q tensorflow
+  #       apt-get -qq install curl
+
+Then, TensorFlow and Curl would be automatically installed upon running the
+Jupyter notebook. This works for any cell magic (not just those mentioned above)
+and only affects the creation of Jupyter notebooks.
+
+.. warning::
+  It is good practice to ensure the ``.py`` and ``.html`` files match the ``.ipynb``
+  files as closely as possible. This flag (and Jupyter cell magics) should only
+  be used when the relevant code is already intended to be executed by the end
+  user.
 
 .. _without_execution:
 
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1093,10 +1093,10 @@ for an example that uses the `public Binder server <http://mybinder.org>`_.
 Making cell magic executable in notebooks
 =========================================
 
-Often times, tutorials will include bash scripts for the user to copy/paste into
-their terminal. These scripts are not intended to be run when someone is
-building the documentation, as they will already have those dependencies in
-their environment. Hence they are normally written as code blocks inside text::
+Often times, tutorials will include bash code for the user to copy/paste into
+their terminal. This code should not be run when someone is building the
+documentation, as they will already have those dependencies in their
+environment. Hence they are normally written as code blocks inside text::
 
   #%%
   # Installing dependencies
@@ -1108,7 +1108,7 @@ their environment. Hence they are normally written as code blocks inside text::
 
 This works fine for the ``.py`` and ``.html`` files, but causes problems when
 rendered as an Jupyter notebook. The downloaded ``.ipynb`` file will not have
-those dependencies installed, and thus will not work without running the script.
+those dependencies installed, and will not work without running the bash code.
 
 To fix this, we can set the ``promote_jupyter_magic`` flag in ``conf.py``::
 
@@ -1118,7 +1118,7 @@ To fix this, we can set the ``promote_jupyter_magic`` flag in ``conf.py``::
   }
 
 If this flag is ``True``, then when a Jupyter notebook is being built, any code
-block starting with a `Jupyter cell magic <https://ipython.readthedocs.io/en/stable/interactive/magics.html>`_ (e.g. ``%%bash`` or ``%%writefile``)
+block starting with `Jupyter cell magics <https://ipython.readthedocs.io/en/stable/interactive/magics.html>`_ (e.g. ``%%bash`` or ``%%writefile``)
 will be turned into a runnable code block.
 
 For our earlier example, we could change the Markdown text to::
@@ -1132,15 +1132,14 @@ For our earlier example, we could change the Markdown text to::
   #       pip install -q tensorflow
   #       apt-get -qq install curl
 
-Then, TensorFlow and Curl would be automatically installed upon running the
+meaning TensorFlow and Curl would be automatically installed upon running the
 Jupyter notebook. This works for any cell magic (not just those mentioned above)
 and only affects the creation of Jupyter notebooks.
 
 .. warning::
   It is good practice to ensure the ``.py`` and ``.html`` files match the ``.ipynb``
-  files as closely as possible. This flag (and Jupyter cell magics) should only
-  be used when the relevant code is intended to be executed by the end
-  user.
+  files as closely as possible. This functionality should only be used when the
+  relevant code is intended to be executed by the end user.
 
 .. _without_execution:
 
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -79,6 +79,7 @@ DEFAULT_GALLERY_CONF = {
     'thumbnail_size': (400, 280),  # Default CSS does 0.4 scaling (160, 112)
     'min_reported_time': 0,
     'binder': {},
+    'promote_jupyter_magic': False,
     'image_scrapers': ('matplotlib',),
     'compress_images': (),
     'reset_modules': ('matplotlib', 'seaborn'),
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by guberti** [comment]
Would love a look from @larsoner!

**Comment by larsoner** [comment]
> These commands (like %%bash) allow code in other languages to be executed in Jupyter. If the notebook converter sees a code block that starts with a Jupyter cell magic command, we will now turn it into a runnable code block.

Is this necessary? I mean I see why it's perhaps cool/useful, but so far in SG we try to keep ``.py == .ipynb == .html`` as closely as possible in terms of what can be run, what actually gets run, and what the output looks like. The more we deviate from this, the less clear all these relationships become.

For example, in the BASH case, you turn this into a runnable code block for ipynb. But it could in principle be run in Python via a `subprocess` call, and also run during tutorial generation. So it's weird to have it in one place and not the others. It's also weird to take what's currently "all comments/text" and turn parts of it runnable sometimes.

So I think I'm -1 on making code-blocks-inside-comment-blocks runnable.

**Comment by guberti** [comment]
You're right that in general, keeping `.py == .ipynb == .html` is sensible, and turning text into code is a little weird. However, this change to the way `.ipynb` notebooks are generated will sometimes make the difference between the notebooks working when you click "run all", and the notebooks failing.

Consider a use case where the tutorial relies on some dependencies to work (for example, we need to install the Arduino command line interface, or install some packages with `apt`):
```bash
apt-get -qq install imagemagick curl
curl -fsSL https://raw.githubusercontent.com/arduino/arduino-cli/master/install.sh | sh
```

Usually for a tutorial, the environment being used to build the documentation already has these things installed, while someone following the tutorial is expected to copy/paste them into their shell. That's fine for the `.py` and `.html` use cases, but it means that the dependencies will not be installed in the `.ipynb`, meaning it won't just "work out of the box".

We *could* use `subprocess`, as you suggest, in a Python block to run these commands:
```python
subprocess.run("apt-get -qq install imagemagick curl; "
  curl -fsSL https://raw.githubusercontent.com/arduino/arduino-cli/master/install.sh | sh
, capture_output=True, shell=True)
```
However, this could screw up the `.py` file (which was working fine before, since these dependencies are already installed and it does not make sense to re-install them every time the script is run), and it makes the `.html` file uglier too.

For use cases like this, *having the option* to run this Bash code makes the difference between some `.ipynb` files working, and failing. Most folks won't use this, but they can just leave out the `%%bash` cell magic. And it's useful enough when needed to justify its inclusion, in my opinion.

lucyleeow:
On the fence about this. I do like the `.py == .ipynb == .html` principal.

> Consider a use case where the tutorial relies on some dependencies to work

I would have thought having to install the correct dependencies to be able to run a tutorial is pretty standard :thinking:

guberti:
I'm not sure about other projects, but for a *lot* of the Apache TVM tutorials, running the `.ipynb` notebook fails for no other reason than because the `bash` install scripts are not runnable. Here are the first three examples I found:

- [Auto-tuning a Convolutional Network for NVIDIA GPU](https://tvm.apache.org/docs/how_to/tune_with_autotvm/tune_relay_cuda.html) (view the `.ipynb` notebook on [Google Colab](https://colab.research.google.com/github/apache/tvm-site/blob/asf-site/docs/_downloads/d1434e80dd27eef6b1c9cbaa13f1197b/tune_relay_cuda.ipynb#scrollTo=6ElxzhzgVWc6))
- [Deploy a Framework-prequantized Model with TVM - Part 3](https://tvm.apache.org/docs/how_to/deploy_models/deploy_prequantized_tflite.html) (view the `.ipynb` notebook on [Google Colab](https://colab.research.google.com/github/apache/tvm-site/blob/asf-site/docs/_downloads/1a26d790f7b98309d730181290dae3ee/deploy_prequantized_tflite.ipynb))
- [Compile PyTorch Models](https://tvm.apache.org/docs/how_to/compile_models/from_pytorch.html) (view the `.ipynb` notebook on [Google Colab](https://colab.research.google.com/github/apache/tvm-site/blob/asf-site/docs/_downloads/1f4943aed1aa607b2775c18b1d71db10/from_pytorch.ipynb))

For all of these tutorials, the `.py` and `.html` versions run fine. The Python notebooks, however, will fail if run, for no reason other than the bash scripts not being run. If my pull request were merged, these would all be fixable by just adding `%%bash` above Bash blocks we want run.

Note that there's not really another good way to get these Python notebooks working again. We could use Sphinx-Gallery to add a header block on top of all of them, but since they all need different dependencies, this would not do much good.

Lastly, I'm of the opinion that this change *improves* the `.html == .ipynb` equivalence. Someone following the `.html` tutorial would be expected to run those Bash scripts, so it makes sense for them to be run in the `.ipynb` file too.

areusch:
just to chime in here (also working with @guberti on some related TVM work), I do think that it's fairly common for Jupyter interpreters to be provided in the cloud with some pre-installed environment that can be missing packages. The most natural way to let users modify those environments is via a shell script. Understand the desire to maintain simplicity here, but I do think that `subprocess.run(...)` is a lot harder for new users to understand than e.g. a bash block, and one of the common uses of SG to render tutorials.

Another more general solution that might be interesting here is to provide a plugin to SG for each code block based on the language specified, and then allow folks to customize how those code blocks are executed and how the output is created. Then I think the principle of equality between all three formats may be easier to consider maintained at a project level. I'm not sure if this is already a feature--apologies for spamming if so.

GaelVaroquaux:
> Another more general solution that might be interesting here is to provide a plugin to SG for each code block based on the language specified,

Those Python scripts than cease to be actual Python scripts, and the vision of sphinx-gallery was to operate on valid Python scripts, as these could be run outside of the framework.

I would worry about the maintenance burden of both the plugin and the scripts. A situation with that many moving part will break, and people will come to sphinx-gallery looking for solutions that we don't have.

guberti:
I still think that allowing tutorial writers to make `bash` blocks executable is the most Pythonic thing to do. Those bash blocks are meant to be run by users following the `.html` tutorial, so allowing them to be runnable from the `.ipynb` notebook improves the `.html == .ipynb` equivalence. And of course, the old behavior will still be the default - making the block runnable requires explicitly specifying `%%bash`.

larsoner:
> And of course, the old behavior will still be the default - making the block runnable requires explicitly specifying %%bash.

If someone already (today) has a code-inside-comment-block that starts with `%% bash` or whatever just as a hint to how you *could* do something in a notebook, then we merge this PR, then they will (on their next build with our `master` branch) all of a sudden have a change to their `.ipynb` file, right? So in this sense this behavior is opt-out and they'd have to do so by removing the `%% bash` from their RST `.. code-block::`.

I think we should instead make this behavior opt-in with something like `sphinx_gallery_conf['code_in_comments_notebook_executable'] = False` (by default) or so. And like #941, we should perhaps `.. warning::` to discourage its use because it makes the `.py` scripts less complete/capable compared to the `.ipynb` notebooks.

guberti:
Making this behavior opt-in by default does seem prudent. I've made that change (calling the relevant flag `promote_jupyter_magic`), and fixed the affected tests.

I've also added it to the configuration documentation, with a warning as @larsoner suggested above.

sphinx_gallery/notebook.py
```diff
@@ -241,6 +267,25 @@ def add_markdown_cell(work_notebook, markdown):
     work_notebook["cells"].append(markdown_cell)
 
 
+def promote_jupyter_cell_magic(work_notebook, markdown):
+    # Regex detects all code blocks that use %% Jupyter cell magic
```
lucyleeow, line 271: docstring

lucyleeow (commented):
Thanks @guberti , just a quick first pass.

doc/configuration.rst
```diff
@@ -1079,6 +1080,60 @@ Binder links will point to these notebooks.
 See the Sphinx-Gallery `Sphinx configuration file <https://github.com/sphinx-gallery/sphinx-gallery/blob/master/doc/conf.py>`_
 for an example that uses the `public Binder server <http://mybinder.org>`_.
 
+.. _promote_jupyter_magic:
+
+Make Jupyter cell magic compile as code
```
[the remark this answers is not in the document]
guberti: How about "Making cell magic executable in notebooks"? Most of the other titles have the verb as a present participle (e.g. `Compressing images`, `Building without executing examples`, `Choosing the thumbnail image`).

sphinx_gallery/notebook.py
```diff
@@ -69,6 +70,29 @@ def directive_fun(match, directive):
...
+def convert_code_to_md(text):
+    code_regex = r'[ \t]*\.\. code-block::[ \t]*([a-z]*)\n[ \t]*\n'
+    indent_regex = re.compile(r'[ \t]*')
+    while True:
+        code_block = re.search(code_regex, text)
+        if not code_block:
+            break
+        start_index = code_block.span()[1]
+        indent = indent_regex.search(text, start_index).group(0)
+        if not indent:
```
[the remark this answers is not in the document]
guberti: Added! Also rewrote the indent checking logic to be a tad cleaner.

sphinx_gallery/notebook.py
```diff
@@ -69,6 +70,29 @@ def directive_fun(match, directive):
                     match.group(1).strip()))
 
 
+def convert_code_to_md(text):
+    code_regex = r'[ \t]*\.\. code-block::[ \t]*([a-z]*)\n[ \t]*\n'
```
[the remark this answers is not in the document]
guberti: Nah you're right, we should convert that. I've changed `[a-z]*` to `\S*` (all non-whitespace characters).

guberti:
> Thanks @guberti , just a quick first pass.

Thanks for the comments @lucyleeow! The things you mentioned have been fixed.

guberti:
Not sure why the docs build is failing - the error seems to be unrelated to my changes:
```
Exception occurred:
  File "/home/circleci/.local/lib/python3.8/site-packages/sphinx/setup_command.py", line 169, in run
    raise DistutilsExecError(
distutils.errors.DistutilsExecError: caused by html builder.
The full traceback has been saved in /tmp/sphinx-err-zmpy9rn2.log, if you want to report the issue to the developers.
```
Any thoughts about the cause?

guberti:
Mind taking another look @lucyleeow?

guberti:
@larsoner anything else I need to do here? Would love to get this finished and merged.

doc/configuration.rst
```diff
@@ -1087,6 +1088,60 @@ Binder links will point to these notebooks.
 See the Sphinx-Gallery `Sphinx configuration file <https://github.com/sphinx-gallery/sphinx-gallery/blob/master/doc/conf.py>`_
 for an example that uses the `public Binder server <http://mybinder.org>`_.
 
+.. _promote_jupyter_magic:
+
+Making cell magic executable in notebooks
+=========================================
+
+Often times, tutorials will include bash scripts for the user to copy/paste into
```
lucyleeow, line 1096:
```suggestion
Often times, tutorials will include bash code for the user to copy/paste into
```
not sure, script just has connotations of file to me.

doc/configuration.rst
```diff
@@ -1087,6 +1088,60 @@ Binder links will point to these notebooks.
...
+  # Installing dependencies
+  #
+  #     .. code-block:: bash
+  #
+  #       pip install -q tensorflow
+  #       apt-get -qq install curl
+
+This works fine for the ``.py`` and ``.html`` files, but causes problems when
+rendered as an Jupyter notebook. The downloaded ``.ipynb`` file will not have
+those dependencies installed, and thus will not work without running the script.
```
lucyleeow, line 1111:
```suggestion
those dependencies installed, and thus will not work without running the bash code.
```

doc/configuration.rst
```diff
@@ -1087,6 +1088,60 @@ Binder links will point to these notebooks.
...
+
+To fix this, we can set the ``promote_jupyter_magic`` flag in ``conf.py``::
+
+  sphinx_gallery_conf = {
+      ...
+      'promote_jupyter_magic': True,
+  }
+
+If this flag is ``True``, then when a Jupyter notebook is being built, any code
+block starting with a `Jupyter cell magic <https://ipython.readthedocs.io/en/stable/interactive/magics.html>`_ (e.g. ``%%bash`` or ``%%writefile``)
```
lucyleeow, line 1121:
```suggestion
block starting with `Jupyter cell magic <https://ipython.readthedocs.io/en/stable/interactive/magics.html>`_ (e.g. ``%%bash`` or ``%%writefile``)
```

doc/configuration.rst
```diff
@@ -1087,6 +1088,60 @@ Binder links will point to these notebooks.
...
+  #%%
+  # Installing dependencies
+  #
+  #     .. code-block:: bash
+  #
+  #       %%bash
+  #       pip install -q tensorflow
+  #       apt-get -qq install curl
+
+Then, TensorFlow and Curl would be automatically installed upon running the
```
lucyleeow, line 1135:
```suggestion
meaning, TensorFlow and Curl would be automatically installed upon running the
```

doc/configuration.rst
```diff
@@ -1087,6 +1088,60 @@ Binder links will point to these notebooks.
...
+Jupyter notebook. This works for any cell magic (not just those mentioned above)
+and only affects the creation of Jupyter notebooks.
+
+.. warning::
+  It is good practice to ensure the ``.py`` and ``.html`` files match the ``.ipynb``
+  files as closely as possible. This flag (and Jupyter cell magics) should only
```
lucyleeow, line 1141:
```suggestion
  files as closely as possible. This functionality should only
```

sphinx_gallery/gen_gallery.py
```diff
@@ -79,6 +79,7 @@ def __call__(self, gallery_conf, script_vars):
     'thumbnail_size': (400, 280),  # Default CSS does 0.4 scaling (160, 112)
     'min_reported_time': 0,
     'binder': {},
+    'promote_jupyter_magic': False,
```
lucyleeow, line 82: Maybe some type checking to ensure the `promote_jupyter_magic` input is correct?

lucyleeow (approved):
Just few nits, looks good. I'd feel more comfortable if @larsoner took a look but happy to merge if not.

doc/configuration.rst
guberti: Changed to "blocks starting with Jupyter cell magics", as [the IPython docs use "magics" as the plural](https://ipython.readthedocs.io/en/stable/interactive/magics.html).

sphinx_gallery/gen_gallery.py
```diff
@@ -376,6 +376,12 @@ def call_memory(func):
         if gallery_conf['app'] is not None:  # can be None in testing
             gallery_conf['app'].add_css_file(css + '.css')
 
+    # jupyter magics
+    if not isinstance(gallery_conf['promote_jupyter_magic'], bool):
```
lucyleeow, line 380: We should probably use `_bool_eval`, see: https://github.com/sphinx-gallery/sphinx-gallery/blob/66bc0f2b617ddd5d27258cc577574f16bd72650b/sphinx_gallery/gen_gallery.py#L171

lucyleeow (approved):
LGTM! Will wait a few days to see if anyone else review and merge if not.

sphinx_gallery/gen_gallery.py
guberti: You're right, that's a cleaner approach. Fixed!

guberti:
Thanks for the review @lucyleeow! I've implemented your suggestions.

> LGTM! Will wait a few days to see if anyone else review and merge if not.

Sounds like a good plan.

Issue #361, opened 2018-03-20
ignore code-blocks


ajtritt:
Please add the ability to ignore code blocks.

I am trying out sphinx-gallery as a way of building a set of tutorials for [PyNWB](https://github.com/NeurodataWithoutBorders/pynwb) that can also serve as integration tests. Since the Python code I am writing is going to be an integration test, I would like to clean up any files that get written after running the test; however, I don't want this code to show up in the tutorial, so I would like sphinx-gallery to ignore these code-blocks. 

I've done some snooping through the sphinx-gallery documentation and code base, and I don't see anything that would suggest this functionality exists. 

I am happy to add this feature myself, but I'm not sure what your procedure around contributing is.

lesteve:
I am afraid I don't understand what you mean. Can you give us a better idea of what you have in mind when you say a tutorial that serves as integration test? How would a `.py` look like?

ajtritt:
The integration test part of this is orthogonal to the problem, you can ignore that. 

I want to be able to add a comment to a code-block that tells sphinx-gallery to ignore this code-block.

```python
"""
Header Docstring
================

This is a test
"""


import os

####################
# First RST block
# ---------------
# This is the first ReStructured Text block
#


with open("test.txt") as f:
    f.write("Hello World\n")

####################
# Final RST block
# ---------------
# This is the final ReStructured Text block
#


# IGNORE THIS FOLLOWING CODE BLOCK WHEN GENERATING RST
os.remove("test.txt")


####################
# the final ReStructured Text block picks up here.
```

ajtritt:
If it helps, I added code to accomplish this. Here's a PR I made on my own fork https://github.com/ajtritt/sphinx-gallery/pull/1

lesteve:
Thanks a lot for the details! The reason I asked about integration test is I wanted to understand better what your motivation for hiding blocks is. When we need to clean up files we tend to do it towards the end of the example with a comment. The alternative is to leave the generated file alone. Maybe you can explain why these two approaches are lacking in your particular use case?

One of the primary goal of sphinx-gallery is to make sure that the output HTML is synchronized with the python code that generated it. This is quite at odds with hiding blocks.

ajtritt:
> When we need to clean up files we tend to do it towards the end of the example with a comment. 
If the clean up is in a comment, then it will not actually remove the file when the script gets executed.

> The alternative is to leave the generated file alone. 
Sometimes I want to clean up the generated file. 

[...]

PR #690, opened 2020-05-20
ENH: Use Sphinx errors

larsoner:
Use Sphinx errors rather than built-in ones when something goes wrong. This coupled with https://github.com/sphinx-doc/sphinx/pull/7699 should make sphinx dev work again. Should probably wait for that one to be merged before merging this one, hence the WIP label.

Closes #600

PR #941, opened 2022-04-11
Add flag to ignore code blocks in Python source parser

guberti:
## What this PR does
This PR adds the flags `# sphinx_gallery_start_ignore` and `# sphinx_gallery_end_ignore`, which can be used in Python source files. For example, consider the following snippet of a source Python file:
```python
# Finding Connected Arduino Boards
# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
# Suppose we want a tutorial showing how to get a list of all connected Arduino 
# boards. We'd like it to work for the user, but we would also like people to
# be able to build the docs even if they don't have an Arduino. Using these
# flags, we can add code to mock the output of subprocess so the docs build
# correctly, all without confusing the end user.
#

import subprocess
# sphinx_gallery_start_ignore
from unittest.mock import create_autospec, MagicMock
subprocess.run = create_autospec(subprocess.run, return_value="nano33ble\n")
# sphinx_gallery_end_ignore

arduino_board = subprocess.run(
    ["arduino-cli", "board", "list], check=True, stdout=subprocess.PIPE
)
```
In this example, the code between `start_ignore` and `end_ignore` **will** be run when the `.py` file is executed. However, when an HTML page is built from the `.py` script, those flags and the lines between them will be ignored. Hence the code block as seen from the webpage would look like this:
```python
import subprocess

arduino_board = subprocess.run(
    ["arduino-cli", "board", "list], check=True, stdout=subprocess.PIPE
)
```

## Why is this useful?
Previously, there was no good way to do "mocking" in Python source files with Sphinx Gallery. For example, imagine the Python source file was a tutorial, and consider the following use cases:
- The tutorial requires some connected hardware (say, an Arduino board), but you don't want to force people to connect an Arduino to build the documentation.
- The tutorial has some non-Python component - for example, running some `bash` commands or downloading a dataset from the Internet. You *could* use Python to do these things, I suppose (use `subprocess` for `bash` commands and `requests` to download the dataset), but it would be much easier to allow the user to do them the "normal way" and mock them in the `.py` file.
- After building the docs, we want some `assert` sanity checks to make sure the tutorial ran correctly and isn't bugged. However, we don't expect the user to run these test cases, and thus should not include them in the `.html` page.

I'm part of the [Apache TVM](https://github.com/apache/tvm) community (which [uses Sphinx Gallery for tutorials](https://tvm.apache.org/docs/how_to/work_with_microtvm/index.html)), and we've run into all of the issues mentioned above. This feature is a small change that would alleviate all of these issues. We're also not the first ones to request this feature - [here's a previous issue](https://github.com/sphinx-gallery/sphinx-gallery/issues/361) and [a pull request](https://github.com/ajtritt/sphinx-galleria/pull/1), though that user fixed his use case a different way.

Yes, it's true that in general using `sphinx_gallery_start_ignore` might be a little gross, as we do want the generated HTML to match the Python source file. However, Apache TVM has a lot of use cases it would really help with, and I imagine other projects might have similar needs. 

## How is it tested?

I've added unit tests for this feature to `test_py_source_parser.py`, and integration tests to `test_gen_rst.py`. I've also verified that `flake8` lints without errors.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -67,6 +67,10 @@ Some options can be set on a per-code-block basis in a file:
 See also :ref:`removing_config_comments` to hide config comments in files from
 the rendered examples.
 
+Some options can be set on a per-line basis in a file:
+- ``# sphinx_gallery_start_ignore`` (:ref:`hiding_code_blocks`)
+- ``# sphinx_gallery_end_ignore`` (:ref:`hiding_code_blocks`)
+
 Some options can be set during the build execution step, e.g. using a Makefile:
 
 - ``make html-noplot`` (:ref:`without_execution`)
@@ -1288,6 +1292,46 @@ further deferred, if desired).  The following produces only one plot::
 
   plt.plot([2, 2])
   plt.show()
+
+.. _hiding_code_blocks:
+
+Hiding code blocks
+==================
+
+Normally, Sphinx-Gallery will render every line of Python code when building
+HTML and iPython notebooks. This is usually desirable, as we want to ensure the
+Python source files, HTML, and iPython notebooks all do the same thing.
+
+However, it is sometimes useful to have Python code that runs, but is not
+included in any user-facing documentation. For example, suppose we wanted to
+add some ``assert`` statements to verify the docs were built successfully, but
+did not want these shown to users. We could use the ``sphinx_gallery_start_ignore``
+and ``sphinx_gallery_end_ignore`` flags to achieve this::
+
+    model.compile()
+    # sphinx_gallery_start_ignore
+    assert len(model.layers) == 5
+    assert model.count_params() == 219058
+    # sphinx_gallery_end_ignore
+    model.fit()
+
+When the HTML or iPython notebooks are built, this code block will be shown as::
+
+    model.compile()
+    model.fit()
+
+The ``sphinx_gallery_start_ignore`` and ``sphinx_gallery_end_ignore`` flags may
+be used in any code block, and multiple pairs of flags may be used in the same
+block. Every start flag must always have a corresponding end flag, or an error
+will be raised during doc generation. These flags and the code between them are
+always removed, regardless of what ``remove_config_comments`` is set to.
+
+Note that any output from the ignored code will still be captured.
+
+.. warning::
+  This flag should be used sparingly, as it makes the ``.py`` source files less
+  equivalent to the generated ``.html`` and ``.ipynb`` files. It is bad practice
+  to use this when other methods that preserve this relationship are possible.
 
 .. _dummy_images:
 
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -9,6 +9,11 @@ dictionary specified in your ``conf.py`` file. A list of the possible
 keys are listed :ref:`below <list_of_options>` and explained in
 greater detail in subsequent sections.
 
+When using these flags, it is good practice to make sure the source Python files
+are equivalent to the generated HTML and iPython notebooks (i.e. make sure
+``.py == .html == .ipynb``). This principle should be violated only when
+necessary, and on a case-by-case basis.
+
 .. _list_of_options:
 
 List of config options
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -73,8 +73,7 @@ See also :ref:`removing_config_comments` to hide config comments in files from
 the rendered examples.
 
 Some options can be set on a per-line basis in a file:
-- ``# sphinx_gallery_start_ignore`` (:ref:`hiding_code_blocks`)
-- ``# sphinx_gallery_end_ignore`` (:ref:`hiding_code_blocks`)
+- ``# sphinx_gallery_start_ignore`` and ``# sphinx_gallery_end_ignore`` (:ref:`hiding_code_blocks`)
 
 Some options can be set during the build execution step, e.g. using a Makefile:
 
@@ -1300,8 +1299,8 @@ further deferred, if desired).  The following produces only one plot::
 
 .. _hiding_code_blocks:
 
-Hiding code blocks
-==================
+Hiding lines of code
+====================
 
 Normally, Sphinx-Gallery will render every line of Python code when building
 HTML and iPython notebooks. This is usually desirable, as we want to ensure the
--- a/sphinx_gallery/py_source_parser.py
+++ b/sphinx_gallery/py_source_parser.py
@@ -37,8 +37,15 @@ Example script with invalid Python syntax
 #     # sphinx_gallery_thumbnail_number = 2
 #
 #     b = 2
+FLAG_START = r"^[\ \t]*#\s*"
 INFILE_CONFIG_PATTERN = re.compile(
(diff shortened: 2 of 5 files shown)


larsoner:
> Previously, there was no good way to do "mocking" in Python source files with Sphinx Gallery.

An alternative would be to wrap anything problematic in a helper (in your codebase) like `setup_hardware(...)`. Then have something like a `MYLIB_MOCK_HARDWARE=true` during doc generation that is noticed by `setup_hardware` that turns it into a noop for example. I have used things like this before I think.

Have you considered this alternative?

lucyleeow:
I'm -1 on this too. I think you could solve these problems with existing functionality. Also having to add and remove these comments for when you do/don't want to fully build the documentation seems a bit cumbersome...

guberti:
For @larsoner's suggestion, it's really hard to create a universal hardware mock that works for all use cases. Simulating things like a microcontroller or FPGA in their entirety would work, but would be really complicated. Instead, I'd really like to be able to mock device inputs in a per-tutorial way, and I don't know of a good way to do that without this change.

I do agree with @lucyleeow though - it's *really* gross to make users remove the ignores if they want to do a "full build", and I no longer think this flag is appropriate for cases like that. I've edited my original pull request comment to reflect that. 

I still think there are still use cases, though, where a `sphinx_gallery_start_ignore` would be really helpful. The main use case is definitely small (but important) mocks that would not affect the script's output:

```python
WORK_FOLDER = "~/ai_models"
# sphinx_gallery_start_ignore
WORK_FOLDER = tempfile.mkdtemp()
# sphinx_gallery_end_ignore
```

It could also be used to automate a step where the user would be expected to provide input when the tutorial is being followed in `.ipynb` or `.html` format:
```python
# %%
# To test out the facial recognition model, save a photo of your own face to `~/ai_test/my_face.jpg`
# sphinx_gallery_start_ignore
test_photo_data = requests.get("https://i.imgur.com/Zq0iBJK.jpeg")
with f as open("~/ai_test/my_face.jpg", "wb"):
  file.write(test_photo_data.content)
# sphinx_gallery_end_ignore
```
Or a tutorial writer who wanted to make sure the tutorial ran correctly could use it to add `assert` statements to the end of their script, without confusing users. E.G:

```python
# sphinx_gallery_start_ignore
assert len(model.layers) == 5
assert model.count_params() == 219058 # Only 219,058 of these are trainable
assert os.path.isfile(FOLDER + "/models/project.zip")
# sphinx_gallery_end_ignore
```
And of course, users who do not find the feature useful could simply not use the `sphinx_gallery_start_ignore` flag.

larsoner:
Fair enough. However, I think it's worth perhaps discouraging the use of this, as we'd like to encourage SG "users" to maintain the `.py == .ipynb == html` relationship as much as possible. So how about in the docs for this option (which still need to be added) we add a `.. warning::` that this should ideally be used sparingly because it breaks this relationship?

And if the `.py == .ipynb == html` principle is not documented anywhere clearly, could you add it early in the config docs or homepage or something?

guberti:
I think a warning in the docs makes sense (per @larsoner's suggestion) makes sense, as the option should be used sparingly.

I've added documentation for this flag to `configuration.rst` and also added a line to the top of the file mentioning the `.py == .html == .ipynb` equivalence. It might be worth expanding on that principle elsewhere in the docs, but I think that's for another pull request.

larsoner:
@lucyleeow can you review and merge if happy?

doc/configuration.rst
```diff
@@ -1289,6 +1298,46 @@ further deferred, if desired).  The following produces only one plot::
   plt.plot([2, 2])
   plt.show()
 
+.. _hiding_code_blocks:
+
+Hiding code blocks
```
lucyleeow, line 1303: I wonder if 'code lines' would be better?

sphinx_gallery/tests/test_py_source_parser.py
```diff
@@ -82,3 +83,33 @@ def test_extract_file_config(content, file_conf, log_collector):
 ])
 def test_remove_config_comments(contents, result):
     assert sg.remove_config_comments(contents) == result
+
+
+def test_remove_ignore_comments():
+    print(sg.IGNORE_BLOCK_PATTERN)
```
lucyleeow, line 89: left over cruft?

lucyleeow (commented):
Looks good, just some nitpicks.

My only other comment is that currently the whole doc building will fail if there are different numbers of start and end flags. Another option is to throw a warning and not remove anything if they do not match, not sure what would be best, WDYT @larsoner ?

sphinx_gallery/py_source_parser.py
```diff
@@ -203,6 +210,26 @@ def split_code_and_text_blocks(source_file, return_node=False):
...
+
+    Parameters
+    ----------
+    code_block : str
+        A code segment.
+    """
+    num_start_flags = len(re.findall(START_IGNORE_FLAG, code_block))
+    num_end_flags = len(re.findall(END_IGNORE_FLAG, code_block))
+
+    assert num_start_flags == num_end_flags, "start/end ignore block mismatch!"
```
lucyleeow, line 229: Also, let's have a more detailed error message. Maybe "All 'sphinx_gallery_start_ignore' flags must have a matching 'sphinx_gallery_end_ignore' flag"

doc/configuration.rst
guberti: Agreed, changed to `Hiding lines of code`

doc/configuration.rst
```diff
@@ -67,6 +72,10 @@ Some options can be set on a per-code-block basis in a file:
 See also :ref:`removing_config_comments` to hide config comments in files from
 the rendered examples.
 
+Some options can be set on a per-line basis in a file:
+- ``# sphinx_gallery_start_ignore`` (:ref:`hiding_code_blocks`)
+- ``# sphinx_gallery_end_ignore`` (:ref:`hiding_code_blocks`)
```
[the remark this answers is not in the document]
guberti: You're right, that's cleaner.

sphinx_gallery/py_source_parser.py
guberti: Error message has been updated.

guberti:
Thanks for the review @lucyleeow! Your comments should be addressed by 71889e1. 

> My only other comment is that currently the whole doc building will fail if there are different numbers of start and end flags. Another option is to throw a warning and not remove anything if they do not match, not sure what would be best

IMO we should fail the build if there is a mismatch of flags. Throwing a warning and not removing anything is not what the documentation author was expecting `sphinx-gallery` to do, and almost guarantees the generated docs would look strange. Curious to hear @larsoner's opinion tho.

lucyleeow:
> IMO we should fail the build if there is a mismatch of flags

I'm okay with this, for people who do use this flag, it can fail.

I just remembered this PR: https://github.com/sphinx-gallery/sphinx-gallery/pull/690, to be consistent and have cleaner output, do you mind using a Sphinx error instead of assert? Happy to merge after

guberti:
@lucyleeow changed to use a Sphinx `ExtensionError`. Thanks!

lucyleeow:
Thanks @guberti !

PR #943, opened 2022-04-15
Fix rendering of embedded URIs in Python notebooks

guberti:
Currently, embedded URIs do not render correctly when converted to `.ipynb` notebooks. Here's one example of the `.ipynb` you can download from the [Executing a Tiny Model with TVMC Micro]:(https://tvm.apache.org/docs/how_to/work_with_microtvm/micro_tvmc.html#sphx-glr-how-to-work-with-microtvm-micro-tvmc-py)

image

The fix for this is pretty straightforward, and just involves using a Regex to find and replace these with the Markdown format of links. Should be a pretty straightforward suggestion!


larsoner:
Thanks @guberti !

guberti:
Thanks for reviewing @larsoner!

PR #944, opened 2022-04-19
Merge toctrees containing subcategories indices and examples without …

### LINKED ISSUE #935: Problem in section and example title level in subgalleries
Hello,

Since I'm using the master branch of sphinx-gallery in POT I get to test all your new features ;).

I have a problem because the examples in the subgalleries that where defined as subsubsection in the original doc:
https://pythonot.github.io/auto_examples/index.html
image
are now repeated moved at the end of the list with the wrong level
https://pythonot.github.io/master/auto_examples/index.html
image

There was no change in the subgalleries files. I'm OK with changing stuff to move back to the original behavior but I can't find the information in the doc.

I also noted that you have the same problem in the dev version of your doc:
https://sphinx-gallery.github.io/dev/auto_examples/index.html#no-image-output-examples
image
so I think it might be a bug.

alexisthual:
Closes #935.

This PR:
- modifies `generate_dir_rst()` from `gen_rst.py` such that (a) it doesn't always include the generated toctree to the generated index rst file and (b) it returns a list containing aforementioned toctree's items
- modifies `generate_gallery_rst()` from `gen_gallery.py` so that it generates only one toctree and that the latter correctly includes examples which do not belong to a subcategory
- updates tests because outputs' order changed in `generate_dir_rst()`


sphinx_gallery/gen_rst.py
```diff
@@ -440,12 +465,28 @@ def generate_dir_rst(src_dir, target_dir, gallery_conf, seen_backrefs):
...
             findex.write(subsection_index_toctree)
 
-    return subsection_index_content, costs, subsection_index_toctree, \
-        subsection_index_path
+    return (
+        subsection_index_path,
+        subsection_index_content,
+        costs,
+        subsection_toctree_filenames,
+    )
```
larsoner, line 489: Can you restore the old order? Technically the API is public

https://sphinx-gallery.github.io/stable/gen_modules/sphinx_gallery.gen_rst.html#sphinx_gallery.gen_rst.generate_dir_rst

So we should maybe change it as little as possible. If you restore the old order and args here and add a `return_toctree_filenames=False` default kwarg (which, when True, means you put `subsection_toctree_filenames` in place of or in addition to the `subsection_index_toctree` that is returned when False) then I think we stay backward compatible.

sphinx_gallery/gen_gallery.py
```diff
@@ -477,21 +477,29 @@ def generate_gallery_rst(app):
...
+            this_content,
+            this_costs,
+            this_toctree_items,
+        ) = generate_dir_rst(
             examples_dir_abs_path,
             gallery_dir_abs_path,
             gallery_conf,
-            seen_backrefs
+            seen_backrefs,
+            include_toctree=False,
```
larsoner, line 490: And then here you do `return_toctree_filenames=True` and change the items that get returned above

larsoner:
Thanks for working on this @alexisthual !

sphinx_gallery/gen_rst.py
larsoner: Oh... in that case, maybe don't bother :) Up to you. If nobody complained so far and we already broke the rule about API stability, breaking it again seems like less of a problem... I'm not sure we really explicitly guarantee a stable API anyway. I just usually try to avoid changing it if it doesn't add too much additional complexity.

alexisthual:
I tried multiple things, but it seems hard (maybe undoable with sphinx `toctree`s) to solve the underlying issue only with `sphinx-gallery` modifications.

However, adding `:titlesonly:` to the `toctree` which generates the part of the navigation bar handling the gallery _almost_ does the trick (see pictures below). From there, I only need to generate one more rst file in order to group together examples which do not belong to a subcategory. 

Before | After
--- | ---
Screenshot from 2022-04-19 18-25-15 | Screenshot from 2022-04-19 18-25-37

@larsoner @rflamary do you think it would be reasonable to ask users to:
- add the `:titlesonly:` directive to the relevant `toctree`
- split `index.rst` in two files: `index.rst` which would contain text related to the gallery description and `default.rst` which would contain text related to examples which do not belong to a subcategory (in the case of `sphinx-gallery`, this would refer to the "General Examples" section ; in the case of `POT`, it would refer to the "OT and regularized OT" section)

sphinx_gallery/gen_rst.py
alexisthual: Maybe it's silly of me, but since the docstring was so concise, I just assumed no one would be using it (I thought it's hard to use a method which is not documented haha)

alexisthual:
@rflamary and I just had a call, here are the main points we discussed:
* aforementioned tasks seem reasonable
* however, this new behaviour should be optional (and maybe the default should be the old behaviour) and well documented
* there could be an option such that subsection titles in the navbar link to anchors in the gallery index file rather than to the subsection index file

larsoner:
The plan sounds good!

> there could be an option such that subsection titles in the navigation bar link to anchors in the gallery index file rather than to the subsection index file

And I like this option. I think in MNE-Python with the latest changes at some point I clicked a subheading link and expected it to stay on the "big page" but it went to one of the subcategory pages instead... so I'd consider switching to this other option if it were available!

alexisthual:
> The plan sounds good!

Ok, this will require more work then, and I won't have a lot of free time in the coming weeks.
When is the next release planned?

larsoner:
I don't think we have a timeline, so no problems waiting for now

rflamary:
@alexisthual you add an option you double the number of tests! I don't envy you ;)

Thank you all for looking into this. I will probably do a release of POT with stable version of sphinx gallery in th meantime but will be ready for when this is implemented .

alexisthual:
Moreover, I added an option to enable/disable nested sections. By default, `nested_sections=True`.

@rflamary I just built `pot`'s latest documentation with this branch and things should now work as expected for you if you set `nested_sections=False` in `./docs/source/conf.py`:
`nested_sections=True` (default) | `nested_sections=False`
--- | ---
Screenshot from 2022-07-05 17-57-15 | Screenshot from 2022-07-05 17-57-28

larsoner:
@alexisthual this sounds good to me, and I agree that the hover behavior is better. Maybe we'll want to add an option for it someday, but let's see if people request it before bothering to add an option. If this is ready to go from you end, can you mark it ready for review? It's still in draft mode. Then I'll take a quick look at the code and hopefully merge!

alexisthual:
> Maybe we'll want to add an option for it someday, but let's see if people request it before bothering to add an option.

Are you speaking about an option to toggle the hovering behaviour? 

> If this is ready to go from you end, can you mark it ready for review? It's still in draft mode. Then I'll take a quick look at the code and hopefully merge!

I still have two more things to do:
1. add tests for when `nested_sections=False`
2. edit the changelog to document this option (some users, similarly to `pot`, will need to set `nested_sections=False` in their conf)

larsoner:
> Are you speaking about an option to toggle the hovering behaviour?

Yes. It's possible people have used custom CSS for example like:

https://github.com/mne-tools/mne-python/blob/6c43b2d398780e2e69ea2f5b2e3db6b59135825c/doc/_static/style.css#L91-L95

It would be great if stuff like this still worked. But even if it doesn't, adding something in the CHANGES.rst would be good. It sounds like this is already on your TODO list, though!

Let me know when to have a look!

rflamary:
That's great, I think that adding this option is a very good idea. Can't wait to do the next release of POT with it ;)

alexisthual:
I tried to add tests to `test_full.py` but it seems much more of a burden than I thought.
Initially, I wanted to run `test_rebuild()` two times with `nested_sections` to `True` and `False` respectively, but it seems that setting `confoverrides` when instantiating `Sphinx` objects triggers weird behaviours (such as considering that some files are out of date when they are not, etc).

Otherwise I think this PR is ready for review, and that we can discuss other interesting tests together!

larsoner:
> Initially, I wanted to run test_rebuild() two times with nested_sections to True and False respectively, but it seems that setting confoverrides when instantiating Sphinx objects triggers weird behaviours (such as considering that some files are out of date when they are not, etc).

Yeah I would expect a full rebuild actually because a config value has changed. We can deal with this later at some point

doc/conf.py
```diff
@@ -395,7 +395,8 @@ def setup(app):
     # each code block
     'capture_repr': ('_repr_html_', '__repr__'),
     'matplotlib_animations': True,
-    'image_srcset': ["2x"]
+    'image_srcset': ["2x"],
+    'nested_sections': False,
```
larsoner, line 399: Why disable this for our own doc builds? It seems like we'd want this cool new feature, no?

larsoner (commented):
Other than one last idea, LGTM +1 for merge!

doc/conf.py
alexisthual: I agree it would be nice to use it in our doc, but nested sections don't play well with gallery items present in the root folder. The main reason is that 2 there will be 2 toctrees in the main `index.rst` (one for these gallery items and one for the gallery's sections) and that they don't interact well.

I think we have 3 ways of handling this:
- disable `nested_sections` (what I chose here)
- create a new folder in the root directory and move root gallery items there (what I chose for `nilearn`) even if this breaks html links (because file paths were changed)
- have users split their root `readme.txt` file into 2 text files: one describing the gallery, and one describing the "root section" (ie the section root items belong to). They could be named `readme_gallery.txt` and `readme.txt` respectively to be consistent across sections. This way, I could have a single toctree in the main `index.rst`, and that would most likely solve our problem!

PR #951, opened 2022-04-22
syntax error fix in sphinx_gallery.downloads

photoniker:
some syntax error in the member downloads.CODE_DOWNLOAD and downloads.CODE_ZIP_DOWNLOAD which were not interpretate right in the resulting *.rst file.

### DIFF
--- a/sphinx_gallery/downloads.py
+++ b/sphinx_gallery/downloads.py
@@ -17,7 +17,7 @@ from .utils import _replace_md5
 CODE_DOWNLOAD = """
 .. _sphx_glr_download_{3}:
 
-\n.. only :: html
+\n.. only:: html
 
  .. container:: sphx-glr-footer
     :class: sphx-glr-footer-example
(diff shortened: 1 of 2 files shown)


lucyleeow:
Can you provide more details of the error you are getting?

photoniker:
Hi, I did not get directly an error. I saw that syntax error for the `.. only :: ` command where the whitespace character must deleted otherwise it is wrong in the *.rst file. 

I can show the differences based on some auto generated *.rst files by using the sublime text editor for syntax highlighting. 

- `.. only :: html`
image


- `.. only:: html`
image

lucyleeow:
I'm reluctant to change our code just based on an editors syntax highlighting, esp if the html renders fine?

photoniker:
Understandable, I had that once that then the '*.rst' was not rendered correctly. 
In 'sphinx_gallery.gen_rst.EXAMPLE_HEADER' it is also correct with ` .. only::`. 

It's just a request ;-) @

photoniker:
OK great this was my request for ;-)

larsoner:
Thanks @photoniker !

PR #952, opened 2022-04-23
Dummy PR to debug test failure

guberti:
Should not be merged - solely exists so I can find out why tests are failing on #940.


guberti:
Apparently the issue was my using backticks in docstrings, which makes sense in retrospect. Closing this PR, as the issue is solved.

PR #961, opened 2022-05-25
Fix Tensorflow/Abseil compatibility

guberti:
Currently, `sphinx-gallery` does not work with [`absl`](https://pypi.org/project/absl-py/) (see [an example of the failure here](https://ci.tlcpack.ai/blue/organizations/jenkins/tvm/detail/PR-10516/9/pipeline/#step-838-log-1604)). This is not ideal, as `absl` a popular library, and is used by other popular libraries like Tensorflow. I have a more detailed writeup for why this happens at https://github.com/apache/tvm/issues/11117, but it has to do with the fact `sphinx-gallery` is not able to fix `absl`'s pointers to its `LoggingTee` object when `restore_std` is called.

This PR fixes the issue and adds a unit test to prevent regressions in the future, allowing `sphinx-gallery` to be used with `tensorflow` without issue.

### DIFF
--- a/sphinx_gallery/tests/test_gen_rst.py
+++ b/sphinx_gallery/tests/test_gen_rst.py
@@ -9,6 +9,8 @@ from __future__ import (division, absolute_import, print_function,
 import ast
 import codecs
 import importlib
+from io import StringIO
+import logging
 import tempfile
 import re
 import os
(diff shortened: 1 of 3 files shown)


sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -726,6 +727,24 @@ def test_output_no_ansi(gallery_conf, script_vars):
     assert output_test_string.split('\n')[-1] == "0.25"
 
 
+def test_absl_logging(gallery_conf, script_vars):
+    """Test using absl logging does not throw error. This is important, as many
+    popular libraries like Tensorflow use absl for logging.
+    """
```
larsoner, line 733: It's easy enough to make the transition a tiny bit easier for other devs who might `git merge upstream/main` or so, so let's do it
```suggestion
    """
    pytest.importorskip('absl')
```

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -726,6 +727,25 @@ def test_output_no_ansi(gallery_conf, script_vars):
...
+    pytest.importorskip('absl')
+
+    gallery_conf.update(image_scrapers=())
+    compiler = codeop.Compile()
+    code = 'from absl import logging\nlogging.info("Should not crash!")'
+    output = sg.execute_code_block(
```
larsoner, line 739:
```suggestion
    sg.execute_code_block(
```

guberti:
Thanks for the review!

guberti:
Hmm, the test failures seem related to `ffmpeg`. Let me make a test PR to see if the tests pass without changes.

guberti:
See #962 - it seems like there is some issue with the `ffmpeg` install that stops the tests from passing on `master`. The test failure is unrelated to the changes here, but it is still probably smart to make sure the tests pass before merging this.

PR #966, opened 2022-06-23
Add multiline flag when searching for ignore flags

driazati:
This follows on #941, I noticed that it wouldn't detect `sphinx_gallery_end_ignore` comments in the middle of a block. I wasn't able to reproduce this in the unit tests however, but I noticed it when editing [this Python file](https://github.com/apache/tvm/blob/be8e92211e3eebb7700c915a00b43c4c7c80a818/gallery/how_to/compile_models/from_oneflow.py)


driazati:
The current regexes work fine if there is a newline before the `# sphinx_gallery_start_ignore` so this change isn't necessary

PR #969, opened 2022-07-08
Avoid matplotlib warnings in seaborn reset_module

lesteve:
Fix #968

I am going to try to reproduce the issue in the CI in my first commit and then push a commit with the fix.

Not sure how to test this in a better way, suggestions welcome!

PR #971, opened 2022-07-11
Remove test examples for seaborn warning

lesteve:
Follow-up of #969. I added test examples to trigger the warnings and test the fix, but forgot to say that they needed to be removed before the PR is merged.

This PR removes the test examples.


larsoner:
Thanks @lesteve

PR #973, opened 2022-07-13
BUG: Fix bug with clicking examples

larsoner:
cc @alexisthual none of the examples here are clickable, this seems to fix it. Do you agree it's the right fix?

https://sphinx-gallery.github.io/dev/auto_examples/index.html


larsoner:
I'll merge to get this fix in but @alexisthual feel free to open another PR with a better fix if you have one!

alexisthual:
Good catch! To me, this feels like an elegant fix :)

PR #513, opened 2019-06-25
MAINT: Remove support for old Python and Sphinx

larsoner:
Closes #405.
Closes #407.

Bumps minimums to Sphinx 1.8.3 (~1.5 years old) and Python 3.5, removing associated code. Also adds `flake8` testing.

PR #985, opened 2022-08-05
Remove old sphinx compatibility code

QuLogic:
This code was for Sphinx < 1.6, and the minimum is now 1.8.3, so this should have been dropped in #513.


larsoner:
Thanks @QuLogic !

PR #986, opened 2022-08-05
Use Mock more in tests

QuLogic:
No need to re-implement things in the standard library, meaning less code to worry about here.

Note, this is based on #985 because it removes the side-effect that the `fakesphinxapp` fixture was really for.

### DIFF
--- a/sphinx_gallery/tests/conftest.py
+++ b/sphinx_gallery/tests/conftest.py
@@ -4,11 +4,11 @@ Pytest fixtures
 """
 from __future__ import division, absolute_import, print_function
 
-import collections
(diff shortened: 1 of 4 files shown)


larsoner:
@QuLogic there are some failures and now you need a rebase/merge, let me know when it's green and I'm happy to look!

larsoner:
(And I see that there are a lot more - lines than + lines in this PR, which is great!)

QuLogic:
Ah, the `.args` and `.kwargs` aliases were only added to `call` objects in 3.8. Should be fixed now.

larsoner:
Thanks @QuLogic !

PR #989, opened 2022-08-15
DO NOT MERGE: POC for name writing

larsoner:
For @alexrockhill. On my machine running `make` in the `sphinx_gallery/tests/tinybuild` directory I get the output file `sg_ref_names.html`:

Screenshot from 2022-08-15 13-06-57

### DIFF
--- a/sphinx_gallery/tests/tinybuild/conf.py
+++ b/sphinx_gallery/tests/tinybuild/conf.py
@@ -105,3 +105,36 @@ sphinx_gallery_conf = {
 nitpicky = True
 highlight_language = 'python3'
 html_static_path = ['_static_nonstandard']
+
+
+from pathlib import Path  # noqa
+names = list()
+
+
+def append_names(app, what, name, obj, options, lines):
+    names.append(name)
+
+
+# Instead of hooking into this event, we should just touch the file in the
+# Python code
+def touch_names(app, env, added, changed, removed):
+    fname = 'sg_ref_names'
+    with open(Path(app.srcdir) / f'{fname}.rst', 'w'):
+        pass
+    return [fname]
+
+
+def write_names(app, docname, source):
+    if docname != 'sg_ref_names':
+        return
+    lines = []
+    lines.append(':orphan:\n\nNames\n=====\n')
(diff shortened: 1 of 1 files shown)


sphinx_gallery/tests/tinybuild/conf.py
```diff
@@ -105,3 +105,36 @@ def __call__(self, gallery_conf, fname):
...
+    lines.append(':orphan:\n\nNames\n=====\n')
+    lines.extend(f'- {name}\n' for name in names)
+    source[0] = '\n'.join(lines)
+
+
+def setup(app):
+    """Set up the Sphinx app."""
+    app.connect('autodoc-process-docstring', append_names)
+    app.connect('env-get-outdated', touch_names)
+    app.connect('source-read', write_names)
```
larsoner, line 140: The trick is to use `source-read` to write the RST source to be whatever you want. This event appears to happen after all `autodoc-process-docstring`s have been completed

alexrockhill: Ah, thanks so much, could not figure that out for the life of me

alexrockhill: Huh, I did it as you suggested in the comment making touching the file in the code first and the `autodoc-process-docstring` happened after the `source-read`... maybe the `env-get-outdated` is necessary... Trying that now

alexrockhill: It's the weirdest thing, when I copy-paste your code and put it in, it works and runs the `autodoc-process-docstring` first but when I do the same thing on the PR, it runs the `source-read` PR. I must be doing something wrong, I'll see if I can figure it out...

alexrockhill: Ah okay, it appears that they interleave per module so I think the solution is to do it as you've done it here at the highest level (i.e. not in `auto_examples` or `auto_tutorials`) so that it runs last (hopefully), fingers crossed that works.

alexrockhill: Eureka that worked :)

PR #993, opened 2022-08-19
BUG: Fix single column example

### LINKED ISSUE #992: Minigallery rendered in column instead of line
I have noted in the last release that the minigalleries do not render in lines but as columns:

image

From the "master" version of the doc
https://pythonot.github.io/master/quickstart.html

Before it was giving this:
image

I don't really know what to do this seems like some css stuff ?

larsoner:
Closes #992

cc @rflamary @mthrok @alexisthual I plan to merge and cut 0.11.1. I tested locally on `tinybuild` and on MNE-Python and it fixed it both places

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,16 @@
 Change Log
 ==========
 
+v0.12.0
+-------
+
+Support for Sphinx < 3 dropped in this release. Requirement is Sphinx >= 3.
+
+**Fixed bugs:**
+
+-  BUG: Fix single column example `#993 <https://github.com/sphinx-gallery/sphinx-gallery/pull/993>`__ (`larsoner <https://github.com/larsoner>`__)
+
+
 v0.11.0
 -------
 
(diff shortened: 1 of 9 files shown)


larsoner:
Our minimal requirement of sphinx 1.8.3 seems excessive. Sphinx 3 was released 2 years ago, so let's go with that at least. Even NumpyDoc nowadays requires Sphinx 4+, so I'm tempted to increase our minimum to that, but if 3 works then we can go with it

alexisthual:
Sorry I was away from keyboard for a few days, but it seems you fixed it @larsoner, thanks :smile:

rflamary:
Yes indeed big thanks to @larsoner , you and all other contributors like @alexisthual  are doing an impressive job with this and we all need you so it is greatly appreciated.

mthrok:
+1 Thanks for the quick fix. The fix is working well for us. https://pytorch.org/audio/main/pipelines.html#torchaudio.pipelines.Wav2Vec2ASRBundle

alexisthual:
> +1 Thanks for the quick fix. The fix is working well for us. https://pytorch.org/audio/main/pipelines.html#torchaudio.pipelines.Wav2Vec2ASRBundle

Ohh, your custom CSS looks sleek!

PR #997, opened 2022-08-26
[MAINT, MRG] A few small leftovers from API usage

alexrockhill:
I wrote a quick function to just basically grep the example files for matches with the functions/classes/methods. There was an issue with methods in particular not being found because they are called from a variable name e.g. (`brain = mne.viz.Brain(...)` `brain.add_skull(...)`) how the example matching is done currently. Searching for just the function/class/method e.g. `add_skull` could be a false-positive because other modules have the same functions/classes/methods. The ubiquitous `fit(...)` comes to mind as something that would cause the method in this PR real trouble with false-positives. However, if you've seen one `fit`, you've kind of seen them all so perhaps for `sg_api_usage`, we could take a less-conservative approach to match methods like in this PR. It actually gives a very rosy account of the MNE example coverage with only 176/644 API entries not in examples (relative to the 204/644 for the current method using a pretty extensive `api_usage_ignore` docstring which this method would not need). What do you think @larsoner?

[test_api_entries.pdf](https://github.com/sphinx-gallery/sphinx-gallery/files/9436151/test_api_entries.pdf)

### DIFF
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -863,6 +863,9 @@ def write_api_entry_usage(app, docname, source):
             assert entry in gallery_conf['api_entries']['function']
             return 'func'
 
+    workdirs = [os.path.join(app.builder.srcdir, examples_dir)
+                for examples_dir in gallery_conf['examples_dirs']]
+
     # find used and unused API entries
     unused_api_entries = list()
     used_api_entries = dict()
@@ -873,10 +876,10 @@ def write_api_entry_usage(app, docname, source):
         # just search for function/class/method, this may lead to a few
         # false-positives for functions/classes/methods with the same
         # name, but for this purpose, that is preferable to false-negatives
-        examples = _search_for_entry(entry, gallery_conf['examples_dirs'],
-                                     gallery_conf['gallery_dirs'])
+        examples = _search_for_entry(
+            entry, workdirs, gallery_conf['gallery_dirs'])
         if examples:
-            # check if backreferences empty
+            # get references from backreferences
             example_fname = os.path.join(
                 backreferences_dir, f'{entry}.examples.new')
             if not os.path.isfile(example_fname):  # use without new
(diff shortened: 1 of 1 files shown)


alexrockhill:
Ah, I found some leftovers when touching this as well. I assumed that ref was `sphx_glr_auto_xxx` when it can be specified by the config and I accidentally put `has_graphviz` as a condition to just write the list of all the examples and when they are used. Happy to put those in a separate PR if this one isn't going to be merged. I'm curious to see if other people think this is a good idea though.

sphinx_gallery/gen_gallery.py
```diff
@@ -843,29 +863,37 @@ def get_entry_type(entry):
...
-            backreferences_dir, f'{entry}.examples.new')
-        if not os.path.isfile(example_fname):  # use without new
-            example_fname = os.path.splitext(example_fname)[0]
-        assert os.path.isfile(example_fname)
-        if os.path.getsize(example_fname) == 0:
-            unused_api_entries.append(entry)
-        else:
+        # just search for function/class/method, this may lead to a few
+        # false-positives for functions/classes/methods with the same
+        # name, but for this purpose, that is preferable to false-negatives
```
larsoner, line 878: >  if you've seen one fit, you've kind of seen them all

I'm not sure, even in `sklearn` the `kwargs` there can vary a bit. A more drastic example of where this is problematic is `inst.plot` in MNE-Python -- we have many of them (dozens?) and if make a new class with a `.plot` method and *don't* have an example for it, we'd want to know

alexrockhill: Yeah, I suppose it's better to wait for the right way to do this by following the variable types in code execution that is much harder to do. I will change this to just the fixes and then it can be merged.

alexrockhill:
This PR 1) Fixes that the target directory was assumed to be `auto_xxx` but can be anything (in the usage graphs) 2) fixes that the text output and links will still be written if graphviz is not installed and 3) removes methods since they are not found properly in the examples (it would be nice to add this back at some point but IMO it's better not to have something with incorrect information).

larsoner:
> 3) removes methods since they are not found properly in the examples

I'm confused by this statement since there are plenty of examples linked in this minigallery for example

https://mne.tools/dev/generated/mne.io.Raw.html#mne.io.Raw.plot

Do you mean they are sometimes found properly, but other times not? This seems like a place we might want to have an option about whether or not to include them, and maybe in the future improve our linking to them. (It seems like we'd want to include these for MNE-Python probably, even if it generates some false alarms...)

alexrockhill:
Ah okay, I assumed the issue was larger in scope that it is. I'm not sure why the `mne.viz.Brain` methods aren't being found, perhaps that's an edge case and a small bug rather than a larger issue.

larsoner:
Thanks @alexrockhill !

PR #1010, opened 2022-09-22
BUG: Update for matplotlib

larsoner:
@alexisthual should fix our error. Found it by looking for the actual link to `matplotlib.pyplot` and seeing that they changed the HTML filename


alexisthual:
Great, thank you for your help! :)

PR #1012, opened 2022-09-22
FIX: role names

### LINKED ISSUE #1008: QUESTION about full role names
The regex for roles of references allows for fully written out names `:function:`, `:method:`, `:attribute:`, `:object:` along with the standard `:func:`, `:meth:`, `:attr:`, `:obj:` role names:

https://github.com/sphinx-gallery/sphinx-gallery/blob/51cd69c71c099d1b3ef50894d9817f9bba9fd3ee/sphinx_gallery/backreferences.py#L210-L216

I couldn't find any sphinx documentation describing this feature, in particular the [Cross-referencing Python objects](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-python-objects) section doesn't mention it.

The regex with the full names was added in 6a1810e38 by @choldgraf. The [documentation](https://sphinx-gallery.github.io/stable/configuration.html#add-mini-galleries-for-api-documentation) doesn't mention the full names.

So I have two questions:

1. is the use of full names documented somewhere?
2. if not, do we need the full role names here?

(the reason for my question is that I'm working on a PR to enhance the regex to allow for default roles, titles etc. and don't want to make it more complex than needed)

StefRe:
The regex in `identify_names` also covered fully spelled out role names (`object` instead of `obj` etc.) which are however not recognized by sphinx.  

The [documentation](https://sphinx-gallery.github.io/stable/configuration.html#add-mini-galleries-for-api-documentation) lists only the correct short forms.

Closes #1008.

### DIFF
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -208,10 +208,10 @@ def _get_short_module_name(module_name, obj_name):
 
 # keep in synch w/ configuration.rst "Add mini-galleries for API documentation"
 _regex = re.compile(r':(?:'
-                    r'func(?:tion)?|'
-                    r'meth(?:od)?|'
-                    r'attr(?:ibute)?|'
-                    r'obj(?:ect)?|'
+                    r'func|'
+                    r'meth|'
+                    r'attr|'
+                    r'obj|'
                     r'class):`~?(\S*)`'
                     )
 


lucyleeow (approved):
LGTM! Thanks.
CI failure is unrelated and due to: https://github.com/sphinx-doc/sphinx/issues/10850
Will give a day and merge.

lucyleeow:
Thanks @StefRe

PR #819, opened 2021-04-23
DOC Use intersphinx references in projects_list.rst

lucyleeow:
closes #755

Have only added projects already included in interpshinx list in our conf.py

PR #754, opened 2020-10-07
DOC update link in projects_list

lucyleeow:
closes #753

Ideally when you are in the dev version of the docs, this link should be https://sphinx-gallery.github.io/dev/auto_examples/index.html, but I am not sure how to make this link change depending on the doc version...any ideas @larsoner ? It's also not that important and I am sure we can leave as stable.

PR #1015, opened 2022-09-23
DOC: fix 'Who uses Sphinx-Gallery' list

StefRe:
the current [link](https://nilearn.github.io/auto_examples/index.html) in https://sphinx-gallery.github.io/stable/projects_list.html results in 404.

- prevent the project lists in README.rst and projects_list.rst from
  becoming out of sync by putting the projects list in a text file
  to include in both rst files (there were already 2 projects more in
  README.rst than in projects_list.rst)
- fix broken and some permanently redirected links reported by linkcheck

Github markup doesn't allow including files into README.rst, so we must
include from README.rst. Github doesn't process :ref: or :doc: roles
(they are shown literally) so we need to handle the Sphinx Gallery link
separately. The intersphinx links were replaced by ordinary links as
it's very unlikely that these targets would be changed without a
redirect in place.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -23,25 +23,30 @@ Who uses Sphinx-Gallery
 =======================
 
 * `Sphinx-Gallery <https://sphinx-gallery.github.io/stable/auto_examples/index.html>`_
-* `Scikit-learn <http://scikit-learn.org/dev/auto_examples/index.html>`_
-* `Nilearn <https://nilearn.github.io/auto_examples/index.html>`_
-* `MNE-python <https://www.martinos.org/mne/stable/auto_examples/index.html>`_
+
+.. projects_list_start
+
+* `Scikit-learn <http://scikit-learn.org/stable/auto_examples/index.html>`_
+* `Nilearn <https://nilearn.github.io/stable/auto_examples/index.html>`_
+* `MNE-python <https://mne.tools/stable/auto_examples/index.html>`_
 * `PyStruct <https://pystruct.github.io/auto_examples/index.html>`_
-* `GIMLi <http://www.pygimli.org/_examples_auto/index.html>`_
-* `Nestle <https://kbarbary.github.io/nestle/examples/index.html>`_
+* `GIMLi <https://www.pygimli.org/_examples_auto/index.html>`_
+* `Nestle <http://kylebarbary.com/nestle/examples/index.html>`_
 * `pyRiemann <https://pyriemann.readthedocs.io/en/latest/index.html>`_
-* `scikit-image <http://scikit-image.org/docs/dev/auto_examples/>`_
-* `Astropy <http://docs.astropy.org/en/stable/generated/examples/index.html>`_
-* `SunPy <http://docs.sunpy.org/en/stable/generated/gallery/index.html>`_
+* `scikit-image <https://scikit-image.org/docs/dev/auto_examples/>`_
+* `Astropy <https://docs.astropy.org/en/stable/generated/examples/index.html>`_
+* `SunPy <https://docs.sunpy.org/en/stable/generated/gallery/index.html>`_
 * `PySurfer <https://pysurfer.github.io/>`_
-* `Matplotlib <https://matplotlib.org/index.html>`_ `Examples <https://matplotlib.org/gallery/index.html>`_ and `Tutorials  <https://matplotlib.org/tutorials/index.html>`__
-* `PyTorch tutorials <http://pytorch.org/tutorials>`_
-* `Cartopy <http://scitools.org.uk/cartopy/docs/latest/gallery/>`_
+* `Matplotlib <https://matplotlib.org/stable/index.html>`_:
+  `Examples <https://matplotlib.org/stable/gallery/index.html>`_ and
+  `Tutorials <https://matplotlib.org/stable/tutorials/index.html>`_
+* `PyTorch tutorials <https://pytorch.org/tutorials>`_
+* `Cartopy <https://scitools.org.uk/cartopy/docs/latest/gallery/>`_
 * `PyVista <https://docs.pyvista.org/examples/>`_
-* `SimPEG <http://docs.simpeg.xyz/content/examples/>`_
+* `SimPEG <https://docs.simpeg.xyz/content/examples/>`_
 * `PlasmaPy <https://docs.plasmapy.org/en/latest/examples.html>`_
-* `Fury <http://fury.gl/latest/auto_examples/index.html>`_
-* `NetworkX <https://networkx.github.io/documentation/stable/auto_examples/index.html>`_
+* `Fury <https://fury.gl/latest/auto_examples/index.html>`_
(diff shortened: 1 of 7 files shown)


lucyleeow:
Could you update it in our [README.rst](https://github.com/sphinx-gallery/sphinx-gallery/blob/master/README.rst) too?

StefRe:
sorry for the many pushes. build_docs error is unrelated to this PR, see #1017 .

README.rst
```diff
@@ -22,35 +22,7 @@ script and puts it into an examples gallery.
...
-* `Optuna <https://optuna.readthedocs.io/en/stable/tutorial/index.html>`_
-* `Auto-sklearn <https://automl.github.io/auto-sklearn/master/examples/index.html>`_
-* `OpenML-Python <https://openml.github.io/openml-python/main/examples/index.html>`_
-* `TorchIO <https://torchio.readthedocs.io/auto_examples/index.html>`_
-* `Neuraxle <https://www.neuraxle.org/stable/examples/index.html>`_
-* `Biotite <https://www.biotite-python.org/examples/gallery/index.html>`_
-* `Apache TVM <https://tvm.apache.org/docs/tutorial/index.html>`_
-* `Tonic <https://tonic.readthedocs.io/en/latest/auto_examples/index.html>`_
-* `Radis <https://radis.readthedocs.io/en/latest/auto_examples/index.html>`_
+.. include:: doc/projects_list.txt
```
lucyleeow, line 25: Thank you! Been meaning to fix this for a while!

doc/projects_list.txt
```diff
@@ -0,0 +1,29 @@
+* :ref:`Sphinx-Gallery <examples-index>`
```
lucyleeow, line 1: Should this file extension be `.rst` since we are using rst syntax?

lucyleeow (approved):
Thank you! Just a nit

doc/projects_list.txt
StefRe: I just used the same file name (I mean the base name of the file, without the extension) `projects_list` for the include file and the corresponding docs file, that's why I used `.txt` for the former. Other solutions would be

1. rename `projects.rst` to `who_uses.rst` and then `projects_list.txt` to `projects_list.rst` to include the latter in `README.rst` and `who_uses.rst`  or   
2. put a label in the existing `projects.rst` where the list starts and include `start-after` this label

I'm open for both options - your choice.

lucyleeow: 2 seems cleaner. Thanks!

StefRe: done

StefRe:
hm, the github rst parser seems to not like the `:start-after:` and `:end-before:` options, see https://github.com/StefRe/sphinx-gallery/tree/fix/who-uses-links, or am I missing something here? Any ideas, @lucyleeow ?

StefRe:
maybe this is caused by https://github.com/github/markup#github-markup?:

> The HTML is sanitized, aggressively removing things that could harm you and your kin—such as script tags, inline-styles, and class or id attributes.

StefRe:
Changing the order of inclusion doesn't work either as github doesn't process the roles:
grafik

StefRe:
@lucyleeow so I guess I'll have to get back to the original version with the included txt file which isn't ideal but works correctly.

lucyleeow:
What about just including the whole `doc/projects_list.rst` file (including the title)? Does the title end up looking weird? Maybe we can amend the title underscore markup?

Thanks again for working on this!

StefRe:
Up to about 20 mins ago I was convinced that including the complete txt file into README.rst did work correctly but when I now try to reproduce it I can't make it work (see https://github.com/StefRe/sphinx-gallery/tree/fix/test_who-uses-links).  

It turns out that including files into rst is blocked for security reasons, see https://github.com/github/markup/issues/172#issuecomment-33241601. Including files into markdown doesn't work either (https://github.com/github/markup/issues/346) and there are obviously no plans to pursue any of this.  

I'll try and think of another solution.

StefRe:
PyPi packages also [don't process sphinx extension](https://packaging.python.org/en/latest/guides/making-a-pypi-friendly-readme/#creating-a-readme-file):
>     - reStructuredText (without Sphinx extensions)

(the entry for SG uses `:ref:` to automatically switch between dev and stable, see #754)
(the entries for Scikit-learn and matplotlib use intersphinx links introduced in #819, the advantage is not clear to me (other than maybe if they would change their docs url without a redirect, which is rather unlikely))

larsoner:
Let's just get rid of `:ref:` and use HTML links for the README

Also in some packages for RST that will only be included (not rendered alone) I've seen `name.inc` (for "include") used. It prevents Sphinx from rendering it twice -- once as an "include" in the correct spot, and once as a standalone file where you don't want it.

StefRe:
This is the best compromise I could achieve: it preserves the `:ref:` for SG from #754 but reverts the intersphinx links from #819.  
The only little blemish is the additional empty line between Sphinx Gallery and the other projects:

in README.rst:
grafik

in the docs:
grafik

README.rst
```diff
@@ -23,25 +23,30 @@ Who uses Sphinx-Gallery
 =======================
 
 * `Sphinx-Gallery <https://sphinx-gallery.github.io/stable/auto_examples/index.html>`_
-* `Scikit-learn <http://scikit-learn.org/dev/auto_examples/index.html>`_
-* `Nilearn <https://nilearn.github.io/auto_examples/index.html>`_
-* `MNE-python <https://www.martinos.org/mne/stable/auto_examples/index.html>`_
+
+.. projects_list_start
+
```
lucyleeow, line 28: Sorry to nit, does it work with this line removed?

StefRe: Sorry, but I'm not sure what you mean exactly: the label `.. projects_list_start` is required to indicate where the range of lines to include starts; it must be _below_ the entry for sphinx gallery because that is treated differently in README.rst and projects_list.rst (simple link vs :ref:).

lucyleeow: Sorry I mean no blank line between '.. projects_list_start' and the start of the bullet points like so:

```rst
.. projects_list_start
* `Scikit-learn <http://scikit-learn.org/stable/auto_examples/index.html>`_
...
```

StefRe: Oh, sorry - my bad, I didn't understand that you were referring to the empty lines. I already tried this but even without any empty lines in-between like so:
```
* `Sphinx-Gallery <https://sphinx-gallery.github.io/stable/auto_examples/index.html>`_
.. projects_list_start
* `Scikit-learn <http://scikit-learn.org/stable/auto_examples/index.html>`_
```
Sphinx (or docutils better to say) starts a new bullet list. I didn't find a way to remove the space between the first line and the rest. That's why I decided to re-insert the empty lines around `.. projects_list_start` for better readability.

lucyleeow: Ah great, thanks for checking, these little things shouldn't annoy me so much :sweat_smile:

lucyleeow:
Thanks @StefRe !

PR #1019, opened 2022-09-26
MNT: Fix required sphinx version

StefRe:
- docs don't build with sphinx 5.2.1
- fix minimum required sphinx version in README.rst acc. to a6317fe7

Closes #1017. 

Not sure if https://github.com/sphinx-gallery/sphinx-gallery/blob/master/azure-pipelines.yml and https://github.com/sphinx-gallery/sphinx-gallery/blob/master/continuous_integration/azure/install.sh need to be updated too?

PR #1021, opened 2022-09-26
MNT: Fix erroneous commit c6ed4e

StefRe:
I'm terribly sorry, but I made a mistake in  #1019: it must read _"...don't build with sphinx 5.2.0"_. Sphinx 5.2.1 fixes the issue, so in requirements it must be of course `sphinx >= 3, != 5.2.0` instead of `!= 5.2.1`.


lucyleeow:
Thanks for the fix, CI failure unrelated but should be investigated elsewhere.

Issue #1024, opened 2022-09-28
Error building matplotlib docs with default value api_usage = 'unused'


StefRe:
matplotlib's `conf.py` doesn't specify `api_usage`. The current default value of `unused` (51cd69c71) leads to the following error (shown with `SPHINXOPTS=-W --keep-going -vv`). No errors for `api_usage = False`.

```
Traceback (most recent call last):
  File "C:\programs\mambaforge\envs\mpl-dev\lib\site-packages\sphinx\events.py", line 94, in emit
    results.append(listener.handler(self.app, *args))
  File "c:\daten\python\sphinx-gallery\sphinx_gallery\gen_gallery.py", line 717, in write_api_entries
    app.config.sphinx_gallery_conf['api_entries'][what].add(name)
KeyError: 'exception'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "C:\programs\mambaforge\envs\mpl-dev\lib\site-packages\sphinx\cmd\build.py", line 279, in build_main
    app.build(args.force_all, filenames)
  File "C:\programs\mambaforge\envs\mpl-dev\lib\site-packages\sphinx\application.py", line 350, in build
    self.builder.build_update()
  File "C:\programs\mambaforge\envs\mpl-dev\lib\site-packages\sphinx\builders\__init__.py", line 302, in build_update
    self.build(to_build,
  File "C:\programs\mambaforge\envs\mpl-dev\lib\site-packages\sphinx\builders\__init__.py", line 318, in build
    updated_docnames = set(self.read())
  File "C:\programs\mambaforge\envs\mpl-dev\lib\site-packages\sphinx\builders\__init__.py", line 425, in read
    self._read_serial(docnames)
  File "C:\programs\mambaforge\envs\mpl-dev\lib\site-packages\sphinx\builders\__init__.py", line 446, in _read_serial
    self.read_doc(docname)
  File "C:\programs\mambaforge\envs\mpl-dev\lib\site-packages\sphinx\builders\__init__.py", line 502, in read_doc
    publisher.publish()
  File "C:\programs\mambaforge\envs\mpl-dev\lib\site-packages\docutils\core.py", line 217, in publish
    self.document = self.reader.read(self.source, self.parser,
  File "C:\programs\mambaforge\envs\mpl-dev\lib\site-packages\sphinx\io.py", line 104, in read
    self.parse()
[...]

PR #1033, opened 2022-10-13
[BUG, MRG] fix issue with api usage dict

alexrockhill:
Fixes https://github.com/sphinx-gallery/sphinx-gallery/issues/1024

### DIFF
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -710,10 +710,9 @@ def write_api_entries(app, what, name, obj, options, lines):
     if app.config.sphinx_gallery_conf['show_api_usage'] is False:
         return
     if 'api_entries' not in app.config.sphinx_gallery_conf:
-        app.config.sphinx_gallery_conf['api_entries'] = \
-            {entry_type: set() for entry_type in
-             ('class', 'method', 'function', 'module',
-              'property', 'attribute')}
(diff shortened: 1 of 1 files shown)


larsoner:
I'm inclined to also make this change I suggested in https://github.com/sphinx-gallery/sphinx-gallery/issues/1024#issuecomment-1261295310:

> In the meantime maybe we should change the default to make it not run at all to make it opt-in. Don't want to break people's builds...

Just in case there are other hidden bugs or slowdowns etc., make it opt-in rather than opt-out by default...

alexrockhill:
> I'm inclined to also make this change I suggested in [#1024 (comment)](https://github.com/sphinx-gallery/sphinx-gallery/issues/1024#issuecomment-1261295310):
> 
> > In the meantime maybe we should change the default to make it not run at all to make it opt-in. Don't want to break people's builds...
> 
> Just in case there are other hidden bugs or slowdowns etc., make it opt-in rather than opt-out by default...

Maybe no one will use or know about it then but that's fine, I'll make a follow-up PR as soon as I have a chance.

alexrockhill:
Ok, all fixed, should be good to merge

alexrockhill:
Ok to merge? It would be nice to have the bug closed

larsoner:
Yes, thanks for the reminder @alexrockhill !

PR #1034, opened 2022-10-13
Extend `'ignore_pattern'` to absolute paths

### LINKED ISSUE #1030: Option to not build subdirectories
I found `'filename_pattern': '^((?!more).)*$'` regex to not build `examples/more/*` (works but no clue on caveats), but if there's multiple subdirectories and multiple `'example_dirs'` it gets messy. A look through `gen_gallery.py` and `gen_rst.py` suggests no existing option for this, so I'd like to request it, alternatively the regex to achieve this (which would help to document).

OverLordGoldDragon:
Fixes #1030 by allowing regex filtering based on not just filenames but absolute paths, hence directories:

```python
'ignore_pattern': r'\{0}more\{0}'.format(os.sep),
```
will ignore `/examples/more/`. Still don't know "how to not build any subdirectories" but maybe there's regex for that.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -403,14 +403,15 @@ def generate_dir_rst(
 
     if not os.path.exists(target_dir):
         os.makedirs(target_dir)
-    # get filenames
-    listdir = [fname for fname in os.listdir(src_dir)
-               if fname.endswith('.py')]
+    # get absolute paths
+    abspaths = [os.path.normpath(os.path.join(src_dir, fname))
+                for fname in os.listdir(src_dir) if fname.endswith('.py')]
     # limit which to look at based on regex (similar to filename_pattern)
-    listdir = [fname for fname in listdir
-               if re.search(gallery_conf['ignore_pattern'],
-                            os.path.normpath(os.path.join(src_dir, fname)))
-               is None]
+    abspaths = [p for p in abspaths
+                if re.search(gallery_conf['ignore_pattern'], p) is None]
+    # get names only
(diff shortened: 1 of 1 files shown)


larsoner:
>  "how to not build any subdirectories" but maybe there's regex for that.

Yes presumably something like `f'^{os.sep}.*{os.sep}.*$`, i.e., more than one `os.sep`, or similar should do it

larsoner:
I looked at the code and it seems reasonable, but it's always risky to merge code that adds functionality but has no test. Are you up for adding one?

OverLordGoldDragon:
Unfamiliar with SG's testing and lack the time to learn, also I think this feature is a passive test on existing features in that other tests still pass despite using absolute paths. So it can exist quietly without breaking existing functionality until SG tests & docs it.

larsoner:
> So it can exist quietly without breaking existing functionality until SG tests & docs it.

Yes but without a regression test that fails on `master` but passes on this PR, a) we don't know it actually does what it says it does, and b) even if it does we have much less of a guarantee that it will continue to do so in the future (i.e., that we won't accidentally break it).

I can push a commit later this week if it would help

OverLordGoldDragon:
Guess I'm saying this PR might as well not exist as far as SG is concerned, _until_ SG acknowledges it does (that's `a)`). It strikes me as overkill to test this as it's just an intermediate essentially equivalent operation.

In fact, taking another look, it's an identical operation, I just renamed things...

Also `'ignore_pattern': r'{0}examples{0}.*{0}.*'.format(re.escape(os.sep))` works.

> I can push a commit later this week if it would help

Thanks, luckily we're good.

PR #1035, opened 2022-10-16
Fix "`subsection_index_toctree` referenced before assignment"

OverLordGoldDragon:
Produced with `'nested_sections': False`:

```
Handler <function generate_gallery_rst at 0x0000021152A7A440> for event 'builder-inited' threw an exception 
(exception: local variable 'subsection_index_toctree' referenced before assignment)
```

Unsure if this reflects intent, feel free to edit.

### DIFF
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -562,17 +562,18 @@ def generate_gallery_rst(app):
 
                 fhindex.write(subsection_index_content)
 
-                # Write subsection toctree in main file
-                # only if nested_sections is False or None
-                if (not gallery_conf["nested_sections"] and
-                        len(subsection_toctree_filenames) > 0):
-                    subsection_index_toctree = """
+                # Write subsection toctree in main file only if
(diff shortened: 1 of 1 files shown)


larsoner:
@alexisthual can you look? I think this has to do with your nesting changes

OverLordGoldDragon:
Can this merge sometime soon? I struggle remote installing from multiple commits. I do think it's a simple bug,

```python
if cond:
    var = 1
print(var)
```
is guaranteed to fail for `cond = False`. I could add `# TODO` in case more code is due for some reason.

OverLordGoldDragon:
Apparently full equivalency mattered, rewrote and added doc to reflect both conds. The failure's again checksums. Also guess I'll install from fork in meantime.

larsoner:
Agreed this makes sense. It would be great to have a regression test, but we clearly didn't hit these lines before so I don't think the onus is on you here :)

Thanks @OverLordGoldDragon !

alexisthual:
Sorry I'm a bit late in the game here, but the fix seems ok to me!

PR #1040, opened 2022-10-18
Use descriptive link text for example page header

betatim:
Instead of using "here" as link text, tell the user what will happen if they click the link.

For example in the scikit-learn documentation the header looks like this:
image

There are two things that I think this PR improves:
1. Tell the user what will happen when they click the link. When I haven't looked at the example gallery for a while and come back as a "new user" (no memory of the past) I read the header and think "ok, so clicking will start a download, I don't want that. But then how do I get to the binder? It can't be downloading the binder. Weird. Well, let's start scrolling". Instead of realising what the link actually does, which is take you to the bottom of the page.
1. it uses the link text to tell the user something, compared to "click here". I think these days people know that they should click things that look like a link, if they want to follow the link. This means we can use the link text to tell the reader something informative.

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -1128,7 +1128,7 @@ EXAMPLE_HEADER = """
     .. note::
         :class: sphx-glr-download-link-note
 
-        Click :ref:`here <sphx_glr_download_{1}>`
+        :ref:`Go to the end <sphx_glr_download_{1}>`
         to download the full example code{2}
 
 .. rst-class:: sphx-glr-example-title


larsoner:
Test failure is unrelated and due to Mayavi so I restarted it. If it continues to fail I think we can ignore it in this PR

alexisthual:
> @alexisthual you've been thinking about usability and we hide these altogether in MNE-Python so I'm happy to defer to your judgment here

I'm no UI/UX expert but I think that what @betatim did in this PR makes a lot of sense and follows today's standards when it comes to this type of links :slightly_smiling_face:

betatim:
I took a quick look at the CI failure (Main Linux conda_python38_sphinxDev) and it looks like the problem is in `sphinx_gallery/tests/test_scrapers.py::test_save_mayavi_figures Fatal Python error: Aborted` but the output doesn't help me understand why it is unhappy. Does someone have an idea what this test does/why it might fail with a change like this?

larsoner:
It's just a `mayavi` problem, restarting and will merge once green, thanks in advance @betatim !

PR #1041, opened 2022-10-21
`w` srcset support

OverLordGoldDragon:
Supports

```
:srcset: example.png, example.png 500w
```

which was being overridden as `500x` in my custom scraper. The motivation's to provide an alternative for high resolution images by using an HTML trick - setting an excessive `w` while having `max-width` (which `sphx-glr-single-img` does provide)

```
<img ... srcset="example.png, example.png 2000w", style="max-width: 200px">
```

which prevents the page from zooming and blurring the image. Said zooming occurs on my system, Windows 10, which has a 125% display zoom set via display settings.

Unsure if I'll manage to include docs & testing if that's required, but I can keep a fork.

<hr>

### HTML MRE

```html
<!DOCTYPE html>
<html>
  <body>
    <img src="example.png", srcset="example.png, example.png 2000w", style="max-width: 200px">
  </body>
</html>
```

example

^ this is blurry for me (on this very page) unless I zoom to 1/1.25 = 0.8.

### DIFF
--- a/sphinx_gallery/directives.py
+++ b/sphinx_gallery/directives.py
@@ -175,7 +175,14 @@ def _parse_srcset(st):
             srcset[0] = spl[0]
         elif len(spl) == 2:
             mult = spl[1][:-1]
-            srcset[float(mult)] = spl[0]
+            symbol = spl[1][-1]
+            if symbol == 'x':
+                srcset[float(mult)] = spl[0]
+            elif symbol == 'w':
+                srcset[mult] = spl[0]
+            else:
+                raise ValueError('`srcset` supports "x" like "2x" or "w" like '
+                                 '"500w", got "%s"' % st)
         else:
             raise ExtensionError('srcset argument "{entry}" is invalid.')
     return srcset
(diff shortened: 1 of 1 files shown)


OverLordGoldDragon:
Unclear test failures, `test_full.py` passes locally with the repository cloned, and changed code is nearly fully equivalent to unchanged unless `w` is used...

larsoner:
Locally on your branch I see:
```
$ pytest sphinx_gallery
...
FAILED sphinx_gallery/tests/test_full.py::test_negative_thumbnail_config - FileNotFoundError: [Errno 2] No such file or directory: '/private/var/folders/q3/m...
FAILED sphinx_gallery/tests/test_full.py::test_image_formats - AssertionError: /private/var/folders/q3/m4sdprkd5658j79sssw9byp00000gn/T/pytest-of-larsoner/py...
FAILED sphinx_gallery/tests/test_full.py::test_rebuild - AssertionError: assert False
```
So at least some of the failures seem legitimate

OverLordGoldDragon:
They're not reproduced on my PC:

 1. Clone repo
 2. `make html` in `tinybuild/`
 3. `pytest.main([__file__, "-s"])` added to bottom of `test_full.py`, run the file through Spyder IDE

Am I doing it wrong?

larsoner:
> Am I doing it wrong?

To run tests it's a standard `pytest` call like:
```
pytest sphinx_gallery
```
it should reproduce the failures

OverLordGoldDragon:
Ok I decided to just go with blurs, feel free to reopen if anyone wants this.

PR #1047, opened 2022-11-02
Fixed a bug where backslashes in paths could show up in RST files

ayshih:
When the gallery directory path has an OS path separator in it, the separators need to be sanitized (on Windows) to be all forward slashes before the path is inserted into RST files.  This PR fixes a missing sanitization.

For comparison, here is another place in the codebase where the sanitization is performed:
https://github.com/sphinx-gallery/sphinx-gallery/blob/15ec1f945f21a8bf38645ffe6aab6e92f2f90170/sphinx_gallery/backreferences.py#L299-L300


ayshih:
Thanks for the quick merge!

PR #1048, opened 2022-11-02
Reorder paragraphs in the minigallery documentation

ayshih:
#908 added a new paragraph (about `exclude_implicit_doc`) to the minigallery documentation, but it was placed in an odd position: the paragraph separates the paragraph introducing the `minigallery` directive from the paragraph describing an option for that directive, and is also somewhat removed from the other configuration discussion.  This PR swaps two paragraphs so that the section reads better.


larsoner:
Thanks @ayshih

PR #1052, opened 2022-11-29
MAINT: Work around IPython lexer bug

larsoner:


larsoner:
Works around https://github.com/ipython/ipython/issues/13845

PR #1053, opened 2022-11-29
BUG: Fix full check

larsoner:
Hopefully this makes it so we don't have to restart half our CI runs

I'll push half a dozen or so empty commits sequentially to help make sure I didn't miss anything else that needs to be special-cased


larsoner:
This now also restores codecov which we forgot to migrate in #779

PR #1057, opened 2022-12-20
Add support for include directive in notebook.py

guberti:
reStructuredText supports [an include directive](https://docutils.sourceforge.io/docs/ref/rst/directives.html#include) that can be used to insert reStructuredText data from another file. This behavior is also [supported by sphinx](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html). This pull request adds support for this directive to `sphinx-gallery`!

An example helps to demonstrate the syntax. Suppose `include.txt` contains the text `Data from included file!\n`. We can use it in an `rst` document like this:

```rst
Include blocks can be used without indentation. They may
contain additional directives inside themselves.

.. include:: inclusion.txt

Back in the main document.

    If an include directive is indented, this indent will
    be preserved.

    .. include:: inclusion.txt

```

The document will then be templated as follows:

```rst
Include blocks can be used without indentation. They may
contain additional directives inside themselves.

Data from included file!

Back in the main document.

    If an include directive is indented, this indent will
    be preserved.

    Data from included file!

```

@larsoner please take a look!

### DIFF
--- a/sphinx_gallery/notebook.py
+++ b/sphinx_gallery/notebook.py
@@ -118,6 +118,17 @@ def rst2md(text, gallery_conf, target_dir, heading_levels):
         Note that ``over_char`` is `None` when only underline is present.
     """
 
+    include_re = re.compile(r"^([ \t]*)\.\. include::\s*(.+)\n", flags=re.M)
+
+    def load_include(match):
+        full_path = os.path.join(target_dir, match.group(2))
+        with open(full_path) as file:
+            lines = file.read()
+        indented = textwrap.indent(lines, match.group(1))
+        return indented
+
+    text = re.sub(include_re, load_include, text)
+
     # Characters recommended for use with headings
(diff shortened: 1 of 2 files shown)


larsoner:
`.. include` supports all sorts of options about line numbers to start/end, text to use to start/end, etc. and I see these commonly used. I'm a bit worried about having a deficient implementation here, and also sad to have to reimplement them...

Have you seen if [using `pypandoc`](https://sphinx-gallery.github.io/stable/configuration.html#using-pypandoc-to-convert-rst-to-markdown) automatically takes care of this issue? If so I'd rather just add a doc note saying "use pypandoc for advanced stuff like `.. include::` directive support"

guberti:
@larsoner that's a fair point - it does seem gross to have such a half-assed implementation of include. I can't get `pypandoc` to work (though it should in theory), but monkey-patching works fine for my use case.

PR #1060, opened 2022-12-30
Fix broken links when using dirhtml builder

mgoulao:
The idea of this PR is to develop a better solution to #921.

* Remove `dirhtml` blocking and associated test
* Add condition to `directives.visit_imgsg_html` to support `dirhtml` special case

I'm unaware of any other problem related to using the `dirhtml` builder.

### DIFF
--- a/doc/getting_started.rst
+++ b/doc/getting_started.rst
@@ -27,7 +27,7 @@ help you get started.
 .. note::
    Working `sphinx
    builders <https://www.sphinx-doc.org/en/master/man/sphinx-build.html#options>`_
-   for sphinx_gallery include `html` and `latex`.
+   for sphinx_gallery include `html`, `dirhtml` and `latex`.
 
 
 .. _set_up_your_project:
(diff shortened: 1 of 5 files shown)


larsoner:
> I'm unaware of any other problem related to using the dirhtml builder.

I think in order to un-block it like this we need to test it to make sure it succeeds without error. The easiest way to add such a test would probably be to run the `tinybuild` HTML build using the `dirhtml` builder.

mgoulao:
I have added a simple `test_error_messages_dirhtml` test which passed. Is this enough? Or should I duplicate every test using the `sphinx_app` fixture?

larsoner:
> Or should I duplicate every test using the sphinx_app fixture?

No I don't think that's necessary. I'll try to push a commit to make CIs happy, at the very least CircleCI is protecting against unregistered/new users...

larsoner:
All green after a few style fixes and making the code more DRY, thanks @mgoulao !

Issue #913, opened 2022-02-08
`html-noplot` does not work with strict build options


pmeier:
The [official instructions](https://sphinx-gallery.github.io/stable/configuration.html#building-without-executing-examples) for building `html-noplot` include setting `-D plot_gallery=0 -b html $(ALLSPHINXOPTS)`. This will emit the following warning:

```
The config value `plot_gallery' has type `str', defaults to `bool'.
```

This effectively prevents us from having `-W` in `ALLSPHINXOPTS`, i.e. running in strict mode that turns warnings into errors. Given that you cannot set anything besides strings from the CLI, this is pretty limiting.

larsoner:
If you put `plot_gallery = 'True'` in your `conf.py` instead of `plot_gallery = True` in your `conf.py` it should work, as Sphinx will see it as a str var, then in SG we try to be smart about converting string values to booleans. Can you try it, and if it works, open a PR to improve the docs here?

pmeier:
That indeed works, thanks. Still, this feels kinda awkward since we didn't have `plot_gallery` config option before. I'm guessing it just assumes `plot_gallery=True` if nothing is set and then complains because the CLI option is a string. Maybe we should fix this behavior rather than documenting the workaround?

larsoner:
> Maybe we should fix this behavior rather than documenting the workaround?

That would be nice. It might just require making the default configuration `'plot_gallery': 'true'` or something. Want to take a look?

PR #1062, opened 2023-01-06
Adds `plot_gallery` as a string by default

melissawm:
Prevents warning about `plot_gallery` being a str but defaulting to bool when invoking with `-D plot_gallery=0 -b html $(ALLSPHINXOPTS)`.

Fixes gh-913

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1197,7 +1197,7 @@ a default::
 
     sphinx_gallery_conf = {
         ...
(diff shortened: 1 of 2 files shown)


larsoner:
Looks good, CIs are green, and it looks like I've been using `'plot_gallery': 'True'` in my own doc builds for a while successfully, so in it goes. Thanks @melissawm !

Issue #1069, opened 2023-01-21
Have a show_code_download that could be set to False to not have "Download Python source code"


LuisBL:

larsoner:
Closing as dup of #949

Issue #949, opened 2022-04-20
disable download buttons


photoniker:
Is it possible to config sphinx-gallery that the download button for jupyter or both are removed from the examples pages?

larsoner:
Have you tried using CSS with `visibility: hidden`? Something like this might work (I know it hides something but I can't remember what):

https://github.com/mne-tools/mne-python/blob/fd0212e5167683a4aca4cd98321af5809a06581f/doc/_static/style.css#L47-L48

photoniker:
ah I have found one possible way. I manipulate the CODE_DOWNLOAD member of sphinx_gallery.downloads. 
Maybe not the best way, but it works. 

`from sphinx_gallery import downloads

downloads.CODE_DOWNLOAD = """
.. _sphx_glr_download_{3}:

\n.. only :: html

 .. container:: sphx-glr-footer
    :class: sphx-glr-footer-example

{2}
\n  .. container:: sphx-glr-download sphx-glr-download-python

     :download:`Download Python source code: {0} <{0}>`\n"""`

GaelVaroquaux:
> ah I have found one possible way. I manipulate the CODE_DOWNLOAD member of sphinx_gallery.downloads.

You should really be using CSS for this. That's what it is made for.

photoniker:
Hi, yes thanks it works mit CSS. 

My last problem is that I build docu for html and Qt. In the Qt help the text of the download buttons are added: 
![image](https://user-images.githubusercontent.com/17592823/164480386-aa627248-ea22-4c42-ad1e-ef86b10189aa.png)

Although I set the buttons in the CSS visibility `visibility: hidden`.

photoniker:
OK yes thanks, now it work

PR #1070, opened 2023-01-21
add show_code_download conf.py option

LuisBL:
Related to https://github.com/sphinx-gallery/sphinx-gallery/issues/1069

provide a show_code_download key to conf.py, if False do not expose "Download Python source code" neither Jupyter notebook

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -59,6 +59,7 @@ file:
 - ``capture_repr`` and ``ignore_repr_types`` (:ref:`capture_repr`)
 - ``nested_sections`` (:ref:`nested_sections`)
 - ``api_usage_ignore`` (:ref:`api_usage_ignore`)
+- ``show_code_download`` (:ref:`show_code_download`)
 - ``show_api_usage`` (:ref:`show_api_usage`)
 
 Some options can also be set or overridden on a file-by-file basis:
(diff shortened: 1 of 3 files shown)


larsoner:
See #949, I think we've decided this is not necessary

SimenZhor:
@larsoner as a user I think it would be a lot more convenient to have this as an option in conf.py rather than having to make custom css rules to hide the buttons. Especially as that would eliminate the need for uploading the `_downloads` folder to my host (this PR should probably make sure that folder does not be generated in the first place).

larsoner:
@lucyleeow WDYT?

lucyleeow:
I understand it would be easier to users but I am reluctant to add to our already large list of config options (though it probably doesn't add much maintenance).
How about adding a part to our documentation about how to hide this via CSS? Maybe expanding the section 'And some things can be tweaked directly in CSS:' ([here](https://sphinx-gallery.github.io/stable/configuration.html#list-of-config-options)). I vaguely recall a few issues about other things that can be tweaked via CSS.

Borda (approved):

larsoner:
I think for simplicity we would just remove both `.py` and `.ipynb` when `show_code_download=False`

@LuisBL would you be interested in adding this behavior? If not someone else could probably take over

Issue #1063, opened 2023-01-07
Image in rst block not rendered correctly


michaelweinold:
I have the following directory structure:

```
examples/
├── image.png
└── script.py
sphinx/
├── conf.py
└── source/
    └── auto_examples
```

and the associated `sphinx-gallery` configuration:

```
sphinx_gallery_conf = {
    'examples_dirs': '../examples',
    'gallery_dirs': './source/auto_examples'
}
```

In my gallery file `scipt.py`, I would like to add a custom thumbnail and an image to an `rst` block like so:

```
"""
Docstring
=======
"""

# sphinx_gallery_thumbnail_path = '../examples/image.png'


# %%
# .. image:: ../examples/image.png
#    :alt: An example image
```

The thumbnail works fine, but the image does not get copied over to `auto_examples` and thus fails to render:

<img width="161" alt="Screenshot 2023-01-07 at 10 37 06" src="https://user-images.githubusercontent.com/23102087/211144059-496dfe0c-c341-4a5d-8d17-7bc945df6b4d.png">

michaelweinold:
Unfortunately my little function for copying images over did not work. It caught only [files starting with `sphx_glr`](https://sphinx-gallery.github.io/stable/advanced.html#id5):

```
def copy_complete_content_of_directory(
    path_dir_src: str = examples_dirs,
    path_dir_dst: str = gallery_dirs
) -> None:
    # remove destination directory if it already exists
    if os.path.exists(path_dir_dst):
        shutil.rmtree(path_dir_dst)
    # create destination directory
    os.makedirs(path_dir_dst)
    # copy files
    for item in os.scandir(path_dir_src):
        if item.is_dir():
            src = item.path
            dst = os.path.join(path_dir_dst, item.name)
            shutil.copytree(src, dst)
```

I then tried to use an approach suggested in the documentation: ["Example 2: detecting image files on disk"](https://sphinx-gallery.github.io/stable/advanced.html#example-2-detecting-image-files-on-disk) that includes some code snippets for building this.

However, this example is currently broken (`from sphinx_gallery.gen_rst import figure_rst` won't work, this likely needs to be `from sphinx_gallery.scrapers import figure_rst`).

The only working implementation I found on GitHub is an adaption used for this wiki page here: https://ricosjp.github.io/siml/examples/00_basic.html#sphx-glr-examples-00-basic-py
Unfortuantely, I was unable to reproduce the image rendering and am still stuck.

[...]

PR #1071, opened 2023-01-23
ENH: allow rst files to pass through

### LINKED ISSUE #1067: Passthrough bare rst, and possibly index.rst
Over in Matplotlib we are trying to improve our narrative user docs.  These have traditionally been written in rst, but we have a bunch of tutorials written as sphinx-galleries that we would like to move into our narrative docs.   

First, it would be quite nice if a gallery that had a bare rst in it would pass this through to the doc build.  So if we have `/gallery/subdir/my_interesting_doc.rst` it would just be copied over to `/doc/gallery/subdir/my_interesting_doc.rst` allowing sphinxs to render it there.  Perhaps this is already available somehow, but I didn't see the ability.

Second, it would be nice if _we_ provided `/gallery/subdir/index.rst` that sphinx-gallery didn't bother making its own.  Yes, appreciated that thumbnails etc would not be in the built index page, but those are not always the most useful in narrative docs where you just want the outline rather than a bunch of snapshots.

Is this already all possible and I'm missing something?  If not, it seems relatively straightforward (at least conceptually, I haven't dug into the code), and non-breaking in that folks who don't want this don't have to do it.   Thanks!

jklymak:
Closes #1067.  

Ready for review 27 Jan...

This allows *.rst files in galleries to pass through as rst.  So in `/examples_with_rst/` `rst_example1.rst` is passed though to the build directory.  Note that the `toctree` does _not_ get an entry for this, so the user will need to add a `toctree` to their `README.txt`

Secondly, it allows `index.rst` to be in a gallery, and then generating the `index.rst` is not done.  Obviously the user is responsible for creating the `toctree` by hand for this.  

Again, the motivation for this is from Matplotlib, where we really like to write long tutorials with lots of plots as python files, but other narrative documentation pages are more natural as rst.  Similarly, we are going to want manual control of our table of contents, so we would like to be able to write our own `index.rst`.  

BTW, I tested that this works well with `nested_sections=True` as well (which is responsible for some of the code complexity)

Open to suggestions as to how to better structure this.  I think I've put things at the natural place to inject the passthroughs, but it does lead to a couple large if/else blocks that could probably be optimized.


jklymak:
Hmm, this is close.  It works fine when invoked from `make html` in `doc/` but not when run as `python setup.py build_sphinx -nW --keep-going` from the project root.  I must have something broken in specifying the root.

jklymak:
OK, I think the CI failures are expected.  

Before I do a bunch of code cleanup, and write more docs (other than the self-documenting of the examples):

1. Does the user API/behaviour for this make sense?  In particular:  If the user adds an rst file it is put it somewhere in the TOC, versus the gallery adding it to the hidden TOC and perhaps using a no-image thumbnail.  The implemented behaviour is my preference, but I don't think it's too difficult to add a thumbnail.
2. Does the general code flow make sense?  I found the generation of the gallery, thumbnails, and index pages pretty intertwined with each other, so I may not be doing the alternate routes as cleanly as possible.  

Thanks!

larsoner:
So to me this PR is actually two *separable* things:

1. Allow `index.rst` as a pass-through/replacement of `README.txt`
2. Allow `*.rst` to be copied over

For (1), I like the current design -- it seems simple enough and I don't think it will break things for people. At least I don't think we allow the source and generated dirs to be the same -- if we did, that's a problem for `README.txt`->`index.rst` generation, as running `make html` once will create `index.rst` in the source dir in that case, thereby making it no longer be regenerated, which is bad.

For (2), I think we might want to think about a proper API. For example, people have asked about being able to copy other stuff (e.g., images) as well (e.g., https://github.com/sphinx-gallery/sphinx-gallery/issues/1063#issuecomment-1375786052) but my thought so far was "just do this yourself if you want it". But now that it's been brought up at least a couple of times, something like `'copyfile_regex: ''` where you can give a non-empty regex of files to copy seems like it would solve both your use case and more general ones (e.g., images within the examples dir) as you could do `'copyfile_regex': r'.*\.rst'`.

jklymak:
> For (2), I think we might want to think about a proper API. For example, people have asked about being able to copy other stuff (e.g., images) as well (e.g., [#1063 (comment)](https://github.com/sphinx-gallery/sphinx-gallery/issues/1063#issuecomment-1375786052)) but my thought so far was "just do this yourself if you want it". But now that it's been brought up at least a couple of times, something like `'copyfile_regex: ''` where you can give a non-empty regex of files to copy seems like it would solve both your use case and more general ones (e.g., images within the examples dir) as you could do `'copyfile_regex': r'.*\.rst'`.

Hmm, that is probably easy, and may actually simplify my exploratory code above now that I think of it that way.  But even simpler, is there any reason to not copy _everything_ in the gallery source directory over to the target directory?  Maybe excluding README.*, particularly README.rst?

larsoner:
Yes, because examples can produce/write junk files, people can have crufty .orig files on their systems, etc. What you're talking about would be the equivalent of something like `'.*'` and I don't think it's a good/safe/backward compatible default

jklymak:
OK, fair.  Adding a `copyfile_regexp` isn't that hard.

jklymak:
Assuming this is working the way folks are happy with, what is the advice on how to document/test.  

Currently I've just added two new galleries: `/examples_with_rst` and `/examples_rst_index`.   I didn't roll either of these into the main examples because you build with `nested_sections = False`, so I'd have to put the `:toctree` for the `rst` file at the top level.  And of course it would not test `/examples_rst_index` at all, since there is no subsections by index levels if `nested_sections=False`.  

I think there are two options - the first is keep `nested_sections=False` and add the two new galleries (probably could get away with one, but it's a little clearer as two).  The other option is to actually change `nested_sections=True` which I believe means a little rewriting of the main example Readme, but nothing too major.

timhoffm:
API suggestion:

1. Use wildcard expressions [fnmatch](https://docs.python.org/3/library/fnmatch.html) instead of regular expressions. These are easier to read and write and very common for files.
2. Use a list instead of a single item.

This reads for example
```
copy_files = ['document.rst', '*.png']
```
instead of
```
copyfile_regex = r'(document\.rst|.*\.png)'
```

3. (optionally) If you really think regex capability is needed you can support passing a compiled regex: `copy_files = ['simple.rst', 're.compile(r'\d{2}\.png')]`. But that could always be added later if there is user request.

larsoner:
Looking through our existing [config values](https://sphinx-gallery.github.io/stable/configuration.html) there are several that use regex and none that use glob/fnmatch, so we should probably stick with regex for within-package consistency

sphinx_gallery/tests/test_full.py
```diff
@@ -26,11 +27,11 @@
...
 import pytest
 
-N_TOT = 13  # examples (plot_*.py in examples/**)
+N_TOT = 18  # examples (plot_*.py in examples/**)
 
 N_FAILING = 2
 N_GOOD = N_TOT - N_FAILING
-N_RST = 17 + N_TOT + 1  # includes module API pages, etc.
+N_RST = 17 + N_TOT + 1 + 12  # includes module API pages, etc.
 N_RST = '(%s|%s)' % (N_RST, N_RST - 1)  # AppVeyor weirdness
```
jklymak, line 35: I'll admit these numbers are completely empirical to me.  I don't understand what is included here and what should not be.  I added files to `tinybuild` and obviously they changed the totals, but I'm not clear where +12 comes from exactly, or how it relates to +5 in N_TOT.  Perhaps whoever architected these tests can tell me how to make this deterministic, and I can try.  (Note I also got screwed up for quite a while as I think you cannot run these tests standalone - you need to run the whole file?)

larsoner: It's likely not up to date anymore, and I agree it's not clear (even though I think I came up with it!). I think we probably want something like
```
N_EXAMPLES = ...  # examples/*.py files that get turned to RST
N_FAILING = 2
N_GOOD = N_EXAMPLES - N_FAILING
N_INDEX = ...  # index.rst files that SG generates
N_OTHER = ...  # other pages (API/modules, etc.)
N_RST = N_EXAMPLES + N_INDEX + N_OTHER
N_RST  = '(%s|%s)' % (N_RST, N_RST - 1)  # AppVeyor weirdness
```
or so, would that make sense to you? It might be a good idea to do this in `master` first to make sure it's working properly, though... I can do this sometime this week or next unless you want to try @jklymak

jklymak: Sure let me try as I looked at it all recently.  I tried doing what you have here but failed, but starting from master makes sense and then after that I can add one file at a time and get it to make sense.

jklymak:
.. this is rebased on #1072

sphinx_gallery/gen_gallery.py
```diff
@@ -505,81 +505,84 @@ def generate_gallery_rst(app):
...
-        # to generate_dir_rst)
-        index_rst_new = os.path.join(gallery_dir_abs_path, 'index.rst.new')
-        with codecs.open(index_rst_new, 'w', encoding='utf-8') as fhindex:
-            # :orphan: to suppress "not included in TOCTREE" sphinx warnings
-            fhindex.write(":orphan:\n\n" + this_content)
-
-            # Write toctree with gallery items from gallery root folder
-            if len(this_toctree_items) > 0:
-                this_toctree = """
+        if this_content:
```
jklymak, line 508: Everything from here to the `else` is just an indentation change of the original code...

sphinx_gallery/gen_gallery.py
```diff
@@ -588,20 +591,42 @@ def generate_gallery_rst(app):
...
+                    )
+                    fhindex.write(download_fhindex)
 
-            if (app.config.sphinx_gallery_conf['show_signature']):
-                fhindex.write(SPHX_GLR_SIG)
+                if (app.config.sphinx_gallery_conf['show_signature']):
+                    fhindex.write(SPHX_GLR_SIG)
+
+            _replace_md5(index_rst_new, mode='t')
+        else:
```
jklymak, line 608: The change is from here, down...

sphinx_gallery/gen_gallery.py
larsoner: Github "Hide whitespace " viewing mode agrees :)

sphinx_gallery/gen_gallery.py
```diff
@@ -104,6 +104,7 @@ def __call__(self, gallery_conf, script_vars):
     'prefer_full_module': [],
     'api_usage_ignore': '.*__.*__',
     'show_api_usage': False,
+    'copyfile_regex': None,
```
larsoner, line 107: Perhaps instead:
```suggestion
    'copyfile_regex': '',
```
I think it's more consistent with our "string but don't use" of `ignore_repr_types` at least. There might be some advantage to having it always be a string in terms of sphinx detecting value types, too

larsoner (commented):
Just a couple of minor things, otherwise LGTM!

@lucyleeow would you be able to look in the next week or so? It would be good to have another set of eyes on this. But if not, I can take one last pass after @jklymak addresses my comments and merge

jklymak:
Thanks for the review @larsoner 

One thing occurs to me is that the index.rst passthrough happens regardless of configuration setting.  Did you want to protect users from having an index.rst and a README.txt?  Or even force them to opt in with a config setting?

larsoner:
> One thing occurs to me is that the index.rst passthrough happens regardless of configuration setting. Did you want to protect users from having an index.rst and a README.txt? Or even force them to opt in with a config setting?

While I think the risk of someone errantly having an `index.rst` is low, it might be worth protecting them.

But rather than a new config value, any reason this couldn't be handled by `r'index\.rst'` being in the list of copied files, i.e., reuse the existing config key? In the function where you currently `op.isfile(index_rst_path)` would become `if op.isfile(index_rst_path) and regex_match(...)` or whatever. Then `index.rst` is then just a special case of of the copy regex, and we document why it's a particularly useful one.

jklymak:
popping to draft to think about the above.  I think its pretty straightforward...

jklymak:
Commit 2 (and 3) are a refactor of how I ignore the README.txt logic if the index.rst exists.  I think I prefer the new way, but it's a bit more intertwined with the README.txt logic.  The advantage is less duplication.  Disadvantage is that the index.rst.new gets made even though it will never get used, which is a bit confusing.

larsoner:
> Disadvantage is that the index.rst.new gets made even though it will never get used, which is a bit confusing.

Sounds worth it given the advantages. And it should be plenty fast anyway!

doc/configuration.rst
```diff
@@ -1865,6 +1866,44 @@ each subsection, and a specific toctree for each subsection.
...
+responsibility to ensure that they are in a sphinx ``toctree`` somewhere
+in your document.  You can, of course, add a ``toctree`` to your
+``README.txt``.
+
+Manually passing ``index.rst``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you want control over the gallery (or subgallery) ``index.rst``, you
+can bypass Sphinx-Gallery creating it from ``README.txt`` by including
+it in the gallery-source instead of the README.  If you do this, you are
```
larsoner, line 1903:
```suggestion
it in the gallery-source instead of the README and setting the `copyfile_regex` to a value that causes it to be copied, like ``r'index\.rst'`` or something more general.  If you do this, you are
```
Right?

larsoner (commented):
Just one minor adjustment to the docs (assuming my understanding is correct), otherwise LGTM!

@timhoffm can you see if you're happy with the API as well?

https://output.circle-artifacts.com/output/job/61f7264b-dd38-4f20-b705-51deb7651ad3/artifacts/0/rtd_html/configuration.html#manual-passthrough

jklymak:
> Sounds worth it given the advantages. And it should be plenty fast anyway!

I think so, as the loop I was bailing out of also did a number of other things, like sg_execution_times, and zipped files.

doc/configuration.rst
jklymak: If I read something twice, I need to rewrite it ;-)

```
You can bypass Sphinx-Gallery automatically creating an index ``index.rst``
from a ``README.txt`` in a gallery directory or subdirectory. If your
``copyfile_regex`` includes ``index.rst``, and you have an ``index.rst`` in the
gallery-source instead of the README, Sphinx-Gallery will use that instead of
the index it automatically makes.  If you do this, you are responsible for
adding your own Sphinx ``toctree`` in that index (or elsewhere in your Sphinx
documentation) that includes any gallery items or other files in that
directory.
```

larsoner: Yeah my version was not great, just wanted to get the idea across :)

jklymak:
@timhoffm has limited time currently due to family obligations.  But happy to wait for him or other folks.  I think the API here is relatively opt-in, so it should not affect anyone who doesn't want this pass-through to happen.

larsoner:
Let's try it -- thanks @jklymak !

jklymak:
Thanks for your help and work maintaining sphinx-gallery!

GaelVaroquaux:
Thank you so much @jklymak. There's nothing better than heavy users (matplotlib) helping to shape the library.

PR #1072, opened 2023-01-29
TST: gallery inventory/re-structure tinybuild

jklymak:
The `test_full.py` test attempts to do an inventory of files.  Where some of the numbers comes from is a little mysterious.  And frankly I could not find two of the files that are noted as needing updating according to sphinx.  🤷 Any help there appreciated.  

Edit: Note that none of the numbers are mysterious now with the re-structure of tinybuild ;-)

However, this should give a more clear baseline for #1071 where I couldn't make heads or tails of what number to put in there when I changed what was in tinybuild.


sphinx_gallery/tests/test_full.py
```diff
@@ -26,11 +26,13 @@
...
-
+# file inventory for tinybuild:
+N_EXAMPLES = 13  # total number of plot_*.py files in tinybuild/examples
 N_FAILING = 2
-N_GOOD = N_TOT - N_FAILING
-N_RST = 17 + N_TOT + 1  # includes module API pages, etc.
+N_GOOD = N_EXAMPLES - N_FAILING  # galleries that run w/o error
+N_INDEX = 2 + 2 + 1  # indices SG generates + execution times + local_module
+N_OTHER = 9 + 2 + 2  # gen_modules + mainindex.rst + minigallery.rst + 2??
+N_RST = N_EXAMPLES + N_INDEX + N_OTHER  # includes module API pages, etc.
```
jklymak, line 35: This looks to work, but I don't know why I need the +2 here.  sphinx is mysterious!

jklymak:
Actually, doing this brings up a possible bug, or at least a behaviour I wouldn't expect.  Or maybe its just how `tinybuild/` is organized:

`examples/future/READEME.rst` gets used to make `auto_examples/future/index.rst` which is good.  

However, when sphinx builds, it finds `examples/future/README.rst` and decides that it is a valid part of the build and makes `_build/html/examples/future/READEME.html` (as well as `_build/html/auto_examples/future/index.html`).  I guess this is just a side-effect of having `tinybuild/` build `auto_examples` next to `examples`?   I do wonder if `tinybuild` should be structured "properly" where the rst from the example files is a level above (eg up in `tests`) or `tinybuild` gets another level of subdirectories?

jklymak:
The second commit re-arranges tinybuild to just contain `tinybuild/doc` and `tinybuild/examples`, which I think is the usual way of arranging galleries and stops rst files from being generated twice.

jklymak:
BTW `mayavi` is failing to install `Could not import backend for traitsui` which is causing CI failures...

larsoner (approved):
I'll try to fix CIs separately then merge

Thanks for cleaning up my mess / unclear code here @jklymak , I think this was mostly my bad doc/organization!

jklymak:
@larsoner - not too messy at all, just when I added new galleries I couldn't get the numbers to line up.  I understand now where they were coming from (copies in both `examples_with_rst/` and `auto_examples_with_rst`) but perhaps more straightforward to set up like we may expect most projects to set up their galleries.

larsoner:
Okay failures look real now

jklymak:
Note I left as two commits, but feel free to squash....

larsoner:
Thanks @jklymak !

PR #1074, opened 2023-01-31
MAINT: Fix CIs

### LINKED ISSUE #1073: CI: `mayavi` is failing to install
`mayavi` is failing to install "Could not import backend for traitsui", which is causing CI failures...

_Originally posted by @jklymak in https://github.com/sphinx-gallery/sphinx-gallery/issues/1072#issuecomment-1407787519_

larsoner:
Closes #1073

Wild guess at the problem...


jklymak:
Saying please worked.  However, I have to say that full-on mayavi with GUI toolkits etc is a pretty brutal test dependency to carry around.   Is there something special SG does with mayavi plots that you could break?

larsoner:
We have a built in Mayavi scraper. However, given the limited maintenance Mayavi gets, I think we should really deprecate it and tell people how to make their own

PR #1077, opened 2023-02-14
FIX: indentation fix

jklymak:
`this_toctree` is not defined if the conditional is false, so don't add it to the index!  Follow up to #1071

### DIFF
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -462,8 +462,7 @@ def _format_toctree(items, includehidden=False):
 
     st = """
 .. toctree::
-   :hidden:
-"""
+   :hidden:"""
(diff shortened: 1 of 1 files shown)


sphinx_gallery/gen_gallery.py
```diff
@@ -462,8 +462,7 @@ def _format_toctree(items, includehidden=False):
 
     st = """
 .. toctree::
-   :hidden:
-"""
+   :hidden:"""
```
jklymak, line 465: This included an extra carriage return, which makes build errors....

larsoner:
Thanks for the quick fix @jklymak !

jklymak:
Yes well, I should have tested on Matplotlib too ;-).

PR #1083, opened 2023-02-24
API: Deprecate mayavi scraper

### LINKED ISSUE #1075: MAINT: Deprecate Mayavi scraper
We have a built in Mayavi scraper. However, given the limited maintenance Mayavi gets, I think we should really deprecate it and tell people how to make their own

_Originally posted by @larsoner in https://github.com/sphinx-gallery/sphinx-gallery/issues/1074#issuecomment-1410931810_

This is motivated by the fact that many of our CI troubles come from Mayavi

larsoner:
Closes #1075

1. Add a note in the documentation about how people can add their own mayavi scraper (copy-pasted version of our scraper, basically)
2. Add a warning in the Mayavi examples README about deprecation
3. Add a FutureWarning if they try to call `mayavi_scraper` themselves (rare probably, but why not)
4. Add a Sphinx warning on config resolution when `'mayavi'` is in the list of scrapers
5. Internally use `_MayaviScraper()` to avoid the deprecation warnings in our own builds

I think we should merge this and cut 0.12.0 (with a warning about the API change). Then the next PR after we change the version number to 0.13.dev0 can actually remove all Mayavi-related code.

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -326,6 +326,8 @@ output formats. To use SVG, you can do::
 You can also use different formats on a per-image basis, but this requires
 writing a customized scraper class or function.
 
+.. _mayavi_scraper:
+
 Example 4: Mayavi scraper
 -------------------------
 Historically, sphinx-gallery supported scraping Mayavi figures as well as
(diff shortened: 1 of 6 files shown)


agramfort (approved):
thx @larsoner 🙏

PR #1086, opened 2023-03-06
MNT: fix subfolder README detection

jklymak:
In #1072, if the top level of a gallery had an index.rst instead of a README.txt, any subfolders _also_ had to have an index.rst instead of a README.rst.  The change here checks subdirectories as well for the presence of a README and allows subfolders to have a different README/index choice than the parent.

### DIFF
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -585,6 +585,9 @@ def generate_gallery_rst(app):
                 ] + [''])
 
                 indexst += subsection_index_content
+                has_readme_subsection = True
+            else:
+                has_readme_subsection = False
 
             # Write subsection toctree in main file only if
             # nested_sections is False or None, and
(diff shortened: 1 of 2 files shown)


larsoner:
Worth adding something to tinybuild to make sure we don't break this in the future?

jklymak:
Yeah, interestingly, the tinybuild did fail to build https://github.com/sphinx-gallery/sphinx-gallery/tree/master/sphinx_gallery/tests/tinybuild/examples_rst_index. So I guess there is something wrong with the tests that test if all the files are built.  I can work on that over the next day or two

jklymak:
OK, I added a test - the execution checks before were passing even though the README.txt was not being translated all the way to an index.html.  This test just hard-tests for the index.html; it fails on master and should pass here.

larsoner:
Thanks @jklymak !

jklymak:
Thank for the review!

PR #1097, opened 2023-03-10
FIX: Only ANSI sanitize non-HTML output

sdhiscocks:
This resolves an issue with HTML being ANSI sanitized, which can cause issues.

For example, this breaks Plotly when rendering with plotly.min.js embedded, due to ANSI escape in JS code, causing plots to not be displayed.

Relates to original #838


larsoner:
- https://github.com/sphinx-gallery/sphinx-gallery/releases/tag/v0.12.2
- https://pypi.org/project/sphinx-gallery/0.12.2/

Thanks @sdhiscocks !

PR #1099, opened 2023-03-13
Remove default '%matplotlib inline' line

### LINKED ISSUE #1091: Remove "%matplotlib inline" line by default
Now that we have jupyterlite included, we should remove the default "%matplotlib inline", as it breaks under jupyterlite and it is no longer necessary under jupyter.

ArturoAmorQ:
Fixes #1091.

> Now that we have jupyterlite included, we should remove the default "%matplotlib inline", as it breaks under jupyterlite and it is no longer necessary under jupyter.

My approach with the test was simply to remove the assertion, but I am sure there might be a better way to do it. Any feedback is welcomed.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -745,24 +745,19 @@ Add your own first and last notebook cell
 Sphinx-Gallery allows you to add your own first and/or last cell to *every*
 generated notebook. Adding a first cell can be useful for including code that
 is required to run properly in the notebook, but not in a ``.py`` file. By
-default, the following first cell is added to each notebook:
-
-.. code-block:: ipython
-
-    %matplotlib inline
+default, no first cell is added.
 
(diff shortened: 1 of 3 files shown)


larsoner:
@GaelVaroquaux @ArturoAmorQ I thought this would make our downloadable notebooks no longer behave the same for end-users when they download and run code locally, but local testing suggests that the plot does still end up inline. So this is okay by me!

larsoner:
... and as such I consider this a sort of bugfix

GaelVaroquaux:
> but local testing suggests that the plot does still end up inline.

Yes, the ecosystem improved and things now work out much better (it's nice, isn't it).

lesteve:
Very nice, I meant to do that as the first follow-up to the Jupyterlite integration #977!

PR #1104, opened 2023-03-16
Add test for setting a non-agg Matplotlib backend

ayshih:
This PR adds a regression test for #1102.  If the Matplotlib backend is reset to `agg` between code blocks, the example will raise an `AssertionError` during the build of `tinybuild` in the test suite.

### DIFF
--- a/sphinx_gallery/tests/tinybuild/examples/plot_matplotlib_backend.py
+++ b/sphinx_gallery/tests/tinybuild/examples/plot_matplotlib_backend.py
@@ -1,0 +1,24 @@
+"""
+Setting the Matplotlib backend
+==============================
+"""
+
+# %%
+# The Matplotlib backend should start as `agg`
+
+import matplotlib
+print(f"Matplotlib backend is {matplotlib.get_backend()}")
+assert matplotlib.get_backend() == "agg"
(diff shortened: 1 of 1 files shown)


larsoner:
I think you need to adjust the number of examples expected in `tinybuild` / `sphinx_gallery/tests/test_full.py`

ayshih:
Oops, number now updated

larsoner:
So far so good! Based on https://output.circle-artifacts.com/output/job/0dc682b1-1c23-4719-81d6-6e26cec1a970/artifacts/0/tiny_html/auto_examples/plot_matplotlib_backend.html it seems like a good test would be to check in `test_full.py` something like `assert html_output.count('Matplotlib backend is svg') == 2, no?

ayshih:
Sure, I can check the example print output in a unit test, but I should presumably then remove the assert statements in the example itself.  If there is a regression, those assert statements would cause to the build to fail before the unit test even has a chance to run.

larsoner:
Ahh you're right, somehow I missed the `assert` statements. Good enough, thanks @ayshih !

ayshih:
Okay, great!

PR #1105, opened 2023-03-17
Fix JupyterLite URL with nested gallery folders

lesteve:
I realised this when trying to use the feature on scikit-learn where there are nested folders 

You can also see it in sphinx-gallery, go to https://sphinx-gallery.github.io/stable/auto_examples/no_output/just_code.html, click on JupyterLite button, you will get a HTTP 404.

~Let me know if you want me to add a test for this!~ This was easy to modify an existing test so I did it!


larsoner:
Thanks @lesteve !

PR #1111, opened 2023-03-21
ENH: Add support for WebP

### LINKED ISSUE #1055: Add WEBP to _KNOWN_IMG_EXTS
I was wondering if we could add "webp" to the list of known image extensions that scrapers may produce. The change would be a one-liner in:

https://github.com/sphinx-gallery/sphinx-gallery/blob/e2ecb35d9049d0df0a1db4f6d15be817745cd565/sphinx_gallery/scrapers.py#L339

which would have to be changed to:

```python
_KNOWN_IMG_EXTS = ('png', 'svg', 'jpg', 'gif', 'webp')
```

--- 

The background for this is that I have recently built a custom scraper for the gallery of a rendering engine. This works great, but now I would like to add animations to some of the examples. Technically, GIF could work but the limitation of 256 colors really shows when you have complex shadows or gradients and that may send the wrong message with respect to the project's rendering capabilities. (It's wgpu/vulkan based, so it has all the bells and whistles and 265 colors don't do this justice.)

[WEBP ](https://developers.google.com/speed/webp) would be a nice alternative here, because it, too, has wide browser support, better compression than GIF and - most importantly - uses YUV420 instead of 8-bit palette RGB as a colorspace. The only thing "preventing" me from using it is sphinx-gallery, which currently hardcodes supported file extensions...

As a current workaround I am monkey-patching `_KNOWN_IMG_EXTS` in the `_get_sg_image_scraper` like so:

```python
def _get_sg_image_scraper():
    # hack webp as supported extension
    import sphinx_gallery.scrapers
    sphinx_gallery.scrapers._KNOWN_IMG_EXTS += ("webp",)

    ...
```

This works, but it would be much nicer if I wouldn't have to rely on a hack that modifies private variables from an upstream project.

StefRe:
and update docs on supported formats.

[WebP](https://en.wikipedia.org/wiki/WebP) yields much smaller file sizes than optimized png in only very little more time. For the Matplotlib docs with 2637 `sphx-glr...` files I got the following results:

|                                          | Total size in MB | Percentage of orginal size | Extra time|
|---------------------------- |:-----------------:|:---------------------------:|------------|
plain png                           | 162.4 | 100 % | -  |
optipng -o7                       | 115.9 | 71 %  | 4:35 hrs |
webp lossless quality=100 | 65.5  | 40 % | 9 mins  |

<br>  
<hr>  

My first thought was to integrate the `_KNOWN_IMG_EXTS` list into the docstring of [`matplotlib_scraper`](https://github.com/sphinx-gallery/sphinx-gallery/blob/b17bf5500a6b3340725cbf01207a2c3d1da14c3b/sphinx_gallery/scrapers.py#L103-L105) but the only solution I found for it is https://stackoverflow.com/a/10308363/3944322 and I think it's a bit of overkill, isn't it?

Do we need tests for it? [`test_custom_scraper_thumbnail_alpha`](https://github.com/sphinx-gallery/sphinx-gallery/blob/b17bf5500a6b3340725cbf01207a2c3d1da14c3b/sphinx_gallery/tests/test_gen_rst.py#L445) could be parameterized with `_KNOWN_IMG_EXTS` but I'm not sure if this really gives more/better test coverage.

Closes #1055

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -154,8 +154,8 @@ Image scrapers are functions (or callable class instances) that do the following
 things:
 
 1. Collect a list of images created in the latest execution of code.
-2. Write these images to disk in PNG, JPEG, SVG, or GIF format (with .png,
-   .jpg, .svg, or .gif extensions, respectively)
+2. Write these images to disk in PNG, JPEG, SVG, GIF, or WebP format (with .png,
+   .jpg, .svg, .gif, or .webp extensions, respectively)
 3. Return reST that embeds these figures in the built documentation.
 
 The function should take the following inputs (in this order):
@@ -197,8 +197,8 @@ reST (see below).
 
 This function will be called once for each code block of your examples.
 Sphinx-Gallery will take care of scaling images for the gallery
-index page thumbnails. PNG images are scaled using Pillow, and
-SVG images are copied.
+index page thumbnails. PNG, JPEG and WebP images are scaled using Pillow, and
+SVG and GIF images are copied.
 
 .. warning:: SVG images do not work with ``latex`` build modes, thus will not
              work while building a PDF version of your documentation.
@@ -305,8 +305,8 @@ Example 3: matplotlib with SVG format
 -------------------------------------
 The :func:`sphinx_gallery.scrapers.matplotlib_scraper` supports ``**kwargs``
 to pass to :meth:`matplotlib.figure.Figure.savefig`, one of which is the
-``format`` argument. Currently Sphinx-Gallery supports PNG (default) and SVG
-output formats. To use SVG, you can do::
+``format`` argument. See :ref:`custom_scraper` for supported formats.
+To use SVG, you can do::
 
     from sphinx_gallery.scrapers import matplotlib_scraper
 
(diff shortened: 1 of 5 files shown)


larsoner:
> Do we need tests for it? [test_custom_scraper_thumbnail_alpha](https://github.com/sphinx-gallery/sphinx-gallery/blob/b17bf5500a6b3340725cbf01207a2c3d1da14c3b/sphinx_gallery/tests/test_gen_rst.py#L445) could be parameterized with _KNOWN_IMG_EXTS but I'm not sure if this really gives more/better test coverage.

I'm less concerned about line coverage as I am about making sure the feature actually works as desired. For example, you could modify some `tinybuild` example to make matplotlib savefig in webp format (or whatever you're going to do in the matplotlib doc build to get it to actually use webp) then `assert` that `sphx_glr_<something>.webp` shows up on disk and in the HTML in `test_full.py` somewhere.

StefRe:
@larsoner I'm a bit at a loss: tests run locally (Win10 / Python 3.10) without errors (218 passed out of 218 collected, nothing skipped). Do you happen to have any idea how to tackle this? Installed Pillow version is 9.4.0 on CI - the same as on my local machine.

larsoner:
You incremented the file counter but I don't see a new example, did you forget a `git add sphinx_gallery/tests/tinybuild/examples/plot_webp_thumb.py` or so?

StefRe:
OMG, yes of course - there should be a `sphinx_gallery/tests/tinybuild/examples/plot_webp.py` that's not included here. Thank you so much! I stared for maybe an hour at the test logs but didn't see the most obvious thing.

Issue #1100, opened 2023-03-13
Some Jupyterlite examples don't work due to missing / not found dependencies


StefRe:
- plot_1_exp.ipynb
`ModuleNotFoundError: No module named 'local_module'`

- plot_2_seaborn.ipynb
`ModuleNotFoundError: No module named 'seaborn'`

- plot_6_function_identifier.ipynb
`ModuleNotFoundError: No module named 'sphinx_gallery'`

- plot_3d.ipynb
`ModuleNotFoundError: No module named 'mayavi'`

- plot_0_plotly.ipynb
`ModuleNotFoundError: No module named 'plotly'`

- plot_collisions.ipynb, plot_glyphs.ipynb, plot_lighting.ipynb, plot_ray_trace.ipynb
`ModuleNotFoundError: No module named 'pyvista'`

lesteve:
Thanks a lot for your feed-back! I am seeing only this now, don't hesitate to ping me directly in the future on sphinx-gallery JupyterLite-related things!

Trying to answer the issues you found:
- plot_1_exp.ipynb
`ModuleNotFoundError: No module named 'local_module'`

This one is actually an issue in Binder too (I noticed it recently while working on the JupyterLite integration). The reason is because we are not copying `local_module.py` in our notebooks folder. It could be a worthwhile fix although the fact that nobody complained about it seems to show that this use case does not happens that often in practice. Personally I am not planning to work on this.

- plot_2_seaborn.ipynb
`ModuleNotFoundError: No module named 'seaborn'`

Yep this is known, in #977 we focused on a MVP. You can add a `%pip install seaborn` cell before importing seaborn and it will work. I have some plans to implement a way to be able to modify the notebook based on its contents to add this kind of cells inside sphinx-gallery, or rather to make it configurable so that people can do what they want in `conf.py`

- plot_6_function_identifier.ipynb
`ModuleNotFoundError: No module named 'sphinx_gallery'`

This one is a bit special in the sense that you expect the package you write examples for to be installed. In the use case I am targetting first (scikit-learn mostly but matplotlib could be another good candidate), the package is available in Pyodide so `import sklearn` and `import matplotlib` would work. For the generic case, this needs a bit more thoughts how it could be made available inside the JupyterLite environment.

- plot_3d.ipynb
`ModuleNotFoundError: No module named 'mayavi'`

[...]

PR #1113, opened 2023-03-23
ENH Add functionality to modify Jupyterlite notebooks based on their content

lesteve:
To be able to run in JupyterLite, notebooks may need additional code, e.g. `%pip install seaborn` since `seaborn` (or any pure-Python code) is not part of Pyodide.

Also this allows to put a warning in a markdown cell to manage user expectations at the beginning of the notebook. We could add a better warning for notebooks that we don't expect to run any time soon, e.g. the ones using pyvista.

No tests or doc for now, but this still may be enough to get some feed-back on the code.

xref https://github.com/sphinx-gallery/sphinx-gallery/issues/1100

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1157,8 +1157,6 @@ You then need to add `jupyterlite_sphinx` to your Sphinx extensions in
 
     extensions = [
         ...,
-        ...,
-        ...,
         'jupyterlite_sphinx',
     ]
 
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1168,8 +1168,9 @@ You can configure JupyterLite integration by setting
     sphinx_gallery_conf = {
       ...
       'jupyterlite': {
-         'jupyterlite_contents': <str> # JupyterLite contents where to copy the example notebooks (relative to Sphinx source directory)
-         'use_jupyter_lab': <bool> # Whether JupyterLite links should start Jupyter Lab instead of the Retrolab Notebook interface.
+         'use_jupyter_lab': <bool>, # Whether JupyterLite links should start Jupyter Lab instead of the Retrolab Notebook interface.
+         'notebook_modification_function': <function>, # function that implements JupyterLite-specific modifications of notebooks
+         'jupyterlite_contents': <str>, # JupyterLite contents where to copy the example notebooks (relative to Sphinx source directory)
          }
     }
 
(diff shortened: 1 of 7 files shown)


larsoner:
I was going to complain that you need to [u pdate the doc](https://sphinx-gallery.github.io/stable/configuration.html#generate-jupyterlite-links-for-gallery-notebooks-experimental) but I see you're not at that stage yet :)

@StefRe can you look to see if this makes sense?

lesteve:
Yep planning to write doc + tests for this when the implementation has settled down a bit.

The main things I am wondering about right now (mentioned in the TODOs), what should the signature be for the function that modify a Jupyterlite notebook (and is user-defined in `conf.py`):
- `modify(json_dict: dict) -> None`: the one I have implemented for now. `json_dict` is what you get when you do `json.load(open(notebook_filename))` and the function modifies the dict in place.
- `modify(json_dict: dict, notebook_filename: str) -> None`. This can be convenient to have the filename to change the notebook according to its name, e.g. all the notebooks in the pyvista folder
- `modify(notebook_filename: str)`. The function is expected to load the notebook and writes a modification to the same file. This implies more boilerplate code in `conf.py` but maybe is simpler? Also you can use additional packages in `conf.py`, e.g. `nbformat` rather than our custom code for creating and writing notebooks. without adding an additional dependency of sphinx-gallery.

StefRe:
@larsoner:  The approach seems reasonable to me, looks good.  

@lesteve: I'd prefer `modify(json_dict: dict, notebook_filename: str) -> None` along with a note in the docs that `notebook_modification_function` should not / does not need to use the filename for writing (or let the function return `True` if it already wrote back the file or `False` / `None` if not and SG should do the writing).

larsoner:
+1 for dict and filename

doc/configuration.rst
```diff
@@ -1179,12 +1180,28 @@ use_jupyter_lab (type: bool, default: ``True``)
...
-  Sphinx source directory. This is used as Jupyterlite contents.
+notebook_modification_function (type: function, default: ``None``)
+  Function that implements JupyterLite-specific modifications of notebooks. By
+  default, it is ``None`` which means that notebooks are not going to be
+  modified. Its signature should be ``notebook_modification_function(json_dict:
+  dict, notebook_filename: str) -> None`` where ``json_dict`` is what you get
+  when you do ``json.load(open(notebook_filename))``. The function is expected
+  to modify ``json_dict`` in place by adding notebook cells. It is not expected
+  to write to the file, since ``sphinx-gallery`` is in charge of this.
+  ``notebook_filename`` is provided for convenience becauses it is useful to
```
StefRe, line 1191: typo: becauses --> because

sphinx_gallery/interactive_example.py
```diff
@@ -332,6 +334,25 @@ def create_jupyterlite_contents(app, exception):
...
+
+    logger.info('Modifying Jupyterlite notebooks ...', color='white')
+    for notebook_filename in notebook_filename_list:
+        with open(notebook_filename) as f:
+            notebook_content = json.load(f)
+
+        notebook_modification_function(notebook_content, notebook_filename)
+
+        with open(notebook_filename, "w") as f:
+            json.dump(notebook_content, f)
```
StefRe, line 354: Maybe we should add here `indent=2` as in https://github.com/sphinx-gallery/sphinx-gallery/blob/7f7087d3ee101c0b71b3a92b101db28f3058d667/sphinx_gallery/notebook.py#L347
so that the modified notebook keeps its readability

doc/configuration.rst
```diff
@@ -1196,9 +1213,11 @@ extra things will happen:
 2. The built Jupyter Notebooks from the documentation will be copied to a
    folder called ``<jupyterlite_contents>/`` (relative to Sphinx source
    directory)
-3. The rST output of each Sphinx-Gallery example will now have a
+3. If ``notebook_modification_function`` is not None, this function is going to
```
StefRe, line 1216: None --> ``None``

StefRe:
@lesteve Perfect, no more comments 👍 .

sphinx_gallery/interactive_example.py
```diff
@@ -332,6 +334,25 @@ def create_jupyterlite_contents(app, exception):
...
                         ignore=_remove_ipynb_files)
 
+    notebook_modification_function = gallery_conf['jupyterlite'][
+        'notebook_modification_function']
+
+    if notebook_modification_function is None:
+        return
+
+    notebook_pattern = os.path.join(contents_dir, "**", "*.ipynb")
+    notebook_filename_list = glob.glob(notebook_pattern, recursive=True)
```
larsoner, line 344: `glob` order is arbitrary, seems like it would be nice to sort for consistency
```suggestion
    notebook_filename_list = sorted(
        glob.glob(notebook_pattern, recursive=True))
```

larsoner (commented):
Just one tiny idea, otherwise LGTM!

sphinx_gallery/interactive_example.py
lesteve: Done

larsoner:
Thanks @lesteve !

PR #1115, opened 2023-03-28
FIX JupyterLite button links

lesteve:
Turns out #1105 was not the right fix, we need to use a relative URL. For example if I copy `_build/html` to a repo like https://github.com/lesteve/sphinx-gallery-jupyterlite, an example html page is 
https://lesteve.github.io/sphinx-gallery-jupyterlite/auto_examples/no_output/plot_raise.html
and I want the link to go to
`https://lesteve.github.io/sphinx-gallery-jupyterlite/lite` so `../../lite`. `/lite` as done in #1105 goes to `https://lesteve.github.io/lite` which is incorrect. If the example is one folder above you want `../lite` instead. This is why I use `os.path.relpath` to get the root URL.

I have added more tests as well with a nested gallery folder (aka sub-gallery I think).

### DIFF
--- a/sphinx_gallery/interactive_example.py
+++ b/sphinx_gallery/interactive_example.py
@@ -339,7 +339,7 @@ def gen_jupyterlite_rst(fpath, gallery_conf):
     Parameters
     ----------
     fpath: str
-        The path to the `.ipynb` file for which a JupyterLite badge will be
+        The path to the `.py` file for which a JupyterLite badge will be
         generated.
 
     gallery_conf : dict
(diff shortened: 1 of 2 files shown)


lesteve:
To make it easier to double-check I have generated the doc locally and made it available on
https://lesteve.github.io/sphinx-gallery-jupyterlite

You can see that the JupyterLite button works both for:
- a file in one of the example directory: https://lesteve.github.io/sphinx-gallery-jupyterlite/auto_examples/plot_0_sin.html
- a file nested one folder below: https://lesteve.github.io/sphinx-gallery-jupyterlite/auto_examples/no_output/just_code.html

larsoner:
Thanks @lesteve !

PR #1116, opened 2023-03-28
MNT: Remove Python 2 leftovers

StefRe:
to improve readability and unify coding style (older code used `%`, newer contributions use `format` etc). Modifications were done using [pyupgrade](https://pypi.org/project/pyupgrade-docs/).

Files in [tinybuild\examples\future](https://github.com/sphinx-gallery/sphinx-gallery/tree/master/sphinx_gallery/tests/tinybuild/examples/future) were deliberately left unchanged.

I know it's a lot of changed files but I think it greatly improves readability and [the Python docs](https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting) also recommend the use of `format` over `%`.


larsoner:
> and [the Python docs](https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting) also recommend the use of format over %.

I actually prefer f-strings nowadays. Is it simple to change to that?

StefRe:
> Is it simple to change to that?

I think so, using the [`--py36-plus`](https://github.com/asottile/pyupgrade#f-strings) option (which will also change some other things but these changes can be reverted if they occur). 

See also the note in the linked docs which sounds good:

> pyupgrade is intentionally timid and will not create an f-string if it would make the expression longer or if the substitution parameters are sufficiently complicated (as this can decrease readability).

I'm going to do the f-strings in a separate commit so you can have a look.

StefRe:
After rebasing I get [two failing tests](https://dev.azure.com/sphinx-gallery/sphinx-gallery/_build/results?buildId=1246&view=logs&j=66ad8d10-cb69-57fe-8782-e0500ab4b7b6&t=b2a44cde-8fb7-5f37-5248-a7b5141501c3&l=699) that pass locally so I assume it's not a real test failure. I can't "Rerun failed jobs" in https://dev.azure.com/sphinx-gallery/sphinx-gallery/_build/results?buildId=1246&view=results due to `TF400813: The user '7dbe823d-48de-6791-9031-8f872678a096' is not authorized to access this resource.` What should I do i such a case? Force push a dummy commit to retrigger the pipeline or is there something more intelligent?

larsoner:
Thanks @StefRe !

PR #1119, opened 2023-03-30
ENH raise error in check_jupyterlite_conf with unknown key

lesteve:
This raises an informative error when `sphinx_gallery_conf['jupyterlite']` contains keys that are not expected.

This is useful to detect typos in the key.


larsoner:
Thanks @lesteve !

PR #1123, opened 2023-04-02
CLN Clean up naming of early config validation

lesteve:
Follow-up of #1120. I tried to clean-up the namings and add a few docstrings.


larsoner:
Thanks @lesteve !

PR #1131, opened 2023-04-15
Correctly hide download buttons

timhoffm:
`visibility:hidden; height: 0px` still keeps margins, creating a large  amount of whitespace. `display: none` is the correct way to remove an HTML element completely.


larsoner (approved):
Failures are unrelated, thanks for the fix!

PR #1135, opened 2023-04-24
MNT: Change % formatting to f-strings

StefRe:
This is a follow-up to #1116:

- change remaining old-style % formatting to f-strings (cases like `'%s' % var` or `'str1' 'str2 %s' % (var,)` which were not handled by `pyupgrade` due to their potential ambiguity)
- change `logging.debug("... %s ..." % x)` to `logging.debug("... %s ...", x)` for uniformity
- remove some `u` string prefixes


larsoner:
Thanks @StefRe !

PR #1137, opened 2023-05-12
Set table.dataframe width in CSS file close #1128

### LINKED ISSUE #1128: large pandas dataframe print
Hi there, 

I'm using sphinx-gallery for an examples chapter in our documentation. When adding a pandas dataframe by printing the dataframe via the head() method, you see big tables are not wrapped at the edge of the site. 

image

Can that be handle somehow? Showing only the columns fitting on the site?

photoniker:
set the table width of dataframe to auto to avoid clipped cell values (e.g. float values).

### DIFF
--- a/doc/_static/theme_override.css
+++ b/doc/_static/theme_override.css
@@ -33,3 +33,7 @@ a.sphx-glr-backref-module-sphinx_gallery {
 .anim-state label {
     display: inline-block;
 }
+
+table.dataframe {
+    width: auto
+}


larsoner:
This will only fix the bug for sphinx-gallery's own API/ref docs, not for projects that *use* SG to generate their docs.

I would have expected the fix to land in this file:

https://github.com/sphinx-gallery/sphinx-gallery/blob/122578b15464be4ec756ae4bf0e0a03367df15f4/sphinx_gallery/_static/sg_gallery-dataframe.css#LL2C33-L2C33

And hopefully just be copy-paste-updated from a newer version of their file that we copied originally. Can you look?

photoniker:
> This will only fix the bug for sphinx-gallery's own API/ref docs, not for projects that *use* SG to generate their docs.
> 
> I would have expected the fix to land in this file:
> 
> https://github.com/sphinx-gallery/sphinx-gallery/blob/122578b15464be4ec756ae4bf0e0a03367df15f4/sphinx_gallery/_static/sg_gallery-dataframe.css#LL2C33-L2C33
> 
> And hopefully just be copy-paste-updated from a newer version of their file that we copied originally. Can you look?

Oh you're totally right. Will Change that!

larsoner:
Thanks @photoniker !

PR #1138, opened 2023-05-13
Add support for capturing svg and png

### LINKED ISSUE #1136: Support more `_repr_*_`
Context: I'm looking for a way to include GraphViz-(run-time)-generated in examples.

One way would be to enable capturing `_repr_png_` and possibly `_repr_svg_`. Is it likely that you will accept a PR for this?

Slightly related: #1134 (as these may also be embedded rather than storing as separate files) https://github.com/xflr6/graphviz/issues/199 (if graphviz includes `_repr_html_` it would also enable it)

oscargus:
Closes #1136

However, I have a doubt that this is a bit "too hacky" as it now checks for both str and bytes to check that a valid representation is obtained. I guess that there is a reason for the explicit str-check? So adding bytes may break things. I was thinking about adding some support to tell if bytes is OK for a given repr (it is required for png), but didn't really find a good way. It is of course possible to drop support for png and this would go away. (It would also be easy to add support for jpeg.)

Also, these are now directly embedded in the source, leading to that they will not show up as thumbnails. I guess that the same thing happens for objects that only have a `_html_repr_` and are not scraped explicitly? Not a big deal for my use case, but would of course be nice even there.

A minor thing, which I sort of realize the reason for, is that these show up at the top. This of course makes sense when it comes to plots, but as I'd like to illustrate the output object in a graphical way, not so much. Easy to solve on my side though. (And I think it only affects the first "cell".)

Should these new representations be on or off by default? I guess having them on may lead to some surprising outputs...

Finally, tests and docs are missing, but I wanted to get some feedback on the bytes-issue before going into that.

(And, yes, not that many lines were needed, at least not with this approach.)


sphinx_gallery/gen_rst.py
```diff
@@ -741,7 +757,7 @@ def _get_last_repr(capture_repr, ___):
             last_repr = None
             repr_meth = None
         else:
-            if isinstance(last_repr, str):
+            if isinstance(last_repr, (str, bytes)):
```
oscargus, line 760:
```suggestion
            if isinstance(last_repr, str) or ininstance(last_repr, bytes) and meth in bytes_repr:
```
Can be an option (and then define `bytes_repr` somewhere, currently only including `_repr_png_`.

sphinx_gallery/gen_rst.py
```diff
@@ -741,7 +757,8 @@ def _get_last_repr(capture_repr, ___):
             last_repr = None
             repr_meth = None
         else:
-            if isinstance(last_repr, str):
+            if (isinstance(last_repr, str) or
+                    isinstance(last_repr, bytes) and meth in ('_repr_png',)):
```
lucyleeow, line 761: Question: could you explain why we need the `isinstance(last_repr, bytes) and meth in ('_repr_png',)` ? Would we not want to stop if `meth` was `'_repr_svg_'` ?

sphinx_gallery/gen_rst.py
```diff
@@ -780,6 +797,11 @@ def _get_code_output(is_last_expr, example_globals, gallery_conf, logging_tee,
     # give html output its own header
     if repr_meth == '_repr_html_':
         captured_html = HTML_HEADER.format(indent(last_repr, ' ' * 4))
+    elif repr_meth == '_repr_png_':
+        captured_html = HTML_PNG_HEADER.format(
+            base64.b64encode(last_repr).decode())
```
lucyleeow, line 802: Just checking `indent` is not required here?

lucyleeow (commented):
Looks good and reasonably simple addition from first glance. Just need test and doc if we're happy to go ahead with adding support for this. Thanks!

sphinx_gallery/gen_rst.py
oscargus: svg will return `str` so it will stop.

oscargus:
Thanks! I'll try to add tests and documentation.

sphinx_gallery/gen_rst.py
lucyleeow: Thanks. My question is more why do we need `isinstance(last_repr, bytes) and meth in ('_repr_png',)` instead of just `(isinstance(last_repr, (str, bytes))`

lucyleeow: What other supported methods could give us `bytes` or is this in case we add new supported methods in future?

oscargus: I had `(isinstance(last_repr, (str, bytes))` first, but then I thought that maybe there was a reason to check for `str` that I could not figure out. (Although unlikely that `_repr_html_` will return bytes?)

Yes, the tuple is for future extensions. Now, I added JPEG as well.

oscargus:
This still needs tests for PNG and JPG output, but I cannot really figure out a good way to generate it without adding dependencies... Will try to generate a minimal binary file and play around a bit.

sphinx_gallery/gen_rst.py
lucyleeow: I had a look at my old PR and I don't think there was any special reason to check for `str` except for maybe safety 'just in case' something weird came out of one of the methods, but as you say it is unlikely `_repr_html_` will return `bytes`, so I think it would be safe to use `(isinstance(last_repr, (str, bytes))` - and even this would be just for safety I think

larsoner:
I would start out by writing the files to disk just like current scrapers, then adding them as `` .mg-sg::` or whatever. Let's see if that's good enough. Then if we want to embed these as base64-encoded images or whatever we can do that separately.

oscargus:
OK, so now it writes files. Also, I added support for `_repr_mimebundle_` in a way transparent for the user. Some objects, like GraphViz graphs, only support this. But asking for `_repr_svg_` will automatically look for an SVG in the MIME bundle (if the method exists etc).

My main problem now is how to test all this. The tests for embedding were quite straightforward as one can simply check the resulting HTML. However, I will need to look more into how to possibly test these ones.

oscargus:
The content of `_repr_mimebundle_` takes precedence if it contains the correct type. This makes sense to avoid generating the same content twice. Also:

https://github.com/scikit-learn/scikit-learn/blob/86541f2b3bc8a96264e265cb810cf79858544340/sklearn/base.py#L636-L638

oscargus:
Another issue, I realize from the tests, is that SVGs are not supported by LaTeX out of the box. Of course it is a bit expected, but I wonder if one should add support for converting SVGs to PDFs? Or mention, e.g., https://pypi.org/project/sphinxcontrib-svg2pdfconverter/ in the documentation?

larsoner:
I would not worry about LaTeX in this PR, we already have the same problem for matplotlib scraping SVGs. Please do open an issue so we don't forget to think about it at some point

For testing I would add a custom class or multiple custom classes with the necessary methods. I'd add them to tests/tinybuild infrastructure as it's really easy and quick to test visually. Then based on that it's easy to add HTML output checks to test_full.py

oscargus:
Thanks! That is what I started to do.

The SVG problem leads to failed doc builds, so it would be nice to at least figure out a way to avoid that.

I noted that `graphviz` was already installed as part of the test suite, so I'll also add a few tests making sure that the figures are saved in correct format etc.

oscargus:
Marking this as ready for review now. Anything else that should be tested?

oscargus:
Managed to get thumbnails etc working...

larsoner:
Error on CircleCI:

```
! LaTeX Error: Unknown graphics extension: .svg.
```
One way to avoid this would be to wrap the SVG output in `.. only:: html`. But maybe we should use the `sphinxcontrib-` package you recommend instead? It would be okay to add this to the test requirements file (assuming the module is well maintained)

oscargus:
I've tried to make it work, but to no avail. My theory is that the svg2pdf converter is not working since the sg image is an `imgsgnode` and not an `img` at that stage. Hence, I tried with using `.. image::` rather than `.. sg-image::`. Then, the thumbnails still works, but no image as the rst-code is kept as is. So while the build now passes, it is not correct.

larsoner:
> As seen, the .. image:: is interpreted as a continuation of the code block. I have (currently) no idea how/where to get around this, so any ideas are appreciated.

Locally you could check this, but let's use CircleCI instead. The RST files are actually saved to disk and you can look:

- https://output.circle-artifacts.com/output/job/ee02670a-3e05-444c-9029-0b0c7b60c191/artifacts/0/tiny_html/_sources/auto_examples/plot_mime_bundle.rst.txt

You can see the problematic line:
```
 .. image:: /auto_examples/images/sphx_glr_plot_mime_bundle_001.svg
```
In particular note the leading space

sphinx_gallery/scrapers.py
```diff
@@ -492,7 +498,7 @@ def _single_line_sanitize(s):
    :class: sphx-glr-single-img
 """
 
-# keep around for back-compat:
+# Used for SVGs
 SINGLE_IMAGE = """
  .. image:: /%s
```
larsoner, line 503: ... so this is the bug, not with your code but with code we no longer use anymore
```suggestion
.. image:: /%s
```

oscargus: Ahh, thanks! Sorry for taking this long to come back to it. I sort of assumed that line was correct...

oscargus:
I got it fully working using `.. sg-image::` for SVG. There is a special template for SVGs without srcset so `.. image-sg::` will directly use the parent method here:

https://github.com/sphinx-gallery/sphinx-gallery/blob/1e825be4cd74c6521b7545413821c52f1c66d151/sphinx_gallery/directives.py#L186-L188

Had to install pipwin and from there cairocffi to get the Windows tests passing. But now it seems to work properly: 
https://output.circle-artifacts.com/output/job/9efaf816-edc7-46d7-b119-4c6b4f319928/artifacts/0/tiny_html/auto_examples/index.html

Not sure if this is an acceptable solution for now or if it may break something else?

sphinx_gallery/scrapers.py
```diff
@@ -353,7 +353,8 @@ def save_figures(block, block_vars, gallery_conf):
     return all_rst
 
 
-def figure_rst(figure_list, sources_dir, fig_titles='', srcsetpaths=None):
+def figure_rst(figure_list, sources_dir, fig_titles='', srcsetpaths=None,
+               svg_image=False):
```
larsoner, line 357: Might as well add these because it's easy
```suggestion
               *, svg_image=False):
```

larsoner (commented):
Just a few minor comments otherwise LGTM!

sphinx_gallery/gen_rst.py
```diff
@@ -726,12 +727,40 @@ def _exec_and_get_memory(compiler, ast_Module, code_ast, gallery_conf,
...
+    available for the last expression.
+    """
+    # First try `_repr_mimebundle_`for those representations that may be there
+    capture_repr_set = set(capture_repr)
+    included_mime_types = {MIME_MAPPING[repr] for repr in
+                           capture_repr_set.intersection(MIME_MAPPING_REPR)}
+    excluded_mime_types = {MIME_MAPPING[repr] for repr in
+                           MIME_MAPPING_REPR.difference(capture_repr_set)}
+    mimebundle = {}
+    if included_mime_types and hasattr(___, '_repr_mimebundle_'):
```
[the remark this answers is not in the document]
oscargus: Yeah, you are correct. I think the idea was to try to catch a slightly more narrow class of exceptions, where `AttributeError` wasn't one of them, but never managed to figure out what a suitable subset might be.

sphinx_gallery/scrapers.py
```diff
@@ -492,6 +499,13 @@ def _single_line_sanitize(s):
    :class: sphx-glr-single-img
 """
 
+# Used for SVGs
+SVG_IMAGE = """
+.. image:: /%s
+   :alt: %s
+   :class: sphx-glr-single-img
+"""
```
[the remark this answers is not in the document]
oscargus: Except for that initial space. Since it said it was kept for backwards compatibility I wasn't sure if the space should still be there to not break anything. But maybe the space is always an issue?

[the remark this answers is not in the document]
larsoner: Yeah I think the space is a bug and can safely be removed

lucyleeow:
@moorepants agree about sympy math. I suspect that the @oscargus does not have time to finish this PR but they may be happy for someone else to take over?
Unfortunately there has been a fair bit of refactor in the code since, so the merge conflicts may not be trivial.

PR #1139, opened 2023-05-13
MAINT: Replace build_sphinx with sphinx-build

oscargus:
In Sphinx 7.0.1 support for build_sphinx was removed. I think this is the way to go.

Also installs `imageio` as I think that is not an expected failure...

Edit: ~install `absl-py` for Sphinx-dev test~. Also, sphinx-dev adds `?hash(?)` to href-link so removed the closing " from the test.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ jobs:
           command: |
             pip install --progress-bar off --user --upgrade --only-binary ":all:" pip setuptools
             pip install --progress-bar off --user --upgrade --only-binary ":all:" numpy matplotlib "pyqt5!=5.15.2,!=5.15.3,!=5.15.8" vtk
-            pip install --progress-bar off --user --upgrade seaborn statsmodels pillow joblib sphinx pytest traits pyvista memory_profiler "ipython!=8.7.0" plotly graphviz "docutils>=0.18"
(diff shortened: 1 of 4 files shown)


lucyleeow (commented):
Thanks for the fix, just a few questions

lucyleeow:
>In Sphinx 7.0.1 support for build_sphinx was removed. I think this is the way to go.

For posterity: https://github.com/sphinx-doc/sphinx/issues/9595

continuous_integration/azure/install.sh
```diff
@@ -20,7 +20,7 @@ if [ "$DISTRIB" == "conda" ]; then
         PIP_DEPENDENCIES="${PIP_DEPENDENCIES} sphinx jinja2<=3.0.3"
     elif [ "$SPHINX_VERSION" == "dev" ]; then
         # It is a mystery to me why we need black, but we get an error with sphinx that it's needed at the end of the build...
-        PIP_DEPENDENCIES="${PIP_DEPENDENCIES} https://api.github.com/repos/sphinx-doc/sphinx/zipball/master black"
+        PIP_DEPENDENCIES="${PIP_DEPENDENCIES} https://api.github.com/repos/sphinx-doc/sphinx/zipball/master black absl-py"
```
lucyleeow, line 23: Why is absl-py needed?

oscargus: Ah, maybe it is not really required. I now see that it is a skipped test, not an error. Will change it back.

larsoner:
Thanks @oscargus , and @lucyleeow ignore my ping from the other PR, sorry I wasn't following properly! :facepalm:

PR #1140, opened 2023-05-15
MAINT: Fix Zenodo reference

oscargus:
There is a generic Zenodo reference that always points at the latest version. This is now used.

(Btw, it looks like there isn't a Zenodo version for 0.13.0.)


lucyleeow:
Thanks @oscargus !

Issue #1144, opened 2023-05-22
Better support for SVGs when building LaTeX


oscargus:
Another issue, I realize from the tests, is that SVGs are not supported by LaTeX out of the box. Of course it is a bit expected, but I wonder if one should add support for converting SVGs to PDFs? Or mention, e.g., https://pypi.org/project/sphinxcontrib-svg2pdfconverter/ in the documentation?

_Originally posted by @oscargus in https://github.com/sphinx-gallery/sphinx-gallery/issues/1138#issuecomment-1554470055_
            

This should probably be handled both in the doc-test scripts as well as in the documentation.

lucyleeow:
I recall some discussion about support for non-html builders: https://github.com/sphinx-gallery/sphinx-gallery/issues/697

Not sure if our thoughts on this has evolved.

lucyleeow:
> Maybe that can be expanded a bit?

+1 sounds like a good idea!

PR #1145, opened 2023-05-22
DOC: Add link to sphinxcontrib-svg2pdfconverter

oscargus:
Related to #1144 (was not really required to do the edit in the issue, could just have added this, but didn't think like that at the time).

The original issue is still there though.


larsoner:
Thanks @oscargus !

PR #1151, opened 2023-07-19
Make _LoggingTee compatible with TextIO

### LINKED ISSUE #1150: Add an "encoding" attribute to _LoggingTee?
Would it be possible to add an "encoding" property or parameter in the  _LoggingTee class? Its value could be "utf-8" for instance, or computed in set_std_and_reset_position.

For information, I am building a gallery with PyTorch Lightning-based scripts. This library (at least its version 1.9.5) uses "sys.stdout.encoding" to print the test results (in pytorch_lightning/loops/dataloader/evaluation_loop.py). If this function is run into the gallery wrapper, sys.stdout is a _LoggingTee which has no "encoding" attribute.

I am not very "proficient" in sys, but it seems that the encoding parameter is a property that is expected and not an eccentricity from the lightning.ai team.

Thank you very much in advance. I can add details if needed.

o-laurent:
Closes #1150.

Following up on the discussion in issue sphinx-gallery/sphinx-gallery#1150, I'd like to propose this PR to improve the compatibility between _LoggingTee and TextIO. It also includes some tests to keep the coverage complete.

Judging from the previous comments, the non-compatibility was already an issue for other packages (e.g. seaborn). In this PR, the classic attributes of [TextIOBase](https://docs.python.org/3/library/io.html#io.TextIOBase) and of [IOBase](https://docs.python.org/3/library/io.html#io.IOBase) (inherited) are all implemented.

This should prevent further issues on the compatibility between the expected stdout and _LoggingTee. I am, of course, open to reverting to adding only the "encoding" attribute if you find it safer/more appropriate. However, this should not be breaking since I am only adding missing attributes and refactoring previous modifications.

(Sorry for the blank lines issue!)

### DIFF
--- a/sphinx_gallery/tests/test_gen_rst.py
+++ b/sphinx_gallery/tests/test_gen_rst.py
@@ -1049,15 +1049,18 @@ def test_textio_compat(log_collector_wrap):
     assert not tee.seekable()
     assert tee.writable()
 
+
 def test_isatty(monkeypatch, log_collector_wrap):
     _, _, tee, _ = log_collector_wrap
     assert not tee.isatty()
     monkeypatch.setattr(tee.output, 'isatty', lambda: True)
     assert tee.isatty()
 
+
 def test_errors(log_collector_wrap):
     _, _, tee, _ = log_collector_wrap
     assert tee.errors == tee.output.errors
+
 
 def test_newlines(log_collector_wrap):
     _, _, tee, _ = log_collector_wrap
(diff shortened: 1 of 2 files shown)


larsoner:
Looks pretty straightforward, complete, and clean -- thanks @o-laurent !

o-laurent:
Thanks for your help and for maintaining the library!

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -1050,13 +1051,26 @@ def test_textio_compat(log_collector_wrap):
     assert tee.writable()
 
 
+def test_fileno(monkeypatch, log_collector_wrap):
```
o-laurent, line 1054: There's just this useless argument in the test... Sorry!

PR #1154, opened 2023-07-30
MAINT Update backreferences docs and add tests

lucyleeow:
Fixes some nitpicks in backreferences docs.

Edit:

This PR now does the following:

* Minor wording updates in backreferences/linking docs
* Update docstring `SphinxDocLinkResolver.resolve`
* Change `prefer_full_module` to set instead of list, like other similar configs (e.g., exclude_implicit_doc)
* Use the full module (instead of short module) name when matching `prefer_full_module` - gives you finer control and brings it in line with what `exclude_implicit_doc` does.
* Update and add a test to better test `_get_short_module_name`

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -426,12 +426,12 @@ To make this work in your documentation you need to include to the
         # a list of regexes of qualified module names where the full module
         # (e.g., a.b.meth) name should be used for links instead of the short
         # module name (e.g., a.meth)
-        'yourmodule\.submod',
+        'module\.submodule',
         ]
     }
 
 In the above example, all qualified names matching the regex
-``'yourmodule\.submod'`` would use the full module name when creating links. All
+``'module\.submodule'`` would use the full module name when creating links. All
 others will use the (default) way of linking.
 
 .. _references_to_examples:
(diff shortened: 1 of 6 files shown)


doc/configuration.rst
```diff
@@ -410,29 +410,29 @@ point to the directory containing ``searchindex.js``, such as
...
-configuration
-dictionary within your Sphinx ``conf.py`` file::
+``prefer_full_module`` configuration within your Sphinx ``conf.py`` file::
 
     sphinx_gallery_conf = {
         ...
         'prefer_full_module':[
-        # a list of regex command of your module where the full module
-        # name should be used for sphinx_gallery instead of the shortend
-        'yourmodule.*+\d{4}',
```
lucyleeow, line 429: I was not sure exactly what this regex was intended to do, regex101 gave:

image

Have amended to just `mod.submod`  as simple example.

larsoner (approved):
Feel free to merge when green

lucyleeow:
Thanks @larsoner, I had forgotten to push. If you're interested, would be good to get your eye over `test_get_short_module_name` - it was difficult to work out how best to test this.

For posterity, the context here is that while I looked into our backreferences code during investigation of: https://github.com/napari/docs/issues/214 and thought we could better test it.

larsoner:
The backreferences code is always a bit of a nightmare. I usually rely on the tests being good enough to catch any breakages. We do check that some matplotlib things resolve in examples so hopefully it's good enough :(

larsoner (approved):
Feel free to mark for merge-when-green if you're happy @lucyleeow !

Issue #440, opened 2018-12-25
BUG: Spaces in example filenames break image linking


shreyasbapat:
I get to see this in place of the plots. Please guide me where I am wrong.
![screenshot from 2018-12-26 01-18-22](https://user-images.githubusercontent.com/24257914/50426128-554cbb80-08ac-11e9-92ab-5b43842fedd8.png)
Thanks! :)

shreyasbapat:
Perhaps, the path in the output block is wrong. The `auto_examples` is extra. But I have not configured this! Any help?

choldgraf:
@shreyasbapat what are you using to host the site? It looks like you're serving from `localhost` though with sphinx you don't need to "serve" the site do you?

double-check that files are in the expected places in your build folder as well - `auto_examples` should be created in your documentation folder when you build the site - it contains the rST and images that are built from your examples

shreyasbapat:
Thanks @choldgraf for a prompt reply.

That's precisely the issue. I am not controlling the links of the images. The links fo the images are incorrect. 
Its like, There is a folder `auto_examples` where the final rST is stored. There is a folder in the directory named `images` where the images are stored. 
Path to the images are like : `auto_examples/images/...`

while it should be :  `images/...`

shreyasbapat:
> @shreyasbapat what are you using to host the site? It looks like you're serving from localhost though with sphinx you don't need to "serve" the site do you?

Here is what I am trying to simulate. : https://poliastro-py.readthedocs.io/en/sphinx-gallery/auto_examples/plot_Analyzing%20the%20Parker%20Solar%20Probe%20flybys.html#sphx-glr-auto-examples-plot-analyzing-the-parker-solar-probe-flybys-py
This is how the gallery is looking when served by Read The Docs.

choldgraf:
hmm - could you provide a link to the github repo?

shreyasbapat:
Sure @choldgraf  here's the link to the repo.
https://github.com/shreyasbapat/poliastro

shreyasbapat:
Sorry. This is the proper link. 
https://github.com/shreyasbapat/poliastro/tree/sphinx-gallery

Plus this is the link to configuration file.
https://github.com/shreyasbapat/poliastro/tree/sphinx-gallery/docs/source/conf.py
https://github.com/shreyasbapat/poliastro/tree/sphinx-gallery/docs/source

larsoner:
[...]

PR #1155, opened 2023-07-31
Maint: Fix `app.builder.outdir` as Sphinx now using pathlib

lucyleeow:
CI is failing with sphinx dev with error:
(see: https://dev.azure.com/sphinx-gallery/sphinx-gallery/_build/results?buildId=1368&view=logs&j=18f51a03-7fab-57d7-f724-30d2d83bc222&t=7c53da0f-8616-5987-83c8-9378d265c3e8&l=1124)
```
ERROR at setup of test_timings ________________________
/usr/share/miniconda/lib/python3.10/site-packages/sphinx/events.py:96: in emit
    results.append(listener.handler(self.app, *args))
sphinx_gallery/docs_resolv.py:471: in embed_code_links
    _embed_code_links(app, gallery_conf, gallery_dir)
sphinx_gallery/docs_resolv.py:324: in _embed_code_links
    doc_resolvers[this_module] = SphinxDocLinkResolver(
sphinx_gallery/docs_resolv.py:146: in __init__
    if doc_url.startswith(('http://',/ 'https://'))/:
E   AttributeError: 'PosixPath' object has no attribute 'startswith'

The above exception was the direct cause of the following exception:
sphinx_gallery/tests/test_full.py:51: in sphinx_app
    return _sphinx_app(tmpdir_factory, 'html')
sphinx_gallery/tests/test_full.py:83: in _sphinx_app
    app.build(False, [])
/usr/share/miniconda/lib/python3.10/site-packages/sphinx/application.py:355: in build
    self.events.emit('build-finished', None)
/usr/share/miniconda/lib/python3.10/site-packages/sphinx/events.py:107: in emit
    raise ExtensionError(__("Handler %r for event %r threw an exception") %
E   sphinx.errors.ExtensionError: Handler <function embed_code_links at 0x7f163d981990> for event 'build-finished' threw an exception (exception: 'PosixPath' object has no attribute 'startswith')
```
I think sphinx started using pathlib in: https://github.com/sphinx-doc/sphinx/pull/11526

Potentially in future we may want to check if `outdir` is a `Path` object (instead of checking if str starts with http..) but have just used `str` for now.

### DIFF
--- a/sphinx_gallery/docs_resolv.py
+++ b/sphinx_gallery/docs_resolv.py
@@ -323,7 +323,7 @@ def _embed_code_links(app, gallery_conf, gallery_dir):
             if url is None:
                 doc_resolvers[this_module] = SphinxDocLinkResolver(
                     app.config.sphinx_gallery_conf,
-                    app.builder.outdir, src_gallery_dir, relative=True)
+                    str(app.builder.outdir), src_gallery_dir, relative=True)
             else:
                 doc_resolvers[this_module] = SphinxDocLinkResolver(
                     app.config.sphinx_gallery_conf,
--- a/sphinx_gallery/tests/test_full.py
+++ b/sphinx_gallery/tests/test_full.py
@@ -205,7 +205,7 @@ def test_junit(sphinx_app, tmpdir):
 
 def test_run_sphinx(sphinx_app):
     """Test basic outputs."""
-    out_dir = sphinx_app.outdir
+    out_dir = str(sphinx_app.outdir)
     out_files = os.listdir(out_dir)
     assert 'index.html' in out_files
     assert 'auto_examples' in out_files


larsoner:
Failure looks real and related

https://dev.azure.com/sphinx-gallery/sphinx-gallery/_build/results?buildId=1370&view=logs&j=18f51a03-7fab-57d7-f724-30d2d83bc222&t=7c53da0f-8616-5987-83c8-9378d265c3e8&l=262

I think `str(outdir)` is a fine fix! Alternatively we could do `Path(outdir)` and use paths where possible. Whichever is easier I'd go with...

lucyleeow:
Have just gone with `str` here. What do you think about moving from `os` to `pathlib` in another PR? (bigger job than initially though, 380 results...)

lucyleeow:
> moving from os to pathlib

It would be a really big job, but I wonder if it would fix things like:  #440

lucyleeow:
merging as CI is green, will consider migrating to pathlib if you support it @larsoner

larsoner:
Migrating to `pathlib` could be helpful. I'm not sure if it's more likely to resolve/fix bugs or create new ones. But could ultimately be worthwhile!

lucyleeow:
>  create new ones.

The more I think about it, the more I think this is true. I think it is okay to have only some parts of the code use pathlib, I will amend the `app.builder.outdir` code to use `Path` as it is cleaner and we probably want to move this way overall.

lucyleeow:
(In any case should be done when one has the time to fix the resulting bugs...!)

PR #1156, opened 2023-08-01
MAINT: Speed up conda solving

larsoner:
CIs are taking 40+ minutes in https://github.com/sphinx-gallery/sphinx-gallery/pull/1154 which is too long. Let's see if `mamba` goes faster. If not, we should just stop using `conda` since we can just use `pip` instead.

It might be worth moving to GitHub Actions as well at some point...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,7 +58,6 @@ junit-results.xml
 # Sphinx documentation
 doc/_build/
 doc/auto_examples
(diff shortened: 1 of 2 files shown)


larsoner:
Okay, 37 min on the linked PR comes down to 3 min in this one. Marking for merge-when green. Thanks `mamba`!

PR #1159, opened 2023-08-03
Ensure consistent encoding for md5sum generation

sdhiscocks:
This resolves an issue where the system default encoding for opening a file is not UTF-8, whereas the encoding default for string encode method is UTF-8, and such hash differs depending on OS.

Related to #722 which originally attempted to resolve inconsistencies across OSs.


lucyleeow:
LGTM, thanks @sdhiscocks !

PR #1160, opened 2023-08-10
updated mpl gui warning catcher to new error message

story645:
Hi folks, https://github.com/matplotlib/matplotlib/pull/26472 changed the wording of the gui error message, so this PR updates the filters and docs to the new warning message wording.  This is gonna be in the 3.8 release. 
I wonder if there's a better way to do this though.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -102,7 +102,8 @@ For example, to remove the specific Matplotlib agg warning, you can add::
 
     warnings.filterwarnings("ignore", category=UserWarning,
                             message='Matplotlib is currently using agg, which is a'
-                                    ' non-GUI backend, so cannot show the figure.')
+                                    ' non-GUI backend, so cannot show the figure.'
+                                    '|(\n|.)*is non-interactive, and thus cannot be shown')
 
 to your ``conf.py`` file.
 
--- a/sphinx_gallery/scrapers.py
+++ b/sphinx_gallery/scrapers.py
@@ -42,7 +42,7 @@ def _import_matplotlib():
     filterwarnings("ignore", category=UserWarning,
                    message='Matplotlib is currently using agg, which is a'
                            ' non-GUI backend, so cannot show the figure.'
-                           '|(\n|.)*is non-interactive, and thus cannot be' 
+                           '|(\n|.)*is non-interactive, and thus cannot be'
                            ' shown')
 
     if matplotlib_backend != 'agg':
--- a/sphinx_gallery/scrapers.py
+++ b/sphinx_gallery/scrapers.py
@@ -41,7 +41,9 @@ def _import_matplotlib():
 
     filterwarnings("ignore", category=UserWarning,
                    message='Matplotlib is currently using agg, which is a'
-                           ' non-GUI backend, so cannot show the figure.')
+                           ' non-GUI backend, so cannot show the figure.'
+                           '|(\n|.)*is non-interactive, and thus cannot be' 
+                           ' shown')
 
     if matplotlib_backend != 'agg':
         raise ExtensionError(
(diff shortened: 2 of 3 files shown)


larsoner (approved):

larsoner (changes requested):
I just realized that this will *only* work for newer matplotlib. The regex should support both old and new matplotlib verisons, which you can accomplish with `(old|new)` for example

doc/configuration.rst
```diff
@@ -101,8 +101,9 @@ For example, to remove the specific Matplotlib agg warning, you can add::
     import warnings
 
     warnings.filterwarnings("ignore", category=UserWarning,
-                            message='Matplotlib is currently using agg, which is a'
-                                    ' non-GUI backend, so cannot show the figure.')
+                            message=r'''Matplotlib is currently using agg, which is a
+                            non-GUI backend, so cannot show the figure.|
+                            (\n|.)*is non-interactive, and thus cannot be shown''')
```
larsoner, line 106: All of those spaces are going to be problematic without a `dedent` or
```suggestion
                            message='Matplotlib is currently using agg, which is a '
                            'non-GUI backend, so cannot show the figure.|'
                            r'(\n|.)*is non-interactive, and thus cannot be shown')
```

sphinx_gallery/tests/test_full.py
```diff
@@ -987,7 +987,8 @@ def test_matplotlib_warning_filter(sphinx_app):
     with codecs.open(example_html, 'r', 'utf-8') as fid:
         html = fid.read()
     warning = ('Matplotlib is currently using agg, which is a'
-               ' non-GUI backend, so cannot show the figure.')
+               ' non-GUI backend, so cannot show the figure.'
+               '|(\n|.)*is non-interactive, and thus cannot be shown')
     assert warning not in html
```
larsoner, line 992: This one is not regex checked, but needs to be now that you've regex-ified the str

larsoner: ... alternatively it might be easier to just keep the lines from `master` and add these two extra lines:
```
warning = "is non-interactive, and thus cannot be shown"
assert warning not in html
```

larsoner: Yes they almost have to pass regardless of actual behavior of sphinx-gallery and it's warning catching... You are currently not checking a regex match here. That would be something like:
```
assert re.match(...) is None
```
Right now it's just a `str.__in__` check:
```
assert warning not in html
```
where `warning` is a string with a bunch of regex-y characters (like `|(\n|.)`) that would never show up in the rendered HTML string when evaluated with the existing `substr in str` check.

Hence you either need to change it to `assert re.match(...) is None)`, or change it to two checks -- one which is the existing check for "old message string not in html" and an added one "new message string not in html" like:

larsoner (approved):
Should work now, marking for merge when green -- thanks for sticking with it @story645 !

larsoner:
> Is no problem, this isn't even a cranky review cycle.

Looks like the CIs are still cranky about style :smile: 
```
sphinx_gallery/scrapers.py:45:80: E501 line too long (82 > 79 characters)
sphinx_gallery/tests/test_full.py:994:1: W293 blank line contains whitespace
```

sphinx_gallery/scrapers.py
```diff
@@ -41,7 +41,9 @@ def _import_matplotlib():
 
     filterwarnings("ignore", category=UserWarning,
                    message='Matplotlib is currently using agg, which is a'
-                           ' non-GUI backend, so cannot show the figure.')
+                           ' non-GUI backend, so cannot show the figure.'
+                           '|(\n|.)*is non-interactive, and thus cannot be'
```
larsoner, line 45:
```suggestion
                           '|(\n|.)*is non-interactive, and thus cannot be'
```

larsoner: Trailing whitespace!

Screenshot 2023-08-10 at 6 44 09 PM


And yes SG should definitely get on board with `black` and `pre-commit.ci` etc. someday...

larsoner:
... I opened https://github.com/sphinx-gallery/sphinx-gallery/issues/1162 so that hopefully next time it's easier for you @story645 !

PR #1161, opened 2023-08-10
MAINT: Make outdated check better

larsoner:
Improve `tinybuild` by

1. Adding something that prints what's detected as modified when rerunning
2. Only replacing thumbnails when they change (which reduced the number of modified files in the test)
3. Slightly better reporting on failure

It's entirely possible this will fail for older versions of Sphnix... let's see!


larsoner:
Added a workaround for https://github.com/pydata/pydata-sphinx-theme/issues/1404

Issue #1162, opened 2023-08-10
MAINT: Use black and pre-commit


larsoner:
We should:

1. Use `black`
2. Use `ruff` instead of flake8 etc (if we don't already)
3. Add both to `pre-commit`
4. Add `pre-commit.ci`

@lucyleeow let me know if this sounds fun/interesting to you and I'm happy to let you do it and give you some pointers. If not then I can do it fairly quickly I think

lucyleeow:
All yours!

larsoner:
I think this is all set now!

PR #1164, opened 2023-08-13
MAINT: black .

larsoner:
First step in #1162 -- run `black .`. One file fails but it has a SyntaxError so this is expected.

I'll mark this for merge-when-green, then open a follow-up PR that does the rest of what's required for #1162:

1. adds the squash+merged commit hash to git-ignore-revs
2. adds `black` to pre-commit
3. Adds `ruff` to pre-commit
4. runs pre-commit ci (should be enabled and show up as green)
5. adds pre-commit CI as a required CI check


lucyleeow:
Nitpick, maybe we could add to our readme.rst under the 'contributing' section what version of black we use for the code. Just so it's documented somewhere since different versions can format differently. Does not have to be in this PR.

larsoner:
Not sure it's worth it with precommit.ci, it'll get out of date fairly often (and we'll have to update it) and if someone happens to use the wrong/older version that didn't format the same the CI will automatically push fixes

lucyleeow:
Yeah good point, sorry missed that we are doing pre-commit. I'm not used to that actually!

PR #1165, opened 2023-08-14
ENH: Use pre-commit

larsoner:
Should come back green. Follow-up commit will uncomment the yamllint and fix the errors, then another to add the squash-and-merged commit to revs-ignore.

Running `ruff` and `codespell` on the entire codebase found some legitimate stuff. At some point it would be nice to have a (much) shorter ignore list for the docstring stuff, but that can be done later if someone gets bored.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,28 @@
...
+        types_or: [python, bib, rst, inc]
+  - repo: https://github.com/mgedmin/check-manifest
+    rev: "0.49"
+    hooks:
+      - id: check-manifest
+  #- repo: https://github.com/adrienverge/yamllint.git
+  #  rev: v1.32.0
+  #  hooks:
+  #    - id: yamllint
+  #      args: [--strict, -c, .yamllint.yml]
```
lucyleeow, line 28: Was this meant to be uncommented or removed?

larsoner: In a follow up PR I uncommented it, made the necessary changes, merged, then added that commit to revs-ignore because it just indented some stuff more

lucyleeow: Ah sorry did not see that, thanks for replying.

PR #1169, opened 2023-08-17
MAINT Fix lint, clean and expand docstrings

lucyleeow:
Addresses some docstring lint rules previously ignored.

* Ignore "D103" (Missing docstring in public function) for tests - there were many tests without docstrings and it did not seem necessary to add these.
* This PR should only include docstring changes - other changes will be in other PRs

Turned out to be larger than I initially expected. If you're wanting to take a look @larsoner, would be useful to just look at any 'parameter' additions to make sure they are correct.

PR #1172, opened 2023-08-17
MAINT Remove unused parametrize in `test_figure_rst_srcset`

lucyleeow:
Also removes unused parametrization in `test_figure_rst_srcset`, and move private funcs to above their use.

Originally `visit_srcset_other` was here to check adding other builders see: https://github.com/sphinx-gallery/sphinx-gallery/pull/808#issuecomment-818034757, but not used so removed here.


sphinx_gallery/directives.py
```diff
@@ -190,6 +176,28 @@ def _parse_srcset(st):
     return srcset
 
 
+def _copy_images(self, node):
```
larsoner, line 179: Can you move this back down to minimize the `diff`? It helps with `git blame`

lucyleeow: Yeah diff is ugly, I'll revert.

lucyleeow:
I think the removal of unused functions was done in #1169  , I might have done that by mistake since lint doesn't pass if we don't remove.

larsoner:
Thanks @lucyleeow !

Issue #229, opened 2017-04-20
"image file not readable" with 0.1.9


dopplershift:
I'm having problems now building our docs with 0.1.9. The warning is:
```
Warning, treated as error:
/Users/rmay/repos/metpy/docs/examples/Advanced_Sounding.rst:None: WARNING: image file not readable: Users/rmay/repos/metpy/docs/examples/images/sphx_glr_Advanced_Sounding_001.png

make: *** [html] Error 1
```
It looks like for some reason now sphinx-gallery is embedding the full absolute path in the rst file:

```rst
.. image:: /Users/rmay/repos/metpy/docs/examples/images/sphx_glr_Advanced_Sounding_001.png
    :align: center
```

Looking over the PRs in the latest release, I'm guessing that [this block](https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/gen_rst.py#L304) is the problem; `figure_paths` is created twice there, but the first one, which handles making relative paths, isn't used. I'm guess copy/paste problem.

dopplershift:
cc @jrleeman

choldgraf:
I have the same problem in trying to build the MPL docs.

In my case, it's not including the `/` at the beginning of the path. So it starts with `Users/choldgraf/etc`, and doesn't start with `/Users/choldgraf/etc`.

dopplershift:
Yeah, I think that relates to `lstrip()` in the same block above--it's probably fine if you actually have a relative path, but not for absolute.

choldgraf:
FWIW this seems to be the case for the SG website too ;-)

![image](https://cloud.githubusercontent.com/assets/1839645/25251737/33015c1e-25cf-11e7-8d79-56660eb9d8f9.png)

http://sphinx-gallery.readthedocs.io/en/latest/auto_examples/plot_exp.html#sphx-glr-auto-examples-plot-exp-py

Titan-C:
I'm terribly sorry on this one. I did not notice this because cached builds of the gallery in my pc and not having good test and not reviewing the output on travis, very irresponsible on my side. This bug came in by PR #218 to deal with windows path and I only checked on the generated html for windows not even the RST files. I'm working on a fix now

dopplershift:
No worries, it happens.

choldgraf:
is this updated in pip?

Titan-C:
[...]

PR #1173, opened 2023-08-17
MAINT Move `figure_rst` path testing to own unit test

lucyleeow:
Test one thing in `test_figure_rst`. Remove path testing to separate test.


sphinx_gallery/tests/test_scrapers.py
```diff
@@ -200,11 +200,14 @@ def test_figure_rst(ext):
 """
     assert image_rst == image_list_rst
 
-    # test issue #229
-    local_img = [os.path.join(os.getcwd(), "third." + ext)]
+
+def test_figure_rst_path():
```
larsoner, line 204: I think we still want this, right?
```suggestion
@pytest.mark.parametrize("ext", _KNOWN_IMG_EXTS)
def test_figure_rst_path():
```

sphinx_gallery/tests/test_scrapers.py
```diff
@@ -200,11 +200,14 @@ def test_figure_rst(ext):
...
+    """Test figure path correct in figure reSt."""
+    # Tests issue #229
+    local_img = [os.path.join(os.getcwd(), "third.png")]
     image_rst = figure_rst(local_img, ".")
 
-    single_image = SG_IMAGE % ("third." + ext, "", "/third." + ext)
```
larsoner, line 207: ... then you'll need to revert this

lucyleeow:
AFAICT #229 and the fix https://github.com/sphinx-gallery/sphinx-gallery/commit/69f11a149b670c81e979ad3a749170525b374be9 was just due to forgetting to replace `/` with `os.sep`, so I don't think extensions make a difference? Happy to revert if not though.

PR #1177, opened 2023-08-17
Avoid importing new modules in backrefs

### LINKED ISSUE #1158: Searching for short module names in backreferences has side effects
This came up via a weird problem we were having building [napari docs](https://github.com/napari/docs) on certain platforms. I think this is a pretty niche problem, and in our case was caused partly by [setuptools allowing case-insensitive imports from editable installations](https://github.com/pypa/setuptools/issues/3994) on certain platforms (in our case macOS). I think that will eventually be fixed, but regardless it suggests a possible improvement in sphinx-gallery.

The basic problem is that sphinx-gallery does some importing when searching for short(er) module names in backreferences:
https://github.com/sphinx-gallery/sphinx-gallery/blob/2e1e2a8a23ed8becb737dac15a733f8b93318cda/sphinx_gallery/backreferences.py#L165

This has potential side effects that may affect examples, scrapers, and reset functions.

Unfortunately I don't really know a good way to clean up after such imports or perform them in isolation (subprocesses would be way too slow). You can remove things from `sys.modules`, but imports also get inserted into the namespace of the parent module. There are also ways to use `importlib` to execute the module without importing it, but none of this prevents the module itself from importing others.

All this said, my current best idea is to _only import modules that are already imported_ when looking at backreferences. That is, something like this:
```diff
diff --git a/sphinx_gallery/backreferences.py b/sphinx_gallery/backreferences.py
index 2d070f5..dfc2d9d 100644
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -14,6 +14,7 @@ from html import escape
 import inspect
 import os
 import re
+import sys
 import warnings
 
 from sphinx.errors import ExtensionError
@@ -163,6 +164,7 @@ class NameFinder(ast.NodeVisitor):
 
 
 def _from_import(a, b):
+    assert a in sys.modules, f'{a} not in sys.modules, not importing it'
     imp_line = f'from {a} import {b}'
     scope = dict()
     with warnings.catch_warnings(record=True):  # swallow warnings
```

This should be safer as it leaves all initial importing up to the user/example code. I think the cost is that it may be a little less aggressive at finding shorter names. I don't think this will generally make a big (or any) difference, but I am also not that familiar with the feature. I did a build with and without this change on the napari docs and couldn't find a meaningful difference in the backrefs. If needed it could be added as a configuration, or maybe an existing configuration already covers this behavior?

I'm happy to hear other ideas here or thoughts on whether this is a real problem at all 😅.

See https://github.com/napari/docs/issues/214 for more discussion - also thanks to @lucyleeow for listening to all my wild theories as I investigated this.

aganders3:
This fixes #1158 (or, is intended to help at least) based on the proposed solution in that issue and discussion elsewhere.

Hopefully this solution makes sense but I'm happy to rework as I came up with a lot of ways to do effectively the same thing. I am not too intimately familiar with the import system (at least I was not when I started on this) and definitely a little lost in some of this sphinx-gallery code.

Either way I know @lucyleeow had also suggested this should possibly be a config and I am happy to add that if you want it. I will also read though to see if there's a good spot to put some of this info in the docs.

It's also worth noting that now setuptools 68.1.0 has been released, which should fix the specific problem (case-insensitive imports from editable installs) that brought this issue to light for me.

### DIFF
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -12,6 +12,7 @@ from html import escape
 import inspect
 import os
 import re
+import sys
 import warnings
 
 from sphinx.errors import ExtensionError
@@ -178,42 +179,27 @@ class NameFinder(ast.NodeVisitor):
         return options
 
 
-def _from_import(a, b):
-    imp_line = f"from {a} import {b}"
-    scope = dict()
-    with warnings.catch_warnings(record=True):  # swallow warnings
-        warnings.simplefilter("ignore")
-        exec(imp_line, scope, scope)
-    return scope
-
-
 def _get_short_module_name(module_name, obj_name):
     """Get the shortest possible module name."""
     if "." in obj_name:
         obj_name, attr = obj_name.split(".")
     else:
         attr = None
-    scope = {}
+
     try:
-        # Find out what the real object is supposed to be.
-        scope = _from_import(module_name, obj_name)
-    except Exception:  # wrong object
+        real_obj = getattr(sys.modules[module_name], obj_name)
+        if attr is not None:
+            getattr(real_obj, attr)  # wrong class
+    except (AttributeError, KeyError):  # wrong object
         return None
-    else:
-        real_obj = scope[obj_name]
-        if attr is not None and not hasattr(real_obj, attr):  # wrong class
-            return None  # wrong object
 
     parts = module_name.split(".")
     short_name = module_name
     for i in range(len(parts) - 1, 0, -1):
         short_name = ".".join(parts[:i])
-        scope = {}
         try:
-            scope = _from_import(short_name, obj_name)
-            # Ensure shortened object is the same as what we expect.
-            assert real_obj is scope[obj_name]
-        except Exception:  # libraries can throw all sorts of exceptions...
+            assert real_obj is getattr(sys.modules[short_name], obj_name)
+        except (AssertionError, AttributeError, KeyError):  # wrong object
             # get the last working module name
             short_name = ".".join(parts[: (i + 1)])
             break
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -13,7 +13,6 @@ import inspect
 import os
 import re
 import sys
-import warnings
 
 from sphinx.errors import ExtensionError
 import sphinx.util
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -186,10 +186,14 @@ def _get_short_module_name(module_name, obj_name):
         attr = None
 
     try:
+        # look only in sys.modules to avoid importing the module, which may
+        # otherwise have side effects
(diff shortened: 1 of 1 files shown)


larsoner:
I think in principle we shouldn't need a config option for this. I'd rather try using it globally and if we find problems with it fix them (maybe by use of a config var but hopefully not)

sphinx_gallery/backreferences.py
```diff
@@ -162,42 +162,27 @@ def get_mapping(self):
...
     short_name = module_name
     for i in range(len(parts) - 1, 0, -1):
         short_name = ".".join(parts[:i])
-        scope = {}
         try:
-            scope = _from_import(short_name, obj_name)
-            # Ensure shortened object is the same as what we expect.
-            assert real_obj is scope[obj_name]
-        except Exception:  # libraries can throw all sorts of exceptions...
+            assert real_obj is getattr(sys.modules[short_name], obj_name)
```
lucyleeow, line 184: I think this is no longer needed (as we assign `real_obj` the same way in 173).
This was here due to: https://github.com/sphinx-gallery/sphinx-gallery/blob/309e5be03088410e495c32272edd09cc9548e50c/sphinx_gallery/tests/test_backreferences.py#L118-L121

sphinx_gallery/backreferences.py
```diff
@@ -162,42 +162,27 @@ def get_mapping(self):
...
-    scope = {}
+
     try:
-        # Find out what the real object is supposed to be.
-        scope = _from_import(module_name, obj_name)
-    except Exception:  # wrong object
+        real_obj = getattr(sys.modules[module_name], obj_name)
+        if attr is not None:
+            getattr(real_obj, attr)  # wrong class
+    except (AttributeError, KeyError):  # wrong object
```
lucyleeow, line 176: I think:
`AttributeError` means wrong class 
`KeyError` means wrong object

Maybe we could move both comments to the except line and expand with the info above?

sphinx_gallery/backreferences.py
```diff
@@ -162,42 +162,27 @@ def get_mapping(self):
...
         obj_name, attr = obj_name.split(".")
     else:
         attr = None
-    scope = {}
+
     try:
-        # Find out what the real object is supposed to be.
-        scope = _from_import(module_name, obj_name)
-    except Exception:  # wrong object
+        real_obj = getattr(sys.modules[module_name], obj_name)
```
lucyleeow, line 173: Maybe add a comment about getting `real_obj` from already imported modules to avoid side effects?

lucyleeow (approved):
A few nits but looks good otherwise!

sphinx_gallery/backreferences.py
lucyleeow: Whoops never mind, missed that we used `module_name` above. Ignore this!

lucyleeow:
@larsoner will merge this soon unless you want another look?

larsoner:
Thanks @aganders3 !

PR #1179, opened 2023-08-18
MAINT Lint - fix ast node type in docstrings

lucyleeow:
Also expands `ast.Module` typing as we use 'ast node' everywhere else.


larsoner:
Thanks @lucyleeow !

PR #196, opened 2017-01-16
Multiple Langauge support

chsasank:
Hi @Titan-C,

Following discussion #194, I've made a working version with support for multiple languages. In the process, I've done quite a bit of refactoring. 

I also 
- made some functions 'private' with prepending `_` to function names. 
- made some code pep8 complaint
- Added comments here and there

What I've done was I've forked `func_rst` branch of your fork from #160 and added my commits. Unfortunately, you've made the top few commits in your branch obsolete since my fork, probably using `commit --amend / squash` and `push --force`. There is a bug (`:overflow:` being visible in gallery page) which you've fixed, but I couldn't merge these changes easily because of the above :(

For further commits you add to #160, please add your commits one by one rather than modifying existing commits. We can rebase/package these commits into 1 after PR is ready to merge.

Please have a look at the implementation and give me your feedback.

Right now, I've written only lua parser, adding other languages shouldn't be hard at this point.

TODO:
- [ ] Merge changes from Titan-C/sphinx-gallery:func_rst carefully.
- [ ] Add other language support
- [ ] Write Tests

PR #1192, opened 2023-09-09
ENH: Support source files in any language

speth:
This PR adds support for reading source files in any language supported by Pygments and generating annotated HTML files from them.

I saw the previous work that was attempted on multi-language support (#196/#197), and it's clear that some of the refactoring work that's been done since allowed this feature to be introduced with relatively few changes to the existing code base.

Obviously, this is a fair bit short of being able to execute such examples and collect outputs, but I think it's a useful capability for projects that have multiple language interfaces to be able to present examples this way, even if only the Python ones have all the bells and whistles.

There are a few things that still need to be resolved here, including additional tests and documentation, but I wanted to get some feedback first before going too much further. There is a bit of refactoring of `py_source_parser.py` that could be done to share some of the implementation with the new `BlockParser` class, which I'd be happy to do if desired.

For context, I'm one of the developers of [Cantera](https://github.com/Cantera/cantera), a library for chemical kinetics that provides interfaces in Python, Matlab, C++, and Fortran. I'm currently working toward a replacement for our home-grown [index of examples](https://cantera.org/examples/), which is implemented as a set of extensions to [Nikola](https://github.com/getnikola/nikola) as we move these examples into our [Sphinx-generated site](https://cantera.org/documentation/docs-3.0/sphinx/html/cython/index.html). I think `sphinx-gallery` provides a great set of features for our Python examples, but I don't want to leave our other interfaces behind. I'm hoping this will also be useful for other projects that include interfaces in multiple languages.


sphinx_gallery/gen_rst.py
```diff
@@ -1126,9 +1127,17 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf, seen_backrefs=No
     target_file = os.path.join(target_dir, fname)
     _replace_md5(src_file, target_file, "copy", mode="t")
 
-    file_conf, script_blocks, node = split_code_and_text_blocks(
+    if fname.endswith(".py"):
+        parser = py_source_parser
+        language = "Python"
+    else:
+        parser = BlockParser(fname, gallery_conf)
+        language = parser.language
```
larsoner, line 1135: On the one hand this is nice because it means whatever `py_source_parser` code we have so far will continue to be used for Python flies, so we're unlikely to break stuff for people.

On the other hand it seems like this will lead to code dup and it might be better to build whatever Python-specific stuff we need (if any? maybe the script_blocks part or something?) into `BlockParser` directly. That could be a follow-up PR I guess because maybe it's not so trivial and/or has other disadvantages (and would definitely be easier to review).

sphinx_gallery/gen_rst.py
```diff
@@ -1223,7 +1236,7 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf, seen_backrefs=No
     ref_regex = _make_ref_regex(gallery_conf["app"].config)
     example_code_obj = identify_names(script_blocks, ref_regex, global_variables, node)
     if example_code_obj:
-        codeobj_fname = target_file[:-3] + "_codeobj.pickle.new"
+        codeobj_fname = os.path.splitext(target_file)[0] + "_codeobj.pickle.new"
```
larsoner, line 1239: It would be great to use pathlib as much as possible for stuff like this

sphinx_gallery/gen_rst.py
```diff
@@ -1281,7 +1294,7 @@ def generate_file_rst(fname, target_dir, src_dir, gallery_conf, seen_backrefs=No
 """
 
 
-def rst_blocks(script_blocks, output_blocks, file_conf, gallery_conf):
+def rst_blocks(script_blocks, output_blocks, file_conf, language, gallery_conf):
```
larsoner, line 1297: I doubt anyone has been using these functions (and they probably shouldn't be) but just in case to minimize pain how about:
```suggestion
def rst_blocks(script_blocks, output_blocks, file_conf, gallery_conf, *, language="python"):
```

sphinx_gallery/gen_rst.py
```diff
@@ -1341,7 +1355,7 @@ def rst_blocks(script_blocks, output_blocks, file_conf, gallery_conf):
 
 
 def save_rst_example(
-    example_rst, example_file, time_elapsed, memory_used, gallery_conf
+    example_rst, example_file, language, time_elapsed, memory_used, gallery_conf
```
larsoner, line 1358: same

sphinx_gallery/gen_rst.py
```diff
@@ -1383,12 +1400,10 @@ def save_rst_example(
         + example_rst
     )
 
-    if time_elapsed >= gallery_conf["min_reported_time"]:
+    if time_elapsed > gallery_conf["min_reported_time"]:
```
larsoner, line 1403: This will change the behavior for the default `min_reported_time = 0`... but maybe that's intentional and better behavior?

sphinx_gallery/utils.py
```diff
@@ -157,6 +145,7 @@ def get_md5sum(src_file, mode="b"):
 
 
 def _replace_md5(fname_new, fname_old=None, method="move", mode="b"):
+    fname_new = str(fname_new)
```
larsoner, line 148:
```suggestion
    fname_new = str(fname_new)  # convert possible Path
```

larsoner (commented):
Looks like a clean implementation so far! Just some preliminary comments from me. Just for background, how will this help you(r project) specifically? I assume you have some mix of Python and something else, and being able to link to non-executed something-else code will make your doc better...

sphinx_gallery/gen_rst.py
speth: Yes, the change was intentional; I thought it didn't make much sense to report the execution time for examples that were never executed.

sphinx_gallery/gen_rst.py
speth: Yes, there is definitely some code duplication caused by trying to leave `py_source_parser` as it is. I'd be happy to refactor it as a subclass of `BlockParser` in a follow-up PR, or as part of this one, whichever you prefer.

speth:
> Just for background, how will this help you(r project) specifically? I assume you have some mix of Python and something else, and being able to link to non-executed something-else code will make your doc better...

I updated the PR description to provide some context.

sphinx_gallery/gen_rst.py
larsoner: Let's do it as a follow-up PR. If we decide later it was a bad idea it makes reverting easier, but hopefully we don't end up there!

larsoner:
One last testing idea -- can you modify `sphinx_gallery/tests/tinybuild` to use this functionality? It's really convenient to be able to do `make` in that dir and look quickly to make sure everything renders properly, plus CircleCI renders it for us (so easy to review by eye). After you add it, you can modify `sphinx_gallery/tests/test_full.py` -- maybe make a new function that uses the `sphinx_app` fixture that gives the paths to the source-copied-to-a-tmpdir and rendered site -- to have some lines that make sure that whatever non-Python source file you add gets rendered in at least some way you expect.

speth:
Thanks for the pointers on setting up some integration tests using the `tinybuild` project. I added a couple of examples there with corresponding tests of the resulting `.rst` files.

The `ci/circleci: build_docs` failure:
```
! LaTeX Error: Unicode character π (U+03C0)
               not set up for use with LaTeX.
```
Is apparently caused by the fact that I tried to use π in a Julia example. I'm not sure how to fix this other than by not using Unicode in the example.

The failure on `sphinx-gallery.sphinx-gallery (Main Linux conda_python310_sphinxDev)`:
```
Theme error:
An error happened in rendering the page advanced.
Reason: ImportError("cannot import name '_get_toctree_ancestors' from 'sphinx.environment.adapters.toctree' (/usr/share/miniconda/lib/python3.10/site-packages/sphinx/environment/adapters/toctree.py)")
```
Doesn't seem to have anything to do with the changes in this PR.

larsoner:
Yes sphinx dev is unrelated so you can ignore it for now. Can you avoid Unicode for now and we can fix it later if needed?

sphinx_gallery/tests/tinybuild/doc/Makefile
```diff
@@ -10,7 +10,7 @@ clean:
 	rm -rf gen_modules/
 
 html:
-	sphinx-build -b html -d _build/doctrees $(SPHINXOPTS) . _build/html
+	sphinx-build -b html -v -d _build/doctrees $(SPHINXOPTS) . _build/html
```
larsoner, line 13: I'd rather move the `-v` to SPHINXOPTS at the top

larsoner (commented):
Other than a few minor comments, looks good to me!

@lucyleeow will you have time to look? If not I think we can rely on our unit tests in addition to my review here, but this touches some core code so if you can look that's better if possible

doc/configuration.rst
```diff
@@ -1248,6 +1270,25 @@ See the Sphinx-Gallery `Sphinx configuration file
...
+downloads can be enabled using the ``notebook_extensions`` option. For example::
+
+    sphinx_gallery_conf = {
+        "notebook_extensions": {".py", ".jl"}
+    }
+
+.. note::
+
+    Currently, all generated notebooks specify Python as the kernel. After downloading,
+    the user will need to manually change to the correct kernel.
```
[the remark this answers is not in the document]
speth: Yes, this is on my list for follow-on work.

sphinx_gallery/tests/tinybuild/doc/Makefile
speth: Sorry, I didn't mean to commit that in the first place.

lucyleeow:
I'll have a look next week.

lucyleeow (approved):
Thanks @speth , this is great and well tested, only some nits.

doc/configuration.rst
```diff
@@ -211,6 +214,25 @@ consult the `regular expressions`_ module for more details.
...
+    sphinx_gallery_conf = {
+        ...
+        'example_extensions': {'.py', '.jl', '.cpp'}
+    }
+
+Parsing and syntax highlighting is supported by the Pygments library, with the language
+determined by the file extension. To override Pygments' default file associations, the
+``filetype_parsers`` option can be used to specify a ``dict`` mapping a file extension
+to any of the `pygments language names <https://pygments.org/languages/>`__. For
+example::
```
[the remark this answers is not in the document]
speth: I tried to clarify. This option isn't so much about handling the case of an extension with no corresponding lexer (though it can be used that way), but for the case where an extension corresponds to several lexers and Pygments' defaults to the wrong one. For example, Pygments lists `'.m` as a known extension for Matlab, Objective-C, and Mason, with the default being Objective-C.

doc/syntax.rst
```diff
@@ -137,6 +137,88 @@ For a clear example refer to the rendered example
...
+of the example. Valid special delimiters are:
+
+1. The comment character followed by ``%%``. For example ``//%%`` for C++.
+2. The comment character followed by a space, followed by ``%%``. For example, ``// %%``
+   for C++.
+3. A line of at least 20 comment characters. For example, ``////////////////////`` for
+   C++.
+
+Any text following the special delimiter on the same line will be converted into a reST
+heading (underlined with ``-``). The reST block continues until a line that does not
```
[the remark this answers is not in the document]
speth: It means that the underlining will be added, using the `-` character. I specified this so the user would know what character to use if they wanted other headings at either the same or a different level.

The existing header syntax is also supported. However, in the case of Matlab, which has had the notion of "code cells" for a long time (possibly where this idea originated), the style is that text appearing on the same line as the `%%` delimiter is treated as a heading, rather than ignored as sphinx-gallery currently does for Python scripts. Beyond Matlab, I thought it was a handy syntax especially in the case where all you want to add is the heading.

speth:
Thanks for the review, @lucyleeow and @larsoner. I think I've addressed all the review comments.

doc/syntax.rst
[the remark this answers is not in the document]
lucyleeow: Ah yes, sorry I think I got confused.

lucyleeow:
Thank you @speth ! This is a great addition!

PR #1193, opened 2023-09-09
MAINT Fix check for mismatched "ignore" blocks

speth:
The checks for mismatched "ignore" blocks were not detecting either "start" or "end" flags, and just comparing `0 == 0`. This PR:
- fixes these regex evaluations
- corrects the name of "start" flag in the `remove_ignore_blocks` docstring
- updates the test to simulate more realistic use case with a multi-line code block


larsoner (approved):
Thanks for the fix @speth !

PR #1197, opened 2023-09-19
ENH: Write all computation times

larsoner:
For a while I've wanted a full listing of example computation times. This PR writes them to the `src_dir` root as `sg_execution_times.html`, which I think is hopefully safe enough.

Snuck in some cleanups of `pytest.mark.parametrize` so our test names can be a bit shorter in the terminal, too.

### DIFF
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -742,7 +742,7 @@ def _sec_to_readable(t):
 
 
 def _cost_key(cost):
-    """Cost (time_elapsed, memory_used) sorting function."""
+    """Cost sorting function."""
     # sort by descending computation time, descending memory, alphabetical name
     return (-cost["t"], -cost["mem"], cost["src_file"])
 
@@ -752,14 +752,13 @@ def _format_for_writing(costs, *, src_dir, kind="rst"):
 
     Parameters
     ----------
-    costs: List[Tuple[Tuple[float], str]]
-        List of tuples of computation costs and absolute paths to example, of format:
-        ((time_elapsed, memory_used), example_path).
+    costs: List[Dict]
+        List of dicts of computation costs and paths, see gen_rst.py for details.
     src_dir : pathlib.Path
         The Sphinx source directory.
     kind: 'rst', 'rst-full' or 'console', default='rst'
         Format for printing to 'console' or for writing `sg_execution_times.rst' ('rst'
-        and 'rst-full').
+        for single galleries and 'rst-full' for all galleries).
 
     Returns
     -------
@@ -802,19 +801,18 @@ def write_computation_times(gallery_conf, target_dir, costs):
         Sphinx-Gallery configuration dictionary.
     target_dir : str | None
         Path to directory where example python source file are.
-    costs: List[Tuple[Tuple[float], str, str]]
-        List of tuples of computation costs and absolute paths to example, of format:
-        ((time_elapsed, memory_used), example_path).
+    costs: List[Dict]
+        List of dicts of computation costs and paths, see gen_rst.py for details.
     """
     total_time = sum(cost["t"] for cost in costs)
     if total_time == 0:
         return
-    if target_dir is None:  # global
+    if target_dir is None:  # all galleries together
         out_dir = gallery_conf["src_dir"]
         where = "all galleries"
         kind = "rst-full"
         ref_extra = ""
-    else:  # local
+    else:  # a single gallery
         out_dir = target_dir
         where = os.path.relpath(target_dir, gallery_conf["src_dir"])
         kind = "rst"
@@ -1155,8 +1153,7 @@ def write_junit_xml(gallery_conf, target_dir, costs):
     target_dir : Union[str, pathlib.Path]
         Build directory.
     costs: List[Tuple[Tuple[float], str]]
-        List of tuples of computation costs and absolute paths to example, of format:
-        ((time_elapsed, memory_used), example_path).
+        List of dicts of computation costs and paths, see gen_rst.py for details.
     """
     if not gallery_conf["junit"] or not gallery_conf["plot_gallery"]:
         return
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -686,6 +686,9 @@ def generate_gallery_rst(app):
                 fhindex.write(indexst)
             _replace_md5(index_rst_new, mode="t")
 
+    # Write a single global sg_execution_times
+    write_computation_times(gallery_conf, None, costs)
+
     if gallery_conf["show_api_usage"] is not False:
         _init_api_usage(app.builder.srcdir)
     _finalize_backreferences(seen_backrefs, gallery_conf)
@@ -693,7 +696,7 @@ def generate_gallery_rst(app):
     if gallery_conf["plot_gallery"]:
         logger.info("computation time summary:", color="white")
         lines, lens = _format_for_writing(
-            costs, os.path.normpath(gallery_conf["src_dir"]), kind="console"
+            costs, src_dir=gallery_conf["src_dir"], kind="console"
         )
         for name, t, m in lines:
             text = (f"    - {name}:   ").ljust(lens[0] + 10)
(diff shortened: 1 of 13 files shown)


sphinx_gallery/gen_gallery.py
```diff
@@ -738,25 +741,25 @@ def _sec_to_readable(t):
...
 
 
-def _format_for_writing(costs, path, kind="rst"):
+def _format_for_writing(costs, *, src_dir, kind="rst"):
     """Provide formatted computation summary text.
 
     Parameters
     ----------
     costs: List[Tuple[Tuple[float], str]]
         List of tuples of computation costs and absolute paths to example, of format:
```
lucyleeow, line 756: Is it still 'absolute paths' ?

sphinx_gallery/gen_gallery.py
```diff
@@ -738,25 +741,25 @@ def _sec_to_readable(t):
...
         ((time_elapsed, memory_used), example_path).
-    path: str
-        Source directory.
-    kind: 'rst' or 'console', default='rst'
-        Format for printing to 'console' or for writing `sg_execution_times.rst' ('rst')
+    src_dir : pathlib.Path
+        The Sphinx source directory.
+    kind: 'rst', 'rst-full' or 'console', default='rst'
+        Format for printing to 'console' or for writing `sg_execution_times.rst' ('rst'
+        and 'rst-full').
```
lucyleeow, line 762: Could we add a comment about 'rst' = single gallery and 'rst-full' (all galleries)?

sphinx_gallery/gen_rst.py
```diff
@@ -442,9 +442,8 @@ def generate_dir_rst(
     index_content: str,
         Content which will be written to the index rst file
         presenting the current example gallery
-    costs: list,
-        List of costs (time_elapsed, memory_used) for building each element of the
-        gallery and absolute path to the example file
+    costs: dict
+        Dict of costs for building each element of the gallery.
```
lucyleeow, line 446: Could you give an example of format?

sphinx_gallery/gen_gallery.py
```diff
@@ -738,25 +741,25 @@ def _sec_to_readable(t):
...
+    return (-cost["t"], -cost["mem"], cost["src_file"])
 
 
-def _format_for_writing(costs, path, kind="rst"):
+def _format_for_writing(costs, *, src_dir, kind="rst"):
     """Provide formatted computation summary text.
 
     Parameters
     ----------
     costs: List[Tuple[Tuple[float], str]]
```
lucyleeow, line 755: I think this needs to be updated now.

sphinx_gallery/gen_rst.py
lucyleeow: Is this list of dicts, one for each example?

sphinx_gallery/gen_gallery.py
```diff
@@ -791,29 +800,39 @@ def write_computation_times(gallery_conf, target_dir, costs):
     ----------
     gallery_conf : Dict[str, Any]
         Sphinx-Gallery configuration dictionary.
-    target_dir : str
+    target_dir : str | None
         Path to directory where example python source file are.
-    costs: List[Tuple[Tuple[float], str]]
+    costs: List[Tuple[Tuple[float], str, str]]
         List of tuples of computation costs and absolute paths to example, of format:
         ((time_elapsed, memory_used), example_path).
```
lucyleeow, line 807: List of Dicts?

sphinx_gallery/gen_gallery.py
```diff
@@ -791,29 +800,39 @@ def write_computation_times(gallery_conf, target_dir, costs):
...
     """
-    total_time = sum(cost[0][0] for cost in costs)
+    total_time = sum(cost["t"] for cost in costs)
     if total_time == 0:
         return
-    target_dir_clean = os.path.relpath(target_dir, gallery_conf["src_dir"]).replace(
-        os.path.sep, "_"
-    )
-    new_ref = f"sphx_glr_{target_dir_clean}_sg_execution_times"
+    if target_dir is None:  # global
```
lucyleeow, line 812: what about: ` # global - all galleries` and `# local - single gallery` ?

sphinx_gallery/gen_gallery.py
```diff
@@ -738,25 +741,25 @@ def _sec_to_readable(t):
     return t
 
 
-def cost_name_key(cost_name):
+def _cost_key(cost):
     """Cost (time_elapsed, memory_used) sorting function."""
```
lucyleeow, line 745:
```suggestion
    """Costs sorting function."""
```

sphinx_gallery/tests/test_full.py
```diff
@@ -861,9 +867,9 @@ def _rerun(
     # - auto_examples/index
     # - auto_examples/plot_numpy_matplotlib
     if how == "modify":
-        n_ch = "[3-7]"
+        n_ch = "[3-9]"
```
lucyleeow, line 870: Could we amend the comment above to add the global execution file to "Ones that can change on stale:". Also I'm lost, why is this not +1 (8) ?

lucyleeow (approved):
Thanks @larsoner ! Doc nitpicks.

I think we could add a note to the 'getting started section' - here ish: https://github.com/sphinx-gallery/sphinx-gallery/blob/11812e6ce39dd64d952ba9590a7a1a09c519d20f/doc/getting_started.rst?plain=1#L150

about the root `sg_execution_times.rst` file?

sphinx_gallery/gen_gallery.py
lucyleeow: Also list of dicts

sphinx_gallery/tests/test_full.py
larsoner: Not totally sure why I needed the extra one here but I also think it's okay to be lenient

larsoner: ... actually updating the comment made it clear I think, pushing

larsoner:
I went to fix the unrelated `graphviz` error and found we're still only requiring 3.7 so I bumped that to 3.8 in a couple of places. I also saw some cruft in the `doc/Makefile` about `theme` so I got rid of that, and in doing so renamed `rtd_html` to `html` in CircleCI since we don't use RTD or change our theme anymore. I'll keep iterating until CIs are happy but I think I incorporated the requested changes so you can look @lucyleeow !

larsoner:
All green!

lucyleeow:
> found we're still only requiring 3.7 so I bumped that to 3.8 in a couple of places

Could we add to our release notes that we are dropping support for 3.7?

sphinx_gallery/gen_rst.py
```diff
@@ -442,9 +442,9 @@ def generate_dir_rst(
     index_content: str,
         Content which will be written to the index rst file
         presenting the current example gallery
-    costs: list,
-        List of costs (time_elapsed, memory_used) for building each element of the
-        gallery and absolute path to the example file
+    costs: dict
```
lucyleeow, line 445:
```suggestion
    costs: List[Dict]
```

sphinx_gallery/gen_rst.py
```diff
@@ -442,9 +442,9 @@ def generate_dir_rst(
...
+        Dict of costs for building each element of the gallery
```
lucyleeow, line 446:
```suggestion
        List of dicts of costs for building each element of the gallery
```

sphinx_gallery/tests/test_full.py
```diff
@@ -849,21 +855,26 @@ def _rerun(
...
+    #
+    # Sometimes it's not all 8, for example when the execution time and
     # memory usage reported ends up being the same.
     #
     # Modifying an example then adds these two:
     # - auto_examples/index
     # - auto_examples/plot_numpy_matplotlib
     if how == "modify":
-        n_ch = "[3-7]"
+        n_ch = "([3-9]|10)"
```
lucyleeow, line 875: Why not `[3-10]` ?

lucyleeow (approved):
Thank you for fixing the CI failure!

sphinx_gallery/tests/test_full.py
larsoner: I had that originally! But it's bad regex :) `[3-9]` is a character range. If you write `[3-10]` regex complains because it sees this as "the range of characters from 3-1, and also the character zero" -- and the range `3-1` doesn't make sense (it would need to be 1-3 and it's also not what we want). So by doing `([3-9]|10)` it will match any single digit between 3 and 9, *or* the literal string 10, which is what we want.

lucyleeow: Ah yes of course, thanks!

PR #1198, opened 2023-09-21
ENH: Always write sg_execution_times

larsoner:
In MNE-Python we only execute examples that have been changed. I'd like to link CircleCI artifacts redirector action to `sg_execution_times.htmtl`, but this only exists if any examples have been executed. It's easy enough to just *always* write this file (the cost is very low), so I figure we might as well do it.

### DIFF
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -805,8 +805,6 @@ def write_computation_times(gallery_conf, target_dir, costs):
         List of dicts of computation costs and paths, see gen_rst.py for details.
     """
     total_time = sum(cost["t"] for cost in costs)
-    if total_time == 0:
-        return
     if target_dir is None:  # all galleries together
         out_dir = gallery_conf["src_dir"]
         where = "all galleries"


larsoner:
Okay I also made it much more usable by injecting some JS and scoped CSS to make it a DataTable. I've always wanted to be able to sort by computation time or memory usage and now I can! And it comes with nice stuff like pagination and search for free:

https://output.circle-artifacts.com/output/job/ffe304c1-792d-40b9-8710-60d0d6656f6f/artifacts/0/html/sg_execution_times.html

larsoner:
@lucyleeow ready for review/merge from my end

sphinx_gallery/tests/test_full.py
```diff
@@ -695,7 +695,7 @@ def test_rebuild(tmpdir_factory, sphinx_app):
     if sys.platform.startswith("win"):
         assert (
             re.match(
-                ".*[0|1] added, [1-9][0-3]? changed, 0 removed$.*",
+                ".*[0|1] added, ([1-9]|1[0-4]) changed, 0 removed$.*",
```
lucyleeow, line 698: I think we have to change the above comment to "XXX on AppVeyor this can be 14"

lucyleeow (approved):
the datatable looks great!

larsoner:
Thanks @lucyleeow pushed a commit to update the comments and marking for merge when green

PR #1203, opened 2023-09-26
ENH: Improve API usage graph

larsoner:
After working on https://github.com/mne-tools/mne-python/pull/12013#issuecomment-1736018069 I wanted to port over some changes that I think will be generally helpful. After some simplifications and tweaks to the diagram to make it more readable, esp. with a dense layout with overlaps, this is what I see:

| | before | after |
| -- | -- | -- |
| SG | image | image |
| MNE | image | image |

@drammock feel free to test out locally (`PATTERN=no_examples make html-pattern` in MNE or `make html` in `doc` of this repo) and tweak if you want to try changing stuff.

### DIFF
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -911,6 +911,17 @@ def _init_api_usage(gallery_dir):
         pass
 
 
+# Colors from https://personal.sron.nl/~pault/data/colourschemes.pdf
+# 3 Diverging Colour Schemes, Figure 12, plus alpha=AA
+API_COLORS = dict(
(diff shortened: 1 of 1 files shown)


drammock:
I don't really have time to tweak/test right now. This is already a big improvement, thanks! My only feedback is that the black text on the dark reddish color is too low-contrast, so I'd tweak that red color to be lighter

larsoner:
Switched to the Figure 13 colormap and reduced the alpha (88 vs AA) and I think it's more readable now:

image

I'll push and mark for merge-when-green, thanks for the quick feedback @drammock !

PR #1207, opened 2023-10-04
BUG: Fix bug with orphan sg_api_usage

larsoner:
Fixes this bug when running in noplot mode:
```
checking consistency... /Users/larsoner/python/mne-python/doc/sg_api_usage.rst: WARNING: document isn't included in any toctree
```
The `:orphan:` and title should always be added.

Also snuck in some `changelog_generator` stuff to speed things up that I've been using elsewhere that should help at release time.


.github_changelog_generator
```diff
@@ -1,5 +1,7 @@
 project=sphinx-gallery
 user=sphinx-gallery
 add-sections={"api":{"prefix":"**API changes**","labels":["api"]},"documentation":{"prefix":"**Documentation**","labels":["documentation"]},"maintenance":{"prefix": "**Project maintenance**","labels":["maintenance"]}}
-max-issues=200
+max-issues=2
```
lucyleeow, line 4: Why do we only want to fetch 2 issues? We should want a few months worth right?

lucyleeow (approved):
LGTM, thanks! Just one question.

.github_changelog_generator
larsoner: We actually already have `issues=no` meaning "don't put any issues . It's a mystery to me why it even bothers fetching issues when this is the case but it did (at least at some point).

lucyleeow: Ah right, did not realise. So for the changelog we only use PRs? I thought we, at some point, used issues as well?

larsoner: Yes we used to but in practice they didn't seem to add much value compared to properly listing PRs so I removed them

lucyleeow: Good to know, I tend to agree. Some issues are just quickly resolved without anything needing to be done and thus did not really add anything to the changelog.

PR #1213, opened 2023-10-23
FIX Copy JupyterLite contents early so it runs before jupyterlite_sphinx build-finished

### LINKED ISSUE #1211: Order dependence with Jupyterlite
I am used to assume that the order in which Sphinx extensions are listed does not matter. That does not seem to be the case with extensions "jupyterlite_sphinx" and "sphinx_gallery.gen_gallery".  At least for me "jupyterlite_sphinx" needs to be listed after "sphinx_gallery.gen_gallery".

Either this dependence on the order should be removed, or it should be clearly noted in the documentation (maybe with a warning box). I wasted too much time on figuring this on my own.

lesteve:
Fix #1211.

I am not sure how to add a tests for this.

You want the notebooks copied to `jupyterlite_contents` copied before the `jupyterlite_sphinx` runs `jupyterlite build --contents jupyterlite_contents`.

I could also copy the contents in `builder-inited` after `generate_gallery_rst` not sure which ones is less brittle.


lesteve:
The CI errors are not related to this PR, probably an issue with libmamba ...

```
ImportError: /usr/share/miniconda/lib/python3.11/site-packages/libmambapy/../../../libmamba.so.2: undefined symbol: solver_ruleinfo2str, version SOLV_1.0
```

I did not find anything too related about these errors, https://github.com/conda-incubator/conda-store/issues/632, that was worked around in https://github.com/conda-incubator/conda-store/pull/634 by not mixing not mixing `defaults` and `conda-forge`.

sphinx_gallery/gen_gallery.py
```diff
@@ -1503,8 +1503,9 @@ def setup(app):
     # Early update of sphinx_gallery_conf at builder-inited
     app.connect("builder-inited", update_gallery_conf_builder_inited, priority=10)
     app.connect("builder-inited", generate_gallery_rst)
-    app.connect("build-finished", copy_binder_files)
-    app.connect("build-finished", create_jupyterlite_contents)
+    # Setting a small priority so that create_jupyterlite_contents runs before
```
lucyleeow, line 1506:
```suggestion
    # Setting a low priority so that create_jupyterlite_contents runs before
```

Just to avoid confusion.

lucyleeow (approved):
LGTM! Thanks

sphinx_gallery/gen_gallery.py
lesteve: So I used small here on purpose rather than low. Callbacks are run in order of ascending priority so small priority means high priority (`priority=100` means higher priority than `priority=500` which is the default). 

Maybe "setting priority to a small number" is less confusing. There is a similar in a comment a few lines above so it could be improved as well ...

lucyleeow: Ah I got myself confused, even after reading the docs.

"setting priority to a small number" sounds good!

lesteve: OK I tweaked the comments hopefully avoiding the confusion a bit more.

For reference, the `connect` [doc](https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.connect) says:

> priority: The priority of the callback. The callbacks will be invoked in order of priority (ascending).

lesteve:
Hmmm actually thinking about it I think we also want copying the notebooks to be after generating the notebooks, I think this needs a bit more tweaking ...

lesteve:
> Hmmm actually thinking about it I think we also want copying the notebooks to be after generating the notebooks, I think the priority needs a bit more tweaking ...

I think everything is good to go actually after thinking about it. The notebooks are generated by `generate_gallery_rst` at `builder-inited` and notebooks are copied at `builder-finished` time by `create_jupyterlite_contents`.

PR #1214, opened 2023-10-23
MAINT: Move to GHA

larsoner:
Long-overdue migration to GitHub Actions. Should help with maintainability and stuff like https://github.com/sphinx-gallery/sphinx-gallery/pull/1213#issuecomment-1775219014


larsoner:
Highlights:

- Looks like Windows at least is a little bit faster using GHA, which is nice:

  | Main | PR |
  | -- | -- |
  | Screenshot from 2023-10-23 12-30-04 | Screenshot from 2023-10-23 14-40-45 |

- There should also should be better coverage because we now test on macOS.

- Flow seems cleaner to me, and the net -100 lines also suggests this is the case.

- We no longer test on Python nightlies but I think it's good enough to just quickly bump once a new Python is released and NumPy is available etc. Nowadays this happens quickly on conda-forge which powers most CI runs in this PR so I think we should be good.
 
@lucyleeow feel free to merge if you're happy!

.github/workflows/tests.yml
```diff
@@ -0,0 +1,105 @@
...
+            python: '3.12'
+            sphinx_version: dev
+            distrib: mamba
+            locale: C
+          - os: ubuntu-latest
+            python: '3.11'
+            sphinx_version: '6'
+            distrib: mamba
+          - os: ubuntu-latest
+            python: '3.8'
```
lucyleeow, line 42: I assume 3.8 is the min python version we support? Not essential, but is it easy to make this a variable? Makes it easier to update and maybe easier to find/more explicit what our min support python version is.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,105 @@
...
+          init-shell: bash
+          create-args: 'python=${{ env.PYTHON_VERSION }} pip numpy setuptools matplotlib pillow pytest pytest-cov coverage seaborn statsmodels plotly joblib wheel libiconv pygraphviz memory_profiler ipython pypandoc'
+        if: matrix.distrib == 'mamba'
+      - name: Set env vars
+        shell: bash -e {0}
+        run: |
+          echo "LC_ALL=C" >> $GITHUB_ENV
+          echo "LANG=C" >> $GITHUB_ENV
+          echo "LC_CTYPE=C" >> $GITHUB_ENV
+        if: matrix.locale == 'C'
```
lucyleeow, line 88: Could we add back this comment:

```
          # Make sure that things work even if the locale is set to C (which
          # effectively means ASCII). Some of the input rst files have unicode
          # characters and we need to deal with this gracefully.
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,105 @@
...
+      matrix:
+        include:
+          - os: ubuntu-latest
+            python: '3.12'
+            sphinx_version: dev
+            distrib: mamba
+            locale: C
+          - os: ubuntu-latest
+            python: '3.11'
+            sphinx_version: '6'
```
lucyleeow, line 39: Just a question, why do want to test this sphinx version?

continuous_integration/install.sh
```diff
@@ -0,0 +1,38 @@
...
+        sudo apt install graphviz
+    fi
+else
+    echo "invalid value for DISTRIB environment variable: $DISTRIB"
+    exit 1
+fi
+pip install --upgrade $PIP_DEPENDENCIES pytest pytest-cov coverage pydata-sphinx-theme
+
+if [[ "$PLATFORM" == "Windows" ]]; then
+    pip uninstall -y graphviz
```
lucyleeow, line 35: Maybe it's obvious but could we add a note as to why we do this?

continuous_integration/install.sh
```diff
@@ -0,0 +1,38 @@
...
+    PIP_DEPENDENCIES="jupyterlite-sphinx>=0.8.0,<0.9.0 jupyterlite-pyodide-kernel<0.1.0 libarchive-c"
+    if [ "$SPHINX_VERSION" == "dev" ]; then
+        PIP_DEPENDENCIES="--pre ${PIP_DEPENDENCIES} https://api.github.com/repos/sphinx-doc/sphinx/zipball/master"
+    elif [ "$SPHINX_VERSION" == "old" ]; then
+        PIP_DEPENDENCIES="${PIP_DEPENDENCIES} sphinx<5"
+    elif [ "$SPHINX_VERSION" != "default" ]; then
+        PIP_DEPENDENCIES="${PIP_DEPENDENCIES} sphinx==${SPHINX_VERSION}"
+    fi
+    if [ "$PLATFORM" == "Darwin" ]; then
+        conda remove -y memory_profiler
```
lucyleeow, line 19: As below, why does this not work on mac?

lucyleeow (approved):
Thank you for doing this, I'm not well versed in CI but it seems much cleaner. Just some nits.

lucyleeow:
> We no longer test on Python nightlies but I think it's good enough to just quickly bump once a new Python is released and NumPy is available etc. Nowadays this happens quickly on conda-forge which powers most CI runs in this PR so I think we should be good.

Should we add a note to our release documentation to check if there is a new python release? I don't like relying on my memory to do things.

.github/workflows/tests.yml
larsoner: > but is it easy to make this a variable?

You mean in YAML? Not super easy. I think it's not too bad to find it here. I'll just update `developers.rst` to mention this alongside "latest Python" in the release instructions/reminders.

.github/workflows/tests.yml
larsoner: Really just to try to hit all supported major Sphinx versions but I'll make it clearer

lucyleeow:
Thank you!

PR #1125, opened 2023-04-05
FEA Add examples recommender system

ArturoAmorQ:
Fixes https://github.com/sphinx-gallery/sphinx-gallery/issues/1081

For a proof of concept using the scikit-learn library navigate through:
https://output.circle-artifacts.com/output/job/22589aa6-cfe8-4daf-a9c0-b5abf94098b6/artifacts/0/doc/auto_examples/index.html

This PR is ready for review.

PR #1218, opened 2023-10-27
MAINT: pydata-sphinx-theme

larsoner:
This should fix the failure from #1125, let's see...


drammock:
> This should fix the failure from #1125, let's see...

oops yeah, I should have warned you all about that 🙊

PR #1225, opened 2023-11-09
ENH: Improve logging visibility of errors and filenames

larsoner:
I recently had a build that I knew would have a bunch of failures. It seemed harder than it should have been to find the failing examples when I browsed CircleCI. So this improves logging of failures by:

1. Highlighting the filename in color, leaving the traceback in regular (white)
2. Indenting the traceback (makes filenames act like a definition list)
3. Using relative paths for filenames, since it's almost always something relative to `src_dir` and the potentially very long leading path typically doesn't matter

## in-build

master:

Screenshot from 2023-11-09 13-59-43

PR:

image

## end-of-build

master:
Screenshot from 2023-11-09 14-00-11

PR:

Screenshot from 2023-11-09 14-07-43 |

### DIFF
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -13,6 +13,7 @@ from datetime import timedelta, datetime
 from difflib import get_close_matches
 from importlib import import_module
 from pathlib import Path
+from textwrap import indent
 import re
 import os
 import pathlib
@@ -20,7 +21,7 @@ from xml.sax.saxutils import quoteattr, escape
 
 from sphinx.errors import ConfigError, ExtensionError
 import sphinx.util
-from sphinx.util.console import red
+from sphinx.util.console import blue, red, purple, bold
 from . import glr_path_static, __version__ as _sg_version
 from .utils import _replace_md5, _has_optipng, _has_pypandoc, _has_graphviz
 from .backreferences import _finalize_backreferences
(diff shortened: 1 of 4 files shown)


larsoner:
FYI color difference during build vs end of build is intentional -- during build it's nice to have a larger block of colorful text so you're more likely to see it, plus good / non-failing examples will appear in between in white. At the end of the build when the exception is raised your attention is already down there, so you're looking for relevant information, and having just the filenames in color helps break up any multiple failures better.

GaelVaroquaux (approved):
Super nice feature :+1:

lucyleeow:
This is great! I hate to be that person but I wonder if the red/purple are colour deficient friendly. I know there is a association with red = bad so I don't know what to do about that but we could amend the purple to be yellow - most people with colour deficiencies are red/green deficient (protanopia/deuteranopia) and yellow would be [seen as very different to black](https://davidmathlogic.com/colorblind/#%23D81B60-%231E88E5-%23FFC107-%23004D40-%23E8E817)  (I used to be an optometrist)

sphinx_gallery/gen_gallery.py
```diff
@@ -1380,9 +1387,14 @@ def summarize_failing_examples(app, exception):
     )
 
     if fail_msgs:
-        fail_message = (
-            "Here is a summary of the problems encountered "
-            "when running the examples\n\n" + "\n".join(fail_msgs) + "\n" + "-" * 79
+        fail_message = bold(
+            purple(
```
lucyleeow, line 1391: maybe a yellow-ish colour?

lucyleeow (approved):
Thanks for this!

larsoner:
Pushed yellow, which is okay but not quite as nice I think:

Screenshot 2023-11-09 at 9 28 24 PM

Thinking about purple/red, putting my image above into https://pilestone.com/pages/color-blindness-simulator-1# for example and looking at the protanopia/deuteranopia simulations look differentiable to me, do you think that's sufficient? FWIW people can remap ANSI terminal color codes to other colors as they want (even my Ubuntu screenshot above is solarized or something, I'd have to check), but we probably shouldn't rely on that.

But if you think yellow is the way to go I can live with it!

lucyleeow:
Yeah fair, go back to purple. Thanks for checking.

PR #1226, opened 2023-11-10
[ENH] Minigallery can take arbitrary files/glob patterns as input

### LINKED ISSUE #1029: Create mini-gallery with an arbitrary list of gallery items
Hi there :slightly_smiling_face: 

For now, mini-galleries filter out gallery items based on whether they use a specified method / class / etc or not.
I think it'd be very nice to be able to generate mini-galleries from an arbitrary list of items specified by the user.
Specifically, I'm planning to use this in `nilearn`'s user guide to reference examples using the nice `sphinx-gallery` thumbnail.

It could look something like this:
```
.. minigallery::
    :add-heading: A list of examples relevant to the section above
    :sg-items: section_1/item_3.py section_4/item_2.py
```

Let me know what you think!

story645:
Closes #1029

- [x] still have to finish up tests and docs, but wanted some feedback if this was in the right direction.

- [x] If so, I'm also wondering if I could strip the thumbnail parent div out of the backref files so that I could do it in the loop.
- [x] get .py files in .rst folders working - seems to work by default and issue was sorting
- [x] file sort order

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -48,6 +48,7 @@ file, inside a ``sphinx_gallery_conf`` dictionary.
 **Cross-referencing**
 
 - ``reference_url``, ``prefer_full_module`` (:ref:`link_to_documentation`)
+- ``minigallery_sort_order`` (:ref:`minigallery_order`)
 - ``backreferences_dir``, ``doc_module``, ``exclude_implicit_doc``,
   and ``inspect_global_variables`` (:ref:`references_to_examples`)
 
@@ -418,6 +419,8 @@ which sorts the files based on the number of code lines::
         'within_subsection_order': NumberOfCodeLinesSortKey,
     }
 
+.. _sort_classes:
+
 In addition, multiple convenience classes are provided for use with
 ``within_subsection_order``:
 
@@ -426,7 +429,7 @@ In addition, multiple convenience classes are provided for use with
 - :class:`sphinx_gallery.sorting.FileSizeSortKey` to sort by file size.
 - :class:`sphinx_gallery.sorting.FileNameSortKey` to sort by file name.
 - :class:`sphinx_gallery.sorting.ExampleTitleSortKey` to sort by example title.
-
+- :class:`sphinx_gallery.sorting.FunctionSortKey` to sort using an arbitrary function.
 
 .. _link_to_documentation:
 
@@ -500,10 +503,68 @@ In the above example, all fully qualified names matching the regex
 (e.g., module.submodule.meth) when creating links, instead of the short module
 name (e.g., module.meth). All others will use the (default) way of linking.
 
+.. _minigalleries_to_examples:
+
+Add mini-galleries
+==================
+
+Sphinx-Gallery provides the :class:`sphinx_gallery.directives.MiniGallery`
+directive so that you can easily add a reduced version of the Gallery to
+your sphinx documentation ``.rst`` files. The mini-gallery directive therefore
+supports passing a list (space separated) of any of the following:
+
+* full qualified name of object
+* pathlike strings to example Python files
+* glob-style pathlike strings to example Python files
+
+
+For example, the rst below adds the reduced version of the Gallery for all
+examples that use the specific function ``numpy.exp``, the example
+``examples/plot_sin_.py``, and all files matching the string
+``/examples/plot_4*``:
+
+.. code-block:: rst
+
+    .. minigallery::
+        :add-heading:
+
+        numpy.exp
+        ../examples/plot_0_sin.py
+        ../examples/plot_4*
+
+
+Listing multiple items, separated by spaces, merges all the examples into a
+single mini-gallery. The mini-gallery will only be shown if the files exist or
+the items are actually used or referred to in an example.
+
+.. minigallery::
+    :add-heading:
+
+    numpy.exp
+    ../examples/plot_0_sin.py
+    ../examples/plot_4*
+
+
+The ``add-heading`` option adds a heading for the mini-gallery, which will be a
+default generated message "Examples using *{full qualified object name}*" if no
+string is provided as an argument. Specifying a custom heading message is
+recommended for a gallery with multiple items because the default message is
+"Examples of one of multiple objects".
+
+The example mini-gallery shown above uses the default heading level ``^``. This
+can be changed using the ``heading-level`` option, which accepts a single
+character (e.g., ``-``).
+
+The following form of passing inputs to minigallery is also supported:
+
+.. code-block:: rst
+
+    .. minigallery:: numpy.exp , ../examples/plot_0_sin.py
+
 .. _references_to_examples:
 
 Add mini-galleries for API documentation
-========================================
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 When documenting a given function/method/attribute/object/class, Sphinx-Gallery
 enables you to link to any examples that either:
@@ -569,34 +630,60 @@ To exclude the functions mentioned above you would use
 instead of any character, if the name is unambiguous you can also write
 ``pyplot.show`` or just ``show``).
 
-Within your sphinx documentation ``.rst`` files, you can easily
-add this reduced version of the Gallery. For example, the rst below adds
-the reduced version of the Gallery for ``numpy.exp``, which includes all
-examples that use the specific function ``numpy.exp``:
+
+.. _file_based_minigalleries:
+
+Create mini-galleries using file paths
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Sometimes you may want to explicitly create a
+:class:`mini-gallery <sphinx_gallery.directives.MiniGallery>` using files that
+do not have functions in common, for example a set of tutorials. The
+mini-gallery directive therefore also supports passing in:
+
+* pathlike strings to sphinx gallery example files
+* glob-style pathlike strings to sphinx-gallery example files
 
 .. code-block:: rst
 
-    .. minigallery:: numpy.exp
-        :add-heading:
+    .. minigallery:: numpy.exp ../examples/plot_0_sin.py ../examples/plot_4*
 
-The ``add-heading`` option adds a heading for the mini-gallery, which will be a
-default generated message if no string is provided as an argument. The example
-mini-gallery shown above uses the default heading. The level of the heading
-defaults to ``^``, but can be changed using the ``heading-level`` option, which
-accepts a single character (e.g., ``-``). The mini-gallery will only be shown
-if the item (here ``numpy.exp``) is actually used or referred to in an example.
+.. minigallery:: numpy.exp ../examples/plot_0_sin.py ../examples/plot_4*
+    :add-heading: Mini-gallery using paths
+    :heading-level: ^
 
-You can also list multiple items, separated by spaces, which will merge all
-examples into a single mini-gallery, e.g.:
+.. _minigallery_order:
 
-.. code-block:: rst
+Sort mini-gallery thumbnails from files
+"""""""""""""""""""""""""""""""""""""""
 
-    .. minigallery:: numpy.exp numpy.sin
-        :add-heading: Mini-gallery using ``numpy.exp`` or ``numpy.sin``
-        :heading-level: -
+The mini-gallery directive generates a gallery of thumbnails corresponding to
+the input file strings or object names. The ordering of these thumbnails can be
+set using the ``minigallery_sort_order`` configuration key, which is any
+sphinx-gallery :ref:`sorting <sort-warning>` method. The minigallery directive
+sorts on a list of backreference ``{input}.example`` and file
+``{gallery_example}.py`` path strings.
 
-For such a mini-gallery, specifying a custom heading message is recommended
-because the default message is vague: "Examples of one of multiple objects".
+Here the sorting method :class:`sphinx_gallery.sorting.FunctionSortKey` is used
+to wrap a function that puts the backreference thumbnails at the end of the
+list.
+
+.. code-block:: python
+
+    sphinx_gallery_conf = {
+    #...,
+    "minigallery_sort_order": FunctionSortKey(
+        lambda x: (not x.startswith("plot_4b"), x)),
+    #...
+    }
+
+Sorting the set of thumbnails generated from
+:ref:`API backreferences <references_to_examples>`, i.e. the thumbnails linked
+to a qualified object name input is not supported, but the sorting function can
+be used to position the entire set of backreference thumbnails since it sees
+the ``{input}.example`` file naming pattern. Thumbnails may be duplicated if
+they correspond to multiple object names or an object name and file/glob input.
+
 
 Auto-documenting your API with links to examples
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -687,11 +687,11 @@ list of arguments:
 Sort mini-gallery thumbnails from files
 """""""""""""""""""""""""""""""""""""""
 
-The minigallery directive generates a gallery of thumbnails corresponding to
-the input file strings or object names.
+The :ref:`minigallery <minigalleries_to_examples>` directive generates a gallery of
+thumbnails corresponding to the input file strings or object names.
 You can specify minigallery thumbnails order via the ``minigallery_sort_order``
-configuration, which gets passed to :py:func:`sorted` ``key`` parameter when sorting
-all minigalleries.
+configuration, which gets passed to the :py:func:`sorted` ``key`` parameter when
+sorting all minigalleries.
 Sorting uses the gallery example filenames (e.g., ``plot_example.py``) and
 backreference filenames (e.g., ``numpy.exp.examples``)
 corresponding to all inputs.
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -709,7 +709,7 @@ corresponding to all inputs.
 See :ref:`own_sort_keys` for details on writing a custom sort key. Below is an
 example of using :class:`sphinx_gallery.sorting.FunctionSortKey` to put
 backreference thumbnails at the end. Note that
-backreference filenames do not start with 'plot_' and ``False`` gets sorted ahead
+backreference filenames do not start with "plot\_" and ``False`` gets sorted ahead
 of ``True`` (as 0 is less than 1).
 
 .. code-block:: python
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -51,6 +51,7 @@ file, inside a ``sphinx_gallery_conf`` dictionary.
 - ``minigallery_sort_order`` (:ref:`minigallery_order`)
 - ``backreferences_dir``, ``doc_module``, ``exclude_implicit_doc``,
   and ``inspect_global_variables`` (:ref:`references_to_examples`)
+- ``minigallery_sort_order`` (:ref:`minigalleries_to_examples`)
 
 **Images and thumbnails**
 
@@ -384,23 +385,8 @@ the converse does not hold.
 
 If you so desire you can implement your own sorting key. It will be
 provided the relative paths to `conf.py` of each sub gallery folder.
-
-.. warning:: If you create your own class for ``'subsection_order'``, ensure
-             that the ``__str__`` of your class is stable across runs.
-             Sphinx determines if the build environment has changed
-             (and thus if *all* documents should be rewritten)
-             by examining the config values using
-             ``md5(str(obj).encode()).hexdigest()`` in
-             ``sphinx/builders/html.py``. Default class instances
-             in Python have their memory address in their ``__repr__`` which
-             will in general change for each build. For ``ExplicitOrder``
-             for example, this is fixed via::
-
-                 def __repr__(self):
-                     return '<%s: %s>' % (self.__class__.__name__, self.ordered_list)
-
-             Thus the files are only all rebuilt if the specified ordered list
-             is changed.
+Note that it will be passed the subfolder path relative to the ``conf.py`` file.
+See :ref:`own_sort_keys` for more.
 
 .. _within_gallery_order:
 
@@ -419,8 +405,6 @@ which sorts the files based on the number of code lines::
         'within_subsection_order': NumberOfCodeLinesSortKey,
     }
 
-.. _sort_classes:
-
 In addition, multiple convenience classes are provided for use with
 ``within_subsection_order``:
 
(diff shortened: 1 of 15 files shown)


larsoner:
> If so, I'm also wondering if I could strip the thumbnail parent div out of the backref files so that I could do it in the loop.

Not sure if this will break anything or not. But you could do it and we could check to see if things still render properly.

> And if it makes sense to add a ... arg that will look in the folder for a readme that defines the sort order?

I'm not sure how it would work to use RST to define the order, but agree it makes sense to define it somewhere somehow.

A solution that would be more consistent with other order setting mechanics would be to provide a `sphinx_gallery_conf["minigallery_sort_order"]` that's a `callable` that takes whatever it needs to take (probably the current source file and list of examples or something?) and returns the sorted order or so.

lucyleeow:
> I have gone down way too deep an html/xml rabbit hole

I have nothing to contribute here but I just thank you for your hard work.

Also the CIs are failing because scipy docs site certificate has expired https://github.com/scipy/docs.scipy.org/issues/80

larsoner:
> does it matter if the heading is before or in the thumbnail parent div? my bias is for in b/c then it stays together as part of the DOM

I am not really sure but if I had to guess I would say no it shouldn't matter, so feel free to try moving it

larsoner:
> I think a standalone/separate PR would be to make something like a FileOrder(ExplicitOrder) object that can be configured with the file it should look for the ordering

Thinking about this more... I'm not sure we can/should implement this at the SG end, at least to start. You could already implement and use this in `matplotlib` because you can define your own sortkey if you want. For example following this code:

https://sphinx-gallery.github.io/stable/_modules/sphinx_gallery/sorting.html#ExampleTitleSortKey

I'd expect something like the following, where the `...` is whatever you need to do to parse the RST in whatever way matplotlib wants to specify the information:

```
class RSTSortKey:

    def __init__(self, src_dir):
        self.src_dir = src_dir
        readme_txt = Path(src_dir) / "README.rst").read_text("utf-8")
        self.fname_list_orderered = ...  # something you pull from the readme

    def __call__(self, filename):
        return self.fname_list_ordered.index(filename)
```

larsoner:
... in other words, I think the hard/interesting work here is in the `...` above, and that's going to be a bit specific to how matplotlib does things. But assuming that works, it would be great to link from the SG config docs to the matplotlib code showing how the `SortKey`-based framework is general enough to support this use case!

larsoner:
> Is it okay to just apply it to the files? I can't tell how the sorting mechanism is handled for the backreferences and I'm concerned that implementing it would require parsing each backreferences file and reordering and I think that's out of scope for this PR.

Can you clarify what you mean? I don't see why this PR would need to mess with the backreferences files

doc/configuration.rst
```diff
@@ -598,6 +598,33 @@ examples into a single mini-gallery, e.g.:
...
+
+The order of the minigallery file based entries can be set using the 
+``minigallery_sort_order`` configuration key. This configuration key supports
+any callable that is a valid *key* argument to ``sorted``::
+
+    sphinx_gallery_conf = {
+    ...,
+    'minigallery_sort_order' : lambda x: return len(x)%4,
+    ... 
+    }
```
larsoner, line 625: Instead of a trivial example here, could you make one that actually does something interesting related to the example above with `numpy.exp ../examples/...`? I think it will be easier for people to understand the use case

doc/configuration.rst
```diff
@@ -598,6 +598,35 @@ examples into a single mini-gallery, e.g.:
...
+
+The order of the gallery entries is determined by the argument order. The ordering of 
+files matching a specific glob pattern, e.g. ``../examples/plot_4*`` can be set using the 
+``minigallery_sort_order`` configuration key. This configuration key supports
+any callable that is a valid *key* argument to ``sorted``::
+
+    sphinx_gallery_conf = {
+    ...,
+    # put duplicate figure at end 
+    'minigallery_sort_order' : lambda path: not path.startswith('plot_4b')
```
larsoner, line 625: Wouldn't this be better, i.e., sort first based on "plot_4b"-ness then on the filename itself?
```suggestion
    'minigallery_sort_order' : lambda path: (not path.startswith('plot_4b'), path)
```

sphinx_gallery/directives.py
```diff
@@ -11,11 +12,26 @@
...
+from .gen_rst import extract_intro_and_title
+from .py_source_parser import split_code_and_text_blocks
+
+
 class MiniGallery(Directive):
     """Custom directive to insert a mini-gallery.
 
-    The required argument is one or more fully qualified names of objects,
-    separated by spaces.  The mini-gallery will be the subset of gallery
+    The required argument is a one or more of the following:
```
larsoner, line 27: Think this is required for RST to render properly
```suggestion
    The required argument is a one or more of the following:

```
and https://output.circle-artifacts.com/output/job/ff1728af-0451-4fef-bab5-967229bb60df/artifacts/0/html/gen_modules/sphinx_gallery.directives.html#sphinx_gallery.directives.MiniGallery seems to suggest this as well

larsoner (commented):
There are now a lot of empty sections like

https://output.circle-artifacts.com/output/job/ff1728af-0451-4fef-bab5-967229bb60df/artifacts/0/html/gen_modules/sphinx_gallery.directives.html#examples-using-sphinx-gallery-directives-visit-imgsg-latex

I don't see them on `dev`

https://sphinx-gallery.github.io/dev/gen_modules/sphinx_gallery.directives.html#sphinx_gallery.directives.MiniGallery

doc/configuration.rst
```diff
@@ -598,6 +598,42 @@ examples into a single mini-gallery, e.g.:
...
+the input ``obj`` and the filepath ``path``:
+
+.. code-block:: rst
+    
+    sphinx_gallery_conf = {
+    ...,
+     "minigallery_sort_order": lambda x: (
+        not (x[0] == "numpy.exp"),
+        not x[1].startswith("plot_4b"),
+        x[1],
```
larsoner, line 629: Okay one last very annoying (and tricky!) thing -- it's important that whatever is in the `sphinx_gallery_conf` have a stable `__repr__` from run to run, otherwise rebuilds will happen for all files on sequential calls to `sphinx_build` even if no files have changed. Unfortunately, we can't guarantee that `lambda`'s `__repr__` will be the same between subsequent runs of Sphinx. So really this needs to be something like:
```
class MinigallerySortOrder:

    def __repr__(self):
         return "MinigallerySortOrder"

    def __call__(self, fname):
        return ...
```
then you pass `"minigallery_sort_order": MinigallerySortOrder(),` as the conf entry. See the `warning` in https://sphinx-gallery.github.io/stable/configuration.html#sorting-gallery-subsections for example

larsoner (commented):
Code looks good to me. Does this get what you need for `matplotlib`?

larsoner:
Okay, let me know once the last doc update is done and I'll merge!

lucyleeow (commented):
Just a flyby review and a lot of things are nitpicks.

I noted in tinybuild that the thumbnails are one per line:

image

Would it be possible to make it a few per line, like backreferences does?

doc/configuration.rst
```diff
@@ -598,6 +598,42 @@ examples into a single mini-gallery, e.g.:
...
+configuration key. This configuration key supports any callable that is a valid 
+*key* argument to ``sorted`` and is provided two parameters to sort on, 
+the input ``obj`` and the filepath ``path``:
+
+.. code-block:: rst
+    
+    sphinx_gallery_conf = {
+    ...,
+     "minigallery_sort_order": lambda x: (
+        not (x[0] == "numpy.exp"),
```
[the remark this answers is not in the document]
lucyleeow: Oh maybe an explanation about what we are doing (i'm probably just bad with `key`) ?

lucyleeow:
Ahh it could be me, let me try again!

doc/configuration.rst
```diff
@@ -598,6 +598,42 @@ examples into a single mini-gallery, e.g.:
 For such a mini-gallery, specifying a custom heading message is recommended
 because the default message is vague: "Examples of one of multiple objects".
 
+Create mini-galleries using file paths
+""""""""""""""""""""""""""""""""""""""
+Sometimes you may want to explicitly create a mini-gallery for files that may be 
+in the same location but not have functions in common, for example a set of 
+tutorials. The mini-gallery directive supports passing in a string path and
```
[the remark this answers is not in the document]
lucyleeow: Maybe this could be clearer?

sphinx_gallery/tests/test_full.py
```diff
@@ -1069,69 +1070,112 @@ def test_backreference_labels(sphinx_app):
...
 
-            # Check for examples
-            assert explicitorder_example.search(text) is not None
-            if filenamesortkey:
-                assert filenamesortkey_example.search(text) is not None
             else:
-                assert filenamesortkey_example.search(text) is None
+                assert not (img or href)
+
+    print(f"{test}: {lxml.html.tostring(text)}")
```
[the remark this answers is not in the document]
lucyleeow: Ahh no worries, lets keep!

lucyleeow:
> let me try again!

Hmm I got the same result... This is the html

```html
</div><p>Test 2-F-G</p>
<div class="sphx-glr-thumbnails"><div class="sphx-glr-thumbcontainer" tooltip="This example tests that the alt text is generated correctly for matplotlib figures."><img alt="" src="_images/sphx_glr_plot_matplotlib_alt_thumb.png" />
<p><a class="reference internal" href="auto_examples/plot_matplotlib_alt.html#sphx-glr-auto-examples-plot-matplotlib-alt-py"><span class="std std-ref">Matplotlib alt text</span></a></p>
  <div class="sphx-glr-thumbnail-title">Matplotlib alt text</div>
</div><div class="sphx-glr-thumbcontainer" tooltip="Setting the Matplotlib backend"><img alt="" src="_images/sphx_glr_plot_matplotlib_backend_thumb.png" />
<p><a class="reference internal" href="auto_examples/plot_matplotlib_backend.html#sphx-glr-auto-examples-plot-matplotlib-backend-py"><span class="std std-ref">Setting the Matplotlib backend</span></a></p>
  <div class="sphx-glr-thumbnail-title">Setting the Matplotlib backend</div>
```

What's the difference between mine and yours?

lucyleeow:
So in mine, I didn't have the last line:

```html
</div></div>
```

it looks like

```html
<p><a class="reference internal" href="auto_examples/plot_matplotlib_backend.html#sphx-glr-auto-examples-plot-matplotlib-backend-py"><span class="std std-ref">Setting the Matplotlib backend</span></a></p>
  <div class="sphx-glr-thumbnail-title">Setting the Matplotlib backend</div>
</div><p>Test 3-F-G-B</p>
```

Which only shows where the difference comes from. I just pulled your branch down so it's up to date :shrug:

Could the difference (between this and other mini-galleries) be because we're using the default `is_backref=False` in `_thumbnail_div` ?

Maybe if it works @larsoner, I'll just put it down to me?

doc/configuration.rst
```diff
@@ -598,6 +598,42 @@ examples into a single mini-gallery, e.g.:
...
+
+
+.. minigallery:: numpy.exp ../examples/plot_0_sin.py ../examples/plot_4*
+   :add-heading: Mini-gallery using paths
+   :heading-level: -
+
+The ordering of thumbnails can be set using the ``minigallery_sort_order`` 
+configuration key. This configuration key supports any callable that is a valid 
+*key* argument to ``sorted`` and is provided two parameters to sort on, 
+the input ``obj`` and the filepath ``path``:
```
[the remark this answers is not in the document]
lucyleeow: > Also, should I just keep it simple and only sort on paths?

Sorry I don't follow. As in only provide `path` and not `obj`?

lucyleeow:
Okay false alarm, it was totally my environment, not sure how I managed to stuff it up. Works fine in new clean env.

doc/configuration.rst
```diff
@@ -598,6 +600,61 @@ examples into a single mini-gallery, e.g.:
 For such a mini-gallery, specifying a custom heading message is recommended
 because the default message is vague: "Examples of one of multiple objects".
 
+.. _file_based_minigalleries:
+
+Create mini-galleries using file paths
+======================================
+
+Sometimes you may want to explicitly create a :class:`mini-gallery <sphinx_gallery.directives.MiniGallery>` 
+for files that may be in the same location but not have functions in common, for example a set of
```
lucyleeow, line 609: nit - line length?

doc/configuration.rst
```diff
@@ -598,6 +600,61 @@ examples into a single mini-gallery, e.g.:
...
+   :heading-level: ^
+
+Sort mini-gallery thumbnails from files
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The ordering of thumbnails can be set using the ``minigallery_sort_order`` 
+configuration key. Minigallery provides two values to sort on:
+
+* input: object name, file name, or glob string
+* path: backreferences ``{input}.example`` or ``{gallery_example}.py`` file
```
lucyleeow, line 632:
```suggestion
* path: backreferences ``{input}.example`` or ``{gallery_example}.py`` file string
```

doc/configuration.rst
```diff
@@ -598,6 +600,61 @@ examples into a single mini-gallery, e.g.:
...
+
+
+.. minigallery:: numpy.exp ../examples/plot_0_sin.py ../examples/plot_4*
+    :add-heading: Mini-gallery using paths
+   :heading-level: ^
+
+Sort mini-gallery thumbnails from files
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The ordering of thumbnails can be set using the ``minigallery_sort_order``
```
lucyleeow, line 628: Can we make explicit that all minigalleries created with the `minigallery` directive will use the single `minigallery_sort_order` config value?

doc/configuration.rst
```diff
@@ -598,6 +600,61 @@ examples into a single mini-gallery, e.g.:
...
+Sort mini-gallery thumbnails from files
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The ordering of thumbnails can be set using the ``minigallery_sort_order`` 
+configuration key. Minigallery provides two values to sort on:
+
+* input: object name, file name, or glob string
+* path: backreferences ``{input}.example`` or ``{gallery_example}.py`` file
+
+The sorting key object is passed to :py:func:`sorted` must have the following:
```
lucyleeow, line 634:
```suggestion
The sorting key object passed to :py:func:`sorted` must have the following:
```

doc/configuration.rst
```diff
@@ -598,6 +600,61 @@ examples into a single mini-gallery, e.g.:
...
 because the default message is vague: "Examples of one of multiple objects".
 
+.. _file_based_minigalleries:
+
+Create mini-galleries using file paths
+======================================
+
+Sometimes you may want to explicitly create a :class:`mini-gallery <sphinx_gallery.directives.MiniGallery>` 
+for files that may be in the same location but not have functions in common, for example a set of 
+tutorials. The mini-gallery directive therefore supports passing in:
```
lucyleeow, line 610:
```suggestion
tutorials. The mini-gallery directive therefore supports passing in
a list (space separated) of any of the following:
```

doc/configuration.rst
```diff
@@ -598,6 +600,61 @@ examples into a single mini-gallery, e.g.:
...
+    "minigallery_sort_order": FunctionSortKey(lambda x: (
+        not (x[0] == "numpy.exp"), # input
+        not x[1].startswith("plot_4b"), # path
+        x[1],
+    )),
+    #... 
+    }
+
+Sorting is not supported for thumbnails generated from references, as described in 
+:ref:`references_to_examples`.
```
lucyleeow, line 655:
```suggestion
Sorting is not supported for a set/subset of thumbnails generated from
 :ref:`backreferences <references_to_examples>` i.e., the thumbnails
 linked to a qualified object name input.
```

I get what this means now. It's hard to document this clearly, I tried my best.

lucyleeow (commented):
More nits!

doc/configuration.rst
lucyleeow: Ohh sorry I was not clear and got mixed up. I did not mean to include backreferences. I mean to make it clear that the same `minigallery_sort_order` is used for all file minigalleries , but now that I think about it I don't think you'd ever want use different sort keys for different galleries. Resolving.

doc/configuration.rst
```diff
@@ -569,34 +615,66 @@ To exclude the functions mentioned above you would use
...
+
+
+.. minigallery:: numpy.exp ../examples/plot_0_sin.py ../examples/plot_4*
+    :add-heading: Mini-gallery using paths
+    :heading-level: ^
+
+Sort mini-gallery thumbnails from files
+"""""""""""""""""""""""""""""""""""""""
+
+The mini-gallery directive generates a gallery thumbnails corresponding to the
```
lucyleeow, line 644: "a gallery of thumbnails" ? or "generates gallery thumbnails"

lucyleeow:
> change minigallery to only sort on path?

I don't use SG to really have an opinion on this. I guess maybe you may want to put backreference thumbnails at the end or start ?

> expand sorting to backreferences? 

I would go with YAGNI for this one.

larsoner (approved):
Agreed the list is nicer!

lucyleeow:
Sorry this feel off my radar somehow. I will take a look this weekend or early next week. Sorry for holding this up.

lucyleeow (approved):
Thank you for all your work. Just some doc nitpicks from me!

sphinx_gallery/directives.py
```diff
@@ -65,32 +91,78 @@ def run(self):
             heading_level = self.options.get("heading-level", "^")
             lines.append(heading_level * len(heading))
 
+        src_dir = config.sphinx_gallery_conf["src_dir"]
+
         def has_backrefs(obj):
-            src_dir = config.sphinx_gallery_conf["src_dir"]
-            path = os.path.join(src_dir, backreferences_dir, f"{obj}.examples")
-            return os.path.isfile(path) and os.path.getsize(path) > 0
+            path = os.path.normpath(
```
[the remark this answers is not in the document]
lucyleeow: If that's not too much to ask, thank you!

doc/configuration.rst
```diff
@@ -569,34 +625,60 @@ To exclude the functions mentioned above you would use
...
-    .. minigallery:: numpy.exp numpy.sin
-        :add-heading: Mini-gallery using ``numpy.exp`` or ``numpy.sin``
-        :heading-level: -
+    .. minigallery:: numpy.exp ../examples/plot_0_sin.py ../examples/plot_4*
+
+
+.. minigallery:: numpy.exp ../examples/plot_0_sin.py ../examples/plot_4*
+    :add-heading: Mini-gallery using paths
+    :heading-level: ^
+
```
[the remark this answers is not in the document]
lucyleeow: Ah sorry, I didn't see it in context. Ignore in that case, sorry!

doc/configuration.rst
```diff
@@ -569,34 +629,59 @@ To exclude the functions mentioned above you would use
...
+Sort mini-gallery thumbnails from files
+"""""""""""""""""""""""""""""""""""""""
+
+The mini-gallery directive generates a gallery of thumbnails corresponding to 
+the input file strings or object names. The ordering of these thumbnails can be 
+set using the ``minigallery_sort_order`` configuration key. The minigallery 
+directive passes the following information to :py:func:`sorted`:
+
+* sphinx-gallery :ref:`sorting key <sort-warning>` method
```
lucyleeow, line 663: I don't quite follow this. I thought the `sorted` function got the `minigallery_sort_order` function?

lucyleeow (commented):
Sorry, another question!

Also could we add a 'minigallery' / `minigallery_sort_order` directive item [hereish](https://github.com/sphinx-gallery/sphinx-gallery/blob/1bdb4b9a4d2000e025c5acd7e1771239d582ac6b/doc/configuration.rst#L53)

lucyleeow:
Thank you! I have added a section on creating your own sort key function, I thought it fit better.

I just noticed that it seems that what is passed to the sort key is the full path to the backreference 'examples' files e.g.,:
`/home/lucy/Documents/sphinx-gallery/doc/gen_modules/backreferences/numpy.exp.examples`
and some weird amalgamation of full and relative path to the example files:
`/home/lucy/Documents/sphinx-gallery/doc/../examples/plot_0_sin.py` ?

If you can confirm on your system, maybe we could fix and amend the test to catch this?

lucyleeow:
> So pushed into a separate commit in case you want to reject, because I understand why you'd want to frontload the FunctionSortKey but I pulled the repr discussion up 'cause I thought the section flow was jumpy otherwise.

That's fine, I've used this ordering.

Some minor changes:

* I've just used `.name` to get the filename instead of the full path (it's not pretty but we were manipulating the sort elements anyway so :shrug:  - did not find a way to test this directly but I have amended `MockSort` so it will fail if not the filename
* I've allowed `None` as a `minigallery_sort_order` input - could not think of way to test this directly, though i noticed we don't test this for `subsection_order` either
* I don't think the doc config `minigallery_sort_order` in the SG docs was working (as we were passing the full path before) and I didn't think it was necessary. Also now something fails if `minigallery_sort_order` does not allow `None`
* Doc amendments. I have realised that you cannot use `FunctionSortKey` for `within_subsection_order` config as they get instantiated with the source dir

lucyleeow:
Thanks @story645 for all your work. Just made minor fixes to save you time.

lucyleeow:
Ah sorry I missed that. What making the making the paths relative to `src_dir` ?

We could probably do that here:
```python
            elif paths := Path(src_dir).glob(obj):
                file_paths.extend([(obj, p) for p in paths])
```

?

Maybe we could add a test to capture this? And update the documentation so people know what we're sorting on.

Let me know if you don't want to work on it.

lucyleeow:
That's fine. We can do it in the sortkey, maybe adding a `parent` ? I think we only ever go 1 folder deep.

lucyleeow:
Sure I'm okay with it the previous status, but it should be documented so people know what we're sorting on (and that it's not just the filename). And tested if possible...

lucyleeow:
Actually, the filepath input arg to the directive would be okay? Since they have to give the filepath relative to `conf.py` so subfolder structure would be retained?

lucyleeow:
Oh yeah good point.

I realise now why I thought it should be the filename, the example we give is along the lines of `x.startswith('plot')`. That would be trickier with the full path, they would have split the path but that's unavoidable if we want the subdir in there.

lucyleeow:
> Ehh, Matplotlib does some wonky nested nonsense so I'd rather be as general as possible.

I meant that I think sphinx gallery only looks 1 level down for subdirs, though matplotlib may do funny things to get around this?

sphinx_gallery/directives.py
```diff
@@ -67,34 +96,67 @@ def run(self):
...
+                        config.sphinx_gallery_conf["examples_dirs"],
+                        config.sphinx_gallery_conf["gallery_dirs"],
+                    )
+                    if (obj.find(e) != -1)
+                ]
+                if len(dirs) != 1:
+                    raise ExtensionError(
+                        f"Error in gallery lookup: input={obj}, matches={dirs}, "
+                        f"examples={config.sphinx_gallery_conf['examples_dirs']}"
+                    )
```
lucyleeow, line 142: @story645 I stumbled on this error accidentally but I don't fully follow what situation this is meant to catch?

I managed to contrive a case where I added a file to a path not in `examples_dir` to a minigallery directive. In that case there were no matches and the error message didn't really make sense.

I assume it is meant to catch cases when `dirs>1`? So maybe a sub-gallery directory is named the same as a examples gallery?

lucyleeow: A more informative error message would be great. There is no test for this but depending on what this is mean to catch, maybe it would be too difficult to write a test.

lucyleeow: cc @larsoner just in case you remember...? :grimacing:

larsoner: No idea :disappointed:

PR #1229, opened 2023-11-17
MAINT Add `extras_require` in `setup.py`

lucyleeow:
@larsoner let me know if I have missed anything!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,13 @@ description = description.strip()
 with open("requirements.txt") as fid:
     install_requires = [line.strip() for line in fid if line.strip()]
 
-extras_require = {"recommender": ["numpy"]}
+extras_require = {
+    "recommender": ["numpy"],
+    "show_api_usage": ["graphviz"],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -30,7 +30,13 @@
 with open("requirements.txt") as fid:
     install_requires = [line.strip() for line in fid if line.strip()]
 
-extras_require = {"recommender": ["numpy"]}
+extras_require = {
+    "recommender": ["numpy"],
+    "show_api_usage": ["graphviz"],
+    "show_memory": ["memory_profiler"],
+    "thumbnail": ["pillow"],
```
larsoner, line 37: We should make `pillow` a core requirement because I think we always scale thumbnails

lucyleeow: Yes I did notice that. We currently only have sphinx as a dependency. Will add pillow and amend docs.

larsoner:
Thanks @lucyleeow !

PR #1241, opened 2023-12-15
Simplify Matplotlib scraper

QuLogic:
Three things:
- Validate `image_srcset` once during configuration, instead of repeatedly on every scraper run. Note, I also passed the scaling numbers through a set, in order to prevent saving extra times.
- Simplify the animation part of the scraper; instead of looping through all animations to match a figure, use a dictionary for direct lookup.
- Add a note about `kwargs` to `matplotlib_scraper`. I was confused why nothing seemed to call it with `kwargs`, but was able to find it in other parts of the documentation. So I mentioned that in the docstring as well.

### DIFF
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -284,6 +284,30 @@ def _fill_gallery_conf_defaults(sphinx_gallery_conf, app=None, check_keys=True):
     except (ImportError, ValueError):
         pass
 
+    # Check for srcset hidpi images
+    srcset = gallery_conf["image_srcset"]
+    if not isinstance(srcset, (list, tuple)):
+        raise ConfigError(
+            "image_srcset must be a list of strings with the "
+            'multiplicative factor followed by an "x", '
+            'e.g. ["2.0x", "1.5x"]'
+        )
+    srcset_mult_facs = set()
+    for st in srcset:
+        if not isinstance(st, str) or st[-1:] not in ("", "x"):
+            raise ConfigError(
+                f"Invalid value for image_srcset parameter: {st!r}. "
+                "Must be a list of strings with the multiplicative "
+                'factor followed by an "x".  e.g. ["2.0x", "1.5x"]'
+            )
+        if st == "":
+            continue
+        # "2x" = "2.0"
+        srcset_mult_facs.add(float(st[:-1]))
+    srcset_mult_facs -= {1}  # 1x is always saved.
+    gallery_conf["image_srcset"] = [*sorted(srcset_mult_facs)]
+    del srcset, srcset_mult_facs
+
     # compress_images
     compress_images = gallery_conf["compress_images"]
     if isinstance(compress_images, str):
--- a/sphinx_gallery/scrapers.py
+++ b/sphinx_gallery/scrapers.py
@@ -131,27 +131,21 @@ def matplotlib_scraper(block, block_vars, gallery_conf, **kwargs):
     srcset = gallery_conf["image_srcset"]
 
     # Check for animations
-    anims = list()
+    anims = {}
     if gallery_conf["matplotlib_animations"]:
         for ani in block_vars["example_globals"].values():
             if isinstance(ani, Animation):
-                anims.append(ani)
+                anims[ani._fig] = ani
     # Then standard images
     for fig_num, image_path in zip(plt.get_fignums(), image_path_iterator):
         image_path = PurePosixPath(image_path)
         if "format" in kwargs:
             image_path = image_path.with_suffix("." + kwargs["format"])
-        # Set the fig_num figure as the current figure as we can't
-        # save a figure that's not the current figure.
+        # Convert figure number to Figure.
         fig = plt.figure(fig_num)
         # Deal with animations
-        cont = False
-        for anim in anims:
-            if anim._fig is fig:
-                image_rsts.append(_anim_rst(anim, str(image_path), gallery_conf))
-                cont = True
-                break
-        if cont:
+        if anim := anims.get(fig):
+            image_rsts.append(_anim_rst(anim, image_path, gallery_conf))
             continue
         # get fig titles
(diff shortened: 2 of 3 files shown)


QuLogic:
Interesting, the Pillow writer doesn't support `Path`s; I'll put a `str` back in there, but fix it in Matplotlib as well.

sphinx_gallery/scrapers.py
```diff
@@ -128,44 +131,24 @@ def matplotlib_scraper(block, block_vars, gallery_conf, **kwargs):
...
-            )
 
     # Check for animations
-    anims = list()
+    anims = {}
     if gallery_conf["matplotlib_animations"]:
         for ani in block_vars["example_globals"].values():
             if isinstance(ani, Animation):
-                anims.append(ani)
+                anims[ani._fig] = ani
```
larsoner, line 141: I worry about using the `._fig` private attribute here

larsoner (commented):
Just one minor concern, otherwise looks useful!

sphinx_gallery/scrapers.py
QuLogic: That's what the original code used; it's just been moved.

sphinx_gallery/scrapers.py
```diff
@@ -109,10 +109,13 @@ def matplotlib_scraper(block, block_vars, gallery_conf, **kwargs):
...
-        :meth:`~matplotlib.figure.Figure.savefig`, e.g. ``format='svg'``.
-        The ``format`` kwarg in particular is used to set the file extension
-        of the output file (currently only 'png', 'jpg', 'svg', 'gif', and
-        'webp' are supported).
+        :meth:`~matplotlib.figure.Figure.savefig`, e.g. ``format='svg'``. The
+        ``format`` keyword argument in particular is used to set the file
+        extension of the output file (currently only 'png', 'jpg', 'svg',
+        'gif', and 'webp' are supported).
+
+        This is not used internally, but intended to be used when overriding
```
lucyleeow, line 117:
```suggestion
        This is not used internally, but intended for used when overriding
```

Small nitpick

sphinx_gallery/tests/test_scrapers.py
```diff
@@ -80,7 +80,7 @@ def test_save_matplotlib_figures(gallery_conf, ext):
 def test_save_matplotlib_figures_hidpi(gallery_conf):
     """Test matplotlib hidpi figure save."""
     ext = "png"
-    gallery_conf["image_srcset"] = ["2x"]
+    gallery_conf["image_srcset"] = [2]
```
lucyleeow, line 83: Sorry, I'm confused, 'image_srcset' does not have to be a string with 'x' at the end?

lucyleeow (commented):
Thank you! Nitpick and some questions.

sphinx_gallery/scrapers.py
QuLogic: That is not grammatically correct, unless you mean "for use when..."?

sphinx_gallery/gen_gallery.py
```diff
@@ -284,6 +284,30 @@ def call_memory(func):
...
+    srcset = gallery_conf["image_srcset"]
+    if not isinstance(srcset, (list, tuple)):
+        raise ConfigError(
+            "image_srcset must be a list of strings with the "
+            'multiplicative factor followed by an "x", '
+            'e.g. ["2.0x", "1.5x"]'
+        )
+    srcset_mult_facs = set()
+    for st in srcset:
+        if not isinstance(st, str) or st[-1:] not in ("", "x"):
```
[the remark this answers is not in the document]
QuLogic: The previous version did not check types, so this adds that, and I wanted to only have one `raise`.

sphinx_gallery/tests/test_scrapers.py
QuLogic: It does, but now it's normalized in the configuration initialization, to avoid doing that on every scrape. But these tests appear to skip that step. I could add that here (and actually maybe that should be tested as well.)

sphinx_gallery/scrapers.py
lucyleeow: Yes, sorry I've amended. I just want to make clearer that it is API available 'for' the user to use.

sphinx_gallery/tests/test_scrapers.py
lucyleeow: Ahh right, that is confusing. 

> I could add that initialization here instead

+1

> actually maybe that should be tested as well.)

+1

QuLogic: Done, though this spilled over into some other tests. Not sure if that warrants a separate PR.

larsoner (approved):
LGTM, @lucyleeow feel free to merge if you're happy!

lucyleeow:
Thanks @QuLogic !

sphinx_gallery/gen_gallery.py
```diff
@@ -284,6 +284,30 @@ def call_memory(func):
...
+            raise ConfigError(
+                f"Invalid value for image_srcset parameter: {st!r}. "
+                "Must be a list of strings with the multiplicative "
+                'factor followed by an "x".  e.g. ["2.0x", "1.5x"]'
+            )
+        if st == "":
```
lucyleeow, line 303: @QuLogic was just looking at our check config func - why do we want to allow item in `gallery_conf["image_srcset"]` to be an empty string?

QuLogic: That preserved the previous code; I think it's always been that way since its initial PR, but as of https://github.com/sphinx-gallery/sphinx-gallery/pull/808#issuecomment-823736082, I think that was dropped from the documentation or usage, but the config handling part never did.

lucyleeow: Ah thank you :pray: , I have a vague memory of empty string being a config option previously now.

PR #1244, opened 2023-12-18
BUG: `UnicodeDecodeError` in recommender

Charlie-XIAO:
See also scikit-learn/scikit-learn#27969. Without specifying the encoding such as utf-8, some systems would use other codecs (e.g. gbk) which cannot decode some characters (e.g. `–`, `“`, `”`). This PR specifies `encoding="utf-8"`.


larsoner (approved):

larsoner:
Thanks @Charlie-XIAO !

PR #1253, opened 2024-01-31
ENH: minigallery_sort_order on full path

story645:
Follow on from #1226, this pr:
* has the sortkey work on full path to account for subdirectories
* fixes the sortkey test and adds check that test is working
* updates docs


lucyleeow:
Thank you!

PR #1263, opened 2024-02-26
DOC Use 'nested_sections' `True` for docs

### LINKED ISSUE #1262: Sub-folder gallery TOC header missing in PST theme
I noticed that in the LHS contents menubar there is no sub-folder gallery name. E.g. our `examples/no_output` examples start with 'A short Python script', but there is no header to indicate the start of the 'no output' sub-gallery. (The RHS 'on this page' menu does show the 2 galleries but I think that's because that is what the headings in that `index` page are.)

image

Having a look at wayback machine, this is what it used to look like in the old theme:

image

So maybe it's a CSS change? I'm not very familiar with this, maybe @drammock would have an insight?

lucyleeow:
closes #1262

It seems PST does not have the TOC problems described in https://github.com/sphinx-gallery/sphinx-gallery/pull/944#discussion_r916603188 and gives us the sub gallery sub heading in the LHS TOC.

I'll double check the rendered docs to make sure it looks okay.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -474,7 +474,7 @@ sphinx_gallery_conf = {
     "capture_repr": ("_repr_html_", "__repr__"),
     "matplotlib_animations": True,
     "image_srcset": ["2x"],
-    "nested_sections": False,
+    "nested_sections": True,
     "show_api_usage": True,
 }
 


lucyleeow:
Failure seems unrelated, I'll look at it tomorrow.

PR #1264, opened 2024-03-07
lint: switch from Black to Ruff's "Black"

Borda:
As Ruff [claims](https://pypi.org/project/ruff/0.0.47/#:~:text=ruff%20is%20compatible%20with%20Black,that%20are%20obviated%20by%20autoformatting.) full compatibility with Black, it makes sense just to use one tool for both to prevent eventual collisions

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,13 +1,9 @@
 repos:
-  - repo: https://github.com/psf/black
-    rev: 24.2.0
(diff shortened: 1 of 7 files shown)


Borda:
@larsoner @lucyleeow, maybe I get some clarity on the falling label check? :rabbit:

larsoner:
Thanks @Borda !

PR #1265, opened 2024-03-07
ci: associate `install.sh` with used job

Borda:
This is just a minor cleaning to reduce root structure.
The `install.sh` is used in a single place in GH actions. I would suggest moving it to the same folder.


.github/workflows/tests.yml
```diff
@@ -3,12 +3,8 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.ref }}
   cancel-in-progress: true
 on:  # yamllint disable-line rule:truthy
-  push:
-    branches:
-      - '*'
-  pull_request:
-    branches:
-      - '*'
+  push: {}
```
Borda, line 6: Note the meaning is the same; on the other hand, I would suggest running them only against the master to reduce the number of jobs running in parallel the PR comes from the root repo

.github/workflows/tests.yml
```diff
@@ -3,12 +3,8 @@ concurrency:
...
+  pull_request: {}
```
Borda, line 7: Note the meaning is the same; on the other hand, I would suggest running them only against the master to reduce the number of jobs running in parallel the PR comes from the root repo

```suggestion
 push:
    branches: ['master']
  pull_request:
    branches: ['master']
```

lucyleeow (approved):
Makes sense now that we are only using GH actions. I'll let @larsoner merge as he's more familiar with the CIs. 

Thanks @Borda !

larsoner (approved):
Thanks @Borda

PR #1268, opened 2024-03-10
allow call script as pkg entry

Borda:
These additional scripts are quite uneasy to handle and maintain; I would suggest calling it from the package directly, and the one with copy drop with 1:1 explanation in docs (as if you try to call the copy shell command, it messes with your local env while development)

ref: https://stackoverflow.com/a/782984/4521646

would simplify work on #1267 -> https://setuptools.pypa.io/en/latest/userguide/entry_point.html
cc: @lucyleeow


larsoner:
Agreed this is a better approach, and hopefully not too hard to adopt for people that had previously been using the scripts  (which was probably rare in the first place) so I think safe enough to merge without a deprecation cycle

PR #1269, opened 2024-03-11
pyproject: cleaning pytest config

Borda:
Cut part of would #1267 as simplification of `pytest` config improving reliability in IDE
also adding `--color=yes` to make the test logs easier to navigate :rabbit: 

cc: @lucyleeow


larsoner:
Thanks @Borda

PR #1270, opened 2024-03-11
ci: simplify check-type-label [wip]

Borda:
This is just marginal readability simplification...

On the other hand, thinking about how to play with this check since it seems to bring some confusion for external/forked PRs. I fully understand and agree with the purpose, just adding some context/ my perception as a contributor... I came to create a PR, and I have many different projects I usually just see the aggregated list, and on the list, I tend to re-visit PRs with some failing checks to make them :green_circle: so to know all is fine on my side and could be processed further. But in this case, I will always have this check :red_square: until a maintainer adds a label...


larsoner:
This does not appear to work:
```
Error: Unable to resolve action `danielchabr/pr-labels-checker@v3`, unable to find version `v3`
```

.github/workflows/label-check.yml
```diff
@@ -2,24 +2,16 @@ name: Labels
...
+    # skip for draft PRs
+    if: github.event.pull_request.draft == false
     steps:
-      - run: |
-          echo "Labels: $LABELS"
-          if [[ "$LABELS" != *"bug"* ]] && [[ "$LABELS" != *"enhancement"* ]] && [[ "$LABELS" != *"api"* ]] && [[ "$LABELS" != *"maintenance"* ]] && [[ "$LABELS" != *"documentation"* ]]; then
-            echo "Please add a type label to this PR"
-            exit 1
-          fi
+      - uses: danielchabr/pr-labels-checker@v3
```
Borda, line 14:
```suggestion
      - uses: danielchabr/pr-labels-checker@v3.3
```

Borda:
@larsoner changed to another action which seems to be working fine :flamingo:

Borda:
Well, it seems that neither I hoped worked correctly, and the last option I found is based on [dockers](https://github.com/marketplace/actions/label-checker-for-pull-requests), which I am not familiar with, so let's leave this as it is...

larsoner:
Okay, feel free to improve the error message if you think it would help though! Can be part of another PR like the `pyproject` one if you want

PR #1271, opened 2024-03-12
More informative title for 'check label' CI workflow

lucyleeow:
Follows on from #1270, I agree this CI test is confusing and its annoying to have it be the cause of a PR being 'red'.

This is just a temporary, slight improvement, to at least help prevent confusion for new contributors: change the CI workflow name so contributors can easily see why this test has failed. Happy to change change wording. It's similar to what scikit-learn has for their 'Check Changelog'  check.

Also skips drafts and improves message, copied from #1270.


larsoner:
Thanks @lucyleeow

PR #1275, opened 2024-03-20
BUG: Fix errors in example usage of ignore_repr_types and reset_argv

speth:
- The example of how to use `reset_argv` is wrong, or at least misleading on two counts. One, it must return an iterable in all cases, so there should be a default case returning `[]`. Two, the value of `script_vars['src_file']` is an absolute path, not just a filename, so just `sample1.py` won't ever be matched.
- Rewrite the sample regex for `ignore_repr_types` so it actually does what's described in the text.


larsoner:
Thanks @speth !

PR #1278, opened 2024-03-31
Remove leftover config checking of `image_srcset`

lucyleeow:
`image_srcset` config added in #808. Originally in that PR an empty string was meant to mean save the usual image: https://github.com/sphinx-gallery/sphinx-gallery/pull/808#issue-855178549. Later on we changed it to never neededing the empty string: https://github.com/sphinx-gallery/sphinx-gallery/pull/808#issuecomment-823736082

This removes the redundant config checking as `image_srcset` should not be taking an empty string.
 
First noticed: https://github.com/sphinx-gallery/sphinx-gallery/pull/1241/files#r1542174678

### DIFF
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -294,15 +294,13 @@ def _fill_gallery_conf_defaults(sphinx_gallery_conf, app=None, check_keys=True):
         )
     srcset_mult_facs = set()
     for st in srcset:
-        if not isinstance(st, str) or st[-1:] not in ("", "x"):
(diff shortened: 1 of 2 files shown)


lucyleeow:
@larsoner AFAICT your comment is still relevant for this config?

> I wouldn't mind a more general name like srcset, we say the builtin Matplotlib scraper supports it, Mayavi doesn't, and other custom scrapers can poll this attribute and expect a list of float. 

From https://github.com/sphinx-gallery/sphinx-gallery/pull/808#issuecomment-818012601

This isn't in the docs, but I think it should be added if it is relevant?

PR #1288, opened 2024-04-18
readme: add link to docs

Borda:
It took me a while to find the link to docs, and mostly, it is mentioned in the readme, or there is a docs badge, but there is none of these yet...
Also, it may be a good idea to add a section with a simple/quick explanation of how to use it...

cc: @lucyleeow


Borda:
seems that all failing cases are identical with main/default branch

lucyleeow:
Thanks @Borda , Erics fixed the CIs in #1289 but this can go in.

+1 for improving our readme with a quickstart.

PR #1289, opened 2024-04-18
BUG: Fix serialization with Sphinx 7.3

### LINKED ISSUE #1286: Allow configuration of sort orders in a way that does not mess with the sphinx cache
Using a configuration like this:

```
sphinx_gallery_conf = {
    "within_subsection_order": FileNameSortKey,
}
```

Results in unecessary rebuilds and also in a warning using the latest sphinx, because this config is not serializable:

```
from sphinx.confing import is_serializable
assert is_serializable(sphinx_gallery_conf)
```

The warning emitted with sphinx 7.3.5 is:

```

looking for now-outdated files... none found
WARNING: cannot cache unpickable configuration value: 'sphinx_gallery_conf'
pickling environment... done
checking consistency... done
preparing documents... done
```

One way would be using strings with a registry of types:

```
sphinx_gallery_conf = {
    "within_subsection_order": "filename_sort_key",
}
```

to still allow users to pass their own implementations, you could use something like:

```
sort_keys = {}

class SortKey:
    def __init_subclass__(cls, config_name):
        if config_name in sort_keys:
            raise ValueError("config_name must be unique")
       sort_keys[config_name] = cls

class FilenameSortKey(SortKey, config_name="filename_sort_key"):
    def __call__(self, filename):
         return filename
```

larsoner:
Closes #1286

1. Support fully qualified names of classes / callables
2. Add aliases for our own ones like `"FileNameSortKey"` as an alias for `"sphinx_gallery.sorting.FileNameSortKey"`
3. Keep `sphinx_gallery_conf` clean: don't store `app` in there or any functions
    1. Check for existence when filling the conf
    2. Load them on the fly when they need to be used
 4. Add to release notes

Kind of a pain to figure out but ultimately it does seem cleaner!

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,6 +5,7 @@ repos:
       - id: ruff-format
         exclude: plot_syntaxerror
       - id: ruff
+        args: ["--fix"]
 
   - repo: https://github.com/codespell-project/codespell
     rev: v2.2.6
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,22 @@
 Changelog
 =========
 
+v0.16.0
+-------
+Sphinx 7.3.2 and above changed caching and serialization checks. Now instead of passing
+instantiated classes like ``ResetArgv()``, classes like ``FileNameSortKey``, or
+callables like ``notebook_modification_function`` in  ``sphinx_gallery_conf``,
+you should pass fully qualified names to classes (that SG will instantiate for you) or
+instantiated instances that have a stable ``__repr__``. For example, instead of defining
+a ``ResetArgV`` class in ``conf.py``, define it in an importable module somewhere in
+your path. This could be for example:
+
+- ``"mymod.utils._ResetArgv"`` where you define a ``_ResetArgv`` class in
+  ``mymod.utils``
+- ``sphinxext.reset.ResetArgv`` where you define a ``ResetArgv`` class in
+  ``doc/sphinxext/reset.py`` and do something like
+  ``sys.path.insert(0, os.path.dirname(__file__) + "/sphinxext")`` in your ``conf.py``
+
 v0.15.0
 -------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,7 +7,7 @@ Sphinx 7.3.0 and above changed caching and serialization checks. Now instead of 
 instantiated classes like ``ResetArgv()``, classes like ``FileNameSortKey``, or
 callables like ``notebook_modification_function`` in  ``sphinx_gallery_conf``,
 you should pass fully qualified name strings to classes or callables. If you change
-to using name strings, you can simply use function as the use of classes to ensure
+to using name strings, you can simply use a function as the use of classes to ensure
 a stable ``__repr__`` would be redundant.
 
 See :ref:`importing_callables` for details.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,7 +12,7 @@ somewhere in your path and instantiate an instance of it. For ``ResetArgv`` this
 be for example either of:
 
 1. ``"mymod.utils.reset_argv"`` where you define a ``class _ResetArgv`` class in
-   ``mymod.utils`` and instantiate it in your ``conf.py`` as
+   ``mymod.utils`` and instantiate it an instance of it in ``mymod/utils.py`` with
    ``reset_argv = _ResetArgv()``.
 2. ``"sphinxext.reset.ResetArgv"`` where you define a ``ResetArgv`` class in
    ``doc/sphinxext/reset.py``, instantiate an instance with
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,10 +11,10 @@ defining a ``ResetArgV`` class in ``conf.py``, define it in an importable module
 somewhere in your path and instantiate an instance of it. For ``ResetArgv`` this could
 be for example either of:
 
-1. ``"mymod.utils.reset_argv"`` where you define a ``class _ResetArgv`` class in
-   ``mymod.utils`` and instantiate it an instance of it in ``mymod/utils.py`` with
+1. ``"reset_argv": "mymod.utils.reset_argv"`` where you define a ``class _ResetArgv`` class in
+   ``mymod.utils`` and instantiate an instance of it in ``mymod/utils.py`` with
    ``reset_argv = _ResetArgv()``.
-2. ``"sphinxext.reset.ResetArgv"`` where you define a ``ResetArgv`` class in
+2. ``"reset_argv": "sphinxext.reset.ResetArgv"`` where you define a ``ResetArgv`` class in
    ``doc/sphinxext/reset.py``, instantiate an instance with
    ``reset_argv = ResetArgv()``, and do something like
    ``sys.path.insert(0, os.path.dirname(__file__) + "/sphinxext")`` in your ``conf.py``.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,19 +6,11 @@ v0.16.0
 Sphinx 7.3.0 and above changed caching and serialization checks. Now instead of passing
 instantiated classes like ``ResetArgv()``, classes like ``FileNameSortKey``, or
 callables like ``notebook_modification_function`` in  ``sphinx_gallery_conf``,
-you should pass fully qualified names to classes or callables. For example, instead of
-defining a ``ResetArgV`` class in ``conf.py``, define it in an importable module
-somewhere in your path and instantiate an instance of it. For ``ResetArgv`` this could
-be for example either of:
+you should pass fully qualified name strings to classes or callables. If you change
+to using name strings, you can simply use function as the use of classes to ensure
+a stable ``__repr__`` would be redundant.
 
-1. ``"reset_argv": "mymod.utils.reset_argv"`` where you define function
-   ``def reset_argv`` class in ``mymod.utils``.
-2. ``"reset_argv": "sphinxext.reset.reset_argv"`` where you define a ``def reset_argv``
-   in ``doc/sphinxext/reset.py``, and do something like
-   ``sys.path.insert(0, os.path.dirname(__file__) + "/sphinxext")`` in your ``conf.py``.
-
-Built in classes like :class:`sphinx_gallery.sorting.FileNameSortKey` and similar can
-be used with shorter direct alias strings like ``"FileNameSortKey"``.
+See :ref:`importing_callables` for details.
 
 v0.15.0
 -------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,12 +12,12 @@ somewhere in your path and instatiate an instance of it. For ``ResetArgv`` this 
 for example either of:
 
 1. ``"mymod.utils.reset_argv"`` where you define a ``class _ResetArgv`` class in
-  ``mymod.utils`` and instantiate it in your ``conf.py`` as
-  ``reset_argv = _ResetArgv()``.
+   ``mymod.utils`` and instantiate it in your ``conf.py`` as
+   ``reset_argv = _ResetArgv()``.
 2. ``"sphinxext.reset.ResetArgv"`` where you define a ``ResetArgv`` class in
-  ``doc/sphinxext/reset.py``, instantiate an instance with ``reset_argv = ResetArgv()``,
-  and do something like
-  ``sys.path.insert(0, os.path.dirname(__file__) + "/sphinxext")`` in your ``conf.py``.
+   ``doc/sphinxext/reset.py``, instantiate an instance with
+   ``reset_argv = ResetArgv()``, and do something like
+   ``sys.path.insert(0, os.path.dirname(__file__) + "/sphinxext")`` in your ``conf.py``.
 
 Built in classes like :class:`sphinx_gallery.sorting.FileNameSortKey` and similar can
 be used with shorter direct alias strings like ``"FileNameSortKey"``.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,8 +8,8 @@ instantiated classes like ``ResetArgv()``, classes like ``FileNameSortKey``, or
 callables like ``notebook_modification_function`` in  ``sphinx_gallery_conf``,
 you should pass fully qualified names to classes or callables. For example, instead of
 defining a ``ResetArgV`` class in ``conf.py``, define it in an importable module
-somewhere in your path and instatiate an instance of it. For ``ResetArgv`` this could be
-for example either of:
+somewhere in your path and instantiate an instance of it. For ``ResetArgv`` this could
+be for example either of:
 
 1. ``"mymod.utils.reset_argv"`` where you define a ``class _ResetArgv`` class in
    ``mymod.utils`` and instantiate it in your ``conf.py`` as
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,12 +11,10 @@ defining a ``ResetArgV`` class in ``conf.py``, define it in an importable module
 somewhere in your path and instantiate an instance of it. For ``ResetArgv`` this could
 be for example either of:
 
-1. ``"reset_argv": "mymod.utils.reset_argv"`` where you define a ``class _ResetArgv`` class in
-   ``mymod.utils`` and instantiate an instance of it in ``mymod/utils.py`` with
-   ``reset_argv = _ResetArgv()``.
-2. ``"reset_argv": "sphinxext.reset.ResetArgv"`` where you define a ``ResetArgv`` class in
-   ``doc/sphinxext/reset.py``, instantiate an instance with
-   ``reset_argv = ResetArgv()``, and do something like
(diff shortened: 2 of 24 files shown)


larsoner:
@maxnoe can you see if the documented changes here make sense to you?

larsoner:
@lucyleeow ready for review/merge from my end

sphinx_gallery/gen_rst.py
```diff
@@ -1487,3 +1498,141 @@ def save_rst_example(
...
+            f"({repr(val)})"
+        )
+    return val
+
+
+def _get_callables(gallery_conf, key):
+    # the following should be the case (internal use only)
+    singletons = ("reset_argv", "minigallery_sort_order")
+    assert key in ("image_scrapers", "reset_modules", "jupyterlite") + singletons, key
+    which = gallery_conf[key]
```
lucyleeow, line 1541: `gallery_conf` is only used once in this func, here. Could we directly pass the conf value instead? This would avoid needing the
```python
    if key == "jupyterlite":
        which = [which["notebook_modification_function"]]
```

as we could directly pass `gallery_conf['jupyterlite']['notebook_modification_function']`

lucyleeow: Also for readability what about having a constant outside of the function like

```python
CALLABLE_CONFIGS = {
    'single': ("reset_argv", "minigallery_sort_order", "jupyterlite[notebook_modification_function]"),
    'multi': ("image_scrapers", "reset_modules")
}
```

(as we wouldn't need to use the key to get the from `gallery_config`)

lucyleeow (approved):
Thank you for fixing this, does not look like it was a simple fix. Lots of nits that can be fixed later really.

doc/configuration.rst
```diff
@@ -403,12 +403,16 @@ set to
...
 
-    from sphinx_gallery.sorting import NumberOfCodeLinesSortKey
     sphinx_gallery_conf = {
         ...
-        'within_subsection_order': NumberOfCodeLinesSortKey,
+        'within_subsection_order': "NumberOfCodeLinesSortKey",
     }
 
+Note that you must provide a *class* here, not an instance of the class. Because of how
+sphinx caches the environment, this should be provided as a fully qualified string
```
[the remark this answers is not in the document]
maxnoe: That's not true. There was no behavior change in sphinx 7.3.0. there is just a warning now in case the config is not being cached.

sphinx_gallery/gen_rst.py
lucyleeow: Happy to have a crack if you're okay with it but don't have the time

NirobNabil:
`subsection_order` field on the conf still  requires a callable which is [called directly in gen_gallery](https://github.com/sphinx-gallery/sphinx-gallery/blob/ebe93883d684a554598561c7259b14d3de5c5a43/sphinx_gallery/gen_gallery.py#L490) and thus generates the *non pickleable* warning. I also went through the tests of sphinx_gallery and couldn't find any testcase covering the `subsection_order` field making sure having a callable in this field doesn't mess up anything. 

while working on the [same issue at matplotlib](https://github.com/matplotlib/matplotlib/pull/28103) i made some changes in gen_gallery and gen_rst to handle the issue with `subsection_order` and it seemed to build without any issue.

- changed [line 447 in gen_gallery](https://github.com/larsoner/sphinx-gallery/blob/d57e478f042bc6b178b44aa36408dfe1a5e4fea9/sphinx_gallery/gen_gallery.py#L447) to 
  ```python 
  sortkey = _get_class(gallery_conf, "subsection_order")
  ```

- and changed the  `_get_class` function in gen_rst at [line 1503](https://github.com/larsoner/sphinx-gallery/blob/d57e478f042bc6b178b44aa36408dfe1a5e4fea9/sphinx_gallery/gen_rst.py#L1528) to
  ```python
  if not inspect.isclass(val) and not hasattr(val, '__class__'):
          raise ConfigError(
              f"{what} must be 1) a fully qualified name (string) that resolves "
              f"to a class or 2) or a class itself 3) or an instance of a class,"
              f" got {val.__class__.__name__} ({repr(val)})"
          )
  ```
  so that it passes with both a class definition or an instance of a class

larsoner:
I'll probably try modifying it just to use _get_callables instead so that a callable needs to be passed rather than a class

sphinx_gallery/gen_rst.py
larsoner: It will make all the invocations longer and less DRY. What currently shows up as this:
```
mod_fun, = _get_callables(gallery_conf, "notebook_modification_function")
```
would become
```
mod_fun, = _get_callables([gallery_conf["notebook_modification_function"]], "notebook_modification_function")
```
in the place where it's actually used. Which is less DRY and less readable to me. I'd rather add a bit of complication to the function to keep the in-code invocations cleaner.

> (as we wouldn't need to use the key to get the from gallery_config)

Keep in mind we do need the key inside the function to provide a nicely formatted error message.

> Also for readability what about having a constant outside of the function like

The constant is really only used inside the function so to me it's a step backward to move it outside the function

doc/configuration.rst
```diff
@@ -403,12 +403,16 @@ set to
 :class:`NumberOfCodeLinesSortKey(src_dir) <sphinx_gallery.sorting.NumberOfCodeLinesSortKey>`,
 which sorts the files based on the number of code lines::
 
-    from sphinx_gallery.sorting import NumberOfCodeLinesSortKey
     sphinx_gallery_conf = {
         ...
-        'within_subsection_order': NumberOfCodeLinesSortKey,
+        'within_subsection_order': "NumberOfCodeLinesSortKey",
```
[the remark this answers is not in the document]
larsoner: I'd rather highlight the shortcut version here. I think it's mostly what people will want to use. But I'll clarify the language that follows a bit.

CHANGES.rst
```diff
@@ -1,6 +1,27 @@
 Changelog
 =========
 
+v0.16.0
+-------
+Sphinx 7.3.2 and above changed caching and serialization checks. Now instead of passing
```
lucyleeow, line 6:
```suggestion
Sphinx 7.3.0 and above changed caching and serialization checks. Now instead of passing
```
?

doc/configuration.rst
```diff
@@ -403,12 +400,19 @@ set to
...
+Note that you must provide a *class* here, not an instance of the class. Because of how
+sphinx caches the environment, this should be provided as a fully qualified string
+that can be imported (though passing a class directly is supported for backward
+compatibility). Built in sphinx-gallery classes, however, for convenience can be
+specified using just the stem for example as ``"NumberOfLinesSortKey"``, which is
+functionally equivalent to providing the fully qualified name
+``"sphinx_gallery.sorting.NumberOfCodeLinesSortKey"``.
+
 In addition, multiple convenience classes are provided for use with
 ``within_subsection_order``:
```
lucyleeow, line 417:
```suggestion
Built in convenience classes supported by ``within_subsection_order``:
```

sphinx_gallery/gen_rst.py
```diff
@@ -1487,3 +1498,141 @@ def save_rst_example(
...
+        raise ConfigError(
+            f"{what} must be 1) a fully qualified name (string) that resolves "
+            f"to a class or 2) or a class itself, got {val.__class__.__name__} "
+            f"({repr(val)})"
+        )
+    return val
+
+
+def _get_callables(gallery_conf, key):
+    # the following should be the case (internal use only)
```
lucyleeow, line 1538: Not sure what this means.

Maybe we could add a docstring one liner?

sphinx_gallery/gen_rst.py
lucyleeow: Thats fine. I was thinking it would be nice to be able to update a constant instead of digging inside a function. Much of a muchness though.

sphinx_gallery/gen_rst.py
```diff
@@ -1487,3 +1498,141 @@ def save_rst_example(
...
+    singletons = ("reset_argv", "minigallery_sort_order", "subsection_order")
+    assert key in ("image_scrapers", "reset_modules", "jupyterlite") + singletons, key
+    which = gallery_conf[key]
+    if key == "jupyterlite":
+        which = [which["notebook_modification_function"]]
+    elif key in singletons:
+        which = [which]
+    if not isinstance(which, (tuple, list)):
+        which = [which]
+    which = list(which)
```
lucyleeow, line 1548: Just a question - what does the `list(which)` do? From above, `which` would either be a tuple or a list already..?

doc/configuration.rst
```diff
@@ -428,40 +432,38 @@ You can create a custom sort key callable for the following configurations:
...
-When creating a custom sort key callable, you must ensure that the ``__repr__``
-is stable across runs. Sphinx determines if the build environment has
-changed, and thus if *all* documents should be rewritten, by examining the
-config values using ``md5(str(obj).encode()).hexdigest()`` in
-``sphinx/builders/html.py``. Default class instances in Python have their
-memory address in their ``__repr__`` which is why generally the ``__repr__``
-changes in each build.
+The best way to do this is to make use of a single instance of each class you want
+to use in some fully resolvable path, and then tell sphinx-gallery to resolve the
+callable from the fully qualified name. For example, you could write a function
```
lucyleeow, line 437: This does not seem to marry with the `CHANGES.rst` notes:

> 1. ``"mymod.utils.reset_argv"`` where you define a ``class _ResetArgv`` class in
   ``mymod.utils`` and instantiate it in your ``conf.py`` as
   ``reset_argv = _ResetArgv()``.

.pre-commit-config.yaml
```diff
@@ -5,6 +5,7 @@ repos:
       - id: ruff-format
         exclude: plot_syntaxerror
       - id: ruff
+        args: ["--fix"]
```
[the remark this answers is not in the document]
Borda: maybe make this as separate PR for tracking?

sphinx_gallery/gen_rst.py
larsoner: Makes it mutable (`tuple` is not)

sphinx_gallery/gen_rst.py
larsoner: it refers to the following `assert` line which should always pass because we've ensured it elsewhere

doc/configuration.rst
larsoner: Nice catch, fixed!

.pre-commit-config.yaml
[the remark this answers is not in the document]
larsoner: In principle we could but in practice it's extra CI and reviewer work for what I think is just updating to standard/modern `pre-commit` practices in scientific python repos, so not necessary I think

sphinx_gallery/gen_rst.py
larsoner: (note that at the end of this `for` we do `which[wi] = what`, which would fail for a tuple. Moreover even if we know it's a list, we don't want to modify the one in `sphinx_gallery_config` so a local `list` is better there, too!)

doc/configuration.rst
```diff
@@ -428,40 +431,38 @@ You can create a custom sort key callable for the following configurations:
...
     sphinx_gallery_conf = {
     #...,
-    "minigallery_sort_order": FunctionSortKey(
-        lambda filename: filename[:10]),
+    "minigallery_sort_order": "sphinxext.plotted_sorter",
     #...
     }
 
+And sphinx-gallery would resolve ``"sphinxext.function_sorter"`` to the
+``function_sorter`` object because the ``doc/`` directory is first on the path.
```
lucyleeow, line 464:
```suggestion
And sphinx-gallery would resolve ``"sphinxext.plotted_sorter"`` to the
``plotted_sorter`` object because the ``doc/`` directory is first on the path.
```

doc/configuration.rst
```diff
@@ -326,28 +326,25 @@ needs to be a Callable that accepts the ``gallery_conf`` and ``script_vars``
...
 
 which is included in the configuration dictionary as::
 
     sphinx_gallery_conf = {
         ...
-        'reset_argv': ResetArgv(),
+        'reset_argv': "sphinxext.reset_argv",
     }
 
 which is then used by Sphinx-Gallery as::
```
lucyleeow, line 350: I think the lines below this need to be updated.

doc/configuration.rst
```diff
@@ -326,28 +326,25 @@ needs to be a Callable that accepts the ``gallery_conf`` and ``script_vars``
 dictionaries as input and returns a list of strings that are passed as
 additional command line arguments to the interpreter.
 
-An example could be::
+An example in ``doc/sphinxext.py`` (assuming you have added ``doc/`` to ``sys.path`` so
+that ``sphinxext.reset_argv`` can be resolved by importlib) could be::
```
lucyleeow, line 330:
```suggestion
A ``reset_argv`` example defined in ``doc/sphinxext.py`` (assuming you have added ``doc/`` to ``sys.path`` so
that ``sphinxext.reset_argv`` can be resolved by importlib) could be::
```
nit

CHANGES.rst
```diff
@@ -1,6 +1,27 @@
...
+-------
+Sphinx 7.3.0 and above changed caching and serialization checks. Now instead of passing
+instantiated classes like ``ResetArgv()``, classes like ``FileNameSortKey``, or
+callables like ``notebook_modification_function`` in  ``sphinx_gallery_conf``,
+you should pass fully qualified names to classes or callables. For example, instead of
+defining a ``ResetArgV`` class in ``conf.py``, define it in an importable module
+somewhere in your path and instantiate an instance of it. For ``ResetArgv`` this could
+be for example either of:
+
+1. ``"mymod.utils.reset_argv"`` where you define a ``class _ResetArgv`` class in
```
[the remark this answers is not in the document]
lucyleeow: Maybe another config would be a better example because in `configuration.rst` we suggest to define a function for `reset_argv`.

doc/configuration.rst
```diff
@@ -403,14 +400,20 @@ set to
...
-    from sphinx_gallery.sorting import NumberOfCodeLinesSortKey
     sphinx_gallery_conf = {
         ...
-        'within_subsection_order': NumberOfCodeLinesSortKey,
+        'within_subsection_order': "NumberOfCodeLinesSortKey",
     }
 
-In addition, multiple convenience classes are provided for use with
-``within_subsection_order``:
+Note that you must provide a *class* here, not an instance of the class. Because of how
```
lucyleeow, line 408: Stupid question, but if this is a string, does it matter if this is a class or instance of a class?

lucyleeow (approved):

doc/configuration.rst
larsoner: Yes, it must be a class because we instantiate it internally. Hence the emphasis on it being a class not an instance here

larsoner:
Okay docs hopefully fixed. Realized there were a few to fix in `advanced.rst` as well!

NirobNabil:
Been waiting for this PR to go into a release so that [the fix at matplotlib](https://github.com/matplotlib/matplotlib/pull/28103) can be merged. Any idea when the next release window of sphinx-gallery might be?

lucyleeow:
We'll release after this PR goes in.

doc/advanced.rst
```diff
@@ -299,7 +298,7 @@ package in a module called ``scraper``. Here is the scraper code::
 
 Then, in our ``conf.py`` file, we include the following code::
 
-   from mymodule import PNGScraper
+   from mymodule.scraper import PNGScraper
 
    sphinx_gallery_conf = {
        ...
```
lucyleeow, line 304: Update below to 'mymodule.scraper.PNGScraper' ?

lucyleeow (approved):
Haven't looked deeply in the tests but should we parametrize some to test config value being str and callable (since we still want to support callable atm). Can be done in separate PR. I guess we would eventually remove callable support if Sphinx moves to toml conf file.

Thanks for fixing this, not an easy one.

larsoner:
@lucyleeow are you up for pushing directly? I think it will be faster

lucyleeow:
Can be done in separate PR but what do you think about just updating the `gallery_conf` value in `_get_callables`, so we don't have to do it twice?

lucyleeow:
> @lucyleeow are you up for pushing directly? I think it will be faster

Done, only changed docs.

larsoner:
> what do you think about just updating the gallery_conf value in _get_callables, so we don't have to do it twice (call _get_callables twice

I wish we could but the env (including gallery conf) is cached after SG runs so it would cause the same serializing problems

larsoner:
Yeah let's get this in and I can start testing with MNE-Python for example. Then maybe release Friday?

lucyleeow:
> > what do you think about just updating the gallery_conf value in _get_callables, so we don't have to do it twice (call _get_callables twice
> 
> I wish we could but the env (including gallery conf) is cached after SG runs so it would cause the same serializing problems

Just a question, as I was just thinking about this. When you say after SG runs, does it mean more generally after build? As I understand, we connect functions to various sphinx events, so 'after SG runs' would just mean after sphinx finishes building? Or am I mis-understanding?

larsoner:
I mean after the examples are run in the `build-inited` event, not after the sphinx-build finishes. It caches the env sometime after `build-inited` completes

PR #1292, opened 2024-04-20
DOC Update FFMpeg note in conf animation docs

lucyleeow:
I think we prefer FFMpeg now and the buggy was probably for matplotlib < 3.3.1


lucyleeow:
Ref: #916

Borda (approved):

larsoner:
Thanks @lucyleeow !

PR #1299, opened 2024-04-29
add option for zip downloads

jamiecook:
Our IT department prevents downloads of .py and .ipynb files.

This PR adds an option (documented) that allows for generating .py.zip and .ipynb.zip files instead.


larsoner:
If this is a common enough use case -- that people's internet is restricted -- we should probably make all projects that use SG expose a `.zip` way of downloading files. In this sense it's a bit odd to make it a project-specific config value. We also already have too many config options, so it would be nice to avoid adding another.

So perhaps rather than  a separate config option we could add a third download "entry" for all projects that use SG:

- `Download Jupyter notebook: 10_overview.ipynb`
- `Download Python source code: 10_overview.py`
- `Download both in a zip archive: `10_overview.zip`

That way people who can download directly still can, and people who have more restricted access can get what they need. I understand this is slightly less flexible in that people who have restricted access have to "download both" in all cases, but I think it's an acceptable tradeoff for those folks to have to (very easily!) delete the one small extra file they get once extracting, or more carefully just extract the one file that they want. The `.zip`s should be tiny so no appreciable bandwidth/size problem should come from this either I think.

And if there are projects that hate this new `.zip` option, as long as we have it in part of its own CSS class, people can set `visibility: hidden` to make it go away.

jamiecook:
Thats a good idea - I will modify and resubmit

jamiecook:
@larsoner that is done, much smaller delta on this one now

image

sphinx_gallery/gen_rst.py
```diff
@@ -234,6 +235,12 @@ def __exit__(self, type_, value, tb):
       :download:`Download Jupyter notebook: {0} <{0}>`
 """
 
+ZIP_DOWNLOAD = """
+    .. container:: sphx-glr-download sphx-glr-download-zip
+
+      :download:`Download all: {0} <{0}>`
```
larsoner, line 241: Somewhere else I think we have a "download all" which means "download from all examples", so I think it's less ambiguous to say "download both" here or "download zipped ipynb and py files" or something. So a minimal version would be 
```suggestion
      :download:`Download both (zipped): {0} <{0}>`
```

sphinx_gallery/utils.py
```diff
@@ -167,6 +168,17 @@ def _replace_md5(fname_new, fname_old=None, method="move", mode="b"):
     assert os.path.isfile(fname_old)
 
 
+def zip_files(file_list, zipname, relative_to, extension=None):
```
jamiecook, line 171:
```suggestion
def zip_files(file_list, zipname, relative_to, extension=None):
    """Creates a zip file with name `zipname` containing the files listed in `file_list`. The zip file contents will be stored with their paths stripped to be relative to `relative_to`.
    """
```

jamiecook:
I think this one just needs approval to run all the other CI workflows.

larsoner:
Thanks @jamiecook !

Issue #1283, opened 2024-04-11
Feature: Per-code-block options


timhoffm:
I would like to pass through options to  [`.. code-block::`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block).

Syntax proposal: Support a comment at the start of code blocks

```
# sphinx_gallery_code_block_options
#     :caption: Example
#     :emphasize-lines: 6

import matplotlib.pyplot as plt
import numpy as np

x = np.arange(10)
y = np.exp(x)
plt.plot(x, y)
```
such that the resulting ReST is
```
.. code-block::
    :caption: Example
    :emphasize-lines: 6

    import matplotlib.pyplot as plt
    import numpy as np

    x = np.arange(10)
    y = np.exp(x)
    plt.plot(x, y)
```

lucyleeow:
That syntax won't work with our [`extract_file_config`](https://github.com/sphinx-gallery/sphinx-gallery/blob/fda9674f1d279f26859d0c0221868f6bbdae8b66/sphinx_gallery/py_source_parser.py#L133), what about making it a dict and sphinx gallery deals with the syntax? It also prevent any indenting errors from the user.

timhoffm:
Thanks for the feedback. I'm not wedded to a particular syntax. Do I understand correctly that configurations have to be single-line? Could you give an example how the dict would look like?

lucyleeow:
Hmm maybe, I'll have a look at the regex.

I was thinking maybe:

```
sphinx_gallery_code_block_options = {'caption': 'Example', 'emphasize-lines': 6}
```
WDYT?

timhoffm:
My original intention was to mimic the ReST declaration as close as possible. But since multiple is not supported, a dict seems like a reasonably simple alternative in terms of clarity of expression and parsing effort. Note that there are value-less options and we have to define how these should be described, e.g. `{'linenos': ''}`.

lucyleeow:
> Also, all parameters need to be strings or trivially be convertable to strings:

Good point. I am happy to force strings?

[...]

PR #1303, opened 2024-05-05
MNT: Turn the block tuple into a namedtuple

timhoffm:
It's better to give the "Block" concept its own entity and thus be able to access it's attributes by name.

I anticipate that I will need to expand the Block data to support per-block options for #1283, and that expansion will be simpler when blocks are no longer pure tuples. I've decided to split this out for simpler review and advertise this PR as code cleanup.

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -162,14 +162,14 @@ The function should take the following inputs (in this order):
 
 1. ``block`` - a Sphinx-Gallery ``.py`` file is separated into consecutive
    lines of 'code' and reST 'text', called 'blocks'. For each
-   block, a tuple containing the (label, content, line_number)
-   (e.g. ``('code', 'print("Hello world")', 5)``) of the block is created.
+   block, a `namedtuple` with ("type", "content", "lineno") items
+   (e.g. ``('code', 'print("Hello world")', 5)``) is created.
 
    * 'label' is a string that can either be ``'text'`` or ``'code'``. In this
      context, it should only be ``'code'`` as this function is only called for
      code blocks.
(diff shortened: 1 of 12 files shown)


timhoffm:
Not quite sure what the failing test for 3.11 / pip mean. Could anybody hint at whether they are related to this PR or whether that pipeline is broken for other reasons? Thanks.

lucyleeow:
I couldn't replicate locally, the rerun test has always been a bit brittle but no idea why `plot_scraper_broken` did not seem to execute in `test_error_messages`. I've rerun the test :shrug: 

I think `save_figures` and docstring of `matplotlib_scraper` needs to be updated (even though `block` is never actually used). The scraper docs should be updated too: https://sphinx-gallery.github.io/dev/advanced.html#write-a-custom-image-scraper

lucyleeow:
Seems spurious, passing now

larsoner (approved):
Merged `main` in just to make sure everything is still green, good to go from your end @timhoffm ?

lucyleeow:
I think we have to update `scraper.py` and scraper docs, but otherwise LGTM!

timhoffm:
Code-wise this is good to go. It's only documentation that could/should be improved. There are a number of places still describing `block` as tuple. Technically, that is still correct, because a named tuple is still a tuple. But describing it as namedtuple would be more helpful.

I'm a bit tight with time now, so it would take a bit until I get around updating the docs as well. If you want to take this as is or update the docs yourself, that would also be very welcome.

lucyleeow:
Done, @larsoner feel free to merge.

larsoner:
Thanks @timhoffm @lucyleeow !

PR #1312, opened 2024-05-21
ENH integrate download/launcher links into `pydata-sphinx-theme` secondary sidebar

### LINKED ISSUE #1258: FEA Integrate download links and binder/juputerlite buttons with `pydata-sphinx-theme` secondary sidebar
Given the popularity of `pydata-sphinx-theme` within the community nowadays, I'm requesting to see if it is possible to have some theme-specific configurations for better integration. In particular, `pydata-sphinx-theme` has a secondary sidebar where we can put the download links (Python source code and Jupyter notebook) and Binder/JupyterLite buttons. This way we no longer need the "Go to the end" note at the top and the footer at the bottom. This idea comes up when switching `scikit-learn` docs to `pydata-sphinx-theme`, see https://github.com/scikit-learn/scikit-learn/pull/28415 for discussion and check its rendered docs [here](https://output.circle-artifacts.com/output/job/49c4c555-2d4f-4110-988f-bcd8ac73e1dd/artifacts/0/doc/auto_examples/index.html).

<p align="center">

</p>

Is it possible to:

(1) Provide a configuration option to remove the top note and the footer.

(2) Expose API to get Binder/JupyterLite/download links (href) (and maybe badge image links (src) as well). I notice that there are things like `interactive_example.gen_binder_url`, but in particular I mean functions/variables added to the page context during `sphinx` build that can be directly called/accessed from Jinja templates (see the [documentation](https://www.sphinx-doc.org/en/master/development/theming.html#defining-custom-template-functions)). In this way users can easily create their own `pydata-sphinx-theme` components to put into the secondary sidebar.

<details>
<p></p>

Ideally, one can create templates like:

```html
<div>
  <a href="{{ sg_binder_url }}">
    <img alt="Launch Binder" src="{{ sg_binder_badge_src }}">
  </a>
  <a href="{{ sg_lite_url }}">
    <img alt="Launch JupyterLite" src="{{ sg_lite_badge_src }}">
  </a>
</div>
{% endif %}
```

and

```html
<div>
  <div title="{{ sg_python_name }}">
    <a download href="{{ sg_python_download_path }}">
      <i class="fa-solid fa-download"></i>
      Download source code
    </a>
</div>
<div>
  <div title="{{ sg_notebook_name }}">
    <a download href="{{ sg_notebook_download_path }}">
      <i class="fa-solid fa-download"></i>
      Download Jupyter notebook
    </a>
</div>
```

</details>

(3) An alternative to (2) is that, directly create `pydata-sphinx-theme` components (templates) and ship them with the library. Then figure out a way for `pydata-sphinx-theme` to discover that template (see the [guide](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/layout.html#add-your-own-html-templates-to-theme-sections)). This saves users' effort but I'm not sure if this is possible.

Charlie-XIAO:
Closes #1258.

This PR intends to integrate the download links (Python and Jupyter) and launcher badges (JupyterLite and Binder) into the secondary sidebar of `pydata-sphinx-theme`. See the *Basics Gallery with Matplotlib* ([here](https://output.circle-artifacts.com/output/job/599669d2-a07f-49d5-9076-89b71263d2d9/artifacts/0/html/auto_examples/index.html)) for how it looks like. I'm not sure if this is the correct way to do in particular since it (may) currently fit only into `pydata-sphinx-theme`, but I think the template + component approach should in fact work with other themes (to some extent) as well. Regardless, I will explain a bit what this PR is doing:

- `setup_pst_secondary_sidebar_links`: This adds `get_download_links` and `get_launcher_links` to per-page context which can be called in the jinja templates. Essentially they work by grabbing information from `doctree`, especially because I find no way to get the unique hash to the downloadable files ([ref](https://www.sphinx-doc.org/en/master/usage/referencing.html#role-download)).
- `sphinx_gallery/components/`: This path is appended to `templates_path` ([ref](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-templates_path)). See `html_theme_options.secondary_sidebar_items` for an example usage with `pydata-sphinx-theme`. It can also be used with e.g. `html_sidebars` ([ref](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_sidebars)) but it is currently designed to best integrate with the secondary sidebar in `pydata-sphinx-theme`.
- As suggested in https://github.com/sphinx-gallery/sphinx-gallery/issues/1258#issuecomment-1948394816, removing the original download links and launcher badges can be done via CSS. And in fact since the components work by grabbing from `doctree`, they can indeed not be actually removed.

@larsoner who involved in the original issue if you have time for a review.

A screenshot of [this page](https://output.circle-artifacts.com/output/job/599669d2-a07f-49d5-9076-89b71263d2d9/artifacts/0/html/auto_examples/plot_0_sin.html):

image

I haven't added documentation or tests yet (the former I want to do after we do reach an agreement on this PR, the latter I'm not sure how to do 🤣 so some guidance is really appreciated).


Charlie-XIAO:
I added a configuration option `template_links_pattern` (maybe it's not a good name). I know that `sphinx-gallery` already has plenty of configurations and may not want to add a new one, but the case here is:

```python
# This will cause the default templates (pagetoc, etc.) to be gone
html_theme_options.secondary_sidebar_items = {"auto_examples/*": [xxx]}
# auto_examples/* match multiple wildcard patterns which is invalid
# Moreover, I believe in most cases people want to add the templates to a whole gallery
# so it would almost always involve wildcard patterns
html_theme_options.secondary_sidebar_items = {"**": DEFAULTS, "auto_examples/*": [xxx]}
```

By adding `template_links_pattern` that accepts a list of glob patterns (same as what one would use for `html_theme_options.secondary_sidebar_items`), one can safely put `sg_download_links` and `sg_launcher_links` templates in a superset of the pages and then use `template_links_pattern` for restriction. See `doc/conf.py`.

Charlie-XIAO:
Documentation also updated :)

sphinx_gallery/components/sg_launcher_links.html
```diff
@@ -0,0 +1,11 @@
+{# Displays the launcher badges for the example. #}
+{% set launcher_links = get_launcher_links() %}
+{% for key, link_info in launcher_links.items() %}
+  {% if link_info %}
+    <div class="{{ key }}-badge-sidebar">
+      <a href="{{ link_info.link }}">
+        <img src="{{ link_info.img_src }}" alt="{{ link_info.img_alt }}" height="20">
```
larsoner, line 7: Rather than have a static `height` here it would probably be better to use a CSS class on the image or a parent (like the `div`), and add an entry in [SG's CSS](https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/_static/sg_gallery.css) to set the height. That way end-users can override the value if they want using CSS.

sphinx_gallery/components/sg_download_links.html
```diff
@@ -0,0 +1,12 @@
+{# Displays the download link for the Python source code of the example. #}
+{% set download_links = get_download_links() %}
+{% for key, link_info in download_links.items() %}
+  {% if link_info %}
+    <div class="sphx-glr-download-{{ key }}-sidebar" title="{{ link_info.title }}">
```
larsoner, line 5: Like maybe 
```suggestion
    <div class="sphx-glr-download-sidebar sphx-glr-download-{{ key }}-sidebar" title="{{ link_info.title }}">
```
then images inside divs with class `sphx-glr-download-sidebar` can have `max-height: 20px` by default in SG's CSS?

larsoner (commented):
> By adding components_pattern that accepts a list of glob patterns (same as what one would use for html_theme_options.secondary_sidebar_items), one can safely put sg_download_links and sg_launcher_links templates in a superset of the pages and then use components_pattern for restriction.

So the use case is that a project has, say, 3 galleries but only wants to enable it on 1 or 2 of them? If so then I'm not sure we actually need to support this. I'd rather assume YAGNI and go with a simpler implementation that is all-or-nothing.

Charlie-XIAO:
> So the use case is that a project has, say, 3 galleries but only wants to enable it on 1 or 2 of them? If so then I'm not sure we actually need to support this. I'd rather assume YAGNI and go with a simpler implementation that is all-or-nothing.

Yep I agree that this is a very rare use case. However I also want an _early return mechanism_, otherwise for all normal pages we will try to grab all those links, find nothing, and add no sidebar, which I think is not ideal.

Is there something in the `doctree` I can grab to easily identify whether a page is a `sphinx-gallery` page?

Charlie-XIAO:
Thanks for the review! I've updated the PR to:

- Remove the `components_pattern` config option that is useful only in rare cases
- Fix the failing tests (in particular how to deal with `templates_path` properly in older sphinx versions)
- Move the docs to "advanced usage" since they are unrelated to "configuration options" now
- Add tests to check that the sidebar links are consistent with the original links

larsoner (approved):
I checked a project build time and it seemed unaffected by the change which is great! And the [rendered doc looks good](https://output.circle-artifacts.com/output/job/18842ae2-dbe9-43f1-abe4-8056932180bb/artifacts/0/html/auto_examples/plot_0_sin.html).

@lucyleeow did you want to look? If not I'm happy to merge

doc/advanced.rst
```diff
@@ -495,3 +495,74 @@ the ``sphinx_gallery.gen_gallery`` extension, you can use in ``conf.py``::
...
+.. code-block:: python
+
+    html_static_path = ["_static"]
+
+    def setup(app):
+        # ...
+        app.add_css_file("hide_links.css")
+
+The above will include the CSS in all pages. If one wants to add the CSS only for
+certain pages, one can alternatively do the following:
```
lucyleeow, line 556:
```suggestion
certain pages, one can alternatively do following (only
add to pages that start with 'auto_examples/'):
```

doc/advanced.rst
```diff
@@ -495,3 +495,74 @@ the ``sphinx_gallery.gen_gallery`` extension, you can use in ``conf.py``::
...
+<https://www.sphinx-doc.org/en/master/development/templating.html>`__):
+``sg_download_links`` and ``sg_launcher_links``. They are primarily designed for use in
+`secondary sidebars <https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/layout.html#secondary-sidebar-right>`__
+of ``pydata-sphinx-theme``, but also work with any other configuration that uses
+templates, e.g., `html_sidebars
+<https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_sidebars>`__.
+
+- ``sg_download_links``: The download links of source code, Jupyter notebook, and the
+  zip of them for an example.
+- ``sg_launcher_links``: The launcher links for JupyterLite and Binder for an example.
```
lucyleeow, line 512: ?
```suggestion
- ``sg_download_links``: Displays the download links of source code, Jupyter notebook, and the
  zip of them for an example.
- ``sg_launcher_links``: Displays the launcher links for JupyterLite and Binder for an example.
```

lucyleeow (approved):
LGTM, thanks this is a nice addition.
Nits only

Charlie-XIAO:
Thanks for the reviews @larsoner @lucyleeow! I've resolved the conversations.

lucyleeow:
Thank you @Charlie-XIAO !

lesteve:
Super nice to see this kind of improvements in sphinx-gallery, thanks everyone!

For a bit of history, back in the days, there was an attempt at floating the buttons to the right https://github.com/sphinx-gallery/sphinx-gallery/pull/336 that I think eventually was reverted because the rendering looked off in some of sphinx themes.

Issue #1220, opened 2023-11-02
Add option to set thumbnails for expected failing examples


tsbinns:
Currently, if an example is listed in `expected_failing_examples`, its thumbnail will be set automatically to the "BROKEN" stamp. For example,
```
sphinx_gallery_conf = {
    ...
    'expected_failing_examples': ['../examples/my_failing_example.py'],
}
```
would appear in the gallery as
![image](https://github.com/sphinx-gallery/sphinx-gallery/assets/56922019/14b772ab-0527-4602-9640-4d7ab8bc53d1)
even if a thumbnail is explicitly specified, e.g. by setting `# sphinx_gallery_thumbnail_path = '_static/my_preferred_thumbnail.png'` in the example file.

It would be nice to have the option to override the automatic "BROKEN" thumbnail behaviour, for instance if there is only one part of your example which is designed to fail but you don't want it to appear as if the entire example is about code failure from the thumbnail.

Thanks!

lucyleeow:
This is reasonable. Are you interested in making a PR?

PR #1313, opened 2024-05-24
[ENH] Add custom thumbnails for failing examples

tsbinns:
## Problem

Belated follow-up of #1220.

In short, currently expected failing examples use the "BROKEN" stamp as their thumbnail, which cannot be overwitten by including e.g. `# sphinx_gallery_thumbnail_path = '_static/my_preferred_thumbnail.png'` in the example file.

Setting a different thumbnail could be useful if there is only one part of your example which is designed to fail, but you don't want it to appear as if the entire example is about code failure from the thumbnail.

## Possible Solution

This is just one idea I had. I am very open to suggestions if someone thinks there is a better solution.

In the `save_thumbnail()` function of `gen_rst.py`, the path to the thumbnails is set for the examples.

Currently, in the case where it is an expected failing example, the `sphinx_gallery_thumbnail_path` specified in the file will be carried over and the proper path to the thumbnail will be generated, however this will be ignored and the path to the "BROKEN" stamp file will be set.

Summary of the changes:

- adds a flag `ignore_thumbnail` that is default `False`
  https://github.com/sphinx-gallery/sphinx-gallery/blob/fc1d4231f7a033d9d2d14e9e0aab4c4e7c5c1747/sphinx_gallery/gen_rst.py#L372
- if no thumbnail is specified **and** it is a failing example, `ignore_thumbnail` is set to `True`
  https://github.com/sphinx-gallery/sphinx-gallery/blob/fc1d4231f7a033d9d2d14e9e0aab4c4e7c5c1747/sphinx_gallery/gen_rst.py#L380-L381
- when the thumbnail image is finally set...
  - if it is a failing example and `ignore_thumbnail==True` (i.e. no thumbnail was specified), the "BROKEN" stamp is used
    https://github.com/sphinx-gallery/sphinx-gallery/blob/fc1d4231f7a033d9d2d14e9e0aab4c4e7c5c1747/sphinx_gallery/gen_rst.py#L397-L398
  - otherwise even if it is a failing example, if `ignore_thumbnail==False` (i.e. a custom thumbnail was specified), this thumbnail will be used in place of the "BROKEN" stamp
    https://github.com/sphinx-gallery/sphinx-gallery/blob/fc1d4231f7a033d9d2d14e9e0aab4c4e7c5c1747/sphinx_gallery/gen_rst.py#L399-L400

In the case that a thumbnail is specified for a failing example and the image does not exist, the default "no_image.png" is used, like it would for a non-failing example.

No changes have been introduced for non-failing examples.

Not sure if this is worthy of a new example, but I included one since it helps here to demonstrate the behaviour. I have no problem if you want this removed before merging.
https://github.com/sphinx-gallery/sphinx-gallery/blob/fc1d4231f7a033d9d2d14e9e0aab4c4e7c5c1747/examples/no_output/plot_raise_thumbnail.py#L1-L28

## Unit tests

The big thing this is missing is unit tests. I am not super familiar with the inner workings of `sphinx-gallery` so I do not know where to start with this.

I noticed in `test_gen_rst.py` the following notes, so it seems like there is not an existing test which I could use as a template for checking the new behaviour (but I may have missed something). Would anyone have some ideas for this?
https://github.com/sphinx-gallery/sphinx-gallery/blob/fc1d4231f7a033d9d2d14e9e0aab4c4e7c5c1747/sphinx_gallery/tests/test_gen_rst.py#L1225-L1227


tsbinns:
> Yeah I think that makes sense!

Sorry, just to clarify, is this a thumbs up just for the new solution or also for adding to the user guide?

larsoner:
To both -- any new (inline or otherwise) configuration option should be documented in our docs :+1:

tsbinns:
Updated the name of the new example to better reflect what it now demonstrates.

Also added an entry into the user guide explaining the new option.

larsoner (approved):
Rendered HTML and docs look good, thanks @tsbinns !

larsoner (changes requested):
Argh actually I just realized that there is no modified test. One option would be to modify an example in `sphinx_gallery/tests/tinybuild` to have this new line and then assert that `"broken"` does not show up in the rendered HTML for example

doc/configuration.rst
```diff
@@ -1316,6 +1317,26 @@ Note that ``sphinx_gallery_thumbnail_number`` overrules
...
+
+Controlling thumbnail behaviour in failing examples
+===================================================
+
+By default, expected failing examples will have their thumbnail image as a
+stamp with the word "BROKEN". This behaviour is controlled by
+``sphinx_gallery_failing_thumbnail``, which is by default ``True``. In cases
+where control over the thumbnail image is desired (e.g. the default behaviour
+of using the first figure created, using a specific image, etc...), this should
+be set to ``False``::
```
lucyleeow, line 1330: maybe:
```suggestion
where control over the thumbnail image is desired, this should be set to
``False``. This will return thumbnail behaviour to 'normal', whereby
thumbnail will be either the first figure created (or the
:ref:`default thumbnail <custom_default_thumb>` if no figure is created)
or :ref:`provided thumbnail <providing_thumbnail>`::

```

lucyleeow (commented):
Nice addition, LGTM, just needs the test as Eric suggested.

larsoner:
@tsbinns don't forget about the test when you get a chance!

tsbinns:
@larsoner Sorry, have it on my to-do list but haven't gotten around to it yet. Will try to sort soon!

lucyleeow:
Fixed a merge conflict - hopefully correctly!

tsbinns:
Sorry again for the delay. Have now added two examples to `tinybuild` that mirror what's in the actual docs:
- an expected failing example where default failing behaviour is used (i.e. the "BROKEN" stamp)
- an expected failing example where default thumbnail behaviour is used (i.e. the plot from the example).

The thumbnails are tested here:
https://github.com/sphinx-gallery/sphinx-gallery/blob/7c33f225f26a2479eedefa379af474fc59c3466a/sphinx_gallery/tests/test_full.py#L333-L371

I can't seem to properly run `tinybuild` on my Windows machine, so I'm not sure if there are some errors for other things I didn't correctly adjust, but the new test passes at least.

sphinx_gallery/tests/test_full.py
```diff
@@ -41,8 +41,8 @@
 #
 # total number of plot_*.py files in
 # (examples + examples_rst_index + examples_with_rst + examples_README_header)
-N_EXAMPLES = 15 + 3 + 2 + 1
-N_FAILING = 2
```
tsbinns, line 45: E.g. not sure if I properly adjusted all of these...

sphinx_gallery/tests/test_full.py
```diff
@@ -330,6 +330,47 @@ def test_negative_thumbnail_config(sphinx_app, tmpdir):
...
+    # (i.e. thumbnail should be the plot from the example, not the "BROKEN" stamp)
+    thumb_fname = op.join(
+        sphinx_app.outdir,
+        "_images",
+        "sphx_glr_plot_failing_example_thumbnail_thumb.png",
+    )
+    thumbnail = np.asarray(Image.open(thumb_fname))
+    assert broken_stamp.shape[:2] == thumbnail.shape[:2]
+    corr = np.corrcoef(broken_stamp[..., :3].ravel(), thumbnail[..., :3].ravel())[0, 1]
+    assert corr < 0.99  # i.e. thumbnail and "BROKEN" stamp are not identical
```
tsbinns, line 371: Also not sure what you feel is an appropriate threshold for non-identical images. Is <0.99 too lenient?
The actual correlation here was ~-0.3.

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -1239,6 +1239,5 @@ def test_newlines(log_collector_wrap):
     assert tee.newlines == tee.output.newlines
 
 
-# TODO: test that broken thumbnail does appear when needed
```
tsbinns, line 1242: Since the new test also looks at the default behaviour for expected failing examples, can remove this `TODO` from elsewhere.

sphinx_gallery/tests/test_full.py
```diff
@@ -241,18 +241,42 @@ def test_junit(sphinx_app, tmp_path):
...
+            "tag": "failure",
+            "message": "Passed even though it was marked to fail",
+        },
+    ]
+    assert len(suite) == len(skips_and_fails)
+    for this_suite, this_example in zip(suite, skips_and_fails):
+        assert this_suite.attrib["classname"] == this_example["classname"]
+        assert this_suite[0].tag == this_example["tag"], this_suite.attrib["classname"]
+        if this_example["message"] is not None:
+            assert this_suite[0].attrib["message"].startswith(this_example["message"])
```
tsbinns, line 279: Needed to add new items for the 2 new examples. Thought perhaps this was easier to read.

tsbinns:
That should now be all the tests working.

sphinx_gallery/tests/test_full.py
larsoner: Let's split the difference (roughly) between a perfect correlation and the actual observed one then
```suggestion
    assert corr < 0.7  # i.e. thumbnail and "BROKEN" stamp are not identical
```

larsoner (approved):

larsoner:
Looks good marking for merge-when-green, thanks in advance @tsbinns !

Issue #1316, opened 2024-05-30
requirements.txt not completely removed


AlexSzatmary:
Several sphinx-gallery source files mention `requirements.txt` but that file is not in the repo. I think it was removed in connection to adopting `pyproject.toml`. As a result, some instructions do not work, such as,

`pip install -r requirements.txt -r dev-requirements.txt `
in `README.rst`.

Instructions could be revised to instead say something like, `pip install sphinx-gallery "sphinx>=4.0" pillow` (also in `README.rst`).

It would be better if the requirements were only listed in one place. I think that would be possible and if done correctly, you could tell devs to do something like, `python -m pip install --verbose --no-build-isolation --editable ".[dev]"` (which is how Matplotlib does it).

lucyleeow:
+1

drammock:
this was fixed by #1317 and can be closed.

PR #1317, opened 2024-06-02
Include dev dependencies in pyproject.toml and update docs on requirements

AlexSzatmary:
sphinx-gallery migrated to pyproject.toml (see #1267) but this migration was not fully implemented. In particular, `requirements.txt` was removed but it is still referenced in several files. The developer install instructions were not updated (#1316) but this PR should resolve that issue. Also, the quickstart in README.rst says, "Sphinx-Gallery will not manage its dependencies" but with dependencies listed in pyproject.toml, pip now takes care of that.

I moved the contents of dev-requirements.txt into pyproject.toml and updated docs on how to install for contributors.

I also noticed a lot of "Package would be ignored" warnings. I resolved that by setting `namespaces = true` in pyproject.toml, which resolves that warning and is the default. I diffed the trees for the installed sphinx-gallery with and without that change; there was no difference.


pyproject.toml
```diff
@@ -65,7 +82,7 @@ where = [
 include = [
   "sphinx_gallery*",
 ] # package names should match these glob patterns (["*"] by default)
-namespaces = false # to disable scanning PEP 420 namespaces (true by default)
+namespaces = true # to disable scanning PEP 420 namespaces (true by default)
```
larsoner, line 85:
```suggestion
namespaces = true # allow scanning PEP 420 namespaces (true by default)
```

larsoner (approved):
Just tweaked a comment to reflect the new behavior, making for merge when green, thanks @AlexSzatmary

larsoner:
Oh CIs are unhappy, looks like there are still some references to dev-requirements. Maybe `git grep` can help here

AlexSzatmary:
@larsoner Thanks for the speedy review! I see that CI needed dev-requirements.txt . I will look into telling CI to install using the dev option.

larsoner:
Okay let me know whenever you need me to approve CIs. GH is going to ask on every commit you push as a new contributor 😬

AlexSzatmary:
I looked at.github/install.sh . There are a lot of possible combinations of installed dependencies. Do they still work as intended since the migration to pyproject.toml? It seems pip would install dependencies even for cases where that is not intended, such as minimal. It would be nice if there were at least comments explaining all of the variants.

larsoner:
Yeah it is not super well documented. Want me to try pushing a commit?

In short I would try changing any instance of `dev-requirements.txt` install to later `pip install .[dev]` or so, and otherwise leave the existing combinations alone.

AlexSzatmary:
@larsoner feel free to try that.

Alternatively, I think that,
```
PIP_DEPENDENCIES=`python -c 'import tomllib; data = tomllib.load(open("pyproject.toml", "rb")); print(" ".join(data["project"]["optional-dependencies"]["dev"]))'``
```
is the most one-for-one comparable to the `-r requirements-dev.txt`.

You would know better than I would which is the better approach.

AlexSzatmary:
I'm also not sure if it's better to include --no-build-isolation in 
`python -m pip install --editable --no-build-isolation ".[dev]"`

larsoner (approved):

larsoner:
Thanks @AlexSzatmary ! I went with the minimalist approach in 4178d841d5eee4539f4715989deed45ecc25fe29 of just replacing the `-r` with equivalent `[...]` in the eventual install line. That way we can leave refactoring for later. Though I think in principle we maybe had all those variants for a reason at some point, so probably not worth it for now!

AlexSzatmary:
Awesome, thanks @larsoner !

PR #1318, opened 2024-06-03
DOC: Remove definition lists from contribution guide

QuLogic:
These appear to have been accidental.

I just re-wrapped these paragraphs so that no lines started with a space (hopefully wrapped at the right width for you.)


larsoner:
Thanks @QuLogic !

PR #1320, opened 2024-06-03
Fix handling of multi-module intersphinx registries

QuLogic:
Some registries may contain multiple modules (e.g., Python contains all the standard library modules, or Matplotlib contains modules in the `mpl_toolkits` namespace), and the previous lookup by the intersphinx mapping _only_ could cause these to be missed. There was already some special casing for Python, but this is now extended to all inventories, fixing the case for Matplotlib as well.

Fixes matplotlib/matplotlib#28250

### DIFF
--- a/sphinx_gallery/docs_resolv.py
+++ b/sphinx_gallery/docs_resolv.py
@@ -307,6 +307,34 @@ def _sanitize_css_class(s):
     return s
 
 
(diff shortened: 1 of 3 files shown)


larsoner (approved):
Awesome work, thanks @QuLogic !

Issue #1315, opened 2024-05-29
DOC: Better input and output folder names


timhoffm:
The current docs describle the following folder layout and config:

```
.
├── doc
│   ├── conf.py
│   ├── index.rst
|   ├── make.bat
│   └── Makefile
├── my_python_module
│   ├── __init__.py
│   └── mod.py
└── examples
    ├── plot_example.py
    ├── example.py
    └── README.txt (or .rst)
```

```
sphinx_gallery_conf = {
     'examples_dirs': '../examples',   # path to your example scripts
     'gallery_dirs': 'auto_examples',  # path to where to save gallery generated output
}
```

I find this not ideal for the following reasons.
1. the input folder (`examples`) is at the top level in the project. I rather see it as part of the documentation, so it should be in `docs`.
2. The naming of the input folder `examples` does not give any hint that this is input.
3. The output folder name becomes part of the sphinx document structure and hence URL. While `auto_example` is descriptive on how this is generated, its irrelevant for the user of the docs. `htttps://myproject.org/examples/hello_world.html` reads much better than `https://myproject.org/auto_examples/hello_worlds.html`.

I thus propose to place both folders under `docs` and name the ouput folder `examples` and name the input folder `examples_inputs`, i.e.

```
.
├── doc
│   ├── examples_inputs
│   │   ├── plot_example.py 
│   │   ├── example.py
│   │   └── README.txt (or .rst)
│   ├── conf.py
│   ├── index.rst
|   ├── make.bat
│   └── Makefile
├── my_python_module
    ├── __init__.py
    └── mod.py
```

```
sphinx_gallery_conf = {
     'examples_dirs': 'examples_inputs',   # path to your example scripts
     'gallery_dirs': 'examples',  # path to where to save gallery generated output
}
```

If you agree with the change, I can make a PR.

drammock:
> I thus propose to place both folders under `docs` and name the ouput folder `examples` and name the input folder `examples_inputs`

I like this idea a lot, but would suggest a small tweak: maybe `examples_source_files` instead of `examples_input`?  To me `examples_input` is still a bit unclear (e.g., it could be "examples about the kinds of input my package accepts"). Another idea is to call it `sg_examples` or similar (to more clearly signal these are files meant for sphinx gallery's consumption).

(sorry if this is too bikesheddy, if you don't immediately agree that either suggestion is an improvement, feel free to disregard).

lucyleeow:
[...]

PR #1321, opened 2024-06-05
MNT: Rename README.rst to GALLERY_HEADER.rst

### LINKED ISSUE #1209: Rename README.rst to gallery_header.rst
Using the name `README` for the gallery header is slightly misleading. READMEs typically contain instructions/context information directed at the user of the source code. But as it is used in sphinx gallery, it’s just regular content to be inserted into the gallery overview page.

I suggest to introduce a more telling name, e.g. `gallery_header.txt/rst` and advertise it as the recommended name. README may continue as a fallback to maintain backward compatibility.

timhoffm:
Closes #1209.

Comments:
- All existing README variants are still supported without deprecation. Documentation is rewritten to consistently use
  GALLERY_HEADER and mention backward compatibility with README.
- I've chosen the capitalized form `GALLERY_HEADER` because it stands out more between all the example files. If you rather keep it lowercase, that'd be fine as well, but to keep it more standardized, I suggest not to support both.
- I've chosen to still support all extensions that README supported ('txt' plus `sphinx_gallery_conf["source_suffix"]`), but except for a full definition in `configuration.rst`, I only use `GALLERY_HEADER.rst` everywhere in the docs. This is to converge to one default, and .rst seems better because the content must be ReST. (Not quite whether Markdown could be added as an alternative, but if so, it's still clearer that you switch to `GALLERY_HEADER.md` rather than keep using `GALLERY_HEADER.txt` but having to switch it's content to Markdown.

Open issues:
- [ ] Tests still only contain the README. Is it ok to just switch to to `GALLERY_HEADER` or do we need to test both variants (which would be more work?
- [ ] Is there a place to write a specific change note on this or is the PR title in https://sphinx-gallery.github.io/stable/changes.html all we have? - If the latter, do you want any specific wording?


lucyleeow:
For these name change improvements (also #1315) I wonder if we should go through a deprecation cycle and then deprecate the old names?

timhoffm:
I wasn't sure how much effort you want burden on existing projects. Sure, the change is quite easy and projects should be able to control which version they use for building.
Deprecation would make the docs simpler, because we don't have to keep the filename variants around. It also makes it easier not to support `GALLERY_HEADER.txt`, which I've basically accepted for simplicity of the code. And finally, we may get away with just switching the tests to `GALLERY_HEADER` and leave `README` untested for the deprecation period.

If a deprecation is acceptable, I'm happy to add it.

timhoffm:
Ok, let's not deprecate for now. Could be a v1.0 topic.

doc/configuration.rst
```diff
@@ -2427,15 +2431,15 @@ across rst files. ::
...
 
 Manually passing ``index.rst``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 You can bypass Sphinx-Gallery automatically creating an  ``index.rst`` from a
-``README.txt`` in a gallery directory or subdirectory. If your
+``GALLERY_HEADER.rst`` in a gallery directory or subdirectory. If your
 ``copyfile_regex`` includes ``index.rst``, and you have an ``index.rst`` in the
-gallery-source instead of the README, Sphinx-Gallery will use that instead of
+gallery-source instead of the GALLERY_HEADER, Sphinx-Gallery will use that instead of
```
lucyleeow, line 2442:
```suggestion
gallery-source instead of the 'GALLERY_HEADER' file, Sphinx-Gallery will use that instead of
```
nit

doc/configuration.rst
```diff
@@ -2409,7 +2412,8 @@ Manually passing files
 
 By default, Sphinx-Gallery creates all the files that are written in the
 sphinx-build directory, either by generating rst and images from a ``*.py``
-in the gallery-source, or from  creating ``index.rst`` from ``README.txt``
+in the gallery-source, or from  creating ``index.rst`` from
+``GALLERY_HEADER.rst`` (or ``README.rst/.txt`` for backward-compatibility)
```
lucyleeow, line 2416: What do you think about this instead? (nit)

```suggestion
``GALLERY_HEADER.rst`` (or ``README.[rst/txt]`` for backward-compatibility)
```

lucyleeow (approved):
LGTM, thanks! Just a nit

lucyleeow:
Sorry forgot about this. Thanks @timhoffm !

Issue #1323, opened 2024-06-09
Navigation bar TOC rendering for nested galleries


remy-abergel:
Hi, 

I am trying to document using sphinx a Python package with nested example galleries. I made a simple package to reproduce the issue I am facing. This package can be downloaded here: [pypkg.zip](https://github.com/user-attachments/files/15752783/pypkg.zip), it has the following structure:
```
.
├── docs/
│   ├── conf.py
│   ├── index.md
│   ├── installation.md
│   ├── make.bat
│   └── Makefile
├── examples/
│   ├── experiments/
│       ├── plot_exp_demo1.py
│       ├── plot_exp_demo2.py
│       └── README.txt
│   └── tutorials/
│       ├── 3d/
│           ├── plot_tuto_3d_demo1.py
│           ├── plot_tuto_3d_demo2.py
│           └── README.txt
│       ├── plot_tuto_2d_demo1.py
│       ├── plot_tuto_2d_demo2.py
│       └── README.txt
├── pyproject.toml
├── README.md
└── src/
    └── pypkg/
        ├── foo.py
        └── __init__.py

``` 

The sphinx build works fine and the rendering of the nested gallery seems nice at a first glance: 

![screenshot1](https://github.com/sphinx-gallery/sphinx-gallery/assets/60599693/b1f2414d-9df9-4f3d-8933-4c23c9825f8c)

However, I end up with a strange rendering of the `tutorials/3D` gallery in the navigation bar: 

![screenshot2](https://github.com/sphinx-gallery/sphinx-gallery/assets/60599693/870717f6-f6b1-4261-8b30-d16a21380861)

I am probably using sphinx-gallery in a wrong way, could you please let me know how I can get rid of the undesired `3D/3D` heading in the navigation bar? 

Many thanks!

lucyleeow:
So AFAICT this is intentional, see: https://github.com/sphinx-gallery/sphinx-gallery/pull/944#discussion_r916603188 and https://github.com/sphinx-gallery/sphinx-gallery/issues/1262#issuecomment-1958751206

In https://github.com/sphinx-gallery/sphinx-gallery/issues/1323#issuecomment-2156641586, when you set `'nested_sections': True` you get one toctree at the end of the `index.rst` file, which gives the spurious headings:

```
.. toctree::
   :hidden:
   :includehidden:


   /_gallery/tutorials/2d/index.rst
   /_gallery/tutorials/3d/index.rst
```

[...]

PR #1332, opened 2024-06-19
MNT: Refactor `generate_dir_rst` and `generate_gallery_rst`

lucyleeow:
There has been a few PRs recently adding a fair bit of stuff to `generate_dir_rst` and `generate_gallery_rst`, making them quite long and difficult to understand.

Prompted by trying to read the code after https://github.com/sphinx-gallery/sphinx-gallery/issues/1323.

A lot of simply pulling code out into helper functions, removing redundant things. Notable changes:

* `generate_dir_rst` now only writes index.rst file if `src_dir` is a subsection file
   * previously we over-wrote it later and had a confusing note in  `generate_gallery_rst` about this
   * the old `include_toctree` already indicated whether it was a subsection or not, we just weren't fully utilising it
* some re-naming of ambiguous terms like 'has_header', which were all just code for whether the user provided an `index.rst` file or not
* Refactor of `collect_gallery_files` - this is opinionated so I am happy to revert or amend
   * Moves this helper to `utils/` - to prevent circular import as I wanted to use it in `gen_rst.py`. This is technically a public function but a [quick search on github](https://github.com/search?q=collect_gallery_files&type=code&p=1) seems to suggest only one or two cases of its use, most are forks of SG
   * We now use it twice:
      * get all example files and check for spaces and duplicates (original use)
         * we previously recursed through all sub folder depths. I've amended this to check one level of subfolder as we only make galleries one level deep, so we shouldn't care how the user has named deeper files (?)
         * we only checked .py files previously but with `example_extensions` it makes sense to check all supported exts
      * get example files to iterate through and generate rst for
         * it makes sense that this is done with the same function as the above check


larsoner (approved):
Thanks for working on this refactoring, I agree it should help!

QuLogic:
This appears to have broken (I say broken, as I'm assuming a refactor didn't intend to change behaviour) generating the zip files for the top-level gallery if it has an `index.rst`; [example from Matplotlib build](https://app.circleci.com/pipelines/github/matplotlib/matplotlib/32186/workflows/84335269-9ee2-4667-b929-bb002ed614a5/jobs/84131). Specifically, the PR-internal commit 292d3e37f1c0942756456a7d57788fbf83e5a296 seems to be the root cause. I _think_ this is because the `generate_zipfiles` call in `_finish_index_rst` is now in an `if sg_root_index`? This doesn't generate the zip, but our `index.rst` is referencing it, causing Sphinx to warn/error out.

(On a side note, large refactors like this are why I dislike squash merging PRs; it makes bisecting more difficult than it should be.)

QuLogic:
A change like this does seem to fix it:
```
diff --git a/sphinx_gallery/gen_gallery.py b/sphinx_gallery/gen_gallery.py
index f8b3c7b..89d8998 100644
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -608,13 +608,15 @@ def _finish_index_rst(
         )
         indexst += subsections_toctree
 
-    if sg_root_index:
-        # Download examples
-        if gallery_conf["download_all_examples"]:
-            download_fhindex = generate_zipfiles(
-                gallery_dir_abs_path, app.builder.srcdir, gallery_conf
-            )
+    # Download examples
+    if gallery_conf["download_all_examples"]:
+        download_fhindex = generate_zipfiles(
+            gallery_dir_abs_path, app.builder.srcdir, gallery_conf
+        )
+        if sg_root_index:
             indexst += download_fhindex
+
+    if sg_root_index:
         # Signature
         if app.config.sphinx_gallery_conf["show_signature"]:
             indexst += SPHX_GLR_SIG
```
Should I open a PR, or do you think this should be refactored somewhere else?

Issue #1066, opened 2023-01-18
Make preview text longer


GaelVaroquaux:
The preview text (summary of the example made from the first words of the docstring) could be longer:
![image](https://user-images.githubusercontent.com/208217/213178085-0c95df31-5c38-41b6-84ae-ce0f4746d76e.png)

We should make it much longer and use css to cut (overlap=hidden)

anushkabishnoi:
Hello, this is my first experience with open source contributions. Would really appreciate it if you explained exactly what is needed to be done.

I would love to work on this issue.

Thanks in advance! :)

larsoner:
@anushkabishnoi sorry for the slow response. We generally follow a standard GitHub contribution setup, so feel free to read about how to contribute to scikit-learn, nilearn, matplotlib, MNE-Python, etc. (whichever instructions are clearest/easiest). The TL;DR is that you make a PR with the change, and ideally modify some test that would fail on `master` but pass on your PR (in this case, it could be that the entire > 80 char intro paragraph text shows up in the gallery index for example)

@fgmacedo feel free to give it a shot as well if you want!

Issue #1085, opened 2023-03-04
The gallery intro is truncated and I want to display more content (is not possible to change the size)


fgmacedo:
Hi, first time here, so I want to say thanks for maintaining this project! 

I've found that the `intro` size, the one that is displayed as the tooltip is truncated at `95` chars. 

I've made a [image gallery with large thumbs](https://python-statemachine.readthedocs.io/en/latest/auto_examples/index.html) and It would be nice to have more content on the tooltip.

![image](https://user-images.githubusercontent.com/281007/222931710-202cbf02-5317-4697-850d-545099fd2017.png)

Here's the code that truncates:

Ref: https://github.com/sphinx-gallery/sphinx-gallery/blob/60b4b5b9597efa7957dc4437ae0aa301192f45b3/sphinx_gallery/gen_rst.py#L243-L244

It would be nice if there's some way to get more `intro` content displayed, being a configuration or another way.

Best regards,

larsoner:
I wonder if what we should instead do is use CSS to set overflow text to use ellipsis

https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/

WDYT?

fgmacedo:
I like this solution @larsoner , works for me!

fgmacedo:
Closing as duplicated of https://github.com/sphinx-gallery/sphinx-gallery/issues/1066

PR #1344, opened 2024-07-07
Introduction tooltip corresponds to the first paragraph

fgmacedo:
Closes: #1066 (related to #1085).

- [x] Introduction tooltip corresponds to the first paragraph and is no longer truncated.
- [x] Tooltip representation is handled using CSS.

#### Without adjusting the `padding`, I got this:

image

Note that given the padding was `10px 10px 10px 10px`, the bottom exposes the top content of the next truncated line:

image

#### Adjusting the `padding`, I got this:

image

### With full default CSS

I've adjusted the clamp to 6 lines, so longer texts use full space, and smaller ones are automatically reduced (states forced to :hover):

image

### DIFF
--- a/sphinx_gallery/_static/sg_gallery.css
+++ b/sphinx_gallery/_static/sg_gallery.css
@@ -185,16 +185,18 @@ thumbnail with its default link Background color */
   border-radius: 4px;
   color: var(--sg-tooltip-foreground);
   content: attr(tooltip);
-  padding: 10px;
+  padding: 10px 10px 5px;
   z-index: 98;
   width: 100%;
-  height: 100%;
   position: absolute;
   pointer-events: none;
   top: 0;
   box-sizing: border-box;
   overflow: hidden;
   backdrop-filter: blur(3px);
+  display: -webkit-box;
+  -webkit-box-orient: vertical;
+  -webkit-line-clamp: 6;
 }
 
 .sphx-glr-script-out {
--- a/sphinx_gallery/_static/sg_gallery.css
+++ b/sphinx_gallery/_static/sg_gallery.css
@@ -178,8 +178,27 @@ thumbnail with its default link Background color */
   max-height: 112px;
   max-width: 160px;
 }
-.sphx-glr-thumbcontainer[tooltip]:hover:after {
-  background: var(--sg-tooltip-background);
+
+.sphx-glr-thumbcontainer[tooltip]::before {
+  content: "";
+  position: absolute;
+  pointer-events: none;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  z-index: 97;
+  background-color: var(--sg-tooltip-background);
+  backdrop-filter: blur(3px);
+  opacity: 0;
+  transition: opacity 0.3s;
+}
+
+.sphx-glr-thumbcontainer[tooltip]:hover::before {
+  opacity: 1;
+}
+
+.sphx-glr-thumbcontainer[tooltip]:hover::after {
   -webkit-border-radius: 4px;
   -moz-border-radius: 4px;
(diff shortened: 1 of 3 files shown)


lucyleeow (commented):
I don't love adding another configuration but if it is not achievable via CSS this seems reasonable.
I'll let @larsoner merge.

larsoner:
> if it is not achievable via CSS this seems reasonable

I think this is probably what we want, and looks like FireFox added support in 2019 so probably okay to use nowadays:

https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp#truncating_a_paragraph

It's a bigger PR but in principle then people can just override the number of lines, box width, etc. to their liking rather than us having some artificial character limit.

@lucyleeow do you agree it's a better way forward? If so then @fgmacedo would you be up for trying this?

drammock:
`height` and `padding` are going to play a role here. Probably for the CSS approach to work and still look good, the semitransparent overlay will need to be a separate node from the tooltip text container. By hacking in the browser console like this:

Screenshot_2024-07-09_16-37-43


I can get this:

Screenshot 2024-07-09 at 16-37-23 Basics Gallery with Matplotlib — Sphinx-Gallery 0 17 dev0-git documentation

lucyleeow:
> @lucyleeow do you agree it's a better way forward? 

100% prefer CSS option. We don't want another config.

fgmacedo:
Changing the approach:

1. Introduction tooltip corresponds to the first paragraph and is no longer truncated.
2. Tooltip representation is handled using CSS.

Do you agree with this?

larsoner:
Yes that sounds good!

Looks like there is a tiny visual regression in [this PR](https://output.circle-artifacts.com/output/job/77c0dec4-3193-4ebb-b365-1404cd55c58e/artifacts/0/html/auto_examples/index.html#general-examples) where the mouseover "black box" doesn't cover the whole tile, for example mousing over I see

image

Whereas on [latest dev](https://sphinx-gallery.github.io/dev/auto_examples/index.html#general-examples) I see:

image

Not sure if it's fixable or not...? One solution would be to make one div with `height=100%` that covers the whole tile, and then inside that the next `div` (or whatever) that does the text trucncation.

drammock:
> One solution would be to make one div with `height=100%` that covers the whole tile, and then inside that the next `div` (or whatever) that does the text trucncation.

That's exactly what I was trying to suggest above  when I said

> the semitransparent overlay will need to be a separate node from the tooltip text container

But @larsoner has said it much more clearly :)

fgmacedo:
Still working on this... the last commit I've added the black box as `before` element, but added a regression that the text is leaking the box... not work as expected. 🙈 

image

fgmacedo:
Tests passing locally. Is `sphinx_gallery/tests/test_full.py::test_interactive_example_logo_exists` an intermittent test?

lucyleeow:
real but unrelated to this PR. Its present in #1346 as well.

fgmacedo:
I've currently set the default to 4 lines, but I might have been a bit conservative. How about increasing it to 6 lines?

image

Preview:

https://output.circle-artifacts.com/output/job/1d60ae2a-9e1f-421d-bcf5-3d55c8cb487e/artifacts/0/html/auto_examples/index.html

larsoner:
Failure in sphinx dev was due to https://github.com/sphinx-doc/sphinx/pull/12425 and I pushed a commit to #1346. Once that lands I'll merge `master` into this PR which should get things green, and I'll look at the examples. Agreed 6 is probably better than 4 for line clamping

larsoner:
Rendered doc looks good and locally I tested it on MNE-Python and things looked good so in it goes, thanks @fgmacedo !

PR #1353, opened 2024-07-24
FIX generate zipfiles when index passed by user

lucyleeow:
Fixes https://github.com/sphinx-gallery/sphinx-gallery/pull/1332#issuecomment-2244559782

When top level index.rst is passed by user we were not generating the zipfiles.

### DIFF
--- a/sphinx_gallery/tests/test_full.py
+++ b/sphinx_gallery/tests/test_full.py
@@ -83,7 +83,14 @@ def _sphinx_app(tmpdir_factory, buildername):
     src_dir = Path(__file__).parent / "tinybuild"
 
     def ignore(src, names):
-        return ("_build", "gen_modules", "auto_examples")
+        return (
+            "_build",
+            "gen_modules",
+            "auto_examples",
+            "auto_examples_README_header",
+            "auto_examples_rst_index",
+            "auto_examples_with_rst",
+        )
 
     shutil.copytree(src_dir, temp_dir, ignore=ignore)
     # For testing iteration, you can get similar behavior just doing `make`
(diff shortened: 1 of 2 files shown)


sphinx_gallery/tests/test_full.py
```diff
@@ -83,7 +83,14 @@ def _sphinx_app(tmpdir_factory, buildername):
...
 
     def ignore(src, names):
-        return ("_build", "gen_modules", "auto_examples")
+        return (
+            "_build",
+            "gen_modules",
+            "auto_examples",
+            "auto_examples_README_header",
+            "auto_examples_rst_index",
+            "auto_examples_with_rst",
```
lucyleeow, line 92: Noticed these were missing, doesn't affect tests but I think if someone is running locally, prevents test copying over previously generated gallery files

lucyleeow:
@QuLogic thanks, your fix was perfect.

Do you need a patch release quickly or are you happy to wait a few days in case we broke other things?

larsoner:
@lucyleeow I would say it's worth making a quick point release

QuLogic:
> Do you need a patch release quickly or are you happy to wait a few days, to wait and see if we broke other things?

I am eager to try out some of the new features, but we are not close to a release, I think, so there's no current rush, and could wait a week or two.

lucyleeow:
@larsoner do you think we should wait a few days (sphinx 8.0 has just been released too) or release now?

larsoner:
We already test against Sphinx dev so hopefully we're safe there. My vote is to release now but it depends on how much extra work it creates.

PR #1355, opened 2024-07-25
FIX Allow str path minigallery entries when backreferences off

### LINKED ISSUE #1354: `minigallery` raises cryptic error when `backreferences_dir` is `None`
I encountered this while trying to add a mini-gallery based on explicit file paths, following the documentation at https://sphinx-gallery.github.io/stable/configuration.html#create-mini-galleries-using-file-paths.  That section doesn't mention that a definition of `backreferences_dir` is required, and earlier sections describe the configuration as being "the directory where function/class granular galleries are stored".  Since I wasn't trying to generate function/class granular galleries, I left the configuration empty.

The result is the following error:
```
Exception occurred:
  File "/usr/local/lib/python3.12/pathlib.py", line 373, in __init__
    raise TypeError(
TypeError: argument should be a str or an os.PathLike object where __fspath__ returns a str, not 'NoneType'
```

I thought it was complaining about the paths I'd supplied to the `minigallery` directive, and it took me a long time to work out that this arises due to `backreferences_dir` being set to `None`.

I think one or more of the following would improve UX here:

- Generating a meaningful error message if `backreferences_dir` is `None`.
- Emphasizing in the documentation that `backreferences_dir` must be set even for mini-galleries based on explicit file paths.

lucyleeow:
Closes #1354

Amend the minigallery directive, such that when `backreferences_dir` is `None`, we simply treat all entries as str paths and glob them.

I thought about adding additional check to the minigallery entries, to make sure they are not objects. But it would be very unusual to have a file named e.g., `sphinx_gallery.sorting.ExplicitOrder` in the source dir?
Happy to add a check though, WDYT @larsoner ?

Also fixes a bug where we were not converting `examples_dirs` and `gallery_dirs` to list when given as str.

Note to self: look into: https://github.com/sphinx-gallery/sphinx-gallery/pull/1226#discussion_r1690889760

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -50,7 +50,7 @@ file, inside a ``sphinx_gallery_conf`` dictionary.
 
 - ``reference_url``, ``prefer_full_module`` (:ref:`link_to_documentation`)
 - ``backreferences_dir``, ``doc_module``, ``exclude_implicit_doc``,
-  and ``inspect_global_variables`` (:ref:`references_to_examples`)
+  and ``inspect_global_variables`` (:ref:`minigalleries_to_examples`)
 - ``minigallery_sort_order`` (:ref:`minigallery_order`)
 
 **Images and thumbnails**
@@ -659,9 +659,17 @@ directive so that you can easily add a reduced version of the Gallery to
 your Sphinx documentation ``.rst`` files. The mini-gallery directive therefore
 supports passing a list (space separated) of any of the following:
 
-* full qualified name of object (see :ref:`references_to_examples`)
+* full qualified name of object (see :ref:`references_to_examples`) - this
+  adds all examples where the object was used in the code or referenced in
+  the example text
 * pathlike strings to example Python files, including glob-style
   (see :ref:`file_based_minigalleries`)
+
+To use object names, you must enable backreference generation, see
+:ref:`references_to_examples` for details.
+If backreference generation is not enabled, object entries to the
+:class:`~sphinx_gallery.directives.MiniGallery` directive will be ignored
+and all entries will be treated as pathlike strings.
 
 .. _references_to_examples:
 
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -656,10 +656,10 @@ Add mini-galleries
 
 Sphinx-Gallery provides the :class:`sphinx_gallery.directives.MiniGallery`
 directive so that you can easily add a reduced version of the Gallery to
-your Sphinx documentation ``.rst`` files. The mini-gallery directive therefore
+your Sphinx documentation ``.rst`` files. The minigallery directive therefore
 supports passing a list (space separated) of any of the following:
 
-* full qualified name of object (see :ref:`references_to_examples`) - this
+* fully qualified name of object (see :ref:`references_to_examples`) - this
   adds all examples where the object was used in the code or referenced in
   the example text
 * pathlike strings to example Python files, including glob-style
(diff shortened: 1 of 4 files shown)


doc/configuration.rst
```diff
@@ -659,10 +659,18 @@ directive so that you can easily add a reduced version of the Gallery to
...
 
+To use object names, you must enable backreference generation, see
+:ref:`references_to_examples` for details.
+If backreference generation is not enabled, object entries to the
+:class:`~sphinx_gallery.directives.MiniGallery` directive will be ignored
+and all entries will be treated as pathlike strings.
+
 .. _references_to_examples:
 
 Add mini-galleries for API documentation
```
lucyleeow, line 676: @larsoner reading:

> The former is useful for auto-documenting functions/methods/attributes/objects
that are used and classes that are explicitly instantiated. The generated links
are called implicit backreferences.

> The latter is useful for classes that are
typically implicitly returned rather than explicitly instantiated (e.g.,
:class:`matplotlib.axes.Axes` which is most often instantiated only indirectly
within function calls). Such links are called explicit backreferences.

From the explanation it sounds like code use backreferences should be 'explicit' and text backreferences should be called 'implicit'? This explanation is a bit confusing at the very least.
Do you know the orginal naming details?

lucyleeow:
@larsoner will release after this FYI

lucyleeow:
> "../../examples" as input

As input to the minigallery directive? But would glob find anything if it was the wrong relative path?

lucyleeow:
>  same file inside the examples_dir folder, but the paths don't evaluate to the same path , 

sorry I do not follow what you mean. Can you give an example?

>  I think something like input needs to be relative path but making the thumbnail needed this path

Are you are talking about:

https://github.com/sphinx-gallery/sphinx-gallery/blob/2fc5067f519ee17de066471cf1e1f2eeffb51add/sphinx_gallery/directives.py#L151

Could you give an example? E.g., if I wanted to test this error, what would the test look like?

doc/configuration.rst
larsoner: Not 100% sure what I meant :) But I think the idea was maybe that the things that are explicit in the code (function calls, instantiations) are implicitly/automatically added as backrefs by SG. Things that are only implicit in the code (like `Axes`) might need to be made explicit.

lucyleeow: Cool, I will try to improve the docs. I want to keep the names as this is what they are called in the code.

lucyleeow:
> this will trigger the error: ../../tinybuild/examples/*/*.py

I get this error:

```
sphinx.errors.ExtensionError: Error in minigallery file lookup: input=../../tinybuild/examples/*/*.py, matches=[], examples_dirs=['../examples/', '../examples_with_rst/', '../examples_rst_index', '../examples_README_header']
```

But again we are in the case where there are NO matches - i.e., glob found files, but the files aren't in `examples_dirs`.

What is an example of a case where there are >1 matches?

sphinx_gallery/directives.py
```diff
@@ -67,12 +70,16 @@ def run(self):
         # Retrieve the backreferences directory
         config = self.state.document.settings.env.config
         backreferences_dir = config.sphinx_gallery_conf["backreferences_dir"]
+        if backreferences_dir is None:
+            logger.warning(
```
lucyleeow, line 74: Also possibly silly question, but I note a `self.warning` (using `Directive.warning`) used above - in what case should we be using the `Directive` warning and in what case should we be using our logger?

I tried `self.warning` and it did not output to the build. Where does the warning output to..?

lucyleeow:
Okay I think I'm going to open another issue for the match error, because it probably shouldn't hold up this PR.

This is ready for review/merge. I just had a question about warnings: https://github.com/sphinx-gallery/sphinx-gallery/pull/1355#discussion_r1692324187

lucyleeow:
Let's continue the conversation in #1356

@story645 any insights on https://github.com/sphinx-gallery/sphinx-gallery/pull/1355#discussion_r1692324187 ? I see you use a `self.warning` in the directive?

lucyleeow:
> an improvement/follow up if I have time would be to make a list of (example/thumbnail/intro/title) so examples don't need to get reparsed here and that would probably clean up some of this.

re-reading, could you expand? Where would the list live?

lucyleeow:
Ah okay, maybe with the thumbnail images as image metadata. But we'd have to consider whether it's worth all the extra work generating the files, performance wise (especially as some projects may not use the minigallery directive). Anyway, things to consider.

@larsoner I'll merge this next week, in case you want to take a look.

lucyleeow:
Merging this.

I *think* there is no need to add extra checks to prevent us from glob-ing object inputs:

>  adding additional check to the minigallery entries, to make sure they are not objects. But it would be very unusual to have a file named e.g., sphinx_gallery.sorting.ExplicitOrder in the source dir?

But @larsoner let me know if you disagree!

PR #1360, opened 2024-07-30
Improve minigallery directive path input resolution

### LINKED ISSUE #1356: Minigallery directive file checking
It's unclear what we want to be checking with:

https://github.com/sphinx-gallery/sphinx-gallery/blob/2fc5067f519ee17de066471cf1e1f2eeffb51add/sphinx_gallery/directives.py#L136-L146

It seems originally the check was for cases where `dirs > 1` (see https://github.com/sphinx-gallery/sphinx-gallery/pull/1226#discussion_r1416586216) BUT we don't seem to be able to make this happen (see https://github.com/sphinx-gallery/sphinx-gallery/pull/1355#issuecomment-2251500148). It would be great if we could work out what we are meant to be checking here but :shrug:  if not.

We **can** reach this error by making `dirs = 0`. I think this check would be to ensure the glob files we find live in the examples file, which is a reasonable check to have.

Even this is not working though. Using `../../tinybuild/examples/*/*.py` as a minigallery entry in tinybuild (as suggested: https://github.com/sphinx-gallery/sphinx-gallery/pull/1355#issuecomment-2251484364), `dirs=0` even though the examples found do live within an `examples_dirs`. This is because `'../../tinybuild/examples/*/*.py'.find('../examples/')` gives -1.

We can fix this by amending the check to:

```python
if (obj.find(Path(e).name) != -1)
```

BUT then we get this error:

> ValueError: '/home/<>/sphinx-gallery/sphinx_gallery/tests/tinybuild/doc/../../tinybuild/examples_README_header/plot_exp.py' is not in the subpath of '/home/<>/sphinx-gallery/sphinx_gallery/tests/tinybuild/doc/../examples' OR one path is relative and the other is absolute.

If I recall correctly, @story645 , we wanted to keep the glob paths we find to be relative, to allow sorting. BUT can we pass the relative paths to the sortkey but use abs paths otherwise in the directive? I think this would make the above check (ensuring the glob paths found live within an `examples_dir`) much easier/better as well as avoiding errors like the one above.

Would you be up for tacking this? I have spent more time than I should on SG :grimacing:

lucyleeow:
closes #1356

Uses `relative_to` to get target gallery path - in particular improves the part where we figure out if the file is in a subgallery/subfolder. I think the `relative_to` is more robust and I think it should work as intended.

e.g., 
one glob match: `'.../sphinx-gallery/sphinx_gallery/tests/tinybuild/examples_rst_index/examp_subdir2/plot_sub2.py'`
example dir path: `.../sphinx_gallery/sphinx_gallery/tests/tinybuild/examples_rst_index/`

Uses `resolve` to get abs paths.

cc @story645

### DIFF
--- a/sphinx_gallery/directives.py
+++ b/sphinx_gallery/directives.py
@@ -108,9 +108,9 @@ class MiniGallery(Directive):
         file_paths = []
         for obj in obj_list:
             if backreferences_dir and (path := has_backrefs(obj)):
-                file_paths.append((obj, path))
+                file_paths.append((obj, path.resolve()))
             elif paths := Path(src_dir).glob(obj):
-                file_paths.extend([(obj, p) for p in paths])
+                file_paths.extend([(obj, p.resolve()) for p in paths])
 
         if len(file_paths) == 0:
             return []
@@ -126,7 +126,7 @@ class MiniGallery(Directive):
             )
         for obj, path in sorted(
             set(file_paths),
-            key=((lambda x: sortkey(os.path.abspath(x[-1]))) if sortkey else None),
+            key=((lambda x: sortkey(str(x[-1]))) if sortkey else None),
         ):
             if path.suffix == ".examples":
                 # Insert the backreferences file(s) using the `include` directive.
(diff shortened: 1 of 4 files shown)


sphinx_gallery/directives.py
```diff
@@ -144,21 +144,41 @@ def has_backrefs(obj):
...
+                    )
+                # `path` inside several `examples_dirs`, take the longest match
+                elif n_match > 1:
+                    ex_parents, target_dir = max(
+                        [(match[0].parents, match[1]) for match in gal_matches],
+                        key=lambda x: len(x[0]),
+                    )
+                # `path` is not inside a `examples_dirs`
+                else:
                     raise ExtensionError(
```
lucyleeow, line 170: This is the case where the glob match is not inside an `examples_dirs` (i.e., in old code when `len(dirs) = 0`).

sphinx_gallery/directives.py
```diff
@@ -144,21 +144,41 @@ def has_backrefs(obj):
...
+                    ex_parents, target_dir = (
+                        gal_matches[0][0].parents,
+                        gal_matches[0][1],
+                    )
+                # `path` inside several `examples_dirs`, take the longest match
+                elif n_match > 1:
+                    ex_parents, target_dir = max(
+                        [(match[0].parents, match[1]) for match in gal_matches],
+                        key=lambda x: len(x[0]),
+                    )
```
lucyleeow, line 167: I think this may have been what the old `len(dirs)>1` was testing before. The only case where I think this could come about is e.g.,

file: `/a/b/c/d/file.py`
examples dirs (abs path):
* `/a/b/`
* `/a/b/c/d` (i.e. you have a gallery within a sub-sub-folder of another gallery - this *may* be done if someone wanted to use more nesting for their galleries)

In this case, I think it makes sense that we just use the longest matching example dir.

lucyleeow (commented):
I've described how I think we'd get to the errors below.

I am not sure how best to test it, as this would error on a build. I think I could add a test to `test_gen_gallery.py` where I copy over a new `.rst` file that has a minigallery directive that would error (using the `sphinx_app_wrapper.srcdir` to get the right path).

Would this be the best way to do this @larsoner ? I thought about making a new fixture, like `sphinx_app_wrapper` but copying different src files, but I think if I can just copy over files using the same `sphinx_app_wrapper` fixture, it would be better?

larsoner:
> Would this be the best way to do this @larsoner ? I thought about making a new fixture, like sphinx_app_wrapper but copying different src files, but I think if I can just copy over files using the same sphinx_app_wrapper fixture, it would be better?

Yeah I think reusing the existing `sphinx_app_wrapper` is reasonable

sphinx_gallery/directives.py
lucyleeow: Hmm I don't think that is the same. E.g., `examples_dirs` in mpl would be e.g., `../galleries/examples`, `../galleries/plot_types/`. These folders have sub-gallaries/sub-folders, but these subfolderes do not need to be added to `examples_dirs` as SG auto checks for 1 level of sub-folders. 

It's only in the case where you want 2 levels of sub-folders e.g., you also have `../galleries/examples/animation/moreanimation/` in your `examples_dirs`.

It would be unusual case. And I don't even know if I can be bothered adding a test for this... :thinking:

lucyleeow:
This is ready for merge, would be good to get your eye over this @larsoner !

* added tests for the minigallery cases where the file is not in an examples dir and when its from a file in a gallery nested inside another gallery (here the file path matches 2 examples_dirs)
* moved things to use pathlib

lucyleeow:
Figuring out `sphinx_app_wrapper` was a wild ride.

For posterity, it was confusing that the `conf_file` fixture and the `conf_file` pytest mark are called the same thing. They are separate and would work fine if they were name differently.

larsoner:
> For posterity, it was confusing that the conf_file fixture and the conf_file pytest mark are called the same thing. They are separate and would work fine if they were name differently.

Feel free to open a PR to rename one of them!

larsoner:
Thanks @lucyleeow !

lucyleeow:
Good idea, somehow I am just hesitant about changing existing 'convention'.

PR #1362, opened 2024-08-01
DOC add note on filtering joblib warnings

lucyleeow:
deals with https://github.com/sphinx-gallery/sphinx-gallery/pull/877#issuecomment-2253683803

@larsoner @QuLogic I am not certain I 100% followed your comments so please let me know if the doc update is not right. Thank you

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -2160,8 +2160,8 @@ about a `worker restarting <https://github.com/joblib/joblib/issues/883>`_),
 will be captured by Sphinx-Gallery in a similar way to warnings during example
 execution. This can be filtered out with ``warnings.filterwarnings``
 (see :ref:`removing_warnings`). Note that this
-differs from the ``- W`` / ``--fail-on-warning`` ``sphinx-build`` flag, which
-converts warnings during documentation building into errors.
+differs from warnings affected by the ``- W`` / ``--fail-on-warning`` ``sphinx-build``
+flag, which converts warnings during documentation building into errors.
 
 .. warning::
     Some packages might not play nicely with parallel processing, so this feature
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -2157,8 +2157,9 @@ Sphinx.
 
 Warnings emitted by :mod:`joblib` during gallery generation (e.g., the ``UserWarning``
 about a `worker restarting <https://github.com/joblib/joblib/issues/883>`_),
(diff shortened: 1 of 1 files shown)


doc/configuration.rst
```diff
@@ -2155,6 +2155,14 @@ If an ``int``, then that number of jobs will be passed to :class:`joblib.Paralle
 If ``True``, then the same number of jobs will be used as the ``-j`` flag for
 Sphinx.
 
+Warnings emitted by :mod:`joblib` during gallery generation (e.g., the ``UserWarning``
+about a `worker restarting <https://github.com/joblib/joblib/issues/883>`_),
+will be captured by Sphinx-Gallery in a similar way to warnings during example
+execution. This can be filtered out with ``warnings.filterwarnings``
+(see :ref:`removing_warnings`). Note that this
```
larsoner, line 2162:
```suggestion
(see :ref:`removing_warnings`). This can be particularly important
to do if you have tweaked warning handling in your doc build
to treat warnings as errors, e.g., with a line like
``warnings.filterwarnings("error")``. Note that this
```

larsoner:
Just a tiny tweak to mention the line that @QuLogic and I both have that treats all warnings as errors, thanks @lucyleeow !

doc/configuration.rst
lucyleeow: What does 

> warning handling in your doc build
to treat warnings as errors

mean specifically? It is different from the `-W` flag in `sphinx-build` right?

Is what I said below correct?

> Warnings emitted by :mod:`joblib` during gallery generation,
will be captured by Sphinx-Gallery in a similar way to warnings during example

If so, I would except the warning to be captured and put in the output blocks of examples?

larsoner: > mean specifically? It is different from the -W flag in sphinx-build right?

Yes, if you add your own custom warning filter like for example

https://github.com/matplotlib/matplotlib/blob/adb97e848ca88f0b9abc5b1c5085f548847e3f8f/doc/conf.py#L104

The difference these are *Python* warnings (from any module!) treated as errors. `-W` has to do with treating *Sphinx* warnings (which are otherwise just log messages) as errors.

> If so, I would except the warning to be captured and put in the output blocks of examples?

It might be, not sure. I actually don't know if that part is as important as the point that "if you treat Python warnings as errors, this joblib warning can really break stuff / cause your build to hang"

PR #1363, opened 2024-08-01
Fix linking to class attributes with prefer_full_module

QuLogic:
In the case of class attributes, only the short module name was attempted as fallback, instead of complying with the `prefer_full_module` setting.


timhoffm (approved):

larsoner:
Would be great to have a test that would have caught this. Maybe something you could easily add to #1364 when you're messing with `tinybuild` anyway?

larsoner:
Thanks @QuLogic !

Issue #617, opened 2020-03-09
Master failing with sphinx dev


lucyleeow:
Master is failing with sphinx dev for tinybuild test `def test_embed_links_and_styles(sphinx_app)` assertion:

```
assert 'sphinx_gallery.backreferences.html#sphinx_gallery.backreferences.identify_names' in lines  # noqa: E501
```
This assertion was to test issue #583 

I built the doc with sphinx dev and noticed the link is:
```
gen_modules/sphinx_gallery.backreferences.html#sphinx-gallery-backreferences-identify-names
```
note the `-` instead of `_` and `.`'s

What do you suggest @larsoner? Amend the test to allow for both or wait for now?

PR #1364, opened 2024-08-01
Obey prefer_full_module setting when finding backreferences

QuLogic:
This fixes several cases of missing backreferences in Matplotlib's `mpl_toolkits` namespace, which need the full module name for referencing into Sphinx.

For example, [`Axes3D.plot`](https://matplotlib.org/stable/api/_as_gen/mpl_toolkits.mplot3d.axes3d.Axes3D.plot.html) only has one back reference:
image

but with this change we have all these:
image

### DIFF
--- a/sphinx_gallery/tests/test_full.py
+++ b/sphinx_gallery/tests/test_full.py
@@ -54,7 +54,7 @@ N_INDEX = 2 + 1 + 3 + 1
 # (examples + examples_rst_index + examples_with_rst + examples_README_header + root-level)
 N_EXECUTE = 2 + 3 + 1 + 1 + 1
 # gen_modules + sg_api_usage + doc/index.rst + minigallery.rst
-N_OTHER = 9 + 1 + 1 + 1 + 1
+N_OTHER = 11 + 1 + 1 + 1 + 1
 N_RST = N_EXAMPLES + N_PASS + N_INDEX + N_EXECUTE + N_OTHER
 N_RST = f"({N_RST}|{N_RST - 1}|{N_RST - 2})"  # AppVeyor weirdness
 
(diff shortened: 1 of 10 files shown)


larsoner:
Would be great to have a test that would have caught this. @QuLogic WDYT about adding an Axes3D to one of the `sphinx_gallery/tests/tinybuild` examples? On `master` that wouldn't backref / minigallery properly but it would on this PR, right? If so then adding it there you could check in `sphinx_gallery/tests/test_full.py` somewhere that it works. (Maybe you'd need to add a custom minigallery to the build as well?)

QuLogic:
But for back references, it would have to be something in the documented module, so from `sphinx_gallery` for `tinybuild`, wouldn't it?

lucyleeow:
I think any module listed in `doc_module` would work, see: https://sphinx-gallery.github.io/dev/configuration.html#add-mini-galleries-for-api-documentation

QuLogic:
I added a `_dummy` module, and moved `DummyClass` there, along with a nested `NestedDummyClass`. And the added tests should use that and cover this and #1363.

sphinx_gallery/tests/test_full.py
```diff
@@ -548,25 +548,49 @@ def test_embed_links_and_styles(sphinx_app):
     # issue 617 (regex '-'s)
     # instance
     dummy_class_inst = re.search(
-        r'sphinx_gallery.backreferences.html#sphinx[_,-]gallery[.,-]backreferences[.,-][D,d]ummy[C,c]lass" title="sphinx_gallery.backreferences.DummyClass" class="sphx-glr-backref-module-sphinx_gallery-backreferences sphx-glr-backref-type-py-class sphx-glr-backref-instance"><span class="n">dc</span>',  # noqa: E501
+        r'sphinx_gallery._dummy.html#sphinx[_-]gallery[.-]_dummy[.-][Dd]ummy[Cc]lass" title="sphinx_gallery._dummy.DummyClass" class="sphx-glr-backref-module-sphinx_gallery-_dummy sphx-glr-backref-type-py-class sphx-glr-backref-instance"><span class="n">dc</span>',  # noqa: E501
```
QuLogic, line 551: I also fixed the regex here, as #617 seems to be saying that `_` or `.` may be `-` instead, but these seem to be also allowing `,` which seems unlikely.

larsoner:
Yikes I thought that would be simpler, thanks for working through it @QuLogic !

lucyleeow:
Stupid question, could we have also tested this by adding mpl to `doc_modules` and adding a `minigallery` directive with `Axes3D.plot` ? (obv this seems more difficult than amending dummy)

PR #1365, opened 2024-08-02
MNT Change mark and fixture names for adding files

lucyleeow:
See: https://github.com/sphinx-gallery/sphinx-gallery/pull/1360#issuecomment-2263163511

`conf_file` fixture and mark were name the same which makes it confusing.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -2161,8 +2161,7 @@ will be captured by Sphinx-Gallery in a similar way to warnings during example
 execution. This can be filtered out with ``warnings.filterwarnings``
 (see :ref:`removing_warnings`). This can be particularly important
 to do if you have tweaked warning handling in your doc build
-to treat warnings as errors, e.g., with a line like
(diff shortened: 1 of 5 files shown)


doc/configuration.rst
```diff
@@ -2161,8 +2161,7 @@ will be captured by Sphinx-Gallery in a similar way to warnings during example
 execution. This can be filtered out with ``warnings.filterwarnings``
 (see :ref:`removing_warnings`). This can be particularly important
 to do if you have tweaked warning handling in your doc build
-to treat warnings as errors, e.g., with a line like
-``warnings.filterwarnings("error")``. Note that this
+to treat warnings as errors. Note that this
```
larsoner, line 2164: Please revert, this `filterwarnings("error")` is really the relevant hint that projects like `matplotlib` and `mne` would have likely noticed.

lucyleeow: Ah I did not understand what this filter was doing. Let me update the doc in a new PR and I'll revert here

PR #1366, opened 2024-08-02
ENH: Improve _sanitize_rst

timhoffm:
When using the sphinx config `default_role = 'obj'`, one can leave out the prefixes :meth: and similar and directly reference code objects using backticks. In there, one can use the sphinx prefix modifiers `.` and `~`.

Examples from Matplotlib where these show up:

image image

This PR enhances `_sanitize_rst` to remove the prefix modifiers:

- `.MyClass` --> MyClass
- `~.MyClass.close` --> close


larsoner (approved):

larsoner:
Thanks @timhoffm !

PR #1367, opened 2024-08-03
DOC Improve doc about joblib warnings

lucyleeow:
Follow on from #1362

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -2155,18 +2155,19 @@ If an ``int``, then that number of jobs will be passed to :class:`joblib.Paralle
 If ``True``, then the same number of jobs will be used as the ``-j`` flag for
 Sphinx.
 
-Python warnings emitted by :mod:`joblib` during gallery generation (e.g., the
+Warnings emitted by :mod:`joblib` during documentation building (e.g., the
 ``UserWarning`` about a
-`worker restarting <https://github.com/joblib/joblib/issues/883>`_),
-will be captured by Sphinx-Gallery. These can be filtered out with
+`worker restarting <https://github.com/joblib/joblib/issues/883>`_) are emitted
+during gallery generation at the same time as warnings from example
+code execution. These can be filtered out with
 ``warnings.filterwarnings`` (see :ref:`removing_warnings`). This is particularly
-important to do if you have tweaked python warning handling in your documentation build
+important to do if you have tweaked warning handling in your documentation build
 to treat warnings as errors, e.g., with a line like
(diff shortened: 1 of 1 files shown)


doc/configuration.rst
```diff
@@ -2155,16 +2155,18 @@ If an ``int``, then that number of jobs will be passed to :class:`joblib.Paralle
...
-Warnings emitted by :mod:`joblib` during gallery generation (e.g., the ``UserWarning``
-about a `worker restarting <https://github.com/joblib/joblib/issues/883>`_),
-will be captured by Sphinx-Gallery in a similar way to warnings during example
-execution. This can be filtered out with ``warnings.filterwarnings``
-(see :ref:`removing_warnings`). This can be particularly important
-to do if you have tweaked warning handling in your doc build
+Python warnings emitted by :mod:`joblib` during gallery generation (e.g., the
+``UserWarning`` about a
+`worker restarting <https://github.com/joblib/joblib/issues/883>`_),
+will be captured by Sphinx-Gallery. These can be filtered out with
```
lucyleeow, line 2161: Also not sure if these warnings are included in code execution output, so I've left at a vague 'capture'...?

lucyleeow: Actually thinking again, it may be to do with when the warning is emitted, at the time of gallery generation/example execution which is why it causes examples to fail (when warnings are error)

lucyleeow: I've amended the doc, hopefully it is right and makes sense

doc/configuration.rst
```diff
@@ -2155,16 +2155,19 @@ If an ``int``, then that number of jobs will be passed to :class:`joblib.Paralle
...
+`worker restarting <https://github.com/joblib/joblib/issues/883>`_) are emitted
+during gallery generation at the same time as warnings from example
+code execution. These can be filtered out with
+``warnings.filterwarnings`` (see :ref:`removing_warnings`). This is particularly
+important to do if you have tweaked warning handling in your documentation build
 to treat warnings as errors, e.g., with a line like
-``warnings.filterwarnings("error")``. Note that this
-differs from warnings affected by the ``- W`` / ``--fail-on-warning`` ``sphinx-build``
-flag, which converts warnings during documentation building into errors.
+``warnings.filterwarnings("error)`` which converts all warnings into errors.In
```
larsoner, line 2166:
```suggestion
``warnings.filterwarnings("error)`` which converts all warnings into errors. In
```

larsoner (approved):
Nice, thanks @lucyleeow !

PR #1369, opened 2024-08-06
FIX: Fix stability of stored compiled regex

larsoner:
I couldn't figure out why my MNE doc build was always rebuilding. It's because the `re.compile(<str>.join(<set>))` will not have a stable `repr` run-to-run because `set` does not have a stable order run-to-run. This PR fixes it. Not easy to add a test because it would require invoking `sphinx` twice, and even then would be a Heisenbug to detect. Hopefully https://github.com/sphinx-doc/sphinx/pull/12741 can move forward to make this more easy to detect.

### DIFF
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -159,9 +159,11 @@ def _update_gallery_conf_exclude_implicit_doc(gallery_conf):
 
     This is separate function for better testability.
     """
-    # prepare regex for exclusions from implicit documentation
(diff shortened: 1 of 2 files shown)


drammock (approved):

lucyleeow (approved):

lucyleeow:
Thanks @larsoner

lucyleeow:
@larsoner I think I'll release now, unless there are any more bug fixes coming? :grin:

larsoner:
Go for it!

PR #1383, opened 2024-10-02
Fix `indexst` variable does not exist when own index gallery is first

### LINKED ISSUE #1382: Variable 'indexst' is undefined when using custom index.rst
When using a custom `index.rst` for the gallery, the following exception is thrown:

```
  File "/home/runner/work/ansys-tools-local-product-launcher/ansys-tools-local-product-launcher/.tox/doc/lib/python3.12/site-packages/sphinx/cmd/build.py", line 332, in build_main
    app = Sphinx(args.sourcedir, args.confdir, args.outputdir,
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/ansys-tools-local-product-launcher/ansys-tools-local-product-launcher/.tox/doc/lib/python3.12/site-packages/sphinx/application.py", line 292, in __init__
    self._init_builder()
  File "/home/runner/work/ansys-tools-local-product-launcher/ansys-tools-local-product-launcher/.tox/doc/lib/python3.12/site-packages/sphinx/application.py", line 366, in _init_builder
    self.events.emit('builder-inited')
  File "/home/runner/work/ansys-tools-local-product-launcher/ansys-tools-local-product-launcher/.tox/doc/lib/python3.12/site-packages/sphinx/events.py", line 109, in emit
    raise ExtensionError(
sphinx.errors.ExtensionError: Handler <function generate_gallery_rst at 0x7f98e601d440> for event 'builder-inited' threw an exception (exception: cannot access local variable 'indexst' where it is not associated with a value)
```

From [this CI run](https://github.com/ansys/ansys-tools-local-product-launcher/actions/runs/11121042825/job/30899262996?pr=186), associated code: https://github.com/ansys/ansys-tools-local-product-launcher/tree/maint/remove_upper_limits [1]

I believe this was introduced in #1332, by removing this code block https://github.com/sphinx-gallery/sphinx-gallery/pull/1332/files#diff-016e2fd897b8720412d190b7936fb295e31dc953c73fcb912ca1362ec8ca2191L660-L663

```python
        else:
            # we are not going to use the index.rst.new that gets made here,
            # but go through the motions to run through all the subsections...
            indexst = "Never used!"
```

However, I'm not sure if the correct fix is to simply re-introduce the code block, or if there is a better solution.

[1] this branch updates `sphinx-gallery` from `0.15.0` to `0.17.1`

lucyleeow:
closes #1382

Bug introduced by refactor in https://github.com/sphinx-gallery/sphinx-gallery/pull/1332

When the first gallery in the example gallery loop, in `generate_gallery_rst`, is one with its own `index.rst` file, the `indexst` variable does not exist which is a problem because we pass it to `_finish_index_rst`. This variable is not used in `_finish_index_rst` when the gallery has its own index.rst file but `_finish_index_rst` does various other things so needs to be run always, even when the gallery has its own `index.rst`.

Adds test.

Will do a patch release once this is merged

### DIFF
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -534,7 +534,7 @@ def get_subsections(srcdir, examples_dir, gallery_conf, check_for_index=True):
         subfolders = [
             subfolder
             for subfolder in subfolders
-            # Returns `None` if `index.rst` or gallery header found
+            # Return is not `None` only when a gallery head file is found
             if _get_gallery_header(
                 os.path.join(examples_dir, subfolder), gallery_conf, raise_error=False
             )
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -534,11 +534,18 @@ def get_subsections(srcdir, examples_dir, gallery_conf, check_for_index=True):
         subfolders = [
             subfolder
             for subfolder in subfolders
-            # Returns `None` if no `index.rst` file and no gallery header found
+            # Returns `None` if `index.rst` or gallery header found
             if _get_gallery_header(
                 os.path.join(examples_dir, subfolder), gallery_conf, raise_error=False
             )
             is not None
+        ]
+    else:
+        # just make sure its a directory
+        subfolders = [
+            subfolder
+            for subfolder in subfolders
+            if os.path.isdir(os.path.join(examples_dir, subfolder))
         ]
 
     base_examples_dir_path = os.path.relpath(examples_dir, srcdir)
@@ -743,7 +750,7 @@ def generate_gallery_rst(app):
             is_subsection=False,
         )
 
-        # `this_context` is None when user provides own index.rst
+        # `this_content` is None when user provides own index.rst
         sg_root_index = this_content is not None
         costs += this_costs
         write_computation_times(gallery_conf, gallery_dir_abs_path, this_costs)
--- a/sphinx_gallery/gen_gallery.py
+++ b/sphinx_gallery/gen_gallery.py
@@ -541,11 +541,14 @@ def get_subsections(srcdir, examples_dir, gallery_conf, check_for_index=True):
             is not None
         ]
     else:
-        # just make sure its a directory
+        # just make sure its a directory, that is not `__pycache__`
         subfolders = [
             subfolder
             for subfolder in subfolders
-            if os.path.isdir(os.path.join(examples_dir, subfolder))
+            if (
+                subfolder != "__pycache__"
+                and os.path.isdir(os.path.join(examples_dir, subfolder))
+            )
         ]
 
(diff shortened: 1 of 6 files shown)


lucyleeow:
A `__pycache__` gets added to `sphinx_gallery/tests/testconfs/src` during the new `test_own_index_first` test.

When the folder has it's own `index.rst` file, we do not check that these folders have a `index.rst` or gallery header file when we collect subgallery folders. We do not need a index.rst/header file as we do not write the root gallery index or any nested sub gallery index files when the user provides their own root gallery `index.rst`.

@larsoner I've amended the code such that we do not collect any folders that called `__pycache__`. I think we could also avoid collecting any 'dunder' folder nested galleries as well. WDYT?

sphinx_gallery/gen_gallery.py
```diff
@@ -534,18 +534,23 @@ def get_subsections(srcdir, examples_dir, gallery_conf, check_for_index=True):
         subfolders = [
             subfolder
             for subfolder in subfolders
+            # Returns `None` if `index.rst` or gallery header found
```
larsoner, line 537: Right?
```suggestion
            # Returns `None` if `index.rst` or gallery header not found
```

larsoner (approved):
Just one question about a comment otherwise LGTM!

sphinx_gallery/gen_gallery.py
lucyleeow: Turns out we are both wrong, this whole thing is pretty confusing.

https://github.com/sphinx-gallery/sphinx-gallery/blob/7ddaf923dcb542931e10907ab6e426191b9b48c2/sphinx_gallery/gen_rst.py#L433-L458

If there is an `index.rst` -> return None
If there is a gallery header -> return file path
If there is no `index.rst`, no gallery header and raise is false -> None

So we want a subsection dir to contain a gallery header, so we check that the return `is not None`, which only happens when there is a gallery header. Let me amend the comment to be what happens when return is not None, as this is what we are checking for anyway.

lucyleeow: Actually the `check_for_index` parameter name is pretty confusing here, what about we change it to `check_for_header` or `check_for_gallery_header` but thats a bit verbose.

larsoner: Either is fine by me!

lucyleeow: changed to `check_for_header`

lucyleeow:
I think I should release now with just this PR, unless you think we should wait for #1384 or #1385  @larsoner ?

lucyleeow:
Happy to wait. Should it be a minor release then, or patch still?

larsoner:
Might as well make it a minor release

Issue #1371, opened 2024-08-08
ENH: Allow `sphx-glr-multi-img` images to span multiple lines


tsbinns:
Discussed with @larsoner.

When plotting multiple figures from the same cell, the figures are placed in the html as `sphx-glr-multi-img` images in a `sphx-glr-horizontal` list. Although each figure is stored in a new list element, if the figures are wide, they are shrunk down regardless of the specified figure sizes and placed on (what is essentially) the same line, which can make them illegible, e.g.: https://mne.tools/mne-connectivity/dev/auto_examples/decoding/cohy_decomposition_plotting.html#visualising-the-filters

It would be nice to have some sort of option that if multiple figures are shown from the same cell, the list elements containing each of the figures would be placed beneath one another. E.g.,

Before:
![image](https://github.com/user-attachments/assets/10b8eab1-ac66-434e-9d96-c2368f8bb92f)

After:
![image](https://github.com/user-attachments/assets/34371b4a-38e0-43c6-b807-8d553a3c2814)

Other ideas are also welcome. If people think this is a sensible feature I am happy to write a PR (though I wouldn't have the time until later in the year).

larsoner:
Agreed this would be nice. I know I've wanted something like this before and presumably @drammock has, too, when working on tutorials.

drammock:
> Agreed this would be nice. I know I've wanted something like this before and presumably @drammock has, too, when working on tutorials.

yep, just been too busy to request/implement it...

lucyleeow:
Naive question, is there a clever CSS/javascript way to detect image size/ratio and automatically decide on best layout?

tsbinns:
I have some time to look into this now; wanted to get some thoughts on the best way to proceed.

So currently, the max image width in `sg_gallery.css` for the `multi-img` class is 42%, which gives the 2 columns of images (compared to 100% max width for `single-img` class).

> Naive question, is there a clever CSS/javascript way to detect image size/ratio and automatically decide on best layout?

I'm also not sure about this. A simple approach would just be to allow `multi-img` max width to also be 100%, which would prevent things being shrunk.

However, I'm guessing it would be preferred if the default behaviour of 42% width remains and a new option for allowing wider `multi-img` images is added? In this case, do you think it's better if the option toggles between 42% and 100% max width, or should the user be allowed to set any max width for more flexibility? Or, does someone else have another suggestion they think might work better?

[...]

PR #1384, opened 2024-10-02
[ENH] Add option to render multiple images from same cell as single-img

tsbinns:
Discussed in #1371.

3 ideas were proposed:
1. Global no-width-limit: change the CSS `max-width` for `multi-img`
2. A single example no-width-limit: change set the `# sphinx_gallery_multi_image = single` _somewhere in the file_
3. A single block with an example no-width-limit: set the `# sphinx_gallery_multi_image_block = single` _within that block_

This PR so far addresses point 2 (adding a config setting for the whole example). Will have a go at point 3 as well.


sphinx_gallery/scrapers.py
```diff
@@ -374,7 +383,7 @@ def save_figures(block, block_vars, gallery_conf):
     all_rst = ""
     prev_count = len(image_path_iterator)
     for scraper in _get_callables(gallery_conf, "image_scrapers"):
-        rst = scraper(block, block_vars, gallery_conf)
+        rst = scraper(block, block_vars, gallery_conf, file_conf)
```
tsbinns, line 386: A question I have is how passing `file_conf` to custom scrapers should be handled? I'm guessing we cannot expect everyone using a custom scraper to add a new parameter. Is it enough to check if the scraper function has the corresponding param name?

larsoner: IIRC `block_vars` is a dict so can we just add it to `block_vars` like `file_conf["block_vars"]`? It's a bit of a misnomer to have file-level info in there but I think it's okay.

An alternative would be to use magic to inspect arguments but I don't think it's as clean

tsbinns: Sure, can add this info to `block_vars`.

Just thinking ahead towards the other suggestion of having a block-specific setting, this info would still be in `block` at this point. Does it make sense to read this info. *before* it gets passed to the scraper and also store it in `block_vars`?

sphinx_gallery/gen_rst.py
```diff
@@ -1000,12 +1000,26 @@ def execute_code_block(
...
+    # Check for `sphinx_gallery_multi_image_block` in this block
+    multi_image_match = re.search(
+        r"^[\ \t]*#\s*sphinx_gallery_multi_image_block[\ \t]*\n?",
+        block.content,
+        re.MULTILINE,
+    )
+    if multi_image_match is not None:
+        script_vars["multi_image"] = py_source_parser.extract_file_config(
+            block.content
+        )["multi_image_block"]
```
tsbinns, line 1019: My attempt at reading the block content and passing the relevant setting to (what becomes) `block_vars`. Looking at this, I feel like there's a more streamlined solution, e.g.:
```
script_vars["multi_image"] = py_source_parser.extract_file_config(block.content).get("multi_image_block")
```

sphinx_gallery/scrapers.py
```diff
@@ -204,15 +204,25 @@ def matplotlib_scraper(block, block_vars, gallery_conf, **kwargs):
             )
         )
 
+    # Determine whether single-img should be converted to multi-img
+    convert_to_multi_image = True  # default is to convert
+    if block_vars.get("multi_image") is not None:  # block setting takes precedence
+        convert_to_multi_image = block_vars["multi_image"] != "single"
+    elif block_vars["file_conf"].get("multi_image") is not None:  # then file setting
+        convert_to_multi_image = block_vars["file_conf"]["multi_image"] != "single"
```
tsbinns, line 212: This in the scraper prioritises checking the block-specific setting, then falls back to the file-wide setting.

tsbinns:
Will address the failing tests and start to think about new unit tests for these features tomorrow.

sphinx_gallery/scrapers.py
```diff
@@ -204,15 +204,25 @@ def matplotlib_scraper(block, block_vars, gallery_conf, **kwargs):
...
     plt.close("all")
     rst = ""
     if len(image_rsts) == 1:
         rst = image_rsts[0]
     elif len(image_rsts) > 1:
-        image_rsts = [
-            re.sub(r":class: sphx-glr-single-img", ":class: sphx-glr-multi-img", image)
-            for image in image_rsts
-        ]
+        if convert_to_multi_image:
```
lucyleeow, line 219: I may have mis-read, but should there be an `else` for when `convert_to_multi_image` is True?

tsbinns: So the existing behaviour for when there are multiple images from a single cell is to switch their CSS class from `single-img` (100% max width) to `multi-img` (42% max width). Then afterwards the images are added to a list.

With these changes, `convert_to_multi_image` is default `True` (so the class gets switched from `single-img` to `multi-img`), but if `sphinx_gallery_multi_image` or `sphinx_gallery_multi_image_block` is `"single"`, then `convert_to_multi_image` becomes `False` and the images remain as a `single-img` class. Regardless of whether the image class changes, they still get added to a list afterwards.

Based on the examples in tinybuild this gives the expected behaviour, e.g.:
- File-wide:
image

- Block-specific:
image

examples/plot_1_exp.py
```diff
@@ -2,11 +2,12 @@
...
 =================================
 
-This example demonstrates how to import a local module and how images are
-stacked when two plots are created in one code block. The variable ``N`` from
-the example 'Local module' (file ``local_module.py``) is imported in the code
-below. Further, note that when there is only one code block in an example, the
-output appears before the code block.
+This example demonstrates how to import a local module and how images are stacked when
+two plots are created in one code block (see the :doc:`plot_9_multi_image_separate`
+example for information on controlling this behaviour). The variable ``N`` from the
```
tsbinns, line 7: Noticed that the plot exp example discussed images being stacked side-by-side, so added a link to the new example mentioning this behaviour can be changed.

examples/plot_9_multi_image_separate.py
```diff
@@ -0,0 +1,52 @@
+"""
+Force plots to be displayed on separate lines
+=============================================
+This example demonstrates how the visualisation of multiple plots produced from a single
+code block can be controlled. The default behaviour is to stack plots side-by-side,
+however this can be overridden to display each plot created by the code block on a
+separate line, preserving their size.
```
tsbinns, line 7: Added a new example demonstrating the `sphinx_gallery_multi_image` and `..._block` options.

sphinx_gallery/gen_rst.py
```diff
@@ -1000,12 +1000,21 @@ def execute_code_block(
...
         re.MULTILINE,
     )
-    need_save_figures = match is None
+    need_save_figures = defer_figs_match is None
+
+    # Add `sphinx_gallery_multi_image_block` setting to block variables
+    # (extract config rather than just regex search since the option's value is needed)
+    script_vars["multi_image"] = py_source_parser.extract_file_config(
+        block.content
+    ).get("multi_image_block")
```
tsbinns, line 1014: Would be interested if others think there's a better way to handle this.

Of course just a regex search is possible, but unlike the `defer_figures` flag we need to extract the value of variable, which the `extract_file_config()` func can handle.

tsbinns:
Have added entries for the new config options in the docs.

Also added to the unit tests, however they start from the point where `save_figures()` is called, which overlooks the code where `sphinx_gallery_multi_image_block` and `file_conf` are added to `block_vars` in `execute_code_block()`. Is this okay?

sphinx_gallery/scrapers.py
lucyleeow: Thanks mis-read the code

larsoner:
> Also added to the unit tests, however they start from the point where save_figures() is called, which overlooks the code where sphinx_gallery_multi_image_block and file_conf are added to block_vars in execute_code_block(). Is this okay?

It looks like you added to `tinybuild` which should be able to test the full functionality, assuming you write the `assert` statements correctly I think

tsbinns:
> It looks like you added to tinybuild which should be able to test the full functionality, assuming you write the assert statements correctly I think

Okay, I added to `test_full` checks that the file-wide and block-specific options work using the html output from `tinybuild`.

PR #1385, opened 2024-10-04
Allow to disable writing computation times

bmwiedemann:
Allow to disable writing computation times to allow for reproducible builds.

Default to not write durations. if SOURCE_DATE_EPOCH indicates that a reproducible build is wanted.

Helps with https://github.com/pygraphviz/pygraphviz/issues/541

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -65,6 +65,7 @@ file, inside a ``sphinx_gallery_conf`` dictionary.
 **Compute costs**
 
 - ``min_reported_time`` (:ref:`min_reported_time`)
+- ``write_computation_times`` (:ref:`write_computation_times`)
 - ``show_memory`` (:ref:`show_memory`)
 - ``junit`` (:ref:`junit_xml`)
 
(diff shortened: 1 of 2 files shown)


larsoner:
Makes sense to me, can you add it to the [configuration doc](https://github.com/sphinx-gallery/sphinx-gallery/blob/master/doc/configuration.rst) somewhere?

bmwiedemann:
I added some docs. Feel free to adjust it as you like.

doc/configuration.rst
```diff
@@ -2293,6 +2294,15 @@ The ``min_reported_time`` parameter can be set to a number of seconds. The
 duration of scripts that ran faster than that amount will not be logged nor
 embedded in the html output.
 
+.. _write_computation_times:
+
+Write computation times
+=======================
+
+Set to ``false`` if you want to omit computation times from the output.
```
drammock, line 2302:
```suggestion
Set to ``False`` if you want to omit computation times from the output.
```

doc/configuration.rst
```diff
@@ -2293,6 +2294,15 @@ The ``min_reported_time`` parameter can be set to a number of seconds. The
...
+This helps with reproducible builds.
+If the ``SOURCE_DATE_EPOCH`` environment variable is set, it defaults to false.
+True otherwise.
```
drammock, line 2305:
```suggestion
Default is ``True`` unless the ``SOURCE_DATE_EPOCH`` environment variable is set.
```

doc/configuration.rst
lucyleeow, line 2302:
```suggestion
Set to ``False`` if you want to omit computation times from all gallery outputs.
```

lucyleeow (approved):
LGTM

PR #1391, opened 2024-10-15
Fix custom sort

### LINKED ISSUE #1389: custom within-subsection sorting doesn't work as documented
The [example in the docs](https://sphinx-gallery.github.io/stable/configuration.html#custom-sort-keys) of using `FunctionSortKey`:

```py
"minigallery_sort_order": FunctionSortKey(lambda filename: filename[:10])
```

is guaranteed not to work, because non-string values are tested with `inspect.isclass()` wihtin the `_get_class` helper --- but `isclass()` returns `False` for class *instances* (and the recommended usage always generates a class instance).

If instead I follow [the other recommendation in the docs](https://sphinx-gallery.github.io/dev/configuration.html#importing-callables) to use a FQN of an importable function, I still get an error:

```
In [1]: from sphinx_gallery.gen_rst import _get_class
In [2]: sphinx_gallery_conf = {
   ...:     "doc_module": "mne_bids",
   ...:     "reference_url": {
   ...:         "mne_bids": None,
   ...:     },
   ...:     "backreferences_dir": "generated",
   ...:     "examples_dirs": "../examples",
   ...:     "within_subsection_order": "mne_bids.utils.example_sorter",  # <---- HERE
   ...:     "gallery_dirs": "auto_examples",
   ...:     "filename_pattern": "^((?!sgskip).)*$",
   ...: }
   ...: 
In [3]: _get_class(sphinx_gallery_conf, "within_subsection_order")
---------------------------------------------------------------------------
ConfigError                               Traceback (most recent call last)
Cell In[3], line 1
----> 1 _get_class(sphinx_gallery_conf, "within_subsection_order")

File /opt/sphinx-gallery/sphinx_gallery/gen_rst.py:1645, in _get_class(gallery_conf, key)
   1640         raise ConfigError(
   1641             f"{what} must be a fully qualified name string, could not import "
   1642             f"{attr} from {mod}"
   1643         )
   1644 if not inspect.isclass(val):
-> 1645     raise ConfigError(
   1646         f"{what} must be 1) a fully qualified name (string) that resolves "
   1647         f"to a class or 2) or a class itself, got {val.__class__.__name__} "
   1648         f"({repr(val)})"
   1649     )
   1650 return val

ConfigError: sphinx_gallery_conf['within_subsection_order'] must be 1) a fully qualified name (string) that resolves to a class or 2) or a class itself, got function (<function example_sorter at 0x7d99b40f2160>)
```

I suspect that this error was introduced in #1289 (but haven't dug deeply, maybe it's older)

I'll submit a PR shortly

drammock:
closes #1389 

I *think* no docs update is needed because now things work the way the docs describe them, but feel free to request changes / push commits if you think I'm wrong about that.


larsoner:
@lucyleeow I did not follow all discussion so I'll let you review, feel free to merge if you're happy. I can look deeper if it would help to have another set of eyes or you're busy though!

sphinx_gallery/gen_rst.py
```diff
@@ -1617,16 +1617,18 @@ def save_rst_example(
 
 def _get_class(gallery_conf, key):
     """Get a class for the given conf key."""
+    from .sorting import FunctionSortKey
```
drammock, line 1620: nesting required to avoid circular import

sphinx_gallery/gen_rst.py
```diff
@@ -1641,6 +1643,8 @@ def _get_class(gallery_conf, key):
                 f"{what} must be a fully qualified name string, could not import "
                 f"{attr} from {mod}"
             )
+        if not is_builtin_alias:
+            val = FunctionSortKey(val)
```
drammock, line 1647: just to spell out the logic of this function:

- if it's a string alias of a builtin SortKey class, convert to FQN string
- then, if it's a string, check for ".", split, and try to import
- if that worked, wrap the imported func in `FunctionSortKey` (unless it's already a built-in SortKey class)

sphinx_gallery/sorting.py
```diff
@@ -162,7 +162,7 @@ def __call__(self, filename):
         return title
 
 
-class FunctionSortKey:
+def FunctionSortKey(func, r=None):
```
drammock, line 165: opted for CamelCase on the function name, to avoid changing the API (though if you hate this, it's probably no big deal to change it to snake_case, since it wasn't really working before anyway AFAICT).

drammock (commented):
left a few comments to hopefully speed up reviewing

drammock:
one other important note: now that `FunctionSortKey` returns a type (rather than an instance) it will fail sphinx's `is_serializable` checks. So might be worth highlighting that in the docs / *strongly* encouraging the use of FQN strings as the preferred way to do custom sorting.

lucyleeow:
Half way through review I realised that I already figured out that `subsection_order` and `minigallery_sort_order` worked differently to `within_subsection_order`, when reviewing #1226. Thus I specifically removed `within_subsection_order` from the docs (see: https://sphinx-gallery.github.io/stable/configuration.html#custom-sort-keys and my last note: https://github.com/sphinx-gallery/sphinx-gallery/pull/1226#issuecomment-1918365134

(also which part of the docs were you reading? They can definitely be improved, I was not sure where to put the section on custom sort keys as you can use them for several 'unrelated' configs :weary: )

I do like that this now supports `within_subsection_order` BUT I think it would no longer work for `subsection_order` and `minigallery_sort_order` as these both use `_get_callables`, which can't be a class.

https://github.com/sphinx-gallery/sphinx-gallery/blob/9947c26a7457f7ef35b41cee7d6fa89e1f36c67d/sphinx_gallery/gen_rst.py#L1698-L1701

(obviously we need to improve our testing)

Ideally, I think I would love to get rid off `_get_class`, but this would involve:

* internally make the convenience classes (e.g., `ExampleTitleSortKey`) functions instead of classes, but keep the names as camelcase (since we ask the user to pass as string anyway). I think we only had classes for the stable `__repr__` but as we are passing a string now, this is not necessary. I think this would also aid the `within_subsection_order` and `FunctionSortKey` problem, as we would no longer be doing:

https://github.com/sphinx-gallery/sphinx-gallery/blob/9947c26a7457f7ef35b41cee7d6fa89e1f36c67d/sphinx_gallery/gen_rst.py#L569

* maybe getting rid of `FunctionSortKey` altogether (a quick [search on github](https://github.com/search?q=FunctionSortKey&type=code&p=1) seems to show that it is not really used), it was introduced not long before the serialization stuff. 

Sorry @larsoner, have to bring you in here, you'll be able to come up with a better solution than my tired brain.
Also apologies if I have made a mistake somewhere.

drammock:
> Thus I specifically removed `within_subsection_order` from the docs (see: https://sphinx-gallery.github.io/stable/configuration.html#custom-sort-keys and my last note: [#1226 (comment)](https://github.com/sphinx-gallery/sphinx-gallery/pull/1226#issuecomment-1918365134)

Ah yes, I somehow missed that `within_subsection_order` wasn't mentioned in the "Custom Sort Keys" section of the config docs!  So my earlier statement of "the docs say you can X but it doesn't work" is not strictly correct.

> maybe getting rid of FunctionSortKey altogether (a quick [search on github](https://github.com/search?q=FunctionSortKey&type=code&p=1) seems to show that it is not really used)

-0.5 from me; this PR came about because I want to use it (on subsection orders) and couldn't get it to work. I'd be content though if passing an FQN string of a sorting function worked on subsection orders.  Note that at present (in this PR) `FunctionSortKey` is getting used internally to convert the imported FQN into a callable class (to parallel the other SortKey classes), so even if removed from the public API, it might still be in the codebase... unless:

> I would love to get rid off `_get_class`

...then maybe `FunctionSortKey` could go away entirely. I can look into that.

> I do like that this now supports `within_subsection_order` BUT I think it would no longer work for `subsection_order` and `minigallery_sort_order` as these both use `_get_callables`, which can't be a class.

I'll add a test of whether FunctionSortKey still works on `subsection_order` and `minigallery_sort_order`, and report back.

lucyleeow:
I completely agree that ideally I would like all 3 configs to be able to take a custom sort key and for them to work similarly.

I think a function and string FQN should work on all 3 but we've just got to get the class thing working.

I thought of getting rid of `_get_class` because maybe we could get `within_subsection_order` to work with `_get_callables`, by changing the convenience classes (e.g., `ExampleTitleSortKey`) to be functions as they are now given as strong. It would also be nice to reduce some complexity in the code. Hopefully I haven't missed something important though.

We can keep `FunctionSortKey`, I think we can make it work. But I think we'll  want to recommend using functions.

Thanks for working on this!

lucyleeow:
Thinking about it more, maybe the convenience classes could be a 'callable' (i.e. class instance) instead of a class. This would allow us to use `_get_callables` and keep these CamelCased classes as classes, and not have to change them to functions? Instead of init-ing with the `src_dir` we can just pass it the full path?

The only other thing we would want to make sure works is any existing user's own classes used for `within_section_order` (as we know the `FunctionSortKey` does not work). They could have done this by subclassing `_SortKey` (a [search on github](https://github.com/search?q=%22_SortKey%22&type=code) for this class was not helpful as I don't think you can do a literal string search over all of github) or making their own class that works the same way. I don't know how common this is but I don't think we could get rid of `_get_class` without breaking existing users own `within_subsection_order` classes.

I guess the other option is to loosen, so a class instance or a class are both accepted? Not sure what the best solution is here. I would like to go for getting rid of `_get_class` but I don't want to break existing users' code/force them to change their code.

Apologies for patchiness, I am on holiday but I will try to keep on top of this.

sphinx_gallery/tests/test_gen_gallery.py
```diff
@@ -339,6 +345,45 @@ def test_example_sorting_title(sphinx_app_wrapper):
...
+    "plot_1.py",
+]
+
+def custom_sorter(filename):
+    return ORDER.index(filename)
+
+sphinx_gallery_conf = {
+    "examples_dirs": "src",
+    "gallery_dirs": "ex",
+    "within_subsection_order": FunctionSortKey(custom_sorter),
```
drammock, line 364: I can add here 
```py
    "minigallery_sort_order": FunctionSortKey(custom_sorter),
    "subsection_order": FunctionSortKey(custom_sorter),
```
and the test will still pass, but I think that's not especially informative because (IIUC) the site being built by for this test doesn't have subsections / minigalleries? Could use some guidance on how/where to test that `FunctionSortKey(my_func)` and FQN strings work correctly for minigalleries and subsection ordering... maybe that's already tested here?:

https://github.com/sphinx-gallery/sphinx-gallery/blob/9947c26a7457f7ef35b41cee7d6fa89e1f36c67d/sphinx_gallery/tests/test_full.py#L1350

...but I find that test/parametrization inscrutable, so I'd appreciate some guidance on whether those cases are already covered, and if not, how/where I should add/tweak a test to cover them.

drammock (commented):
OK `_get_class` is now gone, replaced by a tweaked version of `_get_callables`. It was quite a game of whack-a-mole running tests locally (mostly because I didn't realize at first all the different ways `_get_callables` was being used).

> Thinking about it more, maybe the convenience classes could be a 'callable' (i.e. class instance) instead of a class. This would allow us to use _get_callables and keep these CamelCased classes as classes, and not have to change them to functions? Instead of init-ing with the src_dir we can just pass it the full path?

What I ended up doing was detecting whether the "callable" was one of our built-in sorter classes, or the class returned by `FunctionSortKey`, and if so, instantiating them within `_get_callables`. I *think* this means that any user-created callable class instances should still work (and IIUC we don't expect users to pass in an actual *class* right? just instances?)

larsoner:
> and IIUC we don't expect users to pass in an actual class right? just instances?

It looks like at least mpl does:

https://github.com/matplotlib/matplotlib/blob/9489b938556b16633591c2e829462b1aa4e4a46e/doc/conf.py#L305

https://github.com/matplotlib/matplotlib/blob/9489b938556b16633591c2e829462b1aa4e4a46e/doc/conf.py#L187

https://github.com/matplotlib/matplotlib/blob/9489b938556b16633591c2e829462b1aa4e4a46e/doc/sphinxext/gallery_order.py#L125

Does that change things?

drammock:
> > and IIUC we don't expect users to pass in an actual class right? just instances?
> 
> It looks like at least mpl does [...] Does that change things?

ah, yes it would. thanks for the tip, I'll rethink this.

sphinx_gallery/gen_rst.py
```diff
@@ -1695,6 +1682,10 @@ def _get_callables(gallery_conf, key):
                     raise ConfigError(f"Unknown string option for {readable}: {what}")
                 what = _reset_dict[what]
             which[wi] = what
+        # make sure built-in sorter classes get instantiated (so they become callable)
+        if is_custom_sorter or what in BUILTIN_SORTERS:
```
drammock, line 1686: if we change this conditional to

```suggestion
        # make sure classes get instantiated (so they become callable)
        if key == "within_subsection_order" and inspect.isclass(what):
```

then MPL's custom sorter class should work too, and the test suite still passes locally.

sphinx_gallery/sorting.py
lucyleeow: If we allow `within_subsection_order` to take a class instance, then I think this can be reverted and would work on `within_subsection_order` ?

larsoner:
> then MPL's custom sorter class should work too, and the test suite still passes locally.

Do you see an easy enough way to modify `tinybuild` to have mpl's use case? That way we can test the "class" case there but hopefully also test the "class instance" case in more targeted tests like I think you currently have in `test_gen_gallery.py`

sphinx_gallery/sorting.py
drammock: yep! done in 716b180a8ab0228b4f807ef2c564b82c3bd9faaf

drammock:
> > then MPL's custom sorter class should work too, and the test suite still passes locally.
> 
> Do you see an easy enough way to modify `tinybuild` to have mpl's use case?

Done in [2316166](https://github.com/sphinx-gallery/sphinx-gallery/pull/1391/commits/231616654f3bbf8c6ae18039a227524b40620c79)
We lose the fact that tinybuild is testing one of the built-in sorter classes though. Probably not a huge deal.

lucyleeow:
Sorry I got to this late. Would it work to alter `sphinx_gallery/tests/testconfs/src` to add a subgallery? Is it easy to 'fix' the tests that use `sphinx_wrapper_app` such that they pass even with a new subgallery?

I only suggest as it would be nice to have more unit tests, as tinygallery tests do get quite convoluted.

drammock:
> Sorry I got to this late. Would it work to alter `sphinx_gallery/tests/testconfs/src` to add a subgallery? Is it easy to 'fix' the tests that use `sphinx_wrapper_app` such that they pass even with a new subgallery?
> 
> I only suggest as it would be nice to have more unit tests, as tinygallery tests do get quite convoluted.

I think I've achieved what you ask in [d25c7d5](https://github.com/sphinx-gallery/sphinx-gallery/pull/1391/commits/d25c7d510900fe0ffe2a7facf6e573b8dd23cd2e) but please double-check it's what you expected, I'm still not sure I've wrapped my head around your test suite.

lucyleeow:
Yes thats exactly what I was thinking. Just seems like we'd need to update a few more tests, to account for the new subgallery but otherwise seems reasonable. Thank you.

drammock:
ok sorting tests are fixed @lucyleeow (sorry I missed those before, thought I ran full pytest before pushing but I must not have). All green now.

sphinx_gallery/tests/test_gen_gallery.py
lucyleeow: So we could do minigalleries here.

I added a `add_rst` pytest mark here, which will add a `index.rst` file in src with the contents of `file` arg. (This happens in the `sphinx_wrapper_app` in conftest). See this test for example:

https://github.com/sphinx-gallery/sphinx-gallery/blob/13f7e06c7e70cd4c3fa04ab7c317c4faf4f2ce26/sphinx_gallery/tests/test_gen_gallery.py#L647-L662

AFAICT we only test FQN string for `subsection_order` in tinybuild.
It may be a big ask, but what do you think about adding a second subsection to the `sphinx_gallery/tests/testconfs/src` gallery...?

lucyleeow (commented):
So sorry about the delay.

Some nits, mostly around testing (and its mostly our fault, e.g., I didn't realise we hardly test `subsection_order`), otherwise this is looking good.

I think some doc updates should go with this, but I can add them later in a separate PR as there are a few things I want to amend and move around.

lucyleeow:
Let me know when you are done and I'll take another look. Thanks again :pray:

drammock:
> Let me know when you are done and I'll take another look. Thanks again 🙏

OK I think it's ready!  I didn't touch the docs because you said you wanted to tackle that in a separate PR.  I did a somewhat bold refactor of the example sorting tests --- there's now just one test, which is parametrized over the different ways of sorting.  It's a bit wonky looking because part of the parametrization is itself in a `pytest.mark` (the `add_conf` mark) but the `pytest.param` IDs are really explicit so hopefully it's not too hard to figure out what's going on there.

On the plus side: one of the added tests revealed that it wasn't actually working to set subsection order with a plain list, so I've fixed that too (I checked by removing the `ExplicitOrder()` wrapper in MNE-Python's docs --- the doc build fails on SG `master` and succeeds on this PR).

sphinx_gallery/tests/tinybuild/doc/conf.py
```diff
@@ -73,7 +85,7 @@
     ],
     "backreferences_dir": "gen_modules/backreferences",
     "subsection_order": f"{util_root}.noop_key",
-    "within_subsection_order": "FileNameSortKey",
+    "within_subsection_order": CustomSortKey,
```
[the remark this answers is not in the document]
drammock: somehow I missed this comment before, but I think it's now moot?  The string-aliases for builtin sorters are now comprehensively tested in `test_example_sorting` (as are FQN strings of custom funcs, and actual callable functions) so AFAICT it's actually preferable to keep this as-is (since in the parametrized test we only test callable functions, not callable classes).

sphinx_gallery/tests/test_gen_gallery.py
```diff
@@ -268,75 +268,187 @@ def test_spaces_in_files_warn(sphinx_app_wrapper):
...
 
-    Used to test that these keys (in index.rst) appear in a specific order.
+    Used to test that these keys (in index.rst) appear in a specific order. The `.py`
+    files include a source-of-truth line indicating the position in the sort order that
+    the thumbnail for that file should occur, for the various built-in sorters.
+
+    The regex below extracts info from that source-of-truth line to check that the
+    thumbnail ordering was correct --- unless `expected_order` is provided, in which
+    case an alternative regex extracts the digit-part of the filename from the crossref,
+    and compares the order of crossrefs to `expected_order`.
```
lucyleeow, line 281: Thanks for adding this. I actually did not realise the check was this complicated!

sphinx_gallery/tests/tinybuild/doc/conf.py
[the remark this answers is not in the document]
lucyleeow: Yes totally agree, this is good as is.

lucyleeow (approved):
LGTM, thanks for all your work on this. Especially for fixing the subsection order plain list!

I see a few minigallery directives in the new `plot` example files. Did you have plans for adding minigallery sort tests? I am very okay with you not doing this though, as you've already done so much work on this PR already. 

Otherwise, just a question.

I'll merge in a few days in case @larsoner wants to take a look

sphinx_gallery/tests/test_gen_gallery.py
```diff
@@ -268,75 +268,187 @@ def test_spaces_in_files_warn(sphinx_app_wrapper):
...
-
-
-@pytest.mark.add_conf(
-    content="""
-sphinx_gallery_conf = {
-    'examples_dirs': 'src',
-    'gallery_dirs': 'ex',
-    'within_subsection_order': "FileSizeSortKey",
-}"""
+_subsection_explicit_order_list = pytest.mark.add_conf.with_args(
```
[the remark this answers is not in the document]
drammock: nope, I just misunderstood the pytest docs. After re-reading, it seems you only need `with_args` if there's a single positional arg to the custom marker *and* that single positional arg is itself a callable. I'll push a commit removing the `with_args` part.

sphinx_gallery/gen_gallery.py
```diff
@@ -527,9 +528,10 @@ def get_subsections(srcdir, examples_dir, gallery_conf, check_for_header=True):
     if gallery_conf["subsection_order"] is None:
         sortkey = None
     else:
+        subsec_order = gallery_conf.get("subsection_order")
+        if isinstance(subsec_order, list):
+            gallery_conf["subsection_order"] = ExplicitOrder(subsec_order)
```
drammock, line 533: this is the change that makes bare lists work for `subsection_order` (must be wrapped in `ExplicitOrder` before the call to `_get_callables`)

lucyleeow:
Thank you @drammock !

PR #1398, opened 2024-10-29
Add block-level `sphinx_gallery_capture_repr_block` setting

### LINKED ISSUE #1397: [QUESTION] Is it possible to ignore the output of a specific cell?
## Question

I tried looking in the docs for this but I could not find any options on how to hide the output of a specific notebook cells. Did I miss something?

## Motivation

The reason I'm asking is because I'm [currently working on](https://github.com/tpvasconcelos/ridgeplot/pull/267/files#diff-ca4940391ff4423ef59f4c827f3ccc447a492b08698c9860d3efb4c134ca4f17) an examples gallery with Plotly and I sometimes want to have cells with something like:

```python
fig.update_layout(
    foo="Bar",
    bar="Foo",
)
```

but since `update_layout` return a plotly figure, this is automatically captured by Sphinx-Gallery and include in the output.

A workaround would be to assign it back to a variable:

```python
fig = fig.update_layout(
    foo="Bar",
    bar="Foo",
)
```

but I want to be consistent and follow the same standards used throughout the Plotly official docs (they always threat the `Figure` object as mutable and never reassign back to a `fig` variable). See [this](https://plotly.com/python/creating-and-updating-figures/) for instance.

tpvasconcelos:
closes #1397

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -47,6 +47,13 @@ file, inside a ``sphinx_gallery_conf`` dictionary.
 - ``reset_modules`` and ``reset_modules_order`` (:ref:`reset_modules`)
 - ``parallel`` (:ref:`parallel`)
 
+.. admonition:: Diverging from Jupyter
+    :class: danger
+
+    Sphinx-gallery attempts to render examples to HTML in a manner largely consistent with what a user will experience when they download the corresponding ``.ipynb`` notebook file and run it locally.
+    Some options, such as ``'capture_repr': (),``, will make these behaviors less consistent.
+    Consider using these options sparingly as it could lead to confusion or sub-optimal experiences for users!
+
 **Cross-referencing**
 
 - ``reference_url``, ``prefer_full_module`` (:ref:`link_to_documentation`)
@@ -2439,8 +2446,7 @@ captured. For several examples, see :ref:`capture_repr_examples`.
 
 To capture only data directed to standard output, configure ``'capture_repr'``
 to be an empty tuple: ``'capture_repr': ()``. This will imitate the behaviour
-of Sphinx-Gallery prior to v0.5.0. Note however that this behaviour is not
-consistent with how Jupyter notebooks capture output by default!
+of Sphinx-Gallery prior to v0.5.0.
 
 From another perspective, take for example the following code block::
 
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -107,6 +107,7 @@ Some options can also be set or overridden on a file-by-file basis:
 
 Some options can be set on a per-code-block basis in a file:
 
+- ``# sphinx_gallery_capture_repr_block`` (:ref:`capture_repr`)
 - ``# sphinx_gallery_defer_figures`` (:ref:`defer_figures`)
 - ``# sphinx_gallery_multi_image_block`` (:ref:`multi_image`)
 
@@ -2411,11 +2412,15 @@ are:
   present in some objects, for example, pandas dataframes.
 
 Output capture can be controlled globally by the ``capture_repr`` configuration
-setting or file-by-file by adding a comment to the example file, which overrides
+setting, file-by-file by adding a comment to the example file, which overrides
 any global setting::
 
     # sphinx_gallery_capture_repr = ()
 
+, or block-by-block by adding a comment to the code block, which overrides any
+global or file setting::
+
+    # sphinx_gallery_capture_repr_block = ()
 
 The default setting is::
 
@@ -2434,7 +2439,8 @@ captured. For several examples, see :ref:`capture_repr_examples`.
 
 To capture only data directed to standard output, configure ``'capture_repr'``
 to be an empty tuple: ``'capture_repr': ()``. This will imitate the behaviour
-of Sphinx-Gallery prior to v0.5.0.
+of Sphinx-Gallery prior to v0.5.0. Note however that this behaviour is not
+consistent with how Jupyter notebooks capture output by default!
 
 From another perspective, take for example the following code block::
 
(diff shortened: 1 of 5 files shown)


tpvasconcelos:
Hey again @larsoner! It was pretty straightforward to implement what we discussed in #1397 👍 

I opened this PR for you to take a look at and get some feedback first. Happy to work on documenting this in https://sphinx-gallery.github.io/stable/configuration.html once you're happy with the implementation

larsoner (commented):
Awesome, does seem to look correct!

Can you update https://github.com/sphinx-gallery/sphinx-gallery/blob/master/doc/configuration.rst ?

Bonus points if you want to add a `.. warning::` or `.. admonition::` that mentions that options like this one make SG behave differently from how users will experience things in the notebook, and that they are generally discouraged because they makes for an inconsistent experience. There are a few options like this but if you can at least list a few of them, that would already be a good start!

tpvasconcelos:
@larsoner I did't find a clean nice way of introducing the `.. warning::` or `.. admonition::` admonitions. To my eyes the docs look clear to me as they are now. If you think something could be improved maybe it would be simpler for you to push some changes to my branch (I believe you already have access to it).

Let me know what you think

tpvasconcelos:
I ended up adding a note admonition but to connect the suggestions for matplotlib figures to Plotly figures:

image

doc/configuration.rst
```diff
@@ -2434,7 +2439,8 @@ captured. For several examples, see :ref:`capture_repr_examples`.
 
 To capture only data directed to standard output, configure ``'capture_repr'``
 to be an empty tuple: ``'capture_repr': ()``. This will imitate the behaviour
-of Sphinx-Gallery prior to v0.5.0.
+of Sphinx-Gallery prior to v0.5.0. Note however that this behaviour is not
+consistent with how Jupyter notebooks capture output by default!
```
lucyleeow, line 2443: I think it is not necessarily setting `capture_repr` to empty tuple in general that we want to warn about, I think it's more altering it for specific files or code blocks when you've set it to something else globally.

tpvasconcelos: Hey @lucyleeow thanks for the review. I think you and @larsoner obviously have much more experience with this framework and with how users interact with it so please feel free to push the changes you see fit 👍🏽

My 2 cents is that both warnings seem to be redundant. The users will only use these settings knowing what they do so it should be obvious that setting a different value at the block-level will result in different or inconsistent behaviour to other blocks that inherit the global setting. i.e., I don't really see what new information the user is learning but warning them that changing the setting at the block-level will result in a different behaviour when compared to other blocks.. that's the whole point of block-level settings 😄

But again, I don't have strong feeling about this, so feel free to push or let me know what you'd like to see here instead.

Thanks in advance!!

doc/configuration.rst
```diff
@@ -47,6 +47,13 @@ file, inside a ``sphinx_gallery_conf`` dictionary.
 - ``reset_modules`` and ``reset_modules_order`` (:ref:`reset_modules`)
 - ``parallel`` (:ref:`parallel`)
 
+.. admonition:: Diverging from Jupyter
+    :class: danger
+
+    Sphinx-gallery attempts to render examples to HTML in a manner largely consistent with what a user will experience when they download the corresponding ``.ipynb`` notebook file and run it locally.
+    Some options, such as ``'capture_repr': (),``, will make these behaviors less consistent.
+    Consider using these options sparingly as it could lead to confusion or sub-optimal experiences for users!
```
larsoner, line 55: Pushed a commit with this text to cover what I had in mind! @lucyleeow feel free to look and make sure you think it's reasonable

sphinx_gallery/gen_gallery.py
```diff
@@ -1138,7 +1138,7 @@ def _make_graph(fname, entries, gallery_conf):
                 dg.node(node_to, **node_kwargs)
                 dg.edge(node_from, node_to, color=API_COLORS["edge"])
         # add modules with all API entries
-        for module in gallery_conf["_sg_api_entries"]["module"]:
+        for module in gallery_conf["_sg_api_entries"].get("module", []):
```
larsoner, line 1141: Every once in a while when rebuilding MNE-Python doc I'd get an error on this line. No idea why and couldn't figure out how to reproduce it, but it should be safe enough to use `.get` instead of direct dict access here.

larsoner (approved):
LGTM @lucyleeow feel free to merge if you're happy

doc/configuration.rst
larsoner: > The users will only use these settings knowing what they do so it should be obvious that setting a different value at the block-level will result in different or inconsistent behaviour to other blocks that inherit the global setting. i.e., I don't really see what new information the user is learning by warning them that changing the setting at the block-level will result in a different behaviour when compared to other blocks.. that's the whole point of block-level settings

@tpvasconcelos when thinking about this issue it's important to be clear about who "users" are. I think you are using the term to mean "consumers of sphinx-gallery, i.e., (largely) Python library authors building a gallery". My warning has more to do with "end users of a given Python library who are browsing and using the Sphinx-gallery built doc". It's for these latter (end-)users where the inconsistency or surprising behavior can happen. Hopefully the warning I added makes this a bit clearer!

tpvasconcelos: @larsoner gotcha! Sounds reasonable to me 👍🏽

tpvasconcelos:
Thanks @larsoner and @lucyleeow for the prompt feedback and fast turnaround on this issue! 🚀🚀

sphinx_gallery/tests/test_gen_rst.py
```diff
@@ -1043,24 +1043,33 @@ def test_capture_repr(
...
     gallery_conf,
     caprepr_gallery,
     caprepr_file,
+    caprepr_block,
     expected_out,
     req_mpl,
     req_pil,
     script_vars,
 ):
     """Tests that per file capture_repr overrides gallery_conf."""
```
lucyleeow, line 1065: Nitpick, let's update this docstring too.

lucyleeow (approved):
Just a small nitpick

sphinx_gallery/tests/test_gen_rst.py
tpvasconcelos: Nice catch!

larsoner:
```suggestion
    """Tests that per file/block capture_repr overrides gallery_conf."""
```

larsoner:
Pushed nitpick fix and marking for merge-when-green, thanks in advance @tpvasconcelos !

PR #1401, opened 2024-11-07
DOC Improve custom sort key docs

lucyleeow:
Follows after #1391

* Moves the section on custom sort keys, so it does not live with the subsection docs as it applies more generally to subsection order configs as well as minigallery sort
* Adds info on exactly what is being sorted for all 3 configs that allow custom sorting
* Adds note that `within_subsection_order` allows class input, which will be instantiated by SG
* Minor update to the 'Managing multiple gallaries' section
* Adds docstring and comments to `_get_callables`

@drammock if you have time, would you take a look at this as it follows on from your PR, thank you

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -299,8 +299,8 @@ is changed.
 Manage multiple galleries
 =========================
 
-To specify the location of your gallery folder(s), use the Sphinx-Gallery
-configuration dictionary keys:
+To specify the location of your input and output gallery folder(s), use the
+Sphinx-Gallery configuration dictionary keys:
 
 * ``examples_dirs`` (path to folder containing source example files)
 * ``gallery_dirs`` (path to where output files should be placed)
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -225,20 +225,23 @@ Custom sort keys
 
 You can create a custom sort key callable for the following configurations:
 
-* :ref:`subsection_order <sub_gallery_order>` - to sort on paths to subsection folders
-  relative to the ``conf.py`` file
-* :ref:`within_subsection_order <within_gallery_order>` - to sort on full paths to
-  example files
-* :ref:`minigallery_sort_order <minigallery_order>` - to sort on full paths
-  to example files and :ref:`backreference files <minigalleries_to_examples>`
+* :ref:`subsection_order <sub_gallery_order>` - to reorder subsections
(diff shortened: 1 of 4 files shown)


sphinx_gallery/tests/test_gen_gallery.py
```diff
@@ -345,7 +345,7 @@ def _check_order(sphinx_app, key, expected_order=None):
     content=_template_conf.format(
         imports="",
         subsection_order='["src/second-subsection", "src/first-subsection"]',
-        within_subsection_order='"NumberOfCodeLinesSortKey"',  # this is the default
+        within_subsection_order="sphinx_gallery.sorting.NumberOfCodeLinesSortKey",  # this is the default
```
lucyleeow, line 348: This was just because I realise we never test passing the `NumberOfCodeLinesSortKey` object, which we still support for backwards compatibility. I think this is okay as we test 'NumberOfCodeLinesSortKey' as string elsewhere, but let me know @drammock

doc/configuration.rst
```diff
@@ -249,21 +299,30 @@ is changed.
...
+They can be set in your Sphinx ``conf.py`` file::
 
     sphinx_gallery_conf = {
         ...
         'examples_dirs': ['../examples', '../tutorials'],
         'gallery_dirs': ['auto_examples', 'tutorials'],
     }
 
-Keep in mind that both lists have to be of the same length.
+Keep in mind that both lists should be of the same length.
```
lucyleeow, line 317: Technically, it can be of differing lengths, we'll just give a warning.

drammock (approved):
left some minor-ish suggestions for improvement / clarification. But overall I think you've covered all the bases quite nicely, thanks!

doc/configuration.rst
```diff
@@ -218,6 +218,56 @@ be used with shorter direct alias strings like ``"FileNameSortKey"`` (see
...
+
+Custom sort keys
+================
+
+You can create a custom sort key callable for the following configurations:
+
+* :ref:`subsection_order <sub_gallery_order>` - to sort on paths to subsection folders
+  relative to the ``conf.py`` file
+* :ref:`within_subsection_order <within_gallery_order>` - to sort on full paths to
+  example files
```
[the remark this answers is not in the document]
lucyleeow: Yup that makes sense, user does not need to know this detail and its clearer without.

lucyleeow:
Thanks so much for your thorough review!

PR #1402, opened 2024-11-07
MNT Add test for `minigallery_sort_order`

lucyleeow:
Follow on after  #1391

Adds check that callable works for `minigallery_sort_order`

Note I had to import the custom sort function as I needed to build (and not just `create_sphinx_app`) and thus could not define function in config file as it is not it is not pickle-able (for in new sphinx)


sphinx_gallery/utils.py
```diff
@@ -339,3 +339,14 @@ def _custom_example_sorter(filename):
...
+
+def custom_minigallery_sort_order_sorter(file):
+    """Importable custom sorter for minigallery_sort_order, used in our test suite."""
+    ORDER = [
+        "plot_3.py",
+        "plot_2.py",
+        "plot_1.py",
+    ]
+    return ORDER.index(Path(file).name)
+    return file
```
drammock, line 352: cruft

```suggestion
```

drammock (commented):
LGTM other than the one comment

sphinx_gallery/utils.py
lucyleeow: Thank you, that was a github co-pilot addition and I didn't realise it was added. :facepalm:

lucyleeow:
Thanks for the review!

Issue #1409, opened 2024-11-26
disable intersphinx links inside code cells ?


jschueller:
I would like to disable hyperlinks inside code cells because there are too much of them (most lines link to a new object or method) and it takes over the highlighting

There seem to be one option for this stuff: `reference_url` which is a dictionary to add links to module inventories
unfortunately setting this entry to None for the corresponding modules does not void the intersphinx settings, neither setting  invalid links

lucyleeow:
Does this help: https://sphinx-gallery.github.io/stable/configuration.html#stylizing-code-links-using-css ?

For `reference_url`, `None` is just the value for the local module. If you use intersphinx and have added the other modules, we may not look at `reference_url`, so that's probably why you are getting the behaviour described.

drammock:
I can attest that the fine-grained classes for styling code links are very useful, and maybe that's enough for @jschueller. But I too sometimes wish I could fully disable all code cell links except for the ones within my module --- because when almost everything is a link, it's harder to do things like highlight just a portion of a line, or put the "cursor" in a particular spot in a code cell and then use the keyboard to highlight a few lines (when you don't want to copy the whole cell).

All that to say: if it were not too hard to actually disable the links rather than just providing styling handles, I would probably use that functionality.

lucyleeow:
I think I'd be okay adding another config to disable linking in code cells but keep in text cells.

Taking a quick look, I think we embed the code links here:

https://github.com/sphinx-gallery/sphinx-gallery/blob/f5d1fdd5c84871a74f6ae235e32fa5729bc9afae/sphinx_gallery/docs_resolv.py#L496-L518

and I *think* sphinx does the text block linking, so it should be fairly easy to turn off code linking but not text linking. @larsoner can confirm.

PR #1410, opened 2024-11-27
Add embed_code_links parameter

jschueller:
see #1409


doc/configuration.rst
```diff
@@ -654,6 +654,9 @@ point to the directory containing ``searchindex.js``, such as
 If you wish to do the same for ordinary reST documentation,
 see :ref:`plain_rst`.
 
+The parameter ``embed_code_links`` can be used to disable links in code cells
+but keep them in text cells.
+
```
lucyleeow, line 659: I would actually move this up higher, so people can find out about this config quicker, e.g.:

```
By default Sphinx-Gallery adds hyperlinks in your example scripts so that
you can link used functions/methods/attributes/objects/classes to their matching
online documentation. To disable this for code blocks, but keep them for text
blocks, set the configuration ``embed_code_links`` to ``False``::

    sphinx_gallery_conf = {
        ...
        'embed_code_links': False
    }

``embed_code_links`` is ``True`` by default, which causes code snippets within
the gallery will appear like this:
```

lucyleeow (commented):
Seems reasonable.

I would add a `_bool_eval` check in `_fill_gallery_conf_defaults` for the new config.

I think we could add a test in `test_gen_gallery.py`, using `@pytest.mark.add_conf` to set the config. Note that the tests here use `sphinx_app_wrapper` which builds the docs in `sphinx_gallery/tests/testconfs/src`

jschueller:
> Seems reasonable.
> 
> I would add a `_bool_eval` check in `_fill_gallery_conf_defaults` for the new config.
> 
> I think we could add a test in `test_gen_gallery.py`, using `@pytest.mark.add_conf` to set the config. Note that the tests here use `sphinx_app_wrapper` which builds the docs in `sphinx_gallery/tests/testconfs/src`

done

sphinx_gallery/tests/test_gen_gallery.py
```diff
@@ -223,6 +223,24 @@ def test_config_backreferences(sphinx_app_wrapper):
...
+@pytest.mark.add_conf(
+    content="""
+sphinx_gallery_conf = {
+    'embed_code_links': False,
+    'examples_dirs': 'src',
+    'gallery_dirs': 'ex',
+}"""
+)
+def test_config_embed_code_links(sphinx_app_wrapper):
+    """Test no warning is issued under the new configuration."""
```
lucyleeow, line 235: In what situation would you expect a warning?

sphinx_gallery/gen_gallery.py
```diff
@@ -364,6 +365,7 @@ def _fill_gallery_conf_defaults(sphinx_gallery_conf, app=None, check_keys=True):
     for key in (
         "promote_jupyter_magic",
         "run_stale_examples",
+        "embed_code_links",
     ):
         gallery_conf[key] = _bool_eval(gallery_conf[key])
```
lucyleeow, line 370: Reminder to myself, I should add a test checking `_bool_eval` works as expected.

lucyleeow (commented):
I think you can add a test that reads in the generated html file and ensure no links are being added.

(For an example, I think `test_minigallery_sort_order_callable` reads a generated file)

lucyleeow:
Another note to myself - I should update the documentation. I did not realise text block links are added by sphinx and are thus dictated by sphinx intersphinx config. We only add the code block links and by default will add for all modules in intersphinx, but check modules in `reference_url` first. We need to add local module to `reference_url`, to add links to local module in code.

jschueller:
I think I need help to write a test

larsoner:
From https://github.com/sphinx-gallery/sphinx-gallery/issues/1409#issuecomment-2502153575 also instead of a pure boolean I wonder if we should let this be a regex string with default `.*` for things to link. If you set it to an empty string it links nothing. For @drammock's use case it might be `r"mne\..*"` for example to link only to MNE objects or whatever

lucyleeow:
> I think I need help to write a test

I had a quick play and something like this should work:

```python
@pytest.mark.add_conf(
    content=r"""
import sys

intersphinx_mapping = {
    "python": (f"https://docs.python.org/{sys.version_info.major}", None),
}
sphinx_gallery_conf = {
    'examples_dirs': 'src',
    'gallery_dirs': 'ex',
    'embed_code_links': True,
}"""
)
def test_embed_code_links(sphinx_app_wrapper):
    """Check `embed_code_links` works."""
    sphinx_app = sphinx_app_wrapper.build_sphinx_app()
    built_example = Path(sphinx_app.outdir, "ex", "plot_1.html")

    with open(built_example, "r", encoding="utf-8") as fid:
        html = fid.read()
    print(html) # Check link has/has not been added in html
```

I thought that since I added python to intersphinx, `print` function in example `plot_1.py` would be linked but it wasn't? Looking at the SG docs themselves, `print` calls in our examples are also not linked, but variables that are built-ins e.g., string or list are. Maybe I've mis-understood something though?

jschueller:
thanks @lucyleeow I added all in one commit

lucyleeow:
@jschueller sorry not sure what you added? AFAICT the test still is just a `print` ?

Also I think @larsoner 's suggestion of having the config be more flexible with a regex would be nice:
https://github.com/sphinx-gallery/sphinx-gallery/pull/1410#issuecomment-2511849494

jschueller:
giving up

PR #1412, opened 2024-11-29
Doc: mention color css property to hide links

jschueller:
Setting "colors: inherit" in the css helps to hide the links from code cells (#1409) in my case, so it can be useful to mention


lucyleeow:
I'm going to leave this for @larsoner because I am very bad with css

larsoner:
Thanks @jschueller !

PR #1417, opened 2024-12-23
Fix incorrect paths for JupyterLite Notebook interface URLs, unpin `jupyterlite-sphinx`, and update JupyterLite integration docs

agriyakhetarpal:
## Description

I noticed that while JupyterLite dropped the "retro" string in the URL for the Notebook interface at some point in time, `sphinx-gallery` still seems to have it – and I recently hit this on https://github.com/scikit-image/scikit-image/pull/7644 where I wanted to use it for the gallery notebooks. This PR fixes it, so that the buttons inserted by `sphinx-gallery` now open the correct link:

`http://127.0.0.1:8000/lite/notebooks/index.html?path=my_notebook.ipynb`,

instead of the broken one below

`http://127.0.0.1:8000/lite/retro/notebooks/index.html?path=my_notebook.ipynb`

It appears that this was not caught in other projects, such as `scikit-learn`, because the "Lab" interface is used there, for which the URL is correct here.

I also tried to add a minimal test to ensure that this breakage does not happen again, by creating a minimal web server as a pytest fixture, but realised that it would require a JupyterLite site to be built during the test suite – which is quite overkill.

## Additional context

N/A

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1521,21 +1521,21 @@ Sphinx-Gallery automatically generates Jupyter notebooks for any examples built
 with the gallery. `JupyterLite <https://jupyterlite.readthedocs.io>`__ makes it
 possible to run an example in your browser. The functionality is quite similar
 to Binder in the sense that you will get a Jupyter environment where you can
-run the example interactively as a notebook. The main difference with Binder
+run the example interactively as a notebook. The main differences from Binder
 are:
 
 - with JupyterLite, the example actually runs in your browser, there is no need
   for a separate machine in the cloud to run your Python code. That means that
   starting a Jupyter server is generally quicker, no need to wait for the
   Binder image to be built
-- with JupyterLite the first imports take time. At the time of writing
-  (February 2023) ``import scipy`` can take ~15-30s. Some innocuously looking
+- with JupyterLite, the first imports take time. At the time of writing
+  (December 2024) ``import scipy`` can take ~15-30s. Some innocuously looking
   Python code may just not work and break in an unexpected fashion. The Jupyter
   kernel is based on Pyodide, see `here
   <https://pyodide.org/en/latest/usage/wasm-constraints.html>`__ for some
   Pyodide limitations.
-- with JupyterLite environments are not as flexible as Binder, for example you
-  can not use a docker image but only the default `Pyodide
+- JupyterLite environments are not as flexible as Binder. For example, you
+  can not use a Docker image, but only the default `Pyodide
   <https://pyodide.org/en/stable/index.html>`__ environment. That means that
   some non pure-Python packages may not be available, see list of `available
   packages in Pyodide
@@ -1549,8 +1549,12 @@ are:
 
 In order to enable JupyterLite links with Sphinx-Gallery, you need to install
 the `jupyterlite-sphinx <https://jupyterlite-sphinx.readthedocs.io>`_ package.
-For `jupyterlite-sphinx>=0.8` (released 15 March 2023) you also need to install
-`jupyterlite-pyodide-kernel`.
+Versions of `jupyterlite-sphinx` and `sphinx-gallery` should be compatible, but
+we recommend `jupyterlite-sphinx>=0.17.1`.`
+For `jupyterlite-sphinx>=0.8` you also need to install
+`jupyterlite-pyodide-kernel`. The latest released version is recommended, but
+recent versions should work as well, and it depends on the version of Pyodide
+that you are using or planning to use.
 
 You then need to add `jupyterlite_sphinx` to your Sphinx extensions in
 ``conf.py``::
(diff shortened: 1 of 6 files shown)


agriyakhetarpal:
For example, the [Lab URL for `plot_1_exp.ipynb`](https://output.circle-artifacts.com/output/job/221bdb31-0c0e-4fc6-be01-61db393bde79/artifacts/0/html/lite/lab/index.html?path=auto_examples/plot_1_exp.ipynb) from one of the recent PRs (#1410) works, but the [corresponding URL for the Notebook interface](https://output.circle-artifacts.com/output/job/221bdb31-0c0e-4fc6-be01-61db393bde79/artifacts/0/html/lite/retro/notebooks/index.html?path=auto_examples/plot_1_exp.ipynb) does not, which will be fixed in this PR to point to [the correct URL](https://output.circle-artifacts.com/output/job/221bdb31-0c0e-4fc6-be01-61db393bde79/artifacts/0/html/lite/notebooks/index.html?path=auto_examples/plot_1_exp.ipynb).

Thank you!

lucyleeow (approved):
LGTM thanks!
Just for info, I assume this change was bought about when https://github.com/jupyterlab/retrolab was archived and integrated into Jupyter Notebook?

agriyakhetarpal:
Thanks for the review!

Yes, this was indeed the case – the last instance of `retrolab` existing was prior to Notebook 4 here: https://github.com/jupyterlite/jupyterlite/blob/fa5f44ff3f0eec2c75cada62fa659f13c9f26ab0/packages/retro-application-extension/src/index.ts#L73

So your assessment is right, it was dropped with https://github.com/jupyterlite/jupyterlite/pull/1019

larsoner:
CI failures should be fixed by #1416 

We should probably add/update jupyterlite min pin to 0.4.2 then, since it seems like that was the release with the linked PR. Or maybe we at least need some doc update?

Also our CIs (`.github/install.sh` and `.circleci/config.yml`) appear to use `"jupyterlite-sphinx>=0.8.0,<0.9.0"` which is pretty old, they're on 0.17 now. We should probably update that, too, and make sure that CircleCI looks okay. @agriyakhetarpal do you want to give it a try?

agriyakhetarpal:
Ah, thank you for fixing the CI failures before I was able to! :)

I agree with both adding a minimum version for JupyterLite and updating the docs about it – just pushed the relevant changes via 821d17dc5d3f55ec05ab0c116ae9f860f35a46cf and 616e5479580ceb26c95a3657ab82421a1ec919fc. Suggestions on improving the documentation are appreciated. I think it could make sense to move the sentence to an admonition too.

Minor points about `jupyterlite-sphinx` and the surrounding `jupyterlite-pyodide-kernel` dependencies: [we released version `0.17.1`](https://github.com/jupyterlite/jupyterlite-sphinx/releases/tag/v0.17.1) for the former which fixed a metadata issue with `0.17.0`, so I would recommend that as the minimum here; for the latter, I would recommend dropping the upper bound, since both the Pyodide version and `micropip` are bound to `jupyterlite-pyodide-kernel` and users might not be able to find that out themselves without a few rounds of searching online.

doc/configuration.rst
```diff
@@ -1549,8 +1549,12 @@ are:
 
 In order to enable JupyterLite links with Sphinx-Gallery, you need to install
 the `jupyterlite-sphinx <https://jupyterlite-sphinx.readthedocs.io>`_ package.
-For `jupyterlite-sphinx>=0.8` (released 15 March 2023) you also need to install
-`jupyterlite-pyodide-kernel`.
+Versions of `jupyterlite-sphinx` and `sphinx-gallery` should be compatible, but
```
lucyleeow, line 1552:
```suggestion
Versions of ``jupyterlite-sphinx`` and Sphinx-Gallery should be compatible, but
```

doc/configuration.rst
```diff
@@ -1549,8 +1549,12 @@ are:
...
+we recommend `jupyterlite-sphinx>=0.17.1`.`
+For `jupyterlite-sphinx>=0.8` you also need to install
+`jupyterlite-pyodide-kernel`. The latest released version is recommended, but
```
lucyleeow, line 1555:
```suggestion
we recommend ``jupyterlite-sphinx>=0.17.1``.
For ``jupyterlite-sphinx>=0.8`` you also need to install
``jupyterlite-pyodide-kernel``. The latest released version is recommended, but
```

doc/configuration.rst
lucyleeow: I don't quite follow the meaning here. Do you mean that the user should select versions of `jupyterlite-sphinx` and SG that are compatible or that all versions of the two packages are compatible?

doc/configuration.rst
```diff
@@ -1549,8 +1549,12 @@ are:
...
+recent versions should work as well, and it depends on the version of Pyodide
+that you are using or planning to use.
```
lucyleeow, line 1557:
```suggestion
recent versions should work as well, this depends on the version of Pyodide
that you are using or planning to use.
```

lucyleeow (commented):
Thank you for the doc updates. Just some nits, you have to double backtick here (single is just italics) because we don't set a default.

doc/configuration.rst
agriyakhetarpal: Thanks for raising it – I think what I wish to convey here is that recent versions for both packages should work fine. How should we better phrase it?

lucyleeow: 1. Yup that sounds great!

lucyleeow (approved):
Thank you!

PR #1419, opened 2025-01-07
DOC Improve doc on linking code blocks

lucyleeow:
Follows on from https://github.com/sphinx-gallery/sphinx-gallery/pull/1410#issuecomment-2504954954

@larsoner please let me know if I got this part right.

Also any insight into why built-in python functions are not linked, e.g., `print` (but variables that are built-ins e.g., string or list are)? ref: https://github.com/sphinx-gallery/sphinx-gallery/pull/1410#issuecomment-2513478301

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -619,9 +619,11 @@ more information.
 Add intersphinx links to your examples
 ======================================
 
-Sphinx-Gallery enables you to add hyperlinks in your example scripts so that
-you can link used functions/methods/attributes/objects/classes to their matching
-online documentation. Such code snippets within the gallery appear like this:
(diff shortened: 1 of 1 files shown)


larsoner:
Thanks @lucyleeow !

lucyleeow:
> Not sure why print wouldn't show up. It's a py:function in the Python objects.inv

Good to know it's meant to work.

Interestingly it comes up as a gold colour:

image

similar to numbers:

image

from page: https://sphinx-gallery.github.io/stable/auto_examples/plot_3_capture_repr.html

I would like to know why but I probably don't have time to investigate :sweat_smile:

PR #1426, opened 2025-01-24
Fix label checker workflow

lucyleeow:
See: https://github.com/sphinx-gallery/sphinx-gallery/pull/1423#issuecomment-2605382648

I think we have the same problem as in https://github.com/orgs/community/discussions/30923

Not sure on this though due to my poor bash skills, please have a look @larsoner thank you!


larsoner:
Indeed I think we need to pull the latest labels rather than using `event` which can be stale, awesome @lucyleeow !

PR #1430, opened 2025-02-01
Fix: Fix minigallery duplicates

### LINKED ISSUE #1259: `minigallery` with multiple files: deduplicate and should not start new rows
When using `minigallery` with multiple files, each file will start a new row. It would be nice if they do not start new rows. Also, it would be nice to remove duplicates if multiple files share a same example. The old behavior can be easily retained by calling `minigallery` for the files separately.

image

### LINKED ISSUE #1414: Duplicated gallery examples for a given class in the API documentation
On the scikit-learn documentation, I observed that the sphinx-gallery plugin has generated duplicated links for usage examples of the `PartialDependenceDisplay` class:

https://scikit-learn.org/stable/modules/generated/sklearn.inspection.PartialDependenceDisplay.html

image

The example titled "Advanced Plotting With Partial Dependence" is linked 3 times.

I have not investigated what can possibly cause this.

lucyleeow:
closes #1259
closes #1414 

* `.examples` files are still generated as before, no changes there
* create a dict of backreferences (storing title and intro as well for thumbnail creation) and write to json
* Update API usage and minigallery directive to use the backreferences dict file
   * this should fix the minigallery duplicate problem

### DIFF
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -292,7 +292,7 @@ def identify_names(script_blocks, ref_regex, global_variables=None, node=""):
 THUMBNAIL_TEMPLATE = """
 .. raw:: html
 
-    <div class="sphx-glr-thumbcontainer" tooltip="{snippet}">
+    <div class="sphx-glr-thumbcontainer" tooltip="{intro}">
 
 .. only:: html
 
@@ -319,9 +319,33 @@ BACKREF_THUMBNAIL_TEMPLATE = (
 
 
 def _thumbnail_div(
-    target_dir, src_dir, fname, snippet, title, is_backref=False, check=True
+    target_dir, src_dir, fname, intro, title, is_backref=False, check=True
 ):
-    """Generate reST to place a thumbnail in a gallery."""
+    """Generate reST to place a thumbnail in a gallery.
+
+    Parameters
+    ----------
+    target_dir : str
+        Absolute path to output directory, where thumbnails are saved.
+    src_dir : str
+        Absolute path to build source directory.
+    fname : str
+        Filename of example file.
+    intro : str
+        Introductory docstring of example, to show in tooltip
+    title: str
+        Title of example.
+    is_backref : bool
+        Whether the thumbnail is for a backreference/recommendation file (in which
+        case reST added to prevent thumbnails showing in latex docs)
+    check : bool
+        Whether to check if the thumbnail image file exists.
+
+    Returns
+    -------
+    thumbnail : str
+        reST for a thumbnail.
+    """
     fname = Path(fname)
     thumb, _ = _find_image_ext(
         os.path.join(target_dir, "images", "thumb", f"sphx_glr_{fname.stem}_thumb.png")
@@ -341,12 +365,12 @@ def _thumbnail_div(
 
     template = BACKREF_THUMBNAIL_TEMPLATE if is_backref else THUMBNAIL_TEMPLATE
     return template.format(
-        snippet=escape(snippet), thumbnail=thumb, title=title, ref_name=ref_name
+        intro=escape(intro), thumbnail=thumb, title=title, ref_name=ref_name
     )
 
 
 def _write_backreferences(
-    backrefs, seen_backrefs, gallery_conf, target_dir, fname, snippet, title
+    backrefs, seen_backrefs, gallery_conf, target_dir, fname, intro, title
 ):
     """Write backreference file for one example including a thumbnail list of examples.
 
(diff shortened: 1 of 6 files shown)


larsoner:
`test_minigallery_sort_order_callable` failure looks related

sphinx_gallery/directives.py
```diff
@@ -125,67 +127,88 @@ def run(self):
         # Retrieve source directory
         src_dir = config.sphinx_gallery_conf["src_dir"]
 
-        # Parse the argument into the individual objects
-        obj_list = []
+        # Parse the argument into the individual args
```
lucyleeow, line 130: `objects` is very general and a bit confusing for me, changed to `args` as I think of them as directive args (maybe 'inputs' is another option?)

sphinx_gallery/gen_gallery.py
```diff
@@ -835,6 +846,17 @@ def generate_gallery_rst(app):
     # Write a single global sg_execution_times
     write_computation_times(gallery_conf, None, costs)
 
+    # Write backreferences_all to file
+    if gallery_conf["backreferences_dir"]:
+        _write_json(
+            Path(
+                gallery_conf["src_dir"],
+                gallery_conf["backreferences_dir"],
+                "backreferences_all",
```
lucyleeow, line 855: Not 100% on naming file `backreferences_all.json` , happy to take suggestions

sphinx_gallery/backreferences.py
```diff
@@ -341,12 +366,12 @@ def _thumbnail_div(
 
     template = BACKREF_THUMBNAIL_TEMPLATE if is_backref else THUMBNAIL_TEMPLATE
     return template.format(
-        snippet=escape(snippet), thumbnail=thumb, title=title, ref_name=ref_name
+        intro=escape(intro), thumbnail=thumb, title=title, ref_name=ref_name
     )
 
 
 def _write_backreferences(
```
lucyleeow, line 373: This now also returns a dict of backreferences. Thought about changing the function name but maybe just updating the docstring is adequate?

lucyleeow (commented):
Test failures fixed, sorry I just didn't have the time, last week, to get this PR to review state.

More tests to be added, but I just wanted to check with you @larsoner, that you're happy with how this is looking?

sphinx_gallery/backreferences.py
larsoner: I think it's worth changing it even though it's some code churn. A `_write_` function not writing something will be confusing later I think

sphinx_gallery/backreferences.py
```diff
@@ -394,12 +426,16 @@ def _write_backreferences(
...
+                    intro,
                     title,
                     is_backref=True,
                 )
             )
             seen_backrefs.add(backref)
+            backrefs_example[backref].append(
+                (str(Path(target_dir, fname)), intro, title)
+            )
+    return backrefs_example
```
larsoner, line 438: I think changing to a proper `dict` instead of `defaultdict` is a good idea because it will make it raise KeyError rather than add an entry if you try to access a non-existent key
```suggestion
    return dict(backrefs_example)
```

sphinx_gallery/backreferences.py
lucyleeow: It does write, the `.examples` files are still being written, exactly as before. I think we needed to keep these files to be backwards compatible, and I guess it isn't a huge expense at build time.

I kept the name as I think it is not uncommon for functions to both write out and return, and `_write_and_return_backreferences` is a bit unwieldy.

sphinx_gallery/backreferences.py
lucyleeow: Happy to do this but currently we only ever use `_combine_backreferences` on the returned backreferences - which I think returns a normal dict. I guess if we ever use it for something else in the future it would be useful?

sphinx_gallery/directives.py
```diff
@@ -125,67 +127,88 @@ def run(self):
...
+                return examples
+            return False
+
+        file_paths = {}
+        for arg in arg_list:
+            # Backreference arg input
+            if paths := has_backrefs(arg):
+                # Note `ExampleInfo.arg` not required for backreference paths
+                for path in paths:
+                    file_paths[Path(path[0])] = ExampleInfo(
```
lucyleeow, line 173: Note: `path` is the full path to the example file, from `backreferences_all.json` file, and it should by a perfectly resolved path (thus I don't use `resolve()` here) and it should match paths obtained below via `Path(src_dir).glob(arg)` (thus duplication is prevented in the dict keys)

sphinx_gallery/backreferences.py
lucyleeow: Amended just to be safe!

larsoner (approved):
Looks good, thanks @lucyleeow

PR #1435, opened 2025-02-11
Fix minigallery duplicates and add tests and update documenation

lucyleeow:
Follow up to #1430. A few things happening in this PR (sorry!)

**Fixes duplicates between object and file inputs**

In #1430 the dict key used for objects was `Path(target_dir, filename)`, whereas the dict key used for file inputs was `Path(src_dir, filename)`. Thus, files selected via both object and file path would be duplicated.

There were two ways I could have fixed this:

* Change key for file inputs to use `target_dir` - we would have needed to get the `target_dir` anyway to get the thumbnail and it may have been better to raise an error earlier if input path is not in an `examples_dir` (done in `_get_target_dir`) BUT we probably want to sort on `Path(src_dir, filename)` as it's probably easier for users vs `Path(target_dir, filename)`
* Change key for object inputs to use `src_dir` meaning I write BOTH `target_dir` and `src_dir` to `backreferences_all.json` - probably 'safest' as we know `src_dir` and `target_dir` and do not have to do any complex ambiguity checking (as we do in `_get_target_dir`) but probably a bit much more expensive?

I went with the latter option.

**Tests**

* Adds unit tests in `test_gen_gallery.py` - I've had to amend `sphinx_gallery/tests/testconfs/src/plot_1.py` to use a SG function, so I can test an object backreference input
* Add a check in `test_full.py` via `minigallery.rst`
* Add unit tests for the utility functions i added (`_combine_backreferences`, `_read_json`, `_write_json`)

**Docs**

* Moved general minigallery info under the main header, so users don't have to go under "Create mini-galleries using file paths" to find it
* Amended to say duplicates are removed
* Mention new `backreferences_all.json` file and updated to say '.examples' files to be for backward compatibility only
* Improved section on enabling backreferences (personal nitpick changes)
* Update what will be passed to minigallery sortkey functions

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -701,14 +701,18 @@ Add mini-galleries
 ==================
 
 Sphinx-Gallery provides the :class:`sphinx_gallery.directives.MiniGallery`
-directive so that you can easily add a reduced version of the Gallery to
-your Sphinx documentation ``.rst`` files. The minigallery directive therefore
-supports passing a list (space separated) of any of the following:
+directive so that you can easily add a gallery of specific examples,
+a 'mini-gallery', to your reST. This directive works in both reST text blocks in
+examples and ``.rst`` files.
 
-* fully qualified name of object (see :ref:`references_to_examples`) - this
+The minigallery directive supports passing a list, as a space separated directive
+argument or in the body of the directive. There are two ways to specify examples
+to include in the mini-gallery:
+
+* via fully qualified names of object (see :ref:`references_to_examples`) - this
   adds all examples where the object was used in the code or referenced in
   the example text
-* pathlike strings to example Python files, including glob-style
+* via pathlike strings to example Python files, including glob-style
   (see :ref:`file_based_minigalleries`)
 
 To use object names, you must enable backreference generation, see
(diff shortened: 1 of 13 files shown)


sphinx_gallery/gen_gallery.py
```diff
@@ -1691,7 +1692,6 @@ def setup(app):
     app.connect("config-inited", post_configure_jupyterlite_sphinx, priority=900)
 
     if "sphinx.ext.autodoc" in app.extensions:
-        app.connect("autodoc-process-docstring", touch_empty_backreferences)
```
lucyleeow, line 1694: Not sure about this @larsoner 
Now that we don't use the '.examples' files this is not required for us. But is it possible some 3rd party relies on these empty '.examples' files?!

lucyleeow:
Doc build failing due to #1436

sphinx_gallery/gen_gallery.py
larsoner: > But is it possible some 3rd party relies on these empty '.examples' files (created for objects that are never used in any example)?!

Yeah it probably is, otherwise Sphinx will complain about `.. include::` files that are missing

larsoner (commented):
FYI pushed a commit to try to fix CircleCI but it failed, I'll see if I can push another one to fix it

doc/configuration.rst
```diff
@@ -880,38 +884,32 @@ thumbnails corresponding to the input file strings or object names.
...
+
+Sorting is done on the full paths to all the gallery examples (e.g.,
+``path/to/plot_example.py``) that correspond to the inputs.
 
 See :ref:`own_sort_keys` for details on writing a custom sort key.
 
-As an example, to put backreference thumbnails at the end, we could define the function
-below in ``doc/sphinxext.py`` (note that backreference filenames do not start with
-"plot\_" and ``False`` gets sorted ahead of ``True`` as 0 is less than 1)::
+For example, to put all example thumbnails starting with "plot_numpy_" at the start,
```
larsoner, line 893:
```suggestion
For example, to put all example thumbnails starting with ``"plot_numpy_"`` at the start,
```

larsoner:
Okay the remaining CircleCI failure is just from the `pandas` site being down, feel free to address the comment above @lucyleeow !

sphinx_gallery/gen_gallery.py
lucyleeow: Alright i will add this back in!

lucyleeow:
Alright this is green and I'm merging :crossed_fingers: Thanks for the review and fixing the CI @larsoner !

lucyleeow:
@ogrisel I built the docs locally and checked the duplication is fixed in `PartialDependenceDisplay` you mentioned in #1414:

image

But best not to trust me and check yourself as well (is there a CI job that uses SG dev?), before I make a new release! Thank you! 

cc @glemaitre

ogrisel:
Feel free to open a draft PR on the scikit learn repo to test the dev version of sg and use the `[doc build]` commit message marker to trigger a full gallery build in your PR.

ogrisel:
But since it works locally, i would just trust you instead ;)

lucyleeow:
It looks fixed: https://github.com/scikit-learn/scikit-learn/pull/30820

glemaitre:
And thanks for the release

PR #1442, opened 2025-02-27
MNT Add test for `_bool_eval` and add configs to `_bool_eval` check

### LINKED ISSUE #1411: Add test for `_bool_eval`
Reminder to myself, I should add a test checking `_bool_eval` works as expected.

_Originally posted by @lucyleeow in https://github.com/sphinx-gallery/sphinx-gallery/pull/1410#discussion_r1861325317_

lucyleeow:
closes #1411

* Add test for `_bool_eval`
* Adds bool only configurations to `_bool_eval` check
* Improves build options section, as I *think* any configuration can be altered via `-D` build option, as long as it is one of the accepted types by `-D`. As we now accept callables as string path to object, I *think* these can now also be set via `-D`.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -130,8 +130,8 @@ Build options
 Configuration options can be set at build time via the
 `Sphinx build -D <https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-D>`_
 command line option. This overrides the value set in your ``conf.py`` file for that
-configuration. The value seting in your ``conf.py`` is effectively the 'default',
-as it takes lower precedence than the value passed via the ``-D`` build option.
+configuration. Values set in your ``conf.py`` are effectively the 'default',
+as it takes lower precedence than values passed via the ``-D`` build option.
 
 You can also use the ``-D`` option in your Makefile to create useful targets,
 for example:
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -127,10 +127,22 @@ the rendered examples.
 Build options
 ^^^^^^^^^^^^^
 
-Some options can be set during the build execution step, e.g. using a Makefile:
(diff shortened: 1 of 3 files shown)


doc/configuration.rst
```diff
@@ -1746,9 +1758,19 @@ Building without executing examples
 Sphinx-Gallery can parse all your examples and build the gallery
 without executing any of the scripts. This is just for speed
 visualization processes of the gallery and the size it takes your
-website to display, or any use you can imagine for it. To achieve this you
-need to pass the no plot option in the build process by modifying
-your ``Makefile`` with:
+website to display, or any use you can imagine for it.
+
+This can be done by setting the ``plot_gallery`` configuration in the
```
lucyleeow, line 1763: This was a more opinionated change, to talk about the makefile second, but I thought it was more consistent with the rest of the docs which generally talk about setting the conf in the `conf.py` file first

sphinx_gallery/gen_gallery.py
```diff
@@ -365,8 +365,15 @@ def _fill_gallery_conf_defaults(sphinx_gallery_conf, app=None, check_keys=True):
     # XXX anything that can only be a bool (rather than str) should probably be
     # evaluated this way as it allows setting via -D on the command line
     for key in (
+        "download_all_examples",
```
larsoner, line 368: Can we get rid of the `XXX` comment above now that you've made these changes?

larsoner (commented):
Just one idea / question about a comment update

sphinx_gallery/gen_gallery.py
lucyleeow: Good idea, I've amended the comment just so we know why we are doing the `_bool_eval`

larsoner:
Thanks @lucyleeow !

PR #1448, opened 2025-04-07
Write gallery execution times to JSON

user27182:
The gallery executions times are great, but they currently only generate rst versions. This is great for html rendering, but is not great if you want to parse the execution times (e.g. for testing purposes to limit build times).

This PR re-works the `write_computation_times` function to also generate JSON versions.


larsoner:
Build times should already be in https://sphinx-gallery.github.io/stable/configuration.html#junit-xml I think?

user27182:
> Build times should already be in https://sphinx-gallery.github.io/stable/configuration.html#junit-xml I think?

Didn't know that! Thanks, will look into using that first

user27182:
Yup test times are all there.

larsoner:
PR welcome to improve discovery / docs though!

PR #1450, opened 2025-04-08
DOC Update junit doc

lucyleeow:
Follows from #1448

* Add cross reference
* Updates JUnit docs - make path in example consistent


larsoner:
I suggest we let #1451 land and then rebase this PR

PR #1451, opened 2025-04-08
DOC: Add JUnit XML parsing example

user27182:
The documentation for [using junit xml file](https://sphinx-gallery.github.io/stable/configuration.html#using-junit-xml-files) is helpful but is geared towards integration with tools like `CircleCI`. For some users, however, it may be easier and simpler (or otherwise preferable) to parse the XML manually (e.g. to generate test cases with `pytest` for CI).

This PR adds a small code snippet to the docs for parsing the XML.


doc/configuration.rst
```diff
@@ -1327,6 +1331,92 @@ For more information on CircleCI integration, peruse the related
...
+            f"has an execution time: {testcase['time']} seconds\n"
+            f"which exceeds the maximum allowed: {MAX_TIME} seconds."
+        )
+        assert testcase["time"] < MAX_TIME, msg
+
+Execute the tests with the command
+
+.. code-block:: shell
+
+     pytest sphinx_gallery/tests/test_execution_times.py
```
larsoner, line 1416: This is quite a long example to include in the doc. One option would be to use https://sphinx-design.readthedocs.io/en/latest/ with `dropdown` but I think it might be better to throw this in a `gist` and link to that. That way it can be independently updated / adjusted etc.

user27182: I am not familiar with `gist`, so not sure exactly what that process would look like for this example ... please advise on next steps if this is the way we want to move forward.

Some other ideas:
1. First add the actual tests, and add the docs in a follow-up PR with a direct link to the repository's new test module. That way the linked test code is always up to date.
2. Create a new public API for parsing the XML. Then we can shorten the example quite a bit by replacing the parsing code with a simple function call.

larsoner: Go to https://gist.github.com/ it should hopefully make sense how to add code there, and then getting a link to it, etc. But yes if you plan to have it as a test somewhere and want to link to it then that would work, too.

Don't really want to add an XML parsing thing since it seems like a too-general problem

user27182: Or, maybe your tests _do_ generate docs by default, and the tests in this PR just need to be tweaked? I.e. maybe the `PROJECT_DIR` path isn't quite right or differs slightly in CI compared to when I make the docs and run the tests locally.

Basically, I'm not sure if I should try to debug the tests to make them pass or you'd prefer for me to remove them and stick with just having an example in the docs.

user27182: For context, the example in this PR is used by PyVista in https://github.com/pyvista/pyvista/pull/7386, and I thought it would be good to document this use-case here for others to use.

So, we could also broadly link to `PyVista` for an example of using the XML with pytest. Lots of options, not sure what's preferred here by `sphinx-gallery`

larsoner: As far as tests go you should be able to modify

https://github.com/sphinx-gallery/sphinx-gallery/blob/729fed47e5714d1045fb575be53a17a9273f23ba/sphinx_gallery/tests/test_full.py#L199

In the docs I think a link to a PyVista page or a `gist` would be better than a big code block inline here

user27182: > As far as tests go you should be able to modify
> 
> https://github.com/sphinx-gallery/sphinx-gallery/blob/729fed47e5714d1045fb575be53a17a9273f23ba/sphinx_gallery/tests/test_full.py#L199

Thanks for the tip. The key was that a fixure is needed to generate the xml. I tried integrating the tests, but the parametrization does not allow using fixtures, and so things get a bit complicated and then the actual tests would deviate substantially from the ones given as an example in the docs.

And so, I've removed the tests altogether.

> In the docs I think a link to a PyVista page or a `gist` would be better than a big code block inline here

I refactored the code in the docs and the example is much shorter now (23 lines instead of 60). Is this still too long? If so, I'll update it and make use of these other options

doc/configuration.rst
```diff
@@ -1327,6 +1331,56 @@ For more information on CircleCI integration, peruse the related
...
+    test_cases = [dict(case.attrib) for case in xml_root.iterfind("testcase")]
+    test_ids = [case["classname"] for case in test_cases]
+
+    @pytest.mark.parametrize("testcase", test_cases, ids=test_ids)
+    def test_execution_times(testcase):
+        if float(testcase["time"]) > MAX_EXECUTION_TIME:
+            pytest.fail(
+                f"Gallery example {testcase['name']!r} from {testcase['file']!r}\n"
+                f"Took too long to run: Duration {testcase['time']}s > {MAX_EXECUTION_TIME}s",
+            )
```
larsoner, line 1372: This includes a bunch of stuff that I would consider ancillary / unnecessary to answer the essential question, which I think is "how can I parse the XML to get something meaningful"? So if we're going to include a code block I think it should just answer that question, maybe in 3-4 lines. Maybe just show the maximum execution time for example?

Anyone doing `dev` work (which the SG consumers are doing) will be able to work from that hint to easily extend to their own use case that involves `Path`, `pytest`, `pytest.parametrize`, `pytest.fail` etc. which are really a bit too specific here.

If you want to have the 3-4 liner plus a permalink to the PyVista use case that's fine though

PR #1458, opened 2025-05-07
Update instructions for converting script to notebook

ixjlyons:
Running the command as specified gives:

```
/path/to/python: No module named sphinx_gallery_py2jupyter
```


lucyleeow (approved):
Not sure how we missed that in #1268

lucyleeow:
Thanks @ixjlyons !

PR #1499, opened 2025-10-02
Update FAQ with notebook and RST file guidance

timhoffm:
Added FAQ entries about including notebooks and reStructuredText files in galleries.

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -83,7 +83,7 @@ Example structure:
         ├── example_content.rst
         └── GALLERY_HEADER.rst
 
-you can create a minimal placeholder :file:``example.py``
+you can create a minimal placeholder :file:`example.py`
 
 .. code-block:: python
 
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -70,7 +70,7 @@ Can I add a plain reStructuredText file to a gallery?
 -----------------------------------------------------
 Gallery examples can currently only be generated from ``.py`` files. However, as
 a workaround you can create a minimal placeholder ``.py`` file that includes the
-content from a ``.rst`` file. This may be usefull if an example is mainly text
+content from a ``.rst`` file. This may be useful if an example is mainly text
 and you want to take advantage of editor syntax highlighting.
 
 Example structure:
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -59,3 +59,39 @@ for pointers on how to update your css.
 
 .. seealso::
     `Github PR #906 <https://github.com/sphinx-gallery/sphinx-gallery/pull/906>`_
+
+Can I add a notebook to a gallery?
+----------------------------------
+Gallery examples can currently only be generated from ``.py`` files.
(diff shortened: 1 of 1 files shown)


doc/faq.rst
```diff
@@ -59,3 +59,39 @@ for pointers on how to update your css.
...
+
+.. code-block:: none
+
+    .
+    └── examples
+        ├── example.py
+        ├── example_content.rst
+        └── GALLERY_HEADER.rst
+
+you can create a minimal placeholder :file:``example.py``
```
timhoffm, line 86:
```suggestion
you can create a minimal placeholder :file:`example.py`
```

doc/faq.rst
```diff
@@ -59,3 +59,39 @@ for pointers on how to update your css.
 
 .. seealso::
     `Github PR #906 <https://github.com/sphinx-gallery/sphinx-gallery/pull/906>`_
+
+Can I add a notebook to a gallery?
+----------------------------------
+Gallery examples can currently only be generated from ``.py`` files.
+sphinx-gallery needs the text blocks formatted as reStructuredText, but the text
+blocks in notebooks are Markdown.
```
timhoffm, line 67:
```suggestion
Sphinx-Gallery needs the text blocks formatted as reStructuredText, but the text
blocks in notebooks are Markdown.
```

lucyleeow (approved):
Thanks this is useful!

lucyleeow:
I've just realised, you can manually pass .rst files: https://sphinx-gallery.github.io/stable/configuration.html#manually-passing-files

Is this what you are after?

In your 'plain rst' example addition here, I am assuming there is no code to be executed right? In which case the above config would work right? Maybe we can make that config more obvious/clearer though, as I can see how it could be missed.

timhoffm:
Not quite, AFAICS `copyfile_regex` would *only* copies the `.rst` file to the output folder. It would not create an entry in the list of examples in the gallery's index.

lucyleeow:
Ah good point. We could:

1. add a new config for plain rst files, that uses the same code pathways as `copyfile_regex`, but also adds an entry in the gallery index - we would need to be clear in our docs as how these 2 differ
2. make the `copyfile_regex` config more complex, by optionally allowing you to also add the file to the gallery index 

I think the first one would be better, as the config may be too complex/confusing for the second option

lucyleeow:
> Would there be interest in developing rst-to-gallery into a full feature? -

I've never had a need for this, so I will defer to @larsoner 

> I agree with version 1. Just copying additional files is a semantically different task to creating gallery entries from .rst.

I don't have the cycles for this atm, but happy to review a PR.

timhoffm:
Naming: I’ve just used the [existing naming](https://sphinx-gallery.github.io/stable/configuration.html#parsing-and-executing-examples-via-matching-patterns). Though I agree naming is a bit particular. It seems *examples* is used for the input and *gallery* is used for output, see [here](https://sphinx-gallery.github.io/stable/configuration.html#manage-multiple-galleries). But that’s a [separate topic](https://github.com/sphinx-gallery/sphinx-gallery/issues/1315).

Issue #1065, opened 2023-01-09
[Bug] Sidebar links go to different places than tile links


guberti:
## Issue
When sphinx-gallery generates a gallery of examples, you can navigate to examples by _either_ clicking on the left-hand side menu or by clicking on the tiles in the gallery. Here's a screenshot of what that looks like on [sphinx-gallery's own documentation](https://sphinx-gallery.github.io/stable/auto_examples/index.html#general-examples).

![image](https://user-images.githubusercontent.com/3069006/211429366-e17424c0-1b73-4409-8092-88403567a1f6.png)

---

You would expect both links to go to the same place, but this is not the case. The sidebar link for the "Plotting sin" example goes to:

```
https://sphinx-gallery.github.io/stable/auto_examples/plot_0_sin.html
```

while clicking on the tile takes you to:

```
https://sphinx-gallery.github.io/stable/auto_examples/plot_0_sin.html#sphx-glr-auto-examples-plot-0-sin-py
```

These deliver **different** user experiences - the former is what you'd expect, while the latter has the page scrolled down to the title. This means that users who click the latter link are unable to see sphinx-gallery's note about downloading example code or running in Binder.

![image](https://user-images.githubusercontent.com/3069006/211429831-b519c1d9-6eaf-4d47-8dcf-af9eefb12705.png)

---

## Solution

I'd like to change the tile link so that both go to:

```
https://sphinx-gallery.github.io/stable/auto_examples/plot_0_sin.html
```

This should only require modifying `sphinx_gallery/backreferences.py`, but I'd like @larsoner to first confirm that this is a bug and not intended behavior before I make a PR.

larsoner:
I don't think it's an intentional difference, and agree that the root link is better. @drammock agree with the above suggestion?

drammock:
> I don't think it's an intentional difference, and agree that the root link is better. @drammock agree with the above suggestion?

this has bothered me before too. +1 to fix it.

lesteve:
This would be nice to fix but I am not sure how to do it, my main gripe with it is that https://sphinx-gallery.github.io/stable/auto_examples/plot_0_sin.html#sphx-glr-auto-examples-plot-0-sin-py does not show the note "Go to the end to download the full example code or to run this example in your browser via JupyterLite or Binder"

IIRC this is happening because we insert automatic a target in the example rst e.g.
```rst
.. _sphx_glr_auto_examples_plot_1_exp.py:
```

[...]

PR #1510, opened 2025-11-03
ENH: Use :doc: instead of :ref: to link to examples

timhoffm:
The gallery overview page currently links to individual examples via `:ref:`.

Since every example is its own doc, linking via `:doc:` is actually simpler. This has the following advantages:

- The links in HTML becomes shorter. When :ref: generates a URL with an anchor, e.g. `/auto_examples/local_module.html#sphx-glr-auto-examples-local-module-py`, :doc: generates only the file part `/auto_examples/local_module.html`

  This also has a slightly different effect on the page position
  - https://sphinx-gallery.github.io/stable/auto_examples/plot_0_sin.html is at the top of the page
  - https://sphinx-gallery.github.io/stable/auto_examples/plot_0_sin.html#sphx-glr-auto-examples-plot-0-sin-py is slightly scrolled down to the heading
  
  IMHO it is better to start at the very top of the page.

- In preparatation of supporting user-defined ReST files as example inputs: we cannot rely that there's a suitable label available, so linking via :ref: would be cumbersome for pure-ReST inputs.

### DIFF
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -300,7 +300,7 @@ THUMBNAIL_TEMPLATE = """
   .. image:: /{thumbnail}
     :alt:
 
-  :ref:`sphx_glr_{ref_name}`
+  :doc:`{doc_name}`
 
 .. raw:: html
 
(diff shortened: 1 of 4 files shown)


larsoner:
Agreed with the reasoning here and in #1065, awesome that this fixes that issue as well. Thanks @timhoffm !

PR #1516, opened 2025-11-03
Support plain .rst files as example inputs

### LINKED ISSUE #1511: Support plain .rst files as example inputs
Summary of the discussion in https://github.com/sphinx-gallery/sphinx-gallery/pull/1499#issuecomment-3429438275 and following posts:

For the given file structure
```
    └── examples
        ├── script_example.py
        ├── text_example.rst
        └── GALLERY_HEADER.rst
```
I've envisioned that it just picks up any .rst as it picks up any .py, without the need to explicitly mention them.

Basically, the conf
```
sphinx_gallery_conf = {
    ...
    'example_extensions': {'.py', '.rst'}
}
```
- The special files `GALLERY_HEADER.rst` / `README.rst` are always be excluded.
- This is an opt-in feature, i.e. ` 'example_extensions': {'.py',}` stays as the default. and people have to add '.rst' to enable this.

*Note:* I'm currently working on this and have some small perparatory refactoring PRs (#1508, #1509, #1510) - just to let you know in which context they have been created.

timhoffm:
This is a minimal working implementation, which treats the .rst file as single text block.

Testing question: Where would I reasonably add integration tests for this? I don't have full overview on the test architecture. But the point is that this needs a modified config `'example_extensions': {'.py', '.rst'}`, so either it's ok to modify an existing integration test for this, or we'd have to set up a complete separate test with the modified config.

Closes #1511.

### DIFF
--- a/sphinx_gallery/tests/test_full.py
+++ b/sphinx_gallery/tests/test_full.py
@@ -46,7 +46,7 @@ N_FAILING = 4
 N_GOOD = N_EXAMPLES - N_FAILING  # galleries that run w/o error
 # passthroughs and non-executed examples in
 # (examples + examples_rst_index + examples_with_rst + examples_README_header)
-N_PASS = 3 + 0 + 2 + 0
+N_PASS = 4 + 0 + 2 + 0
 # indices SG generates  (extra non-plot*.py file) in
 # (examples + examples_rst_index + examples_with_rst + examples_README_header)
 N_INDEX = 2 + 1 + 3 + 1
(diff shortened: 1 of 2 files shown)


larsoner:
@timhoffm agreed the `tinybuild` scope is no longer so tiny...

A bit late to consider this but in the future I think it's okay to make other, separate variants / sites of full sphinx builds as well. We haven't bothered so far, but given how difficult it sounds like it was to integrate your changes in `tinybuild` it might be worth it next time a feature has complex interactions with the existing `tinybuild` infrastructure

All good to go from your end @timhoffm ?

timhoffm:
Yes, all good from my side. Documentation should be rewritten in a couple of places to reflect that example inputs are not necessarily .py files. But that’s an existing issue since e.g. .cpp is already supported.

I would do this as a follow-up as it may be reasonable to consider the terminology “example” vs. “gallery entry”/“gallery input”.

larsoner:
Okay in it goes, thanks @timhoffm !

Issue #1436, opened 2025-02-11
jupyterlite-sphinx now depends on Jupytext, which requires sphinx-gallery<=0.7.0


lucyleeow:
Docs build in failing in #1435

> /home/circleci/python_env/lib/python3.10/site-packages/jupytext/cell_reader.py:20: UserWarning: Sphinx Gallery in version 0.19.dev0 is not supported by Jupytext. Please use sphinx-gallery<=0.7.0 instead. If that is an issue, feel free to report it at https://github.com/mwouts/jupytext/issues, or even better, prepare a PR to handle the new signature of sphinx_gallery.notebook.rst2md.

ref: https://app.circleci.com/pipelines/github/sphinx-gallery/sphinx-gallery/2884/workflows/76705b1e-9262-4ed3-9c2c-4243eccda4af/jobs/6736

See https://github.com/mwouts/jupytext/issues/614 for the issue in jupytext

I think this is due to the new jupyterlite-sphinx: jupyterlite-sphinx==0.18.0 [requires: jupytext]

Not sure what the best approach is here, maybe to fix this in jupytext? I'm afraid I don't have the spare cycles for this, minigallery was a bigger lift than I expected.

cc @agriyakhetarpal maybe you can comment on this?

lucyleeow:
Yeah it's wild that such an old version of Sphinx-Gallery is pinned but it sounds like that won't change unless someone makes a PR to fix.

@agriyakhetarpal I assume the current intention is that future jupyterlite-sphinx versions will depend on jupytext?

agriyakhetarpal:
Hi @lucyleeow and @larsoner,

Thanks for pinging me! I would say that the warning in its current state is spurious[^1], as the upstream issue for which Sphinx-Gallery was pinned has existed for a long time (almost four years at this point, as mentioned). I have been filtering/ignoring the warning while working on a few downstream projects' documentation websites recently that make use of Sphinx-Gallery, as it does not affect the addition of `jupyterlite-sphinx` or Sphinx-Gallery's JupyterLite integration for them.

[...]

PR #1525, opened 2025-11-27
MNT Update jupyterlite-sphinx version in circle CI

lucyleeow:
Noticed when a comment was removed in #1514

Removes the `!=0.18.0` restriction for  jupyterlite-sphinx in our circle CI - jupytext now supports recent versions of SG (since 1.17 released in April https://github.com/mwouts/jupytext/releases/tag/v1.17.0), so I think we no longer need this restriction, see: #1436


lucyleeow:
Thanks @larsoner ! I updated the github release notes to mention that min python is now 3.10

PR #1541, opened 2025-12-28
MNT: Add type annotations to sphinx_gallery.backreferences

timhoffm:
Follow-up to #1539.

### DIFF
--- a/sphinx_gallery/backreferences.py
+++ b/sphinx_gallery/backreferences.py
@@ -247,15 +247,16 @@ def identify_names(
 
     Returns
     -------
-    example_code_obj : Dict[str, Any]
+    example_code_obj : dict[str, list[dict[str, Any]]]
         Dict with information about all code object references found in an
-        example. Dict contains the following keys:
+        example. Keys are the names as used in the example, values are lists of
+        possible resolved objects with the following keys:
 
-            - example_code_obj['name'] : function or class name (str)
-            - example_code_obj['module'] : module name (str)
-            - example_code_obj['module_short'] : shortened module name (str)
-            - example_code_obj['is_class'] : whether object is class (bool)
-            - example_code_obj['is_explicit'] : whether object is an explicit
+            - code_obj['name'] : function or class name (str)
+            - code_obj['module'] : module name (str)
+            - code_obj['module_short'] : shortened module name (str)
+            - code_obj['is_class'] : whether object is class (bool)
+            - code_obj['is_explicit'] : whether object is an explicit
                 backreference (referred to by sphinx markup) (bool)
     """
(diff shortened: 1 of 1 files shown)


sphinx_gallery/backreferences.py
```diff
@@ -258,7 +259,9 @@ def identify_names(script_blocks, ref_regex, global_variables=None, node=""):
     # Get matches from docstring inspection (explicit matches)
     text = "\n".join(block.content for block in script_blocks if block.type == "text")
     names.extend((x, x, False, False, True) for x in re.findall(ref_regex, text))
-    example_code_obj = dict()  # native dict preserves order nowadays
+    example_code_obj: dict[str, list[dict[str, Any]]] = (
```
timhoffm: can be seen from code below:
```
        cobj = dict(
            name=attribute,
            module=module,
            module_short=module_short or module,
            is_class=is_class,
            is_explicit=is_explicit,
        )
        example_code_obj[name].append(cobj)
```
Actually, cobj would likely better be a namedtuple. But that's a change for a follow-up.

sphinx_gallery/block_parser.py
```diff
@@ -86,10 +98,10 @@ def __init__(self, source_file, gallery_conf):
             self.multiline_cleanup = re.compile(r"\s*")
 
         comment_start = "|".join(self.allowed_comments)
-        allowed_special = "|".join(allowed_special)
-        if allowed_special:
+        allowed_special_str = "|".join(allowed_special)
```
timhoffm: Again, need a new name since the type changes from list[str] to str.

sphinx_gallery/directives.py
```diff
@@ -393,8 +393,7 @@ def _copy_images(self, node):
     # copy image from source to imagedir.  This is
     # *probably* supposed to be done by a builder but...
     # ie myproj/build/html/_images
-    imagedir = os.path.join(self.builder.imagedir, "")
-    imagedir = PurePosixPath(self.builder.outdir, imagedir)
+    imagedir = PurePosixPath(self.builder.outdir, self.builder.imagedir)
```
timhoffm: `os.path.join(... , "")` does nothing, except maybe adding a "/" to the end of the string, but Path does not care wether the string ends with a "/", so we caln leave this step out.

sphinx_gallery/backreferences.py
lucyleeow: Not directly related, but the types here are wrong then?

lucyleeow:
```suggestion
    example_code_obj : dict[str, list[dict[str, Any]]]
```
?

lucyleeow:
Thank you!!

PR #1542, opened 2025-12-29
MNT: Add type annotations to sphinx_gallery.block_parser

timhoffm:


sphinx_gallery/block_parser.py
```diff
@@ -86,10 +95,10 @@ def __init__(self, source_file, gallery_conf):
             self.multiline_cleanup = re.compile(r"\s*")
 
         comment_start = "|".join(self.allowed_comments)
-        allowed_special = "|".join(allowed_special)
-        if allowed_special:
+        allowed_special_str = "|".join(allowed_special)
+        if allowed_special_str:
```
timhoffm: Need a new name to not dynamically change the type of `allowd_sepcial`.

sphinx_gallery/block_parser.py
```diff
@@ -271,10 +287,11 @@ def finalize_block(mode, block):
                     else:
                         block.append(text)
                 else:
-                    block.append(self.continue_text.search(text).group(1))
+                    block.append(self.continue_text.search(text).group(1))  # type: ignore[union-attr]
```
timhoffm: Mypy complains that search may return `Match | None`, and for the None case the `.group(1)` dereferencing would fail. We haven't observed that behavior at runtime. I haven't completely unraveled the logic here, so in theory there could be edge cases that crash. But let's not bother for now.

PR #1543, opened 2025-12-30
MNT: Add type annotations to sphinx_gallery.directives

timhoffm:
Follow-up to #1539.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -349,8 +349,8 @@ intersphinx_mapping = get_intersphinx_mapping(
 )
 
 nitpick_ignore = [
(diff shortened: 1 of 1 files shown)


doc/conf.py
```diff
@@ -348,6 +348,12 @@ def setup(app):
     },
 )
 
+nitpick_ignore = [
+    # autodoc creates this reference, but docutils does not use sphinx, so that
+    # an intersphinx link cannot be created
```
lucyleeow:
```suggestion
    # autodoc creates this reference (for `ImageSg.run`), but docutils does not use
    #  sphinx, so that an intersphinx link cannot be created
```

Issue #839, opened 2021-06-25
Add type annotations


lucyleeow:
Any thoughts on adding type annotations to our code?

cc @larsoner

larsoner:
I haven't worked with them much so I'm not sure what the benefits are or how difficult it will be too implement properly

lucyleeow:
Type annotations allow for static type checking, which can pick up common bugs, mypy have a good intro: https://mypy.readthedocs.io/en/stable/introduction.html I've been working with them recently and have found them helpful, but I realise for SG it is mostly better docstring and parameter documentation that I think would be helpful.

I think I will do some work on just filling out docstrings more and leave formal type annotations for now as they require more work.

timhoffm:
I close this as complete. We have a full initial set of type annotations and mypy running via pre-commit to validate them.

PR #1544, opened 2026-01-09
MNT: Add type annotations to sphinx_gallery.docs_resolv

timhoffm:
Works towards #839.

### DIFF
--- a/sphinx_gallery/docs_resolv.py
+++ b/sphinx_gallery/docs_resolv.py
@@ -2,6 +2,8 @@
 # License: 3-clause BSD
 """Link resolver objects."""
 
+from __future__ import annotations
+
 import gzip
 import json
 import os
@@ -124,7 +126,7 @@ class SphinxDocLinkResolver:
 
     Parameters
     ----------
-    doc_url : str | Path
+    doc_url : str | pathlib.Path
         The base URL of the project website.
     relative : bool
         Return relative links (only useful for links to documentation of this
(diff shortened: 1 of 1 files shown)


larsoner (commented):
Looks totally reasonable! But I expected an update to add `sphinx_gallery.docs_resolv` to
```
[[tool.mypy.overrides]]
module = [
  "sphinx_gallery.backreferences",
  "sphinx_gallery.py_source_parser",
  "sphinx_gallery.utils",
]
check_untyped_defs = true
```
no?

timhoffm:
> But we should have it in the list of files that gets these checks now, no?

To clarify mypy's default behavior: All functions with type signatures are inspected. Functions without type signatures are ignored.

```toml
[[tool.mypy.overrides]]
module = [...]
check_untyped_defs = true
```
makes mypy additionally type-check function bodies without type signatures in the listed modules.

Functions with signatures a inspected no matter whether I put the module in the above list. Since I now type all functions in a module, adding that module to the above override does not have any effect as it does not have untyped functions. When we are fully typed that override is not needed anymore.

You may add a pre-commit hook, but I'm not sure how valuable that is as long as the typing is work-in-progress. It'd only have a value to detect errors if you modify the already typed code. I would have said, it's early enough to systematically bother with typing on regular contributions when the typing is rather complete.

larsoner:
> Since I now type all functions in a module, adding that module to the above override does not have any effect as it does not have untyped functions. When we are fully typed that override is not needed anymore.

True, until the next person who doesn't think about type annotations goes to edit that file and adds an untyped function. Now that module is no longer completely typed and it's up to the reviewer instead of an automatic CI process to catch that error. So without incremental additions to that list plus some automated CI process I think it's too easy to regress.

Issue #1546, opened 2026-01-13
For `'nested_sections': False`, custom labels for the (sub)sections should not be stripped


gregreich:
I ran into similar problems as those described in #1323, and "fixed" them using `'nested_sections': False`. However, this makes it very hard to label and reference those subsection containing the gallery, because sphinx-gallery is very meticulous when removing the labels I have inserted into their respective GALLERY_HEADER.rst files.

lucyleeow:
Labels as in sphinx reference labels?

I had a quick look at the code and tried on a gallery header file in our tinybuild and it does not look like we sanitize the gallery header? 

Can you give a minimal reproducer?

gregreich:
Thanks for the quick reply, @lucyleeow . 

I augmented the example from #1323 by adding a label to `examples/tutorials/2d/README.txt`, and reference it in `examples/tutorials/README.txt`. Compiling with `'nested_sections': True`, the link works; compiling with `'nested_sections': False` the label is not found.

[pypkg.zip](https://github.com/user-attachments/files/24638552/pypkg.zip)

lucyleeow:
Hopefully fixed in [#1550](https://github.com/sphinx-gallery/sphinx-gallery/pull/1550), `nested_sections=False` was not tested at all when that config was added, so I am not surprised there were several bugs.

gregreich:
I installed the PR #1550 and successfully tested it with my real application. Thanks!

PR #904, opened 2022-01-12
[WIP] Handle nested structures

alexisthual:
Fixes #855.

Let's assume the gallery structure is the following (note the new `readme.rst` files in each subsection):

```
- top_level
   - readme.rst
   - subdir
      - readme.rst
      - example.rst
      - example2.rst
   - subdir2
      - readme.rst
      - example1.rst
      - example2.rst
```

This PR:
- generates an `index.rst` file for each subsection; it contains `readme.rst`, the gallery items, and a hidden toctree of the gallery items
- includes each subsection's index file in a hidden toctree appended at the end of the main sections index file

On-going tasks:
- [x] passing tests
- [x] try more levels of nesting (EDIT: outcome is it currently works for 2, doesn't work for more)

PR #1549, opened 2026-01-16
Fix `generate_gallery_rst` when `nested_sections=False` user provides own `index.rst`

lucyleeow:
Edge case from when we added `nested_sections` (#904) and then allowed user to pass their own `index.rst` (#1071).

When user provides their own `index.rst` (`sg_root_index` is False) and `nested_sections=False`, we fall into the `elif` condition, BUT SG would not have created a `index.rst.new` file as `nested_sections=False`.

We need to add the additional condition of `nested_sections=True` and `has_subsection_header`.

https://github.com/sphinx-gallery/sphinx-gallery/blob/afe25ac077ae4fe7e4bfce97748c1b826962fa29/sphinx_gallery/gen_gallery.py#L824-L835

Technically the above problem would also have arisen when `nested_sections=False` and `len(subsection_toctree_filenames)` was 0 (no example files?), but this was too difficult to test

Edit: note I checked that the test fails on main and passes with the changes

Aside, I think in we should push for more unit tests (i.e. `nested_sections=False` should have been tested more extensively in #904, we only check `nested_sections=True` in tiny build and there are no unit tests). The `sphinx_app_wrapper` we use in `test_gen_gallery.py` are quite flexible (though poorly documented) and can be extended. Unit tests would help catch cases like this and also be easier to debug as `test_full.py` is complicated, and is getting more complicated.

(noticed when looking into #1546)

### DIFF
--- a/sphinx_gallery/tests/test_gen_gallery.py
+++ b/sphinx_gallery/tests/test_gen_gallery.py
@@ -298,10 +298,9 @@ def _check_order(sphinx_app, key, expected_order=None):
     else:
         regex = rf".*:{key}=(\d):.*"
         locator = "sphx-glr-thumbcontainer"
-    with open(index_fname, "r", encoding="utf-8") as fid:
-        for line in fid:
-            if locator in line:
-                order.append(re.match(regex, line).group(1))
+    for line in index_fname.read_text(encoding="utf-8").splitlines():
+        if locator in line:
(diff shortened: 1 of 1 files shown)


lucyleeow:
@larsoner fixed conflict and I've amended the test file to always use pathlib instead of `with open`. Hopefully it's right...

PR #1552, opened 2026-01-20
Fix CI label checker

lucyleeow:


lucyleeow:
@larsoner checked that this works (at least for non-bot added labels) - okay to merge if you think adding github token is not a concern?

PR #1557, opened 2026-02-08
MNT: Add type annotations to more modules

timhoffm:
Works towards #839.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -435,3 +435,8 @@ warnings.filterwarnings(
     message="Matplotlib is currently using agg, which is a"
     " non-GUI backend, so cannot show the figure.",
 )
+
+# Workaround to suppress missing link warnings to custom types, e.g. GalleryConfig,
+# created through autodoc.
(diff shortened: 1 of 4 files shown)


doc/conf.py
```diff
@@ -435,3 +435,8 @@ def setup(app):
     message="Matplotlib is currently using agg, which is a"
     " non-GUI backend, so cannot show the figure.",
 )
+
+# Workaround to suppress missing link warnings to custom types, e.g. GalleryConfig,
+# created through autodoc.
+# TODO: There may be better ways to cope with this by adding the types to the docs.
+suppress_warnings = ["ref.class"]
```
timhoffm: Note: Resolution is a bit tricky as sphinx automodule considers the type alias definition as data (appear under [attributes](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html#attributes), but autodoc-generated signatures want to reference to is as class.

IMHO this should not hold up adding annotations. Globally ignoring seems bearable as we do have many :class: refs in the docs anyway.

PR #1568, opened 2026-02-11
Fix `test_tags_in_index_html` docstring

lucyleeow:
Follow up to #1555

cc @Cadair 

Glad my test was helpful, but I should really add some documentation for the sphinx wrapper and decorator stuff


Cadair:
Thanks for fixing this!

The test harness was good, I found it pretty easy to use because I'm familiar with the sphinx one which is similar.

The one thing I reached for but wasn't there was injecting an arbitary file.

PR #1587, opened 2026-03-26
Emit warning when thumbnail not found

### LINKED ISSUE #1462: Incorrect sphinx_gallery_thumbnail_path results in no warning
Following the instructions on [specifying an image as the thumbnail ](https://sphinx-gallery.github.io/stable/auto_examples/plot_4b_provide_thumbnail.html)for Sphinx galleries, I encountered some strange behavior.

- First, the path for rendering the image in the gallery (``.. figure:: ../../_images/agrivoltaics_system.jpg``) differs from the path used for the thumbnail (``# sphinx_gallery_thumbnail_path = '_images/agrivoltaics_system.jpg'``). It seems counterintuitive that the paths for these should be different.

- Second, when the ``sphinx_gallery_thumbnail_path`` does not find an image, it uses a placeholder image but does not raise a warning that the image was not found.

The specific example can be found here: https://github.com/pvlib/pvlib-python/pull/2459

AdamRJensen:
Closes #1462


AdamRJensen:
@larsoner I think this is actually ready for an initial concept review

larsoner (commented):
Looks reasonable to me!

sphinx_gallery/gen_rst.py
```diff
@@ -457,6 +458,13 @@ def save_thumbnail(
     elif os.path.exists(thumbnail_image_path):
         img = thumbnail_image_path
     elif not os.path.exists(thumb_file):
+        if thumbnail_path_set:
+            logger.warning(
+                "sphinx_gallery_thumbnail_path '%s' not found for '%s', "
+                "using default thumbnail.",
+                image_path,
+                src_file,
```
lucyleeow: Could we use f-string here?

sphinx_gallery/gen_rst.py
```diff
@@ -457,6 +458,13 @@ def save_thumbnail(
...
+            )
```
AdamRJensen:
```suggestion
            logger.warning(
                f"sphinx_gallery_thumbnail_path {image_path} not found for "
                f"{src_file}, using default thumbnail.",
            )
```

sphinx_gallery/gen_rst.py
AdamRJensen: The rest of the file uses the older string formatting style. Changing this to f-string also impacts the tests (the assertion of the log_collector should then also be modified).

I think it is better to stick to the convention used in the remainder of the file.

larsoner:
@AdamRJensen we don't get notified on commits, but I see this is green now and comments have been addressed. I see a couple small tweaks to make, so I'll push and mark for merge-when-green if that's okay with you!

AdamRJensen:
> @AdamRJensen we don't get notified on commits, but I see this is green now and comments have been addressed. I see a couple small tweaks to make, so I'll push and mark for merge-when-green if that's okay with you!

@larsoner go for it!

PR #1593, opened 2026-04-10
TST: Prefer Path over os.path in all test code

QuLogic:
But sometimes we need to cast back to `str` because the actual API uses it only. In some cases, there wasn't a huge benefit because of that, but I changed them anyway for consistency across the code base (at least in tests for now).


lucyleeow (approved):
LGTM after a quick skim and trusting the CI, but I'll lett @larsoner take a look and merge

larsoner:
Great, thanks!

Would be great to do something similar to our actual library code as well at some point

QuLogic:
That's the plan, yes.

PR #1594, opened 2026-04-10
Switch from codecs.open to plain open

QuLogic:
There appears to be no reason to use the former in modern Python (perhaps it would be some time ago for Python 2 support), and it is now deprecated in Python 3.14.

With #1593 this fixes all deprecations about `codecs.open`.


lucyleeow (approved):
LGTM thanks!

Issue #1526, opened 2025-11-27
CI with sphinx version 9.0.0 can't resolve environment


lucyleeow:
CI is failing in #1525
https://github.com/sphinx-gallery/sphinx-gallery/actions/runs/19724354946/job/56512953606?pr=1525

The error is:

```
ERROR: Cannot install None, pydata-sphinx-theme==0.6.0, pydata-sphinx-theme==0.6.1, pydata-sphinx-theme==0.6.2, pydata-sphinx-theme==0.6.3, pydata-sphinx-theme==0.8.1, sphinx 9.0.0 (from https://api.github.com/repos/sphinx-doc/sphinx/zipball/master), sphinx-gallery[dev]==0.20.dev0 and sphinx==9.0.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested sphinx 9.0.0 (from https://api.github.com/repos/sphinx-doc/sphinx/zipball/master)
    sphinx-gallery 0.20.dev0 depends on sphinx>=5
    sphinx-gallery[dev] 0.20.dev0 depends on sphinx>=5
    pydata-sphinx-theme 0.8.1 depends on sphinx<5 and >=3.5.4
    sphinx 9.0.0 depends on docutils<0.23 and >=0.20
    pydata-sphinx-theme 0.6.3 depends on docutils<0.17
    pydata-sphinx-theme 0.6.2 depends on docutils<0.17
    pydata-sphinx-theme 0.6.1 depends on docutils<0.17
    pydata-sphinx-theme 0.6.0 depends on docutils<0.17
```

which doesn't really explain why it's gone back to much older versions of `pydata-sphinx-theme` (current version 0.16.1).

Looking at the trace back it seems to start with `sphinx_design` -  sphinx dev is v9.0.0, whereas sphinx-design specifies [`["sphinx>=6,<9"]`](https://github.com/executablebooks/sphinx-design/blob/b3d6eddc5bf436e35c213a29f0dda3c8ffb716f7/pyproject.toml#L31C16-L31C32)

It also tries older versions of `jupyterlite-sphinx` and `jupyterlite_core` but I can't figure out why?

<details open>
<summary>The CI output is:</summary>

INFO: pip is looking at multiple versions of sphinx-design to determine which version is compatible with other requirements. This could take a while.
Collecting sphinx-design (from sphinx-gallery==0.20.dev0)
[...]

PR #1595, opened 2026-04-10
TST Amend `add_rst` to `add_file` and make flexible

lucyleeow:
As discussed in #1568

* changes `add_rst` to `add_file` - as we can add any type of file to src (and we already use it to add non-rst files)
* amends input to allow a dict of format filename: content, allowing any number of files (and any file type) to be added to src

### DIFF
--- a/sphinx_gallery/tests/conftest.py
+++ b/sphinx_gallery/tests/conftest.py
@@ -15,37 +15,6 @@ from sphinx.util.docutils import docutils_namespace
 from sphinx_gallery import docs_resolv, gen_gallery, gen_rst, py_source_parser
 from sphinx_gallery.scrapers import _import_matplotlib
 
-INDEX_RST = """
-=============
-Own index.rst
-=============
-
-Own index.rst file.
-
-.. toctree::
-
-    plot_1
-    plot_2
-    plot_3
-"""
-
-NESTED_PY = """\"\"\"
-Header
-======
-
-Text.
-\"\"\"
-
-a = 1
-"""
-
-GALLERY_HEADER = """
-Gallery header
-==============
-
-Some text.
-"""
-
 
 def pytest_report_header(config, startdir=None):
     """Add information to the pytest run header."""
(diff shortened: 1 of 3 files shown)


sphinx_gallery/tests/conftest.py
```diff
@@ -165,12 +134,16 @@ def conf_file(request):
...
+    """Adds file(s) to environment, see `sphinx_app_wrapper` for details.
+
+    This fixture takes a single `file` kwarg, which should be a dictionary
+    of format {key: <file name to be added>, value: <content to be added to file>}.
+    """
     try:
-        env = request.node.get_closest_marker("add_rst")
+        env = request.node.get_closest_marker("add_file")
     except AttributeError:  # old pytest
-        env = request.node.get_marker("add_rst")
```
lucyleeow: I am wondering if we could remove this since `get_closest_marker` replaced `get_marker` in 2019:

https://docs.pytest.org/en/stable/changelog.html#id768

larsoner: Yes I think so!

lucyleeow: Done!

sphinx_gallery/tests/test_gen_gallery.py
```diff
@@ -562,7 +594,7 @@ def test_binder_copy_files(sphinx_app_wrapper):
     sphinx_app = sphinx_app_wrapper.create_sphinx_app()
     gallery_conf = sphinx_app.config.sphinx_gallery_conf
     # Create requirements file
-    Path(sphinx_app.srcdir, "requirements.txt").open("w")
+    Path(sphinx_app.srcdir, "requirements.txt").touch()
```
lucyleeow, line 597: For posterity, this change was to prevent a pytest warning about not closing the file.

larsoner:
Thanks @lucyleeow !

lucyleeow:
Thanks @larsoner ! What do you think about cutting a new release? Mostly for the new tag feature, though there have been some fixes.
cc @timhoffm

timhoffm:
> Thanks @larsoner ! What do you think about cutting a new release?

Would be good to get sphinx 9 support in. #1526 is now unblocked because pydata-sphinx-theme v0.17.0 released last week supports sphinx 9.

larsoner:
Agreed, probably worth taking care of #1526 then cutting a release

sphinx_gallery/tests/test_gen_gallery.py
QuLogic: Note, this is not strictly the same as `open('w')` will truncate an existing file while `touch()` just updates the timestamp. I suppose I should have checked that this was in a temporary directory and didn't have to worry about tht.

PR #1597, opened 2026-04-14
TYP: Fix typing for back references

QuLogic:
`_write_backreferences` clearly returns a `list` of `Backreferences` in the `dict` (created with `defaultdict(list)`), not a singular `Backreference`.


lucyleeow (approved):
Thank you!

PR #1598, opened 2026-04-14
ENH: Accept Paths in split_code_and_text_blocks

QuLogic:
Tests already pass `Path`s to these functions, so this just codifies that allowance in the types, but also simplifies a few internal cases by assuming there's only `Path`.


QuLogic:
Oops, didn't mean to rebase on top of #1597.

larsoner:
Nice, thanks @QuLogic !

PR #1601, opened 2026-04-15
TST: Use tmp_path fixture instead of tempfile

QuLogic:
This removes the need to manage deletions ourselves, and simplifies some code as we have the Path API available.

Also, note that `BlockParser` accepts `Path`.


larsoner:
Thanks @QuLogic !

PR #1615, opened 2026-06-22
Install and build docs for latest sklearn release

larsoner:
@lucyleeow not sure if this is the fix you had in mind, but this PR uses latest sklearn release and builds docs for the `1.9.X` branch. This will require us to periodically update the sklearn version but that doesn't seem too bad


lucyleeow:
Thank you for looking into this! It was on my list but I didn't get round to it.

They do release every 6 months so I would prefer something more low touch.

Naive question, but why would https://github.com/scikit-learn/scikit-learn/pull/34214 not be incorporated in the nightly wheel...?

larsoner:
Okay I can push a commit to use `pip index versions scikit-learn` to get the latest version... shouldn't require any updates 😄 

Regarding `pip-pre` in principle it should. You could check to see if a wheel has been uploaded in the last two weeks or not. Unfortunately the versions they upload have no numbers or dates beyond `.dev0` so you can't tell how old they are (would be nice if they were `setuptools-scm` or `hatch-vcs` etc. versioned!)

lucyleeow:
Thanks for the fix

I double checked locally and the nightly wheels indeed raise the error so it's not up to date. I will inquire...

larsoner:
Oh actually we do save the pip cache in circle. Since there is no change in the version number, pip could be using the binary from the local cache rather than any new one.

So installing with `--no-cache` would be another option that might have fixed it. Not sure if that'll end up being lower maintenance effort for us or not...

larsoner:
> I double checked locally and the nightly wheels indeed raise the error so it's not up to date. I will inquire...

Though I guess you probably don't have a stale cache problem (though there is a small chance maybe you did?)

lucyleeow:
Nightly wheels are 14 days old because one of the jobs failed and uploader skipped - this has been fixed but we (skl) are/should have a discussion about what should happen with nightly if a job fails.

We *may* want to go back to using nightly wheels, depending on what is decided.
There is still potential for problems with latest version of skl, as we build docs from main...

PR #1617, opened 2026-06-23
Fix docs and reqs for API usage

larsoner:
The `graphviz` Python package does not help with getting the API usage graphs working. Under the hood it uses `.. graphviz`, which command-line builds the graphs using direct calls like `dot` or `neato`. I found this out the hard way when I got mysterious failures in MNE-Python's builds once I moved to 26.04, where `neato` is now no longer installed when installing `sudo apt install graphviz`, you have to add `libgvplugin-neato-layout8` to the list.

This gets rid of `optional-dependencies.show-api-usage`, which I think is a good idea as anyone using it currently will hopefully get a warning that they may need to do something else to get things working. But we should also try to remember to add a note about this removal to our release notes...

### DIFF
--- a/.github/install.sh
+++ b/.github/install.sh
@@ -46,6 +46,12 @@ pip install $EXTRA_ARGS $PIP_DEPENDENCIES \
     -e .${DEP_OPT}
 set +x
 
-if [[ "$DISTRIB" != "mamba" ]] && [[ "$PLATFORM" == "Linux" ]]; then
-    sudo apt install ffmpeg graphviz
+# "pip install pygraphviz" does not guarantee graphviz binaries exist
+if [[ "$DISTRIB" != "mamba" ]]; then
+    if [[ "$PLATFORM" == "Linux" ]]; then
+        sudo apt install ffmpeg graphviz
+    else  # could use brew on macOS pip but it'll take time to install
+        echo "Removing pygraphviz on $PLATFORM when DISTRIB=$DISTRIB"
+        pip uninstall -y graphviz
+    fi
 fi
--- a/.github/install.sh
+++ b/.github/install.sh
@@ -46,12 +46,6 @@ pip install $EXTRA_ARGS $PIP_DEPENDENCIES \
     -e .${DEP_OPT}
 set +x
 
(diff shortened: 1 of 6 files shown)


.github/install.sh
```diff
@@ -46,12 +46,6 @@ pip install $EXTRA_ARGS $PIP_DEPENDENCIES \
     -e .${DEP_OPT}
 set +x
 
-# "pip install pygraphviz" does not guarantee graphviz binaries exist
-if [[ "$DISTRIB" != "mamba" ]]; then
-    if [[ "$PLATFORM" == "Linux" ]]; then
-        sudo apt install ffmpeg graphviz
-    else  # could use brew on macOS pip but it'll take time to install
-        echo "Removing pygraphviz on $PLATFORM when DISTRIB=$DISTRIB"
-        pip uninstall -y graphviz
```
lucyleeow, line 55: Do we no longer need this because neato is no longer installed with graphviz?

lucyleeow (approved):
Thank you!

> But we should also try to remember to add a note about this removal to our release notes...

We could add it to changes.rst in this PR?

.github/install.sh
larsoner: `neato` was never actually installed with the python package `graphviz`. We errantly did `import graphviz` to "check if generating a neato graph will work". But what actually tells you if it will work is (I think) an equivalent of `which neato`, which I implemented with a `subprocess.check_call` here. So we don't need to remove anything from `macos` anymore. It will try `neato --version` which will fail and correctly detect that the graph shouldn't be created there (on our CIs at least)

larsoner:
Docs added!

.github/install.sh
lucyleeow: >  which I implemented with a subprocess.check_call here

here as in this PR?

larsoner: Whoops, thought I pushed that but I hadn't! Pushing now...

larsoner:
Okay based on the error we do need the graphviz package *plus* command line utility. I'll fix tomorrow!

Issue #1621, opened 2026-07-07
Use pathlib.Path internally


timhoffm:
sphinx-gallery makes heavy use of file paths. These are currently mostly implemented as plain strings plus `os` methods.

This is hard to reason about and error prone. I therefore propose to switch to pathlib.Path.

Since this is a substantial change and a not-quite well defined part of the API is public, I propose to tackle this in multiple phases.

1. Expand all functions to accept Paths instead of strings as well. Inside functions switch to pathlib.Path logic.
   - This allows a substantial functional rewrite while keeping backward compatibility.
   - Downside is that we may need additional conversions (coerce inputs to Path, if a path is returned as srting, we need to convert it back). Perfomance impact should still be small, and we'll eventually get rid of most conversions.
2. Start using paths systematically when calling functions (in 1. we still mostly pass strings into the functions), some clearly internal functions may already be converted to drop str support
3. Eventually remove str-paths on most part of the API (may need deprecations)

PR #1622, opened 2026-07-07
MNT: Systematically support pathlib.Path

timhoffm:
This is part 1 of #1621.

> 1. Expand all functions to accept Paths instead of strings as well. Inside functions switch to pathlib.Path logic.
>    - This allows a substantial functional rewrite while keeping backward compatibility.
>    - Downside is that we may need additional conversions (coerce inputs to Path, if a path is returned as srting, we need to convert it back). Perfomance impact should still be small, and we'll eventually get rid of most conversions.

AI disclosure: I've thoroughly refinde  the migration approach with AI. Impelmentation is AI supported. For easier review, different parts of the library are handled in dedicated commits, and I've checked and if needed corrected every step.


timhoffm:
Not yet ready for review. - I've developed on a newer python version and have to make sure this is compatible with the currently min. supported version.

larsoner:
@timhoffm want me to fix the conflict and push any last fixes?

timhoffm:
@larsoner you are very welcome to. I currently don’t have capacity to move this forward.

larsoner:
Thanks @timhoffm !

timhoffm:
@larsoner Thanks for completing this!

PR #1624, opened 2026-07-14
Switch to ty for type checking

larsoner:
The `mypy` failure in https://github.com/sphinx-gallery/sphinx-gallery/pull/1623 is annoying -- let's switch to `ty`. It's much faster and I've been using it successfully in a few projects for months now. It already caught some issues here in the switch!

Worked with Claude Opus 4.8 to make these changes but fully understand them

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,6 @@ optional-dependencies.dev = [
   "jupyterlite-sphinx>=0.17.1",
   "lxml",
   "matplotlib",
-  "mypy",
   "numpy",
   "packaging",
   "plotly",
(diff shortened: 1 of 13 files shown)


larsoner:
... marking for merge-when-green so CIs can stay green

pyproject.toml
```diff
@@ -144,48 +141,37 @@ builtin = "clear,rare,informal,names,usage"
...
   "joblib.*",
+  "matplotlib",
   "matplotlib.*",
+  "memory_profiler",
   "memory_profiler.*",
+  "plotly",
   "plotly.*",
   "pypandoc",
+  "seaborn",
   "seaborn.*",
```
lucyleeow, line 173: There seems to be some duplication here

larsoner: It looks like it, but if you remove the base one you get
```
error[[unresolved-import](https://ty.dev/rules#unresolved-import)]: Cannot resolve imported module `matplotlib`
   --> sphinx_gallery/scrapers.py:157:12
    |
157 |     import matplotlib
    |            ^^^^^^^^^^
```
and putting just `matplotlib*` (no dot) is workable but could then include anything else that starts with that name (not *just* `matplotlib` plus its submodules). The risk of that is low but the two-entry solution seems more correct

PR #1628, opened 2026-07-31
MNT Fix ty errors

lucyleeow:
Fixes the ty errors in: #1623

Details in comments.

### DIFF
--- a/sphinx_gallery/directives.py
+++ b/sphinx_gallery/directives.py
@@ -210,11 +210,16 @@ class MiniGallery(Directive):
             (sortkey,) = _get_callables(
                 config.sphinx_gallery_conf, "minigallery_sort_order"
             )
-        for path, path_info in sorted(
-            file_paths.items(),
-            # `x[0]` to sort on key only
-            key=((lambda x: sortkey(str(x[0]))) if sortkey else None),
-        ):
+        if sortkey:
+            sorted_items = sorted(
+                file_paths.items(),
+                # `x[0]` to sort on key only
+                key=lambda x: sortkey(str(x[0])),
+            )
+        else:
+            sorted_items = sorted(file_paths.items())
+
+        for path, path_info in sorted_items:
             if path_info.intro is not None:
                 thumbnail = _thumbnail_div(
                     path_info.target_dir,
--- a/sphinx_gallery/notebook.py
+++ b/sphinx_gallery/notebook.py
@@ -140,7 +140,7 @@ def rst2md(
     text = re.sub(
         headings,
         lambda match: "{1}{0} {2}".format(
-            "#" * heading_levels[match.group("over", "under")],
+            "#" * heading_levels[(match.group("over"), match.group("under"))],
             *match.group("pre", "heading"),
         ),
         text,


sphinx_gallery/directives.py
```diff
@@ -210,11 +210,16 @@ def has_backrefs(arg: str) -> Any:
             (sortkey,) = _get_callables(
                 config.sphinx_gallery_conf, "minigallery_sort_order"
             )
-        for path, path_info in sorted(
-            file_paths.items(),
-            # `x[0]` to sort on key only
-            key=((lambda x: sortkey(str(x[0]))) if sortkey else None),
-        ):
```
lucyleeow, line 217:
```
error[no-matching-overload]: No overload of function `sorted` matches arguments
     --> sphinx_gallery/directives.py:213:32
      |
  213 |           for path, path_info in sorted(
      |  ________________________________^
  214 | |             file_paths.items(),
  215 | |             # `x[0]` to sort on key only
  216 | |             key=((lambda x: sortkey(str(x[0]))) if sortkey else None),
  217 | |         ):
      | |_________^
      |
  info: First overload defined here
      --> stdlib/builtins.pyi:4720:1
       |
  4720 | / @overload
  4721 | | def sorted(
  4722 | |     iterable: Iterable[SupportsRichComparisonT], /, *, key: None = None, reverse: bool = False
  4723 | | ) -> list[SupportsRichComparisonT]:
  4724 | |     """Return a new list containing all items from the iterable in ascending order.
  4725 | |
  4726 | |     A custom key function can be supplied to customize the sort order, and the
  4727 | |     reverse flag can be set to request the result in descending order.
  4728 | |     """
       | |_______^ First overload defined here
       |
  info: Possible overloads for function `sorted`:
  info:   [SupportsRichComparisonT](iterable: Iterable[SupportsRichComparisonT], /, *, key: None = None, reverse: bool = False) -> list[SupportsRichComparisonT]
  info:   [_T](iterable: Iterable[_T], /, *, key: (_T, /) -> SupportsDunderLT[Any] | SupportsDunderGT[Any], reverse: bool = False) -> list[_T]
```

I guess since `key` could be None or a callable, it couldn't resolve it? The fix is longer but less complex 🤷

sphinx_gallery/notebook.py
```diff
@@ -140,7 +140,7 @@ def rst2md(
     text = re.sub(
         headings,
         lambda match: "{1}{0} {2}".format(
-            "#" * heading_levels[match.group("over", "under")],
+            "#" * heading_levels[(match.group("over"), match.group("under"))],
```
lucyleeow, line 143:
```
  error[invalid-argument-type]: Method `__getitem__` of type `bound method dict[tuple[str | None, str], int].__getitem__(key: tuple[str | None, str], /) -> int` cannot be called with key of type `tuple[str | Any, ...]` on object of type `dict[tuple[str | None, str], int]`
     --> sphinx_gallery/notebook.py:143:19
      |
  143 |             "#" * heading_levels[match.group("over", "under")],
      |                   ^^^^^^^^^^^^^^
      |
```

I guess it doesn't resolve `match.group("over", "under")` to be a 2 item tuple - it resolves it to be a variable length tuple?

Note up in `rst2md` parameter definition we have:

```
heading_levels: dict[tuple[str | None, str], int],
```

larsoner:
Okay for me, thanks @lucyleeow !

PR #1629, opened 2026-08-10
BUG: Fix API usage and execution time generation

larsoner:
Using Claude Opus 5 I looked into something that always bothered me: that `sg_api_usage` and the execution times a) produce an empty file, and b) change every build *even if examples weren't rerun*. This PR fixes those by generating actual files and updating them only if needed, so repeated sphinx-build invocations don't *always* show them as changed (now they only show up as changed if they actually changed). This involves some tedious plumbing into `env` which I wouldn't have had the patience to work out myself from scratch 😄 

Also seeded an `CLAUDE.md` that points to `AGENTS.md`, which we've been doing in MNE-Python.

We don't have an official AI policy, but given that MNE-Python has one now and it mostly mirrors sklearn, I figure we follow suit here. I don't think we need to officially codify it yet unless our AI contribution rate by new contributors etc. goes up a bit.

@lucyleeow feel free to review if you have time. If not, I think our existing unit tests are good enough to get this in and believe nothing broke!


larsoner:
This is mostly a cleanup so I'll go ahead and get this in. Happy to address comments if people get time to review later, though!

PR #1631, opened 2026-08-10
Fix flaky tests

larsoner:
https://github.com/sphinx-gallery/sphinx-gallery/pull/1630 had a flake. Let's see if we can fix it by looking at `md5` instead (should always be produced). Also, try getting `test_full.py` to actually run on Windows, it hasn't been run in quite some time.

Will probably require some iteration but this one is straightforward enough I'll adjust requirements and self-merge once green.

Iterating on this with Opus 5 to draft the changes for me (reviewed by me).


larsoner:
Didn't like that CircleCI was taking ages. Realized it's the MNE's `doc` group install, which we recently sped up (a lot!) by using `uv`. So I'm porting that over here as well.

PR #1633, opened 2026-08-11
Fix parallelization deadlock with recommender

larsoner:
Fixes the deadlock in https://github.com/scikit-learn/scikit-learn/pull/34718


larsoner:
Confirmed to fix things at the sklearn end so in this goes

PR #1635, opened 2026-08-12
Document modern plotly scraper usage

larsoner:
Test for plotly/plotly.py#5701 . Once it lands we can switch to using their `main`, adjust the docs to mention something about which Plotly version is needed to use the `"plotly"` shortcut, then merge here. In the meantime, let's make sure it works!

TODO:

- [x] Get plotly PR merged
- [x] Update docs here with what version they need (dev or X.Y)


plotly_examples/plot_0_plotly.py
```diff
@@ -3,22 +3,24 @@
 Example with the plotly graphing library
 ========================================
 
-Sphinx-Gallery supports examples made with the `plotly library`_.
-Sphinx-Gallery is able to capture the ``_repr_html_`` of plotly figure objects
-(see :ref:`capture_repr`). To display the figure, the last line in your code
-block should therefore be the plotly figure object.
+Sphinx-Gallery supports examples made with the `plotly library`_. To use
+plotly, add its scraper to the list of :ref:`image_scrapers` in the
+``conf.py`` of the project::
```
lucyleeow, line 8: Should we mention the min version of plotly required?

larsoner: Yes! Need to make sure the upstream PR actually lands first. I'll put this into draft mode until then

larsoner:
Okay upstream landed, so I'm going to update the docs here, merge, and cut a release!

PR #1637, opened 2026-08-14
Run the duplicate checks and tag warnings with a type and subtype

larsoner:
1. Fix our deduplication checks (they were silently not working!)
2. Add sphinx-gallery warning types and subtypes so things like the above can be ignored more easily if needed, especially in `-nWT` contexts

Changes drafted by Claude Opus 5, reviewed by me.


larsoner:
This one seems pretty straightforward so I'll mark for merge-when-green

larsoner:
Sklearn failure is `examples/miscellaneous/plot_display_object_visualization.py` gateway timeout. This happens a lot so we should maybe switch to a different example to run at some point

PR #1638, opened 2026-08-15
ENH: Use Sphinx transform mechanic for backreferences

larsoner:
I used Claude Fable 5 to fix some crufty code: in `master` we currently parse backreferences using a bunch of tricky/hacky code that works most but not all the time, isn't complete, has had a number of bugs iteratively worked out over its patchwork functions (I know this because I authored a lot of those hacks...), uses `searchindex`, regex find-and-replace, etc. I always knew there must be a more principled way to do things, and this PR implements it: a `SphinxPostTransform` on the code blocks that goes over the elements and turns them into links in the doctree nodes instead of embedding in HTML via a (buggy) regex find-and-replace.

I rebuilt each of the following, comparing `master` to this PR:

1. `matplotlib`: all 550 examples. It fixed stuff like in `master` we link [`PolarAxes` to `Axes`](https://matplotlib.org/devdocs/gallery/specialty_plots/radar_chart.html), and on this branch it now correctly links to `Axes`.
2. `scikit-learn`: 221 of the examples (avoiding ones with lots of download). 
3. `mne`: 116 of our examples (for speed). 

Claude before-after summary: "zero regressions, three corrected wrong links, and hundreds of new correct ones." 🚀 

The core +/- here is one fewer (now unnecessary) config val (`logger.info`'ed that it's no longer needed), and +400-ish for the new algorithm and -500-ish for the old algorithm.


larsoner:
Let's get this in and see if it works!

lucyleeow:
I tested out an [explain diff skill](https://gist.github.com/geoffreylitt/a29df1b5f9865506e8952488eac3d524) on this PR, just thought I'd share the output:
https://gisthost.github.io/?9ae52f795a508c216e07f5e807bac097

larsoner:
That's neat! Did you pass the test at the end? 😄

drammock:
> That's neat! Did you pass the test at the end? 😄

I did :)

lucyleeow:
Yes!

PR #1645, opened 2026-09-04
Keep the title of an explicit-title cross-reference in tooltips

### LINKED ISSUE #1644: Explicit-title cross-references are mangled in the thumbnail tooltip, two ways
### Summary

A cross-reference written with an explicit title — ``:term:`title <target>``` — does not survive `_sanitize_rst`, so the thumbnail `tooltip=` attribute shows either the wrong words or raw markup. Which of the two depends on whether the title is one word or several.

This is the same class as #136, which is presumably why `_sanitize_rst` exists; this form still slips through it.

### Reproduction

```python
from sphinx_gallery.gen_rst import _sanitize_rst

_sanitize_rst("See the :term:`wind barbs <wind barb>` here.")
# 'See the wind barbs <wind barb> here.'      <- raw markup reaches the reader

_sanitize_rst("See the :term:`ascent <parcel ascent>` here.")
# 'See the parcel ascent here.'               <- the target replaces the title
```

Sphinx renders both as the **title** — `wind barbs` and `ascent`.

### Cause

Two rules in `_sanitize_rst` (`sphinx_gallery/gen_rst.py`, 0.21.0) catch the two cases:

- **`gen_rst.py:323`** — `re.sub(p + r"\S+\s*<([^>`]+)>" + e, r"\1\2\3", string)`. `\S+` cannot span a space, so this matches only a **single-word** title; and the group it keeps is the *target*, so the title is discarded.
- **`gen_rst.py:327`** — `re.sub(p + r"([^`]+)" + e, r"\1\2\3", string)`. A **multi-word** title falls through to here, which keeps the content between the backticks verbatim, `<target>` included — and removes the backticks, so `gen_rst.py:347`, which handles this form correctly, can no longer match.

### A possible disagreement worth settling first

The comment above the first rule says:

```python
# :class:`a.b.c <thing here>`, :ref:`abc <thing here>` --> thing here
```

so keeping the *target* looks intentional. But Sphinx displays the **title** for this form, so the tooltip and the page currently disagree. If the comment's intent is right there is presumably a reason I cannot see, and the multi-word leak is still a bug on its own.

### Suggested fix

Allow a multi-word title, and keep the title:

```python
# before
string = re.sub(p + r"\S+\s*<([^>`]+)>" + e, r"\1\2\3", string)
# after
string = re.sub(p + r"([^`<>]+?)\s*<[^>`]+>" + e, r"\1\2\3", string)
```

Verified through the whole of `_sanitize_rst`, not the rule alone:

| input | current | with the change |
|---|---|---|
| ``:term:`dry adiabats <dry adiabat>``` | `dry adiabats <dry adiabat>` | `dry adiabats` |
| ``:term:`ascent <parcel ascent>``` | `parcel ascent` | `ascent` |
| ``:term:`soundings <sounding>``` | `Two sounding from…` | `Two soundings from…` |
| ``:class:`Sounding <a.b.Sounding>``` | `a.b.Sounding` | `Sounding` |
| ``:term:`dewpoint``` (bare) | unchanged | unchanged |
| ``:class:`~a.b.MyClass```, ` ``literal`` `, `**bold**` | unchanged | unchanged |

Happy to open a pull request with a test if the direction is right — I would rather agree the intent above first, since the change makes the first rule keep the title where the comment says target.

### Versions

sphinx-gallery 0.21.0, Sphinx 9.1.0, Python 3.14.

bjlittle:
Closes #1644.

`_sanitize_rst` mangles ``:role:`title <target>``` three ways, so the thumbnail
`tooltip=` attribute disagrees with what Sphinx renders on the page.

### Before / after

| input | before | after |
|---|---|---|
| ``:term:`wind barbs <wind barb>``` | `wind barbs <wind barb>` | `wind barbs` |
| ``:term:`ascent <parcel ascent>``` | `parcel ascent` | `ascent` |
| ``:py:class:`Sounding <a.b.Sounding>``` | `:pySounding` | `Sounding` |
| ``:std:doc:`the guide <howtos/index>``` | `:stdthe guide` | `the guide` |
| ``:py:mod:`a.b.c``` | `:pya.b.c` | `a.b.c` |
| ``:py:class:`~a.b.MyClass``` | `:py~a.b.MyClass` | `MyClass` |

The last two are not in the issue: widening the role-name group turned out to fix
the bare and `~`-shortened forms of a domain-qualified role as well, not just the
explicit-title one.

### The change

Three edits to the first rule, plus the same role-name widening on the
`:anchor:` catch-all further down (which already kept the title correctly):

- the title pattern `\S+` becomes `[^`<>]+?`, so a **multi-word** title matches.
  Previously it could not span a space, so it fell through to the catch-all rule
  that keeps everything between the backticks verbatim -- leaking the raw
  `<target>` into the tooltip, and stripping the backticks so the correct rule at
  the bottom of the function could no longer match it;
- the group that is kept flips from the target to the **title**, which is what
  Sphinx displays;
- the role-name group `:[^:\s]+:` becomes `:(?:[^:\s`]+:)+`, so a
  domain-qualified role is consumed whole. Matching from the inner colon left
  everything before it -- the `:py` -- untouched as literal text.

### Note for reviewers

This flips one existing assertion. `test_thumbnail_div` had
``:ref:`whatever <better name>``` → `better name`, encoding the old comment's
intent of keeping the target; it is now `whatever`. That is the direction
question raised in the issue, and the one deliberate behaviour change here --
worth a look rather than a skim.

### Testing

New `test_sanitize_rst` in `test_gen_rst.py` covers 18 markup forms across three
positions in the string (start, middle, end). Verified it fails with the fix
reverted -- 27 failed / 27 passed, the failures being exactly the buggy forms and
the previously-working forms staying green.

Full suite: 418 passed, 0 skipped (`test_full.py` included). `pre-commit run -a`
clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013N5ykzsUtz2khPJvLMdEpY

### DIFF
--- a/sphinx_gallery/gen_rst.py
+++ b/sphinx_gallery/gen_rst.py
@@ -322,9 +322,13 @@ def _regroup(match: re.Match[str]) -> str:
 
 def _sanitize_rst(string: str) -> str:
     """Use regex to remove at least some sphinx directives."""
-    # :class:`a.b.c <thing here>`, :ref:`abc <thing here>` --> thing here
-    p, e = r"(\s|^):[^:\s]+:`", r"`(\W|$)"
-    string = re.sub(p + r"\S+\s*<([^>`]+)>" + e, r"\1\2\3", string)
+    # :class:`title <a.b.c>`, :ref:`title <thing here>` --> title
+    # Sphinx renders the explicit title, not the target, so keep the title.
+    # The role name is `(?:[^:\s`]+:)+` rather than a single segment so that a
+    # domain-qualified role (:py:class:, :std:doc:) is consumed whole -- matching
+    # from its inner colon would strand the ":py" as literal text (gh-1644).
+    p, e = r"(\s|^):(?:[^:\s`]+:)+`", r"`(\W|$)"
+    string = re.sub(p + r"([^`<>]+?)\s*<[^>`]+>" + e, r"\1\2\3", string)
     # :class:`~a.b.c` --> c
     string = re.sub(p + r"~([^`]+)" + e, _regroup, string)
     # :class:`a.b.c` --> a.b.c
@@ -348,7 +352,7 @@ def _sanitize_rst(string: str) -> str:
     string = re.sub(r"`([^`<>]+) <[^`<>]+>`\_\_?", r"\1", string)
 
     # :anchor:`the term` --> the term
-    string = re.sub(r":[a-z]+:`([^`<>]+)( <[^`<>]+>)?`", r"\1", string)
+    string = re.sub(r":(?:[a-z]+:)+`([^`<>]+)( <[^`<>]+>)?`", r"\1", string)
 
     # r'\\dfrac' --> r'\dfrac'
     string = string.replace("\\\\", "\\")
--- a/sphinx_gallery/tests/test_backreferences.py
+++ b/sphinx_gallery/tests/test_backreferences.py
@@ -42,7 +42,7 @@ REFERENCE = r"""
         # reST sanitizing
         (
             "1 :class:`~a.b`. 2 :class:`a.b` 3 :ref:`whatever <better name>`",
-            "1 b. 2 a.b 3 better name",
+            "1 b. 2 a.b 3 whatever",
             False,
         ),
         ("use :meth:`mne.io.Raw.plot_psd` to", "use mne.io.Raw.plot_psd to", False),
--- a/sphinx_gallery/tests/test_gen_rst.py
+++ b/sphinx_gallery/tests/test_gen_rst.py
@@ -352,6 +352,39 @@ def test_codestr2rst():
     assert reference == output
 
 
+@pytest.mark.parametrize(
+    "content,expected",
+    [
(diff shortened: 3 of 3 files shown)


bjlittle (commented):
Adversarial review completed: I examined the changed regex behavior, Sphinx explicit-title semantics, qualified-role handling, boundary cases, and the accompanying regression coverage. I found no correctness, compatibility, or test-coverage issues in this change. The implementation matches Sphinx's displayed-title behavior and the tests cover both the reported failures and relevant previously working forms.

🤖 Reviewed by OpenAI Codex: gpt-5.6 sol

bjlittle:
@larsoner You'll need to explicitly approve the CI workflows, thanks!

bjlittle:
The `circleci` failures appear to be caused by unrelated download failures?

larsoner:
Yeah CircleCI has had problems updating Ubuntu stuff in the past 24-48h... hopefully it's better by Monday

larsoner (approved):
Thanks @bjlittle !

PR #1647, opened 2026-09-08
Remove spacing of empty tag lists

PAUL-BERNARD:
#1555 introduces tag filtering, with a `div` on top of every gallery containing the tag filters. That div can be empty if no tag is set, but it is still present, and still adds its `padding-bottom: 2em`, which is usually unwanted and confusing (and not present in previous versions).
This can be seen for example in the matplotlib documentation:
image

This PR proposes to remove styling to that div if no tag is present, so that the padding-bottom is ignored.


larsoner (approved):
Makes sense, thanks!

PR #1649, opened 2026-09-15
Report binder/jupyterlite config errors against the real conf key

hxperl:
`check_binder_conf` and `check_jupyterlite_conf` report configuration errors using their own parameter names — `binder_conf`, `jupyterlite_conf` — but neither is a `sphinx_gallery_conf` key. Users write `binder` and `jupyterlite`, so someone who searches their `conf.py` for the name in the error finds nothing.

On `master` (6f42660):

```
>>> check_binder_conf(["org"])
ConfigError: `binder_conf` must be a dictionary or None.

$ grep -rn 'binder_conf\|jupyterlite_conf' doc/
$
```

The spelling this changes to is the one the module already uses elsewhere — `check_jupyterlite_conf` reports its unknown-keys error as `sphinx_gallery_conf['jupyterlite']`, a few lines below one of the messages changed here, and `gen_gallery.py` does the same for `expected_failing_examples`. So this makes the type and missing-value messages agree with the existing convention rather than introducing one.

Three message strings, plus the test that asserted the old wording.


larsoner:
Nice, thanks @hxperl !

PR #1653, opened 2026-09-17
Fix note/warning admonitions leaking closing tags around nested code-blocks

teonbrooks:
A `.. note::`/`.. warning::` ending in a nested `.. code-block::` renders
with a stray `</p></div>` as visible text after the code, instead of
closing the alert box (mne-tools/mne-python#14311).

`directive_fun` wraps the body in `<p>...</p>`, which isn't valid once the
body contains block content (here, a code-block). Renderers close the tag
early, so the `</p></div>` meant to close the wrapper ends up as literal
text. One solution is to replace the `<p>` wrapper with blank-line wrapping
instead. This PR does that, updates the existing conversion tests for the
new output, and adds a regression test for the nested-code-block case.


drammock (approved):

larsoner:
Thanks @teonbrooks !

